PR #10, opened 2018-10-30
Add an operator env variable for the cluster-node-tuned image.

jmencak:
This PR fixes the issue of fetching external container images.

Further cleanup to follow.

/cc @jeremyeder


jmencak:
@sejug PTAL, thanks!

PR #11, opened 2018-11-05
First steps towards custom profiles.

jmencak:


pkg/controller/objects.go
```diff
@@ -1,159 +0,0 @@
...
-
-import (
-	"os"
-
-	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-)
-
-func (h *Handler) generateDaemonSet(serviceAccount *corev1.ServiceAccount) *appsv1.DaemonSet {
```
sjug: Interesting moving this out of a big struct like you did.

PR #13, opened 2018-11-08
Changing pattern & small build changes

sjug:
WIP

What do you think @jmencak about the syncDaemonSet. I changed the secondary watch as well but it still doesn't work yet. Will expand to monitor the whole cluster.

Suggest rolling out a similar sync to the other objects to short circuit on get, and update after comparison.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ GOFMT_CHECK=$(shell find . -not \( \( -wholename './.*' -o -wholename '*/vendor/
 
 DOCKERFILE=Dockerfile
 IMAGE_TAG=openshift/origin-cluster-node-tuning-operator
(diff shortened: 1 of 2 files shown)


sjug:
At this point all sync functions except the configmaps are following the new get first pattern. Also they do not update as we should to find a way to compare and see if it's necessary. Owner references throughout so deletion of the CRD cleans up nicely enough.

jmencak:
Thanks for the PR!  So at this point I see this a little bit as a step backwards, as the updates are no longer performed.  I can see the benefit of setting the SetControllerReference though.  Can you add the updates to the PR, please?  Thanks.

sjug:
@jmencak Sure, I've added updates back.

PR #14, opened 2018-11-09
Break up the recommend.conf in CR into structs.

jmencak:


jmencak:
@sjug PTAL
/cc @jeremyeder

PR #17, opened 2018-11-15
Add default CR creation under operator control.

jmencak:


jmencak:
@sjug PTAL, thanks

pkg/controller/tuned/tuned_controller.go
```diff
@@ -66,6 +66,12 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	createCustomResource(mgr)
```
sjug: - Why did you pull this into a function?
- Do we want to have the operator create its own CR?

pkg/controller/tuned/tuned_controller.go
```diff
@@ -89,16 +95,17 @@ func syncServiceAccount(r *ReconcileTuned, tuned *tunedv1alpha1.Tuned) error {
...
 
 	sa := &corev1.ServiceAccount{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: saManifest.Name, Namespace: saManifest.Namespace}, sa)
-	if err != nil && errors.IsNotFound(err) {
-		err = r.client.Create(context.TODO(), saManifest)
-		if err != nil {
-			return fmt.Errorf("Couldn't create tuned ServiceAccount: %v", err)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: saManifest.Namespace, Name: saManifest.Name}, sa)
+	if err != nil {
+		if errors.IsNotFound(err) {
```
sjug: Readability changes eh? :smiley:

pkg/controller/tuned/tuned_controller.go
```diff
@@ -234,6 +241,28 @@ func syncDaemonSet(r *ReconcileTuned, tuned *tunedv1alpha1.Tuned) error {
...
+	manifestFactory := manifests.NewFactory()
+
+	// Get the default configuration (CR) for Tuned
+	crManifest, err := manifestFactory.TunedCustomResource()
+	if err != nil {
+		return fmt.Errorf("Couldn't build tuned CustomResource: %v", err)
+	}
+	log.Printf("Built CustomResource manifest for %s/%s\n", crManifest.Namespace, crManifest.Name)
+
+	err = client.Create(context.TODO(), crManifest)
```
sjug: Still want to lead with `Create()`?

jmencak: Yes in this case.  Had issues leading with .Get() in this case.  Prefer code that is tested and works.  Found two other operators that use this too to create the default CR in this way.

PR #23, opened 2018-12-03
Support for multiple CRs.

jmencak:


jmencak:
@sjug PTAL

pkg/controller/tuned/tuned_controller.go
```diff
@@ -278,15 +294,44 @@ func createCustomResource(mgr manager.Manager) error {
 	return nil
 }
 
+func addOwnerReference(meta *metav1.ObjectMeta, tuned *tunedv1alpha1.Tuned) []metav1.OwnerReference {
```
sjug: Your desired functionality of `addOwnerReference()` over the built-in `SetControllerReference()` is to be able to manipulate two owner references (one controller and a secondary)? I would have just wrapped `SetOwnerReferences`.

pkg/controller/tuned/tuned_controller.go
```diff
@@ -278,15 +294,44 @@ func createCustomResource(mgr manager.Manager) error {
...
+	var bController bool = false
+	if tuned.Name == "default" {
+		bController = true
+	}
+
+	if meta.OwnerReferences == nil {
+		meta.OwnerReferences = []metav1.OwnerReference{}
+	} else {
+		for _, owner := range meta.OwnerReferences {
+			if owner.UID == tuned.UID {
```
sjug: Is the UID safe enough to compare versus the `GroupVersion` comparison like in `referSameObject()`?

pkg/controller/tuned/tuned_controller.go
```diff
@@ -278,15 +294,44 @@ func createCustomResource(mgr manager.Manager) error {
 	return nil
 }
 
+func addOwnerReference(meta *metav1.ObjectMeta, tuned *tunedv1alpha1.Tuned) []metav1.OwnerReference {
+	var bController bool = false
```
sjug: The boolean default zero value is already false. Why the name `bController`, maybe `isController`?

pkg/controller/tuned/tuned_controller.go
```diff
@@ -69,7 +69,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 	createCustomResource(mgr)
 	if err != nil {
-		log.Printf("createCustomResource(): %v\n", err)
+		log.Printf("createCustomResource(): %v", err)
```
sjug: Were there changes to the `log.Printf` to add a new line?

pkg/controller/tuned/tuned_controller.go
jmencak: Isn't explicit better than implicit?  isController sounds better, will change, thanks.  b == boolean

pkg/controller/tuned/tuned_controller.go
jmencak: Changes?  Not sure it changed, but the documentation clearly states the '\n' is added if it is not present.  And it definitely makes sense for logging.

pkg/controller/tuned/tuned_controller.go
jmencak: Adding to TODO to look into.

PR #29, opened 2018-12-19
nvme timeout and retry increase per AWS documentation

jeremyeder:
Documented here:  https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html#timeout-nvme-ebs-volumes

@gnufied @eparis


jmencak:
/lgtm

move to cloud-provider specific tuned profiles once they're available in upstream tuned.

PR #30, opened 2019-01-07
Restrict tuned to linux nodes

derekwaynecarr:
Restrict running tuned to nodes that run linux only.

/cc @cuppett @jeremyeder


jeremyeder:
thx. /cc @jmencak

PR #31, opened 2019-01-08
Adding support for tuned "modules" and "scheduler" plugins.

jmencak:
Also adding documentation on supported tuned plugins.


jmencak:
Also see https://github.com/openshift/openshift-tuned/pull/14

@sjug PTAL

PR #36, opened 2019-01-31
Switch to glog and integrate with klog.

jmencak:


pkg/controller/tuned/tuned_controller.go
```diff
@@ -101,12 +101,12 @@ func (r *ReconcileTuned) syncServiceAccount(tuned *tunedv1alpha1.Tuned) error {
 			if err != nil {
 				return fmt.Errorf("Couldn't create tuned ServiceAccount: %v", err)
 			}
-			log.Printf("Created ServiceAccount for %s/%s", saManifest.Namespace, saManifest.Name)
+			glog.V(2).Infof("Created ServiceAccount for %s/%s", saManifest.Namespace, saManifest.Name)
```
sjug: Why the hardcoded `V(n)`? Is that depth?

jmencak: Exactly, it is the depth.  By default verbosity 1 is used at this point, so V(2) will not be shown unless you change the verbosity in the operator's template.  The logging is too chatty by default by my taste, but we still want an optional debugging/tracing verbosity.

sjug: Cool

PR #37, opened 2019-02-04
Rename cluster-node-tuning ClusterOperator

sjug:
As per request from Clayton, we "need to drop “-operator” from the ClusterOperator object name."


jmencak:
So this PR renames not only the ClusterOperator object name, but also other things.  Is this intended?

PR #38, opened 2019-02-05
Upgrading to operator-sdk 0.4.0

jmencak:
@sjug PTAL


sjug:
Tested and working. Didn't seem to need that many changes to the actual operator `pkg/`?

jmencak:
@sjug, not sure which changes you're talking about, but the e2e tests were not working without a few more changes.  If you're talking about the extra cleanup, then yes, this PR is not about just the upgrade, but cleanup too.

sjug:
@jmencak I'm talking about the changes to the operator code which were needed in order to update to the operator-sdk v0.4.0.

PR #44, opened 2019-03-02
assets: specify priority class for operator and tuned

derekwaynecarr:
specifies system-cluster-critical priority for operator.
specifies system-node-critical priority for daemonset.

blocks passing smoke tests for ensuring control plane pods always schedule.

see: openshift/origin#22217


jmencak:
Thank you Derek.

/lgtm

PR #46, opened 2019-03-07
Fix Operator status condition reporting.

jmencak:
Report ClusterOperator status more in line with:
https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md

This PR also improves error reporting for GetConfig() (pulled from
openshift-tuned).


jmencak:
@sjug PTAL, thanks!

sjug:
Helpful documentation, thanks @jmencak.

PR #47, opened 2019-03-12
Prep for landing rhel7/8 containers on the corresponding rhel nodes in 4.1

jmencak:


jmencak:
WIP, but tested, @sjug PTAL

jmencak:
@sjug CI finally happy, please take a look.

pkg/config/config.go
```diff
@@ -15,8 +15,8 @@ const (
 )
 
 // NodeTunedImage returns the operator's operand/tuned image path.
-func NodeTunedImage() string {
-	nodeTunedImage := os.Getenv("CLUSTER_NODE_TUNED_IMAGE")
+func NodeTunedImage(version string) string {
+	nodeTunedImage := os.Getenv("CLUSTER_NODE_TUNED_IMAGE_" + version)
```
sjug: We might want to limit `version` to only `7|8`?

manifests/04-operator.yaml
```diff
@@ -44,7 +44,7 @@ spec:
             value: "node-tuning"
           - name: RESYNC_PERIOD
             value: "600"
-          - name: CLUSTER_NODE_TUNED_IMAGE
+          - name: CLUSTER_NODE_TUNED_IMAGE_RHEL7
```
sjug: Will the config have both keys `CLUSTER_NODE_TUNED_IMAGE_RHEL7` and `CLUSTER_NODE_TUNED_IMAGE_RHEL8`

pkg/controller/tuned/tuned_controller.go
```diff
@@ -375,7 +375,7 @@ func (r *ReconcileTuned) Reconcile(request reconcile.Request) (reconcile.Result,
 		return reconcileResult, err
 	}
 
-	err = r.syncDaemonSet(tunedInstance)
+	err = r.syncDaemonSet(r.manifestFactory.TunedDaemonSetRHEL7, tunedInstance)
```
sjug: This hardcoded version instance is also just for now, like line 32 above in `syncOperatorStatus()`?

pkg/config/config.go
jmencak: Why?

manifests/04-operator.yaml
jmencak: Eventually yes.

pkg/controller/tuned/tuned_controller.go
jmencak: If the decision is made for several daemonsets to be used, then yes, there will be an additional r.manifestFactory.TunedDaemonSetRHEL8

jmencak:
Until it is clear whether 2 daemonsets should really be used or 2 containers one put to sleep.

/hold

PR #52, opened 2019-05-16
assests: fix duplicate resources key in container spec

sjenning:
in service to https://github.com/openshift/origin/pull/22787

PR #53, opened 2019-05-17
assests: fix duplicate resources key in container spec

jmencak:
This is a follow-up PR https://github.com/openshift/cluster-node-tuning-operator/pull/52 which includes the necessary `bindata.go`.  I'll follow up with a PR which will make it unnecessary to run "make generate" when changing resource yamls.

/cc @sjenning


sjenning:
xref https://github.com/openshift/origin/pull/22787

PR #54, opened 2019-05-17
Remove pkg/manifests/bindata.go from the repo, generate it.

jmencak:
This is a Makefile cleanup PR.  File pkg/manifests/bindata.go
is removed from the repository as it was leading to issues as in https://github.com/openshift/cluster-node-tuning-operator/pull/52,
where old bindata.go files were being shipped with modified resources.


sjug:
/lgtm 

Talked to @jmencak about this today and rather than carrying the generated bindata and a script to update it we're just going to generate it as part of the build process.

PR #62, opened 2019-06-06
Add more RelatedObjects for the must-gather tool.

jmencak:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1717739

`openshift-must-gather inspect clusteroperator/node-tuning` now exits successfully (0) and returns all resources created by the operator.

### DIFF
--- a/manifests/04-operator.yaml
+++ b/manifests/04-operator.yaml
@@ -32,9 +32,6 @@ spec:
       containers:
         - name: cluster-node-tuning-operator
           image: registry.svc.ci.openshift.org/openshift/origin-v4.0:cluster-node-tuning-operator
-          ports:
-          - containerPort: 60000
-            name: metrics
           command:
           - cluster-node-tuning-operator
           args:
--- a/pkg/controller/tuned/status.go
+++ b/pkg/controller/tuned/status.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/golang/glog"
 	configv1 "github.com/openshift/api/config/v1"
+	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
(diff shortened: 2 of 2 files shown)


jmencak:
@sjug PTAL, thank you.

manifests/04-operator.yaml
```diff
@@ -32,9 +32,6 @@ spec:
       containers:
         - name: cluster-node-tuning-operator
           image: registry.svc.ci.openshift.org/openshift/origin-v4.0:cluster-node-tuning-operator
-          ports:
```
sjug: Hey @jmencak, why did you get rid of the metrics port?

pkg/controller/tuned/status.go
```diff
@@ -45,11 +53,14 @@ func (r *ReconcileTuned) syncOperatorStatus() (bool, error) {
 		}
 	}
 	coState.Status.RelatedObjects = []configv1.ObjectReference{
-		{
-			Group:    "",
-			Resource: "namespaces",
-			Name:     dsManifest.Namespace,
-		},
+		{Group: "", Resource: "namespaces", Name: tunedManifest.Namespace},
```
sjug: There are a number of `ObjectReference`s without groups like this one. I'm kinda surprised that's the case.

manifests/04-operator.yaml
jmencak: It shouldn't have been there in the first place as it is currently unused.  must-gather tool chokes on this, tries to gather the metrics and exits 1.  Hence the removal.

pkg/controller/tuned/status.go
jmencak: Empty group "" indicates the core API group afaik.  In any case, tested and worked fine.

PR #63, opened 2019-06-07
Bug 1718842: Operator-assisted removal of detached tuned secrets.

jmencak:
This commit adds support for automatic removal of detached tuned
secrets which might have accumulated in the operator namespace
prior to commit 014900fe.

Associated BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1714484

### DIFF
--- a/pkg/controller/tuned/tuned_controller.go
+++ b/pkg/controller/tuned/tuned_controller.go
@@ -3,6 +3,7 @@ package tuned
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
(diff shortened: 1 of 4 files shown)


jmencak:
@sjug  PTAL, thanks!

pkg/controller/tuned/tuned_controller.go
```diff
@@ -122,6 +179,20 @@ func (r *ReconcileTuned) syncServiceAccount(tuned *tunedv1.Tuned) error {
 	} else {
 		// Do not update service account as it leads to accumulating secrets; see BZ1714484
 		glog.V(2).Infof("Tuned ServiceAccount already exists, skipping update")
+
+		isUpgrade, err := r.isOperandUpgrade(tuned)
```
sjug: The secret management is in the `syncServiceAccount()` function because it is the first in the reconcile after the tuned fetch?

pkg/controller/tuned/tuned_controller.go
```diff
@@ -99,6 +99,63 @@ type ReconcileTuned struct {
...
+		if errors.IsNotFound(err) {
+			// Tuned DaemonSet doesn't exist yet; assume it wasn't manually deleted
+			return false, nil
+		} else {
+			return false, fmt.Errorf("Failed to get DaemonSet: %v", err)
+		}
+	}
+	// Tuned DaemonSet exists; compare the manifest and current version of the operand image
+	imageTuned := ntoconfig.NodeTunedImage()
+	return imageTuned != daemonset.Spec.Template.Spec.Containers[0].Image, nil
```
sjug: Is the image the differentiating factor for upgrades?

pkg/controller/tuned/tuned_controller.go
```diff
@@ -99,6 +99,63 @@ type ReconcileTuned struct {
...
+// This is really only necessary to do operator-assisted removal of accumulating secrets prior to commit 014900fe.
+// See BZ1714484
+func (r *ReconcileTuned) removeTunedSecrets(tuned *tunedv1.Tuned) error {
+	glog.V(2).Infof("removeTunedSecrets()")
+
+	opts := &client.ListOptions{}
+	opts.InNamespace(tuned.Namespace)
+	secretsList := &corev1.SecretList{}
+	err := r.client.List(context.TODO(), opts, secretsList)
+	if err == nil {
```
sjug: I would follow the usual:
```
if err != nil {
  return fmt.Errorf("Couldn't get a list of secrets: %v", err)
}

for _, s := range secretsList.Items {
...
}
```

pkg/controller/tuned/tuned_controller.go
```diff
@@ -122,6 +179,20 @@ func (r *ReconcileTuned) syncServiceAccount(tuned *tunedv1.Tuned) error {
 	} else {
 		// Do not update service account as it leads to accumulating secrets; see BZ1714484
 		glog.V(2).Infof("Tuned ServiceAccount already exists, skipping update")
+
+		isUpgrade, err := r.isOperandUpgrade(tuned)
+		if err != nil {
+			glog.Errorf("%v", err)
+			return nil
```
sjug: Probably want to `return err`

pkg/controller/tuned/tuned_controller.go
```diff
@@ -122,6 +179,20 @@ func (r *ReconcileTuned) syncServiceAccount(tuned *tunedv1.Tuned) error {
...
+		}
+		if isUpgrade {
+			// This is really only necessary to do operator-assisted removal of accumulated secrets prior to commit 014900fe
+			glog.V(2).Infof("This is an operand upgrade, removing all tuned secrets.")
+			err := r.removeTunedSecrets(tuned)
+			if err != nil {
+				glog.Errorf("Failed to remove tuned secrets: %v", err)
```
sjug: Do we want to `return err` after this as well?

pkg/controller/tuned/tuned_controller.go
jmencak: Yes.  The secret management is not relevant until you start syncing the daemonset, so anything before that should be good.

pkg/controller/tuned/tuned_controller.go
jmencak: Good question.  I'd say it doesn't have to be in theory, but it will be in practice.  Happy to accept better ways of establishing when the operator is going through an upgrade!  The worst can happen is not removing the secrets if we miss an operator upgrade.

pkg/controller/tuned/tuned_controller.go
jmencak: Ack, will change.

pkg/controller/tuned/tuned_controller.go
jmencak: I'd rather not clean than re-trigger the reconciliation loop.

pkg/controller/tuned/tuned_controller.go
jmencak: Don't think we do.  The reason as above.

jmencak:
Addressing reviewer comments.  The removal of secrets is now performed in syncDaemonSet() where the tuned service account is used and the operator upgrade is now detected via `RELEASE_VERSION`.

/hold cancel

pkg/controller/tuned/tuned_controller.go
```diff
@@ -249,6 +291,25 @@ func (r *ReconcileTuned) syncDaemonSet(tuned *tunedv1.Tuned) error {
 		}
 	} else {
 		glog.V(2).Infof("Tuned DaemonSet already exists, updating")
+
+		operatorReleaseVersion := os.Getenv("RELEASE_VERSION")
+		operandReleaseVersion := ""
+
+		for i, e := range daemonset.Spec.Template.Spec.Containers[0].Env {
```
sjug: Given this new snippet appears twice, may be worth putting in it's own function?

sjug: Also not necessarily an issue, I think using both the key and value is unnecessary.

pkg/manifests/manifests.go
```diff
@@ -136,6 +137,13 @@ func (f *Factory) TunedDaemonSet() (*appsv1.DaemonSet, error) {
 	imageTuned := ntoconfig.NodeTunedImage()
 	ds.Spec.Template.Spec.Containers[0].Image = imageTuned
 
+	for i, e := range ds.Spec.Template.Spec.Containers[0].Env {
```
sjug: Not necessarily an issue again, but highlights the inconsistency in variable names between `ds` here and `daemonset` above.

jmencak:
This needs more time, there might be a potential race.

/hold

pkg/controller/tuned/tuned_controller.go
```diff
@@ -99,6 +100,92 @@ type ReconcileTuned struct {
...
+	saSecrets := map[string]bool{}
+	saManifest, err := r.manifestFactory.TunedServiceAccount()
+	if err != nil {
+		return saSecrets, fmt.Errorf("Couldn't build tuned ServiceAccount: %v", err)
+	}
+
+	sa := &corev1.ServiceAccount{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: saManifest.Namespace, Name: saManifest.Name}, sa)
+	if err != nil {
+		return saSecrets, fmt.Errorf("Failed to get ServiceAccount: %v\n", err)
```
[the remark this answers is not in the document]
jmencak: golint is not being used by the CI to check "style mistakes".  It probably should, but this is likely to come now at a price of consistency.  I've removed all the offending '\n"$' from `tuned_controller.go` to keep golint happy on this particular file.  Going forward, we should probably enforce golint on the entire project even at the expense of making fixes to old branches more troublesome.

jmencak:
Thanks for all the reviews, @sejug!  I believe all comments are addressed now.  One extra thing I've changed is being more defensive and failing when secrets associated with the service account cannot be retrieved.

eparis:
https://bugzilla.redhat.com/1718842 is a 4.1.z BZ but this PR was against master. Thus removing the incorrect linkage from the title.

PR #66, opened 2019-06-11
use GA kubelet OS label

rphillips:
ref: https://jira.coreos.com/browse/POD-114


jmencak:
Thanks Ryan! Looks good, but this shouldn't be merged before 1.15 rebase, right?

rphillips:
@jmencak This is fine to merge now, since 0.14.x also includes the change.

jmencak:
gathered that might be the case looking at k8s PR73048, thanks for the confirmation
/lgtm

PR #67, opened 2019-06-24
Bug 1719967: Implement leader-for-life election.

jmencak:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1719967

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -405,7 +405,7 @@
   revision = "a47c6814c02c8516ee468a2ce68fbb798755a76e"
 
 [[projects]]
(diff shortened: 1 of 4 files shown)


jmencak:
@zvonkok , mind taking a look in Sebastian's absense? Thanks!

zvonkok:
The same approach as here: https://github.com/zvonkok/special-resource-operator/blob/a065b27c1aa9219b3960a8ea47069754a5533920/cmd/manager/main.go#L56 
and this works if multiple instances of the operator are running, one is leader and the other ones are "only"  waiting to become leader. 

/lgtm

PR #69, opened 2019-07-05
Fix custom profiles ordering based on their priorities.

jmencak:
In cases any of the profile priorities were undefined, it was possible to have
incorrect ordering of tuned profiles in tuned-recommend ConfigMap.  This PR
fixes this issue.


jmencak:
@sjug PTAL, thank you!

PR #70, opened 2019-07-11
Basic CR validation using openAPIV3Schema

jmencak:
This is necessary to enable basic custom CR formatting checks.
Outstanding issues:
  - unknown fields are not dropped from CRs
  - only one level of "match:" sections is validated due to recursive
    nature of the tuned CRD spec


jmencak:
@sjug  PTAL

PR #71, opened 2019-07-16
Replace default tuned CR on operator restart.

jmencak:
This commit:
  - replaces default tuned CR on operator restart (e.g. upgrade)
  - makes sure bindata is regenerated out of modified assets
    in case an old version of bindata is already present
  - removes extraneous "for" from logging messages

In the presence of custom CRs not to replace the default CR shipped with the
operator (especially on operator upgrade) is a bug and is addressed by this
commit.  Users should create their own custom CRs instead of modifying the
default one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ PACKAGE_MAIN=$(PACKAGE)/cmd/manager
 
 # Build-specific variables
 GOBINDATA_BIN=./go-bindata
+ASSETS=$(shell find assets -name \*.yaml)
 BINDATA=pkg/manifests/bindata.go
 ENVVAR=GOOS=linux CGO_ENABLED=0
 GOOS=linux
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -22,14 +23,14 @@ build: $(BINDATA)
 	$(GO_BUILD_RECIPE)
 
 # Using "-modtime 1" to make generate target deterministic. It sets all file time stamps to unix timestamp 1
-generate $(BINDATA): $(GOBINDATA_BIN)
+$(BINDATA): $(GOBINDATA_BIN) $(ASSETS)
```
sjug: So the name of this build target is `pkg/manifests/bindata.go`?

jmencak: Yes.  This will trigger `pkg/manifests/bindata.go` to be re-generated when assets change.

sjug: ack, just an odd name is all

PR #73, opened 2019-07-26
Switch to go modules and golang 1.12.

jmencak:
Changes:
- use golang 1.12 instead of 1.10 as a base
- switch to official go modules vs. external tooling (dep) for dependency management
- dropping dependency on operator-sdk
- update k8s.io/{api,apimachinery,client-go} to 1.14 levels
- use controller-gen for API type deepcopy generation
- make it possible to use a cached and non-cached API client
- a complete re-write of e2e tests
- Makefile cleanup


jmencak:
@sjug PTAL

pkg/controller/tuned/status.go
```diff
@@ -67,51 +66,55 @@ func (r *ReconcileTuned) syncOperatorStatus() (bool, error) {
...
-	if err != nil {
+	co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: ntoconfig.OperatorName()}}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: co.Name}, co); err != nil {
 		if errors.IsNotFound(err) {
 			// Cluster operator not found, create it
-			co_created, err := r.cfgv1client.ClusterOperators().Create(co)
-			if err != nil {
-				// Failed to create the cluster operator object
-				return nil, err
+			initializeClusterOperator(co)
```
sjug: Was this done before elsewhere?

pkg/controller/tuned/tuned_controller.go
```diff
@@ -110,15 +112,15 @@ func (r *ReconcileTuned) getTunedSecrets(tuned *tunedv1.Tuned) (map[string]bool,
 	}
 
 	sa := &corev1.ServiceAccount{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: saManifest.Namespace, Name: saManifest.Name}, sa)
+	err = r.cache.Get(context.TODO(), types.NamespacedName{Namespace: saManifest.Namespace, Name: saManifest.Name}, sa)
```
sjug: What happens on a cache miss on `Get()`?

pkg/controller/tuned/tuned_controller.go
```diff
@@ -362,7 +361,8 @@ func (r *ReconcileTuned) syncDaemonSet(tuned *tunedv1.Tuned) error {
 }
 
 func replaceDefaultCustomResource(mgr manager.Manager) error {
-	client := mgr.GetClient()
+	glog.V(1).Infof("ReplaceDefaultCustomResource()")
+	client := ntoclient.GetClient()
```
sjug: I like you that you kept the same general API

test/e2e/operator_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	ntoclient "github.com/openshift/cluster-node-tuning-operator/pkg/client"
+	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func TestOperatorAvailable(t *testing.T) {
+	cl := ntoclient.GetClient()
```
sjug: I like it better as you have it above `client := ntoclient.GetClient()`, no need to save a few characters.

jmencak: Changed as requested.

pkg/controller/tuned/tuned_controller.go
jmencak: You get stale information in exactly the same way you previously received with the split client and reconciliation loop will handle this once the cache is in sync.

pkg/controller/tuned/status.go
jmencak: Good point.  No, this was not done before.  I don't think this is strictly necessary and I have mixed feelings about this myself, but it does initialize the ClusterOperator object rather than keeping it empty.

jmencak:
Thanks for the review, @sjug! I believe I've addressed the concerns.

PR #74, opened 2019-07-29
Documenting recent operator functionality.

jmencak:
Documentation for:
  - profile settings rollback on profile changes
  - the default operator's CR being overwritten
  - a complete example of using custom profiles

Minor fixes and improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Cluster Node Tuning Operator
+# Node Tuning Operator
 
-The Cluster Node Tuning Operator manages cluster node-level tuning for
+The Node Tuning Operator manages cluster node-level tuning for
 [OpenShift](https://openshift.io/).
 
 The majority of high-performance applications require some
@@ -16,8 +16,12 @@ specification](#custom-tuning-specification) is passed to all
 daemons running in the cluster in the format that the daemons understand.
 The daemons run on all nodes in the cluster, one per node.
 
+Node-level settings applied by the containerized tuned daemon are rolled back
+on an event that triggers a profile change or when the containerized tuned
+daemon is terminated gracefully by receiving and handling a termination signal.
 
-## Deploying the node tuning operator
+
+## Deploying the Node Tuning Operator
 
 The operator is deployed by applying the `*.yaml` manifests in the operator's
 `/manifests` directory in alphanumeric order.  It automatically creates a default deployment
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,8 +16,13 @@ specification](#custom-tuning-specification) is passed to all
 daemons running in the cluster in the format that the daemons understand.
 The daemons run on all nodes in the cluster, one per node.
 
+Node-level settings applied by the containerized tuned are rolled back on
+an event that triggers a profile change or when the containerized tuned
+daemon is terminated gracefully by receiving and handling a termination
+signal.
```
ashishkamra: terminated gracefully by "sending a termination signal"

README.md
```diff
@@ -16,8 +16,13 @@ specification](#custom-tuning-specification) is passed to all
 daemons running in the cluster in the format that the daemons understand.
 The daemons run on all nodes in the cluster, one per node.
 
+Node-level settings applied by the containerized tuned are rolled back on
+an event that triggers a profile change or when the containerized tuned
```
ashishkamra: I guess I am little confused on what the settings are rolled back to ?

README.md
```diff
@@ -67,7 +79,7 @@ The `profile:` section lists tuned profiles and their names.
 
       [sysctl]
```
ashishkamra: Is the [sysctl] part of the profile data optional? If the user wants to create a custom CR based on an existing tuned profile such as sap-hana, can they only specify the "include=sap-hana" directive under [main] and skip specifying the [sysctl] part?

README.md
jmencak: Sending?  The containerized tuned daemon itself doesn't send any termination signals.  It *receives* and *handles* termination signals.  openshift-tuned though does send a `TERM` signal to tuned when it receives a `TERM` signal itself.

README.md
jmencak: All the settings changed by a profile (e.g. sysctls) are rolled back upon application of a new profile.  Say profile A sets `net.ipv4.ip_local_port_range="1024 65535"`.  Before profile A is set though, tuned daemon looks at `net.ipv4.ip_local_port_range` and sees the default value `32768 60999`.  It remembers that value and upon a change to profile B `net.ipv4.ip_local_port_range` is unapplied/rolled-back to `32768 60999`.

README.md
jmencak: Yes, that's is correct.  The `[sysctl]` plugin is optional.

jmencak:
@sjug, PTAL, so we can get this merged.

PR #75, opened 2019-08-09
Add initial support for bootloader

### LINKED ISSUE #72: Support for bootloader and script plugin
When optimizing RHEL-based worker nodes for specialized workloads (i.e. to use the real-time kernel) I need to be able to update the cmdline_realtime parameters to reserver cores, memory pages, etc.

For example, parameters like this:

```
cmdline_realtime="+isolcpus=\${isolated_cores} intel_pstate=disable nosoftlockup nmi_watchdog=0 audit=0 mce=off kthread_cpus=0 irqaffinity=0 skew_tick=1 processor.max_cstate=1 idle=poll intel_idle.max_cstate=0 intel_pstate=disable intel_iommu=off default_hugepagesz=\${hugepage_size_default} hugepagesz=\${hugepage_size} hugepages=\${hugepage_num} nohz=on nohz_full=\${isolated_cores} rcu_nocbs=\${isolated_cores}"
```

When comparing with `tuned-profile-realtime`, it also uses the `script` plugin to invoke `tuna` so the `script` plugin may be necesary as well.

yrobla:
Add support for bootloader in tuned. It needs mounting some extra directories for bootloader to be executed. It is not using the functionality yet, as this will be provided on another comit, but it can be tested using a profile that includes [bootloader] section.

Partially resolves #72

### DIFF
--- a/assets/tuned/06-ds-tuned.yaml
+++ b/assets/tuned/06-ds-tuned.yaml
@@ -48,6 +48,16 @@ spec:
         - mountPath: /lib/modules
           name: lib-modules
           readOnly: true
+        - mountPath: /etc/grub.d
+          name: etc-grub
+        - mountPath: /etc/default/grub
+          name: etc-default-grub
+        - mountPath: /boot/grub2
+          name: boot-grub2
+        - mountPath: /etc/grub2.cfg
+          name: boot-grub2-cfg
+        - mountPath: /etc/grub2-efi.cfg
+          name: boot-grub2-cfg
         env:
           - name: OCP_NODE_NAME
             valueFrom:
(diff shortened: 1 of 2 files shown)


assets/tuned/06-ds-tuned.yaml
```diff
@@ -17,7 +17,11 @@ spec:
       serviceAccountName: tuned
       automountServiceAccountToken: true
       containers:
-      - command: ["/var/lib/tuned/bin/run","start"]
+      - command: ["/bin/sh", "-c"]
+        args:
+          - ln -s /boot/grub2/grub.cfg /etc/grub2.cfg;
+            ln -s /boot/grub2/grub.cfg /etc/grub2-efi.cfg;
```
sjug: Are both `grub2.cfg` and `grub2-efi.cfg` needed?

yrobla: I needed to create the symlinks because tuned reads those grub2.cfg and grub2-efi.cfg files instead of reading them from /boot/grub2, that is the original source. So yes, they are needed. In my case i tested with grub2.cfg (bios boot), but i guess grub2-efi.cfg is used for UEFI.

yrobla:
This is not going to be used, as support for kernel args will be in MachineConfig

PR #76, opened 2019-08-12
Realtime profile

yrobla:
Add realtime profile definition, and support for the configmap definition for settings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ REV=$(shell git describe --long --tags --match='v*' --always --dirty)
 
 # Container image-related variables
 DOCKERFILE=Dockerfile
-IMAGE_TAG=openshift/origin-cluster-node-tuning-operator
+IMAGE_TAG=yrobla/origin-cluster-node-tuning-operator
 IMAGE_REGISTRY=quay.io
 
 all: build
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ REV=$(shell git describe --long --tags --match='v*' --always --dirty)
 
 # Container image-related variables
 DOCKERFILE=Dockerfile
-IMAGE_TAG=yrobla/origin-cluster-node-tuning-operator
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -192,6 +193,35 @@ with the exception of dynamic tuning functionality provided by some of the plugi
 
 The following Tuned plugins are currently not supported:
 
-* bootloader
```
jmencak: Why is this being removed in a PR that doesn't implement the `bootloader` plugin?

yanirq: @jmencak  It looks correct since this is removed from the "**not** supported" list.

jmencak: This PR is really a mixture of two PRs and should be split. A) Support for the `bootloader` plugin, B) possibly add an openshift realtime profile as necessary.  I haven't verified this PR actually includes full `bootloader` plugin support.  If this is true, then yes, I agree with this change, but my point to split this PR into two still stands.

yrobla: Ok i left it on hold until the situation is clarified. Tomorrow i will amend this patch to just create a realtime profile without the bootloader part.

jmencak: Right.  The default tuned CR that you're modifying with this PR tries to deliver default out-of-the box tuning for OpenShift control plane and nodes.  The fact it ships 4 profiles (2x tuning for ES on control plane and nodes) is probably already a mistake/bug/issue and should be addressed.  The default tuning should addresses day-1 operations and anything outside day-1 operations (profiles for specific applications, configuring hugepages, sap, nfv, oracle, realtime profiles, ...) should really be set by cluster admins as a day-2 operation.  Adding a realtime profile like that would also only add duplication to what `tuned-profiles-realtime` package ships.  The path forward in my opinion is to add additional `tuned-profiles-*` packages as the operator support for all the tuned plugins improves.

jmencak: /cc @zvonkok

jmencak:
Obsoleted by https://github.com/openshift/cluster-node-tuning-operator/pull/76

PR #77, opened 2019-08-20
Report the "Reason" condition for the ClusterOperator status.

jmencak:
This commit:
- adds the "Reason" condition for the ClusterOperator status
- sets the Degraded condition when the operator is Unavailable
  for an extensive period of time

See: RH BZ1741067 and BZ1717617


jmencak:
@sjug PTAL

PR #79, opened 2019-09-22
[4.3] bump pid limits

rphillips:
@RobertKrawitz was testing some large workloads, and we found pid limits were blocking some pods running. This PR bumps the pid limits to 524288.

### DIFF
--- a/assets/tuned/default-cr-tuned.yaml
+++ b/assets/tuned/default-cr-tuned.yaml
@@ -19,7 +19,7 @@ spec:
 
       [sysctl]
       net.ipv4.ip_forward=1
-      kernel.pid_max=>131072
+      kernel.pid_max=>4194304
       net.netfilter.nf_conntrack_max=1048576
       net.ipv4.neigh.default.gc_thresh1=8192
       net.ipv4.neigh.default.gc_thresh2=32768


assets/tuned/default-cr-tuned.yaml
```diff
@@ -19,7 +19,7 @@ spec:
 
       [sysctl]
       net.ipv4.ip_forward=1
-      kernel.pid_max=>131072
+      kernel.pid_max=>524288
```
jmencak: Thanks for the PR!  Looking at kernel code, `PID_MAX_LIMIT` seems to be the limit, which is likely to be 4*1024*1024 on most today's configs.  From what I read, it is "just" the memory footprint that is effected by this setting.  What's the reasoning that `524288` is a "sane" number?  Can/should we go higher/lower?  /cc @jeremyeder , @zvonkok

RobertKrawitz: That's 4194304 (in the text version of the notification it looks like the asterisks between the 4, 1024, and 1024 got interpreted as formatting).

We could certainly go higher, albeit increasing the memory consumption on smaller nodes (whether that's significant needs testing).

jmencak: Yep, fixed the original comment, thanks! I agree this needs to be tested first.

RobertKrawitz: Based on that, yes, let's go to 4M.

rphillips: works for me... updated.

PR #81, opened 2019-10-07
Use net.ipv4.conf.all.arp_announce=2 in the openshift profile.

jmencak:
Adapting a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1758552 from OCP 3.11.


jmencak:
Also see https://github.com/openshift/openshift-ansible/pull/11935.  /cc @fbac ,  @dcbw

PR #83, opened 2019-10-14
Tolerate only master/etcd taints.

jmencak:
NTO's tuned operand tolerates any taints.  This does not give much room for cases where NTO's operator is not welcome on certain nodes that need a very specific tuning not handled by the operator/operand.  Match MCO's MCD taints and tolerate only: `node-role.kubernetes.io/master` and `node-role.kubernetes.io/etcd`.

### DIFF
--- a/assets/tuned/06-ds-tuned.yaml
+++ b/assets/tuned/06-ds-tuned.yaml
@@ -96,5 +96,9 @@ spec:
       securityContext: {}
       terminationGracePeriodSeconds: 30
       tolerations:
-      # tolerate all taints so that tuned is always present on all nodes
-      - operator: Exists
+      - key: node-role.kubernetes.io/master
+        operator: Exists
+        effect: NoSchedule
+      - key: node-role.kubernetes.io/etcd
+        operator: Exists
+        effect: NoSchedule


sjug:
The effect of this change will be that the NTO will only deploy on master nodes that have etcd?

jmencak:
> The effect of this change will be that the NTO will only deploy on master nodes that have etcd?

No.   The effect of this PR is that the NTO's operand (openshift-tuned) will deploy on non-tainted nodes (e.g. workers unless admins add taints) and on nodes tainted with `node-role.kubernetes.io/master` or/and `node-role.kubernetes.io/etcd`.  So if any node has one (or both) of these taints (typically infrastructure-type nodes), openshift-tuned will run on them.

This is a relatively big change, even though it is a small PR, so I understand the caution.  I follow what MCO already does here.   The advantage of this PR is also it's disadvantage.  Users can taint nodes to prevent openshift-tuned running on certain nodes, but they may also do this unintentionally/unknowingly.

/cc @zvonkok

jmencak:
@zvonkok any feedback? Thanks!

zvonkok:
Since this is an asset the operator manages, there will be no way to customize the deployment of the DS to run on tainted nodes? 

If a cluster-admin creates a pool of nodes with a taint. What would be the way to allow the operand to run on these nodes as well?

jmencak:
Thanks for all the comments!  The functionality will be provided by using a CR in the future.

PR #84, opened 2019-10-21
Add more e2e tests.

jmencak:
The current Node Tuning Operator's e2e tests include only tests for:

  - availability of the ClusterOperator node-tuning object
  - existence of the default tuned CR

The e2e tests are enhanced to check basic functionality of the operator's operands, i.e. the functionality of requested tuning on OpenShift nodes.

The following three tests are added:
  - the basic functionality of NTO and its operands (check for an application of a default sysctl)
  - the application (and rollback) of a custom profile via pod labelling
  - the application (and rollback) of a custom profile via node labelling

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -4,6 +4,9 @@ package e2e
 
 import (
 	"context"
+	"fmt"
+	"os/exec"
+	"strings"
 	"testing"
 	"time"
 
@@ -13,10 +16,14 @@ import (
 	ntoclient "github.com/openshift/cluster-node-tuning-operator/pkg/client"
 	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// Test the ClusterOperator node-tuning object exists and is Available.
 func TestOperatorAvailable(t *testing.T) {
 	client := ntoclient.GetClient()
 	if client == nil {
@@ -38,10 +45,11 @@ func TestOperatorAvailable(t *testing.T) {
 		return false, nil
 	})
 	if err != nil {
-		t.Errorf("Did not get expected available condition: %v", err)
+		t.Fatalf("Did not get expected available condition: %v", err)
 	}
 }
 
+// Test the default tuned CR exists.
 func TestDefaultTunedExists(t *testing.T) {
 	client := ntoclient.GetClient()
 	if client == nil {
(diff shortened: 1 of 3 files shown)


zvonkok:
There are safe and unsafe sysctl for Kubernetes, which one are we supporting in OpenShift? 
Safe sysctls are namespaced, are we going to support the non-namespaced unsafe one? 

Node wide sysctl effect all pods, so its enough to test the setting in one of the containers, right?

jmencak:
> There are safe and unsafe sysctl for Kubernetes, which one are we supporting in OpenShift?

This PR and the Node Tuning Operator (NTO) has nothing to do with (un)safe sysctls.  NTO sets node-level sysctls by using privileged containers and sharing host namespaces.  To answer the question though, unsafe sysctls should work in OpenShift [as of 4.2](https://bugzilla.redhat.com/show_bug.cgi?id=1690754).  Safe sysctl are enabled by default but afaik there's only 3 of them.

> Safe sysctls are namespaced, are we going to support the non-namespaced unsafe one?

Safe sysctls are namespaced, but it should be noted that not all namespaced sysctls are considered safe.  OpenShift [should already support unsafe sysctls for pods](https://bugzilla.redhat.com/show_bug.cgi?id=1690754), but again, this has nothing to do with the NTO and this PR.

> Node wide sysctl effect all pods, so its enough to test the setting in one of the containers, right?

Testing from a container that is sharing host namespaces should be enough to tell that a node-level sysctl was set or not.  If it is not, then I'm missing something and these tests need to be done differently.

test/e2e/operator_test.go
```diff
@@ -56,6 +64,232 @@ func TestDefaultTunedExists(t *testing.T) {
...
+	}
+
+	t.Logf("Labelling pod %s with label %s", pod.Name, podLabelIngress)
+	out, err := exec.Command("oc", "label", "pod", "--overwrite", "-n", ntoconfig.OperatorNamespace(), pod.Name, podLabelIngress+"=").CombinedOutput()
+	if err != nil {
+		t.Fatal(fmt.Errorf("%v", string(out)))
+	}
+
+	t.Logf("Applying the custom ingress profile from %s", profileIngress)
+	out, err = exec.Command("oc", "apply", "-n", ntoconfig.OperatorNamespace(), "-f", profileIngress).CombinedOutput()
```
sjug: Is this step creating the custom profile so that NTO can pick it up and apply it based on the previous label you set? Or how is this being applied?

test/e2e/operator_test.go
```diff
@@ -56,6 +64,232 @@ func TestDefaultTunedExists(t *testing.T) {
...
+	}
+
+	return val, nil
+}
+
+// Makes sure a sysctl value for sysctlVar from inside a (tuned) pod is equal to valExp.
+// Returns an error otherwise.
+func ensureSysctl(sysctlVar string, pod *corev1.Pod, valExp string) (err error) {
+	var val string
+	wait.PollImmediate(5*time.Second, 5*time.Minute, func() (bool, error) {
```
sjug: Not sure if this `PollImmediate()` is necessary since `getSysctl()` always has its own?

test/e2e/operator_test.go
jmencak: That's exactly the correct understadning.

test/e2e/operator_test.go
jmencak: It is not necessary.  I was considering not using it, but then I went for the same pattern and more flexibility (which is probably also questionable).  Would you prefer I removed it in ensureSysctl()?

PR #90, opened 2019-10-30
Do not add ownerReferences for custom CRs.

jmencak:
This PR sets owner references for the default tuned CR only.

Additional owner references in the tuned daemonset were triggering garbage collector to delete the tuned daemonset unnecessarily on deletion of custom tuned CRs.


sjug:
I see the `controllerutil` package has the logic change, what's the rationale behind moving the `setControllerReference()` function call before the `r.cache.Get()`?

jmencak:
> I see the `controllerutil` package has the logic change, what's the rationale behind moving the `setControllerReference()` function call before the `r.cache.Get()`?

The location should not matter.  I believe the code looks more readable this way.  The reason was the `err` variable, but given it has its own scope in the `if` statement, the location should not matter.

sjug:
Looks good otherwise, makes sense that the admin created resources would persist.

PR #94, opened 2019-11-06
Use "preserveUnknownFields: false" for the CRD.

jmencak:
preserveUnknownFields: false is now required on a global level.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1769468


jmencak:
@sjug PTAL

sjug:
/lgtm 

Given we're using 1.16 this should work.

Issue #89, opened 2019-10-29
Is this operator capable of setting sebools?


tylerauerbeck:
I have a workload that requires enabling `container_manage_cgroup` which traditionally I would do by logging onto the system and running `setsebool container_manage_cgroup 1`. I was hoping to be able to do this with this operator as well as I see it's capable of interacting with selinux. However  my attempts so far have been unsuccessful. I've tried adding the following to the `tuned-profiles` configmap:

```
    openshift-node: |
      [main]
      summary=Optimize systems running OpenShift nodes
      include=openshift

      [sysctl]
      net.ipv4.tcp_fastopen=3
      fs.inotify.max_user_watches=65536

      [selinux]
      container_manage_cgroup=1
```

I've also tried setting `container_manage_cgroup` to `true` and `yes` just to try all the options, but still didn't have any luck. Is this possible through this operator? Is there a seperate/better way of handling this? Or is there anything that I need to do after modifying so that this gets put in place?

jmencak:
Thank you for the report and sorry for the delay in response.  This is a valid issue.  Currently the `[selinux]` plugin is not supported.

jmencak:
So I was confused for a moment.  Looking more into the sources of the [tuned daemon's [selinux] plugin](https://github.com/redhat-performance/tuned/blob/master/tuned/plugins/plugin_selinux.py), it has nothing to do with actually setting SELinux booleans.  It only tunes the values in `/sys/fs/selinux/`, `avc_cache_threshold` in particular.  So to answer your question, this operator cannot set sebools.

tylerauerbeck:
Thanks @jmencak . For others interested in this, it looks like the plan is to handle it here: https://github.com/openshift/machine-config-operator/issues/852 and looks to be blocked by https://github.com/ostreedev/ostree/issues/1026

PR #95, opened 2019-11-07
Document the fact SELinux plugin is actually not supported.

jmencak:
See: https://github.com/openshift/cluster-node-tuning-operator/issues/89

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,6 @@ in the `profile:` section of the Tuned CR:
 * net
 * scheduler
 * scsi_host
-* selinux
 * sysctl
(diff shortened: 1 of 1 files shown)


jmencak:
/hold
So this is tricky.  The plugin is actually "partly" supported and the settings in the default CR that use the `[selinux]` tuned plugin work as this is basically a write to a file in `/sys/fs/selinux`.  It needs to be investigated what needs to be done to enable the full `[selinux]` support.

jmencak:
Sorry for the noise, works as designed, please see the original issue.

PR #102, opened 2020-01-10
Add preliminary support for the cpu-partitioning profile.

jmencak:
Changes:
  - allow patches to modify tuned profiles too (see rhbz#1781664)
  - mount host directories to allow:
    * persistent changes to /etc/sysconfig/irqbalance
    * tuned to populate /etc/modprobe.d directory with module
      options (see kvm.rt.tuned.conf)
    * persistent changes to host's /etc/systemd/system.conf
      by tuned setting CPUAffinity
  - create /etc/grub.d/, /boot to make tuned's [bootloader] plugin
    pass and generate /etc/tuned/bootcmdline
  - include tuned patch for rhbz#1781664 (variable expansion)
  - include tuned patch to enable multiple profile inheritance
    (tuned 2.13 feature)
  - include an example custom openshift-cpu-partitioning profile

Note that the [bootloader] plugin does not do any host modifications,
however, the variables in /etc/tuned/bootcmdline should be set
correctly.

### DIFF
--- a/examples/cpu-partitioning.yaml
+++ b/examples/cpu-partitioning.yaml
@@ -1,0 +1,22 @@
+apiVersion: tuned.openshift.io/v1
+kind: Tuned
+metadata:
+  name: openshift-cpu-partitioning
+  namespace: openshift-cluster-node-tuning-operator
+spec:
+  profile:
+  - data: |
+      [main]
+      summary=Configuration for cpu-partitioning
+      include=openshift-node,cpu-partitioning
+      [variables]
+      # {isolated,no_balance}_cores take a list of ranges; e.g. isolated_cores=2,4-7
+      isolated_cores=1
+      no_balance_cores=1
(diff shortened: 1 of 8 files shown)


examples/cpu-partitioning.yaml
```diff
@@ -0,0 +1,21 @@
...
+  name: openshift-cpu-partitioning
+  namespace: openshift-cluster-node-tuning-operator
+spec:
+  profile:
+  - data: |
+      [main]
+      summary=Configuration for cpu-partitioning
+      include=openshift-node,cpu-partitioning
+      [variables]
+      isolated_cores=1-3
```
sjug: Created problems on my dual core worker node, perhaps we should use a smaller range for this example?

jmencak: Thanks for the report!

sjug:
/lgtm

Need more lgtm

PR #108, opened 2020-01-28
Bug 1795665: Do not set owner references on tuned ServiceAccount

sjug:
Fixes bug 1795665 which causes race conditions on upgrade from 4.x -> 4.3.

This PR removes owner reference from the ServiceAccount.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -3,7 +3,7 @@ WORKDIR /go/src/github.com/openshift/cluster-node-tuning-operator
 COPY . .
 RUN make build
 
-FROM registry.svc.ci.openshift.org/ocp/4.0:base
+FROM registry.svc.ci.openshift.org/ocp/4.3:base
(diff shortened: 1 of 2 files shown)


bparees:
@sjug this PR refs a bug which depends on a 4.4 bz (https://bugzilla.redhat.com/show_bug.cgi?id=1791988)

but the 4.4 BZ only links this PR which doesn't match your PR: https://github.com/openshift/cluster-node-tuning-operator/pull/103

Can you point me to where these changes were merge in 4.4 so i can confirm this is the right fix to land in 4.3.z to fix the same issue?

sjug:
@bparees Jiri has done a major rewrite in 4.4 so this race condition isn't present in 4.4+ because the service account is now managed by CVO.

bparees:
@sjug got it, thanks

bparees:
this will probably not make the cut this week, but next patch-manager this should be good to merge.

PR #110, opened 2020-01-29
if end with a return so the else is implied

sjug:
Cleaning up a bit, thoughts?


sjug:
Minor PR, will wait for more changes to push.

PR #111, opened 2020-01-29
Bug 1796633: Adding preliminary realtime profile support.

jmencak:
Changes;
  - add a preliminary support for the realtime profile
  - cleanup custom profiles in `examples/`
  - try-restart irqbalance systemd unit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN INSTALL_PKGS=" \
     yum install --setopt=tsflags=nodocs -y $INSTALL_PKGS $ARCH_DEP_PKGS && \
     rpm -V $INSTALL_PKGS $ARCH_DEP_PKGS && \
     (LC_COLLATE=C cat patches/*.diff | patch -Np1 -d / || :) && \
+    chmod 755 /usr/lib/tuned/*/script.sh && \
     touch /etc/sysctl.conf && \
     yum -y remove patch && \
     yum clean all && \
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -20,6 +20,7 @@ RUN INSTALL_PKGS=" \
     yum install --setopt=tsflags=nodocs -y $INSTALL_PKGS $ARCH_DEP_PKGS && \
     rpm -V $INSTALL_PKGS $ARCH_DEP_PKGS && \
     (LC_COLLATE=C cat patches/*.diff | patch -Np1 -d / || :) && \
+    chmod 755 /usr/lib/tuned/*/script.sh && \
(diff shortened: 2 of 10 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@ WORKDIR ${APP_ROOT}
 COPY --from=builder /go/src/github.com/openshift/cluster-node-tuning-operator/_output/openshift-tuned /usr/bin/
 COPY --from=builder /go/src/github.com/openshift/cluster-node-tuning-operator/assets ${APP_ROOT}
 RUN INSTALL_PKGS=" \
-      tuned tuna tuned-profiles-cpu-partitioning patch socat \
+      tuned tuna tuned-profiles-cpu-partitioning tuned-profiles-realtime patch socat \
```
ArangoGutierrez: ERROR while building Centos-7 image
```bash
package tuned-profiles-realtime is not installed
error building at STEP "RUN INSTALL_PKGS="       tuned tuna tuned-profiles-cpu-partitioning tuned-profiles-realtime patch socat       " &&     ARCH_DEP_PKGS=$(if [ "$(uname -m)" != "s390x" ]; then echo -n hdparm kernel-tools ; fi) &&     mkdir -p /etc/grub.d/ /boot &&     yum install --setopt=tsflags=nodocs -y $INSTALL_PKGS $ARCH_DEP_PKGS &&     rpm -V $INSTALL_PKGS $ARCH_DEP_PKGS &&     (LC_COLLATE=C cat patches/*.diff | patch -Np1 -d / || :) &&     touch /etc/sysctl.conf &&     yum -y remove patch &&     yum clean all &&     rm -rf /var/cache/yum ~/patches &&     useradd -r -u 499 cluster-node-tuning-operator": error while running runtime: exit status 1
ERRO exit status 1                                
make: *** [Makefile:98: local-image] Error 1
```
Running `make local-image`

ArangoGutierrez: Discussed with @jmencak , is ok now

ArangoGutierrez: so if the centos image is not working, should we rm it from the rpo?

jmencak: So far we've been supporting upstream via that.  No, I believe we need to continue doing that.  I've resolved this by backporting the patches for realtime profiles from upstream as patches for both CentOS and RHEL.

jmencak:
It looks like for `irqbalance` to take effect, SIGHUP needs to be sent to the `irqbalance` process.
/hold

jmencak:
/hold cancel
Tested that `irqbalance` service is being restarted and settings take effect for both `realtime` and `cpu-partitioning` profiles.

PR #112, opened 2020-01-30
Bug 1801860: Limit event processing to maxRetries.

jmencak:


pkg/operator/controller.go
```diff
@@ -110,14 +113,20 @@ func (c *Controller) eventProcessor() {
...
-				// Re-enqueue the workqueueKey.  Based on the rate limiter on the queue
-				// and the re-enqueue history, the workqueueKey will be processed later again.
-				c.workqueue.AddRateLimited(workqueueKey)
-				klog.Errorf("unable to sync(%s/%s/%s) requeuing: %v", workqueueKey.kind, workqueueKey.namespace, workqueueKey.name, err)
+				// Limit retries to maxRetries.  After that, stop trying.
+				if c.workqueue.NumRequeues(workqueueKey) < maxRetries {
+					// Re-enqueue the workqueueKey.  Based on the rate limiter on the queue
+					// and the re-enqueue history, the workqueueKey will be processed later again.
+					klog.Errorf("unable to sync(%s/%s/%s) requeuing: %v", workqueueKey.kind, workqueueKey.namespace, workqueueKey.name, err)
+					c.workqueue.AddRateLimited(workqueueKey)
```
ArangoGutierrez: just a nip tic, move this line above the logger, due 118 and 199 lines are comments about this func and not about logging

pkg/operator/controller.go
```diff
@@ -110,14 +113,20 @@ func (c *Controller) eventProcessor() {
...
+					return
+				}
+				klog.Errorf("unable to sync(%s/%s/%s) reached max retries(%d): %v", workqueueKey.kind, workqueueKey.namespace, workqueueKey.name, maxRetries, err)
 			} else {
-				c.workqueue.Forget(obj)
```
ArangoGutierrez: why are we not notifying that the item has finished retrying?

pkg/operator/controller.go
jmencak: Nip tic?  Do you mean nitpick?  The comment is about the entire block of code and if I swapped the lines I'd need to change the "requeuing" into "requed".  If it makes sense to you more and you think this will be more consistent with the entire code within the operator, I can certainly do that.

ArangoGutierrez: sorry, I trusted the spell checker, but yes, nitpick.... [hides under a rock]

ArangoGutierrez:
until 113 passes I can not label here `lgtm`

PR #114, opened 2020-01-31
Remove Docker as image build engine

ArangoGutierrez:
Remove docker as as image build engine, replace for podman
also perperly tag images with the short git commit hash

Signed-off-by: Carlos Eduardo Arango Gutierrez <carangog@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,13 +25,13 @@ API_TYPES_GENERATED:=$(API_TYPES_DIR)/$(API_ZZ_GENERATED).go
 API_GO_HEADER_FILE:=pkg/apis/header.go.txt
 
 # Container image-related variables
-IMAGE_BUILD_CMD:= podman build
-IMAGE_PUSH_CMD:= podman push
+IMAGE_BUILD_CMD=podman build
+IMAGE_PUSH_CMD=podman push
 DOCKERFILE=Dockerfile
 REGISTRY=quay.io
 ORG=openshift
 TAG=$(shell git rev-parse --abbrev-ref HEAD)
-IMAGE=$(REGISTRY)/$(ORG)/origin-cluser-node-tuning-operator:$(TAG)
+IMAGE=$(REGISTRY)/$(ORG)/origin-cluster-node-tuning-operator:$(TAG)
 
 all: build
 
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,6 @@ local-image:
 	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_EXTRA_OPTS) -t $(IMAGE) -f $(DOCKERFILE) .
 
 local-image-push:
-	$(IMAGE_PUSH_CMD) push $(IMAGE)
+	$(IMAGE_PUSH_CMD) $(IMAGE_PUSH_EXTRA_OPTS) $(IMAGE)
 
 .PHONY: all build deepcopy crd-schema-gen test-e2e verify verify-gofmt clean local-image local-image-push
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,13 @@ API_TYPES_GENERATED:=$(API_TYPES_DIR)/$(API_ZZ_GENERATED).go
 API_GO_HEADER_FILE:=pkg/apis/header.go.txt
 
 # Container image-related variables
+IMAGE_BUILD_CMD:= podman build
+IMAGE_PUSH_CMD:= podman push
 DOCKERFILE=Dockerfile
-IMAGE_TAG=openshift/origin-cluster-node-tuning-operator
-IMAGE_REGISTRY=quay.io
+REGISTRY=quay.io
+ORG=openshift
+TAG=$(shell git rev-parse --abbrev-ref HEAD)
+IMAGE=$(REGISTRY)/$(ORG)/origin-cluser-node-tuning-operator:$(TAG)
 
 all: build
 
@@ -110,18 +114,9 @@ clean:
 	rm -rf $(BINDATA) $(OUT_DIR)
 
 local-image:
-ifdef USE_BUILDAH
-	buildah bud $(BUILDAH_OPTS) -t $(IMAGE_TAG) -f $(DOCKERFILE) .
-else
-	sudo docker build -t $(IMAGE_TAG) -f $(DOCKERFILE) .
-endif
+	$(IMAGE_BUILD_CMD) build -t $(IMAGE) -f $(DOCKERFILE) .
 
 local-image-push:
-ifdef USE_BUILDAH
-	buildah push $(BUILDAH_OPTS) $(IMAGE_TAG) $(IMAGE_REGISTRY)/$(IMAGE_TAG)
-else
-	sudo docker tag $(IMAGE_TAG) $(IMAGE_REGISTRY)/$(IMAGE_TAG)
-	sudo docker push $(IMAGE_REGISTRY)/$(IMAGE_TAG)
-endif
+	$(IMAGE_PUSH_CMD) push $(IMAGE)
 
 .PHONY: all build deepcopy crd-schema-gen test-e2e verify verify-gofmt clean local-image local-image-push
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ clean:
 	rm -rf $(BINDATA) $(OUT_DIR)
 
 local-image:
-	$(IMAGE_BUILD_CMD) build -t $(IMAGE) -f $(DOCKERFILE) .
+	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_EXTRA_OPTS) -t $(IMAGE) -f $(DOCKERFILE) .
 
 local-image-push:
 	$(IMAGE_PUSH_CMD) push $(IMAGE)


Makefile
```diff
@@ -20,9 +20,11 @@ API_TYPES_GENERATED:=$(API_TYPES_DIR)/$(API_ZZ_GENERATED).go
 API_GO_HEADER_FILE:=pkg/apis/header.go.txt
 
 # Container image-related variables
-DOCKERFILE=Dockerfile
-IMAGE_TAG=openshift/origin-cluster-node-tuning-operator
-IMAGE_REGISTRY=quay.io
+DOCKERFILE	?= Dockerfile.rhel7
```
sjug: Why the changes in syntax and spacing?

jmencak: Why are you changing the default?

Makefile
```diff
@@ -97,18 +99,9 @@ clean:
 	rm -rf $(BINDATA) $(OUT_DIR)
 
 local-image:
-ifdef USE_BUILDAH
-	buildah bud $(BUILDAH_OPTS) -t $(IMAGE_TAG) -f $(DOCKERFILE) .
-else
-	sudo docker build -t $(IMAGE_TAG) -f $(DOCKERFILE) .
-endif
+	podman build --no-cache -t $(IMAGE) -f $(DOCKERFILE) .
```
sjug: Why the move to podman and not just buildah?

sjug: Also I don't usually use rootless builds, so if it were up to me I'd keep/add sudo as well.

ArangoGutierrez: > Is there a reason we want to remove a support for docker?

OpenShift 4.x is PodMan based, and there is no RHEL support for docker in the upstream releases,so we should move to the tools OCP is based on.

ArangoGutierrez: @sjug supo option added

Makefile
ArangoGutierrez: @sjug syntax back again, it was my VScode linter, and I didn't checked

Makefile
jmencak: I'm always doing rootless builds and I actually prefer them. :)

ArangoGutierrez: update: rootless with buildah and root with podman

jmencak: What would be the reason to use `buildah` and `podman`?  I see the value in supporting `docker` at least for the time being, but I fail to see the reason to use both of `buildah` and `podman`.

jmencak: The sudo/sudo-less problem could be solved by something like: https://github.com/openshift/cloud-credential-operator/blob/1ea34db53e72fefeae29fa1da915644d8281caf1/Makefile#L76 

By default, I'd prefer sudo-less variant though unless there are objections.

ArangoGutierrez: I like that approach, if others agree I can update the PR

sjug: Yeah good call @jmencak, I like that sudo switch.

ArangoGutierrez:
```bash
DISTRO ?= $(shell if which lsb_release &> /dev/null; then lsb_release -si; else echo "Unknown"; fi)
# Default fedora to not using sudo since it's not needed
ifeq ($(DISTRO),Fedora)
	SUDO_CMD =
else # Other distros like RHEL 7 and CentOS 7 currently need sudo.
	SUDO_CMD = sudo
endif
```

the script for telling the os in that link doesn't work for me. looks like `lsb_release` doesn't come by default in fedora32.

thoughts?

jmencak: I think the above is completely unnecessary.  Users can add `SUDO_CMD` from the command line as an "external" `Makefile` variable.  I'd personally keep it empty by default and not add any more unnecessary code (like the one above) to the `Makefile`.  In our CI, the build works fine without sudo.  Never tested with sudo and I'd be reluctant to change this unless there's a strong reason for it.

ArangoGutierrez: say no more, updating :)

Makefile
```diff
@@ -21,8 +21,10 @@ API_GO_HEADER_FILE:=pkg/apis/header.go.txt
 
 # Container image-related variables
 DOCKERFILE=Dockerfile
-IMAGE_TAG=openshift/origin-cluster-node-tuning-operator
-IMAGE_REGISTRY=quay.io
+REGISTRY=quay.io
+ORG=openshift
+TAG=$(shell git branch | grep \* | cut -d ' ' -f2)
```
jmencak: Do we need to use so many filters?  Wouldn't `git rev-parse --abbrev-ref HEAD` work?

Makefile
```diff
@@ -21,8 +21,10 @@ API_GO_HEADER_FILE:=pkg/apis/header.go.txt
...
+IMAGE=${REGISTRY}/${ORG}/origin-cluser-node-tuning-operator:${TAG}
```
jmencak: Why are we using ${} rather than $() to refer to `Makefile` variables here?

Makefile
ArangoGutierrez: for consistency mostly, this is how is being done in other repos

Makefile
ArangoGutierrez: either is a valid reference, if you prefer I can change it.

Makefile
jmencak: To replace one command with three is completely unnecessary and error prone.

Makefile
jmencak: For *consistency* within the `Makefile`, please let's use `$()`

Makefile
```diff
@@ -111,17 +114,16 @@ clean:
 
 local-image:
 ifdef USE_BUILDAH
-	buildah bud $(BUILDAH_OPTS) -t $(IMAGE_TAG) -f $(DOCKERFILE) .
+	buildah bud $(BUILDAH_OPTS) -t $(IMAGE) -f $(DOCKERFILE) .
 else
-	sudo docker build -t $(IMAGE_TAG) -f $(DOCKERFILE) .
+	$(SUDO_CMD) $(CONTAINER_ENGINE) build -t $(IMAGE) -f $(DOCKERFILE) .
```
jmencak, line 119: If we need `$(SUDO_CMD)` here, do we also need it for `buildah`?  Or should we just drop it for podman?  Or is the intent to still support `docker` via this?

jmencak (commented):
On the whole it looks good and I like the new variable name IMAGE, the IMAGE_TAG didn't really make sense.  Thanks!

Makefile
```diff
@@ -25,9 +25,12 @@ API_TYPES_GENERATED:=$(API_TYPES_DIR)/$(API_ZZ_GENERATED).go
 API_GO_HEADER_FILE:=pkg/apis/header.go.txt
 
 # Container image-related variables
+CONTAINER_ENGINE ?= podman
```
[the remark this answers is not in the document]
ArangoGutierrez: You are right, another work around I have seen on another repos is using the var IMAGE_BUILD_CMD 
it can be set to
IMAGE_BUILD_CMD=docker build
IMAGE_BUILD_CMD=buildah bud
IMAGE_BUILD_CMD=podman build

and leave the default to podman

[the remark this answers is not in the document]
jmencak: In that case this can probably be cleaned up even further by removing `USE_BUILDAH` completely.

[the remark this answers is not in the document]
ArangoGutierrez: yup, if you like that approach I can make the proper changes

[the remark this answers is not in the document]
jmencak: I welcome everything that reduces code duplication. :)

Makefile
```diff
@@ -25,9 +25,13 @@ API_TYPES_GENERATED:=$(API_TYPES_DIR)/$(API_ZZ_GENERATED).go
 API_GO_HEADER_FILE:=pkg/apis/header.go.txt
 
 # Container image-related variables
+IMAGE_BUILD_CMD:= podman build
```
jmencak, line 28: Nit, remove space, please.  Also, I know the whole Makefile can do with a cleanup wrt. = vs. := but why add more inconsistencies into this section?

Makefile
```diff
@@ -25,9 +25,13 @@ API_TYPES_GENERATED:=$(API_TYPES_DIR)/$(API_ZZ_GENERATED).go
...
+IMAGE_PUSH_CMD:= podman push
```
jmencak, line 29: Nit, remove space, please

Makefile
```diff
@@ -25,9 +25,13 @@ API_TYPES_GENERATED:=$(API_TYPES_DIR)/$(API_ZZ_GENERATED).go
...
 DOCKERFILE=Dockerfile
-IMAGE_TAG=openshift/origin-cluster-node-tuning-operator
-IMAGE_REGISTRY=quay.io
+REGISTRY=quay.io
+ORG=openshift
+TAG=$(shell git rev-parse --abbrev-ref HEAD)
+IMAGE=$(REGISTRY)/$(ORG)/origin-cluser-node-tuning-operator:$(TAG)
```
jmencak, line 34: What's 'cluser'?

Makefile
```diff
@@ -110,18 +114,9 @@ clean:
 	rm -rf $(BINDATA) $(OUT_DIR)
 
 local-image:
-ifdef USE_BUILDAH
-	buildah bud $(BUILDAH_OPTS) -t $(IMAGE_TAG) -f $(DOCKERFILE) .
-else
-	sudo docker build -t $(IMAGE_TAG) -f $(DOCKERFILE) .
-endif
+	$(IMAGE_BUILD_CMD) build -t $(IMAGE) -f $(DOCKERFILE) .
```
jmencak, line 117: How do I pass additional options to the build now that `$(BUILDAH_OPTS)` are removed?

jmencak (changes requested):
Thanks for the changes, slowly getting there I believe :)

Makefile
ArangoGutierrez: Typo

Makefile
ArangoGutierrez: we could add a var `IMAGE_BUILD_EXTRA_OPTS ?=`  
then it will look something like 
```
$(IMAGE_BUILD_CMD) build -t $(IMAGE) $(IMAGE_BUILD_EXTRA_OPTS) -f $(DOCKERFILE) .
```

and the make can be called

```bash
IMAGE_BUILD_EXTRA_OPTS=some extra args
make build
```
wdyt?

jmencak: The options probably need to go right after "build".  Saying that,
`IMAGE_BUILD_CMD=podman build`  why do we have another "build" on the line?  Same below for push.  Did you test this PR?

jmencak: @ArangoGutierrez I believe this wasn't addressed yet.  Again...  Have you tested this PR?  Also, the extra options need to be passed to the push too.

jmencak (changes requested):
I'm afraid, issues mentioned earlier with this PR still were not addressed.

ArangoGutierrez:
> I'm afraid, issues mentioned earlier with this PR still were not addressed.
5a32159 fix the remaining VAR

Tested as:

```bash
make local-image IMAGE_BUILD_CMD="buildah bud"  ORG=eduardoarango
make local-image-push IMAGE_PUSH_CMD="buildah push"  ORG=eduardoarango
```

jmencak:
Thank you for the PR.  This will work and now it is an improvement over what we had.
/lgtm

PR #115, opened 2020-02-04
Bug 1801797: Pod-label support optimizations.

jmencak:
Changes:
  - Remove leftover profiles on node deletions.
  - On clusters with many thousands of pods, the operator can take large
    amounts of memory.  Also, clusters with a lot of "pod churn" (pods
    started/deleted, pod labels changed, ...) can show CPU utilization spikes
    on a node the operator runs.  Disable pod support by default.  While pod
    support is automatically re-enabled when a custom CR with pod label match
    is detected, it is discouraged to use this feature especially in
    large-scale clusters.
  - A pod label matching fix for pods with the same labels as other pods running 
    on the same node.
  - More consistent logging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,12 @@ overwritten by the operator.  For custom tuning, create your own tuned CRs.
 Newly created CRs will be combined with the default CR and custom tuning
 applied to OpenShift nodes based on node/pod labels and profile priorities.
 
+While in certain situations the support for pod labels can be a convenient
+way of automatically delivering required tuning, this practice is discouraged
+and strongly advised against especially in large-scale clusters.  The default
+tuned CR ships without pod label matching.  If a custom profile is created
+with pod label matching the functionality will be enabled at that time.
+
 
 ## Custom tuning specification
 
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeinformers "k8s.io/client-go/informers"
+	corev1informers "k8s.io/client-go/informers/core/v1"
 	kubeset "k8s.io/client-go/kubernetes"
 	appsset "k8s.io/client-go/kubernetes/typed/apps/v1"
 	coreset "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -54,6 +55,13 @@ type Controller struct {
 
 	listers *ntoclient.Listers
 	clients *ntoclient.Clients
+
+	pod struct {
+		informerFactory kubeinformers.SharedInformerFactory
+		informer        corev1informers.PodInformer
+		informerEnabled bool
+		stopCh          chan struct{}
+	}
 
 	pc *ProfileCalculator
 }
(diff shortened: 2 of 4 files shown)


jmencak:
@sjug, @ArangoGutierrez PTAL, thanks!

README.md
```diff
@@ -41,6 +41,9 @@ the OpenShift platform and any custom changes to the the default CR will be
 overwritten by the operator.  For custom tuning, create your own tuned CRs.
 Newly created CRs will be combined with the default CR and custom tuning
 applied to OpenShift nodes based on node/pod labels and profile priorities.
+While in certain situations the support for pod labels can be a convenient
+way of automatically delivering required tuning, this practice is discouraged
+and strongly advised against especially in large-scale clusters.
```
sjug: Could you add a note on how to enable pod label support in the doc as well?

pkg/operator/controller.go
```diff
@@ -484,6 +514,31 @@ func (c *Controller) informerEventHandler(workqueueKey wqKey) cache.ResourceEven
...
+
+		c.pod.informer = c.pod.informerFactory.Core().V1().Pods()
+		c.listers.Pods = c.pod.informer.Lister()
+		c.pod.informer.Informer().AddEventHandler(c.informerEventHandler(wqKey{kind: wqKindPod}))
+
+		c.pod.informerFactory.Start(c.pod.stopCh)
+	} else {
+		defer close(c.pod.stopCh)
+		c.pod.stopCh <- struct{}{}
+		c.pc.podLabelsDelete()
```
sjug: Was this used somewhere else in order to make a new method?

pkg/operator/controller.go
```diff
@@ -484,6 +514,31 @@ func (c *Controller) informerEventHandler(workqueueKey wqKey) cache.ResourceEven
...
+
+	if enable {
+		c.pod.stopCh = make(chan struct{})
+		c.pod.informerFactory = kubeinformers.NewSharedInformerFactoryWithOptions(c.clients.Kube, ntoconfig.ResyncPeriod(), kubeinformers.WithNamespace(corev1.NamespaceAll))
+
+		c.pod.informer = c.pod.informerFactory.Core().V1().Pods()
+		c.listers.Pods = c.pod.informer.Lister()
+		c.pod.informer.Informer().AddEventHandler(c.informerEventHandler(wqKey{kind: wqKindPod}))
+
+		c.pod.informerFactory.Start(c.pod.stopCh)
```
sjug: Nice way to stop the `infomerFactory`

pkg/operator/controller.go
```diff
@@ -574,13 +625,12 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
...
 	configInformerFactory.Start(stopCh)  // ClusterOperator
 	kubeNTOInformerFactory.Start(stopCh) // DaemonSet
-	kubeInformerFactory.Start(stopCh)    // Pod/Node
+	kubeInformerFactory.Start(stopCh)    // Node
 	tunedInformerFactory.Start(stopCh)   // Tuned/Profile
 
 	// Wait for the caches to be synced before starting worker(s)
 	klog.Info("waiting for informer caches to sync")
 	ok := cache.WaitForCacheSync(stopCh,
-		podInformer.Informer().HasSynced,
```
sjug: Do we still wait for the podInformer cache sync if the pod labels are enabled?

README.md
jmencak: Thank you, will add.

pkg/operator/controller.go
jmencak: We do not at the moment.  We can, however, the question is whether we want to.  On a small cluster I've measured the cache sync wait on enabling the Pod sync to be 0.5-1s.  I cannot think of a good reason to enable it, but there might be one.  Can you think of anything?

sjug: Just wanted to make sure it was still functional as is.

jmencak:
Squashed and pushed.

sjug:
/lgtm

Looks even better than a few minutes ago

PR #119, opened 2020-03-19
Support for MachineConfigs.

jmencak:
Support for MachineConfigs.

Changes:
  - added support for MachineConfig labels in the "recommend:" section
  - augmented Tuned API to publish node's kernel boot parameters
  - tuned patch to rollback bootcmdline settings in containers without grub.cfg
  - consistency changes metaapi -> metav1
  - logging cleanup to show relevant logs to admins for the operator and operand
  - operator cleanup (if/else) statements as suggested by sjug
  - changed "array" to "list" (see yaml terminology) in the README.md and
    documented the new MachineConfig functionality
  - openshift-tuned refactoring and cleanup to:
    * use sharedInformerFactory
    * use caches wherever possible
    * use workqueues for event retries and throttling
    * do a periodic resync
    * compare the profile to be written vs. active data to eliminate unnecessary tuned
      daemon reloads
    * no more global vars (cleanup to use methods vs. functions)

Example usage:

$ oc label node <node-name> node-role.kubernetes.io/worker-rt=
$ oc create -f examples/realtime-mc.yaml
$ oc create -f examples/realtime-mcp.yaml


MarSik:
@jmencak Great, I will take a look. I noticed the PR description uses MachineConfigSelector, that is probably a historical artefact that should be changed to MAchineConfigLabels, right?

jmencak:
> @jmencak Great, I will take a look. I noticed the PR description uses MachineConfigSelector, that is probably a historical artefact that should be changed to MAchineConfigLabels, right?

@MarSik thanks, appreciated.  The code works but still WiP and I'm planning improvements especially to the pruning of unused MCs as it is now can be unnecessarily CPU intensive.  MachineConfigSelector in the PR is only used in an cluster-admin-provisioned MachineConfigPool -- something the cluster administrator would to do create the initial (e.g. `worker-rt`) MachineConfigPool.  So no, it should *not* be changed and works as intended.  MachineConfigLabels are used in the tuned CR in `example/realtime-mc.yaml`.

yanirq:
@jmencak  Can you split the generated and vendored code into different commits so it would be more easy to read ?

yanirq:
We will need to add to e2e tests the coverage needed for these additions.
I wonder if NTO should switch on using Ginko,Gomega for functional tests.

yanirq:
Would it be possible to have additional kernel arguments set in the profile like we did in https://github.com/openshift-kni/performance-addon-operators/pull/133 ?  Doesn't have to be in this PR but adding this question for tracking.

examples/realtime-mc.yaml
```diff
@@ -0,0 +1,21 @@
...
+  namespace: openshift-cluster-node-tuning-operator
+spec:
+  profile:
+  - data: |
+      [main]
+      summary=Custom OpenShift realtime profile
+      include=openshift-node,realtime
+      [variables]
+      # isolated_cores take a list of ranges; e.g. isolated_cores=2,4-7
+      isolated_cores=1
```
yanirq: This will directly utilize https://github.com/redhat-performance/tuned/blob/master/profiles/cpu-partitioning/tuned.conf ?

pkg/operator/controller.go
```diff
@@ -420,39 +478,159 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
...
-	} else {
-		if profile.Spec.Config.TunedProfile == tunedProfileName {
-			klog.V(2).Infof("no need to update Profile %s", nodeName)
+	}
+
+	if mcLabels != nil {
+		// The tuned profile "tunedProfileName" for nodeName matched with MachineConfig
+		// labels set for additional machine configuration.  Sync the operator-created
+		// MachineConfig for MachineConfigPools 'pools'.
+		err := c.syncMachineConfig(getMachineConfigNameForPools(pools), mcLabels, profile.Status.Bootcmdline)
```
yanirq: Is  `profile.Status.Bootcmdline` Represents all tuned kernel arguments that should be added by the mc ? Does it include all argument for the base profiles in use ? (e.g. realtime + network latency arguments)

jmencak:
> @jmencak Can you split the generated and vendored code into different commits so it would be more easy to read ?

I can try to, you insist.  The reason I added vendor dependencies was because I was having dependency issues so I decided to pull in the latest 1.17.1.  I'm actually hoping to change this to 1.18 soon for the 4.5 branch.  Can you not ignore the `/vendor` directory?

jmencak:
> We will need to add to e2e tests the coverage needed for these additions.
> I wonder if NTO should switch on using Ginko,Gomega for functional tests.

I agree that e2e tests need to be added, will add them to this PR.  As for switching to Ginko/Gomega I'm currently not considering the switch.  What would be the benefit?

jmencak:
> Would it be possible to have additional kernel arguments set in the profile like we did in [openshift-kni/performance-addon-operators#133](https://github.com/openshift-kni/performance-addon-operators/pull/133) ? Doesn't have to be in this PR but adding this question for tracking.

Yes, it can be done.  Something like this will work:
```
kind: Tuned
metadata:
  name: openshift-realtime
spec:
  profile:
  - data: |
      [main]
      summary=Custom OpenShift realtime profile
      include=openshift-node,realtime
      [variables]
      # isolated_cores take a list of ranges; e.g. isolated_cores=2,4-7
      isolated_cores=1
      [bootloader]
      cmdline_ocp_realtime=+your_additional_kernel_arguments
    name: openshift-realtime
  recommend:
  - machineConfigLabels:
      machineconfiguration.openshift.io/role: "worker-rt"
    priority: 30
    profile: openshift-realtime
```

examples/realtime-mc.yaml
jmencak: > This will directly utilize https://github.com/redhat-performance/tuned/blob/master/profiles/realtime/tuned.conf?

Yes, see the patch in `/assets/patches/060-realtime.diff`.

pkg/operator/controller.go
jmencak: Yes, it does.

jmencak:
@sjug PTAL

jmencak:
/hold
Waiting for /lgtm from the PSAP team.

jmencak:
@sjug PTAL

sjug:
/lgtm

Great work @jmencak!

PR #124, opened 2020-04-15
Support isolate_managed_irq in the realtime profile.

jmencak:
This PR supports the new isolate_managed_irq functionality in the upstream realtime profile and prepares ground for tuned-2.11.0-9.el7.noarch

### DIFF
--- a/examples/realtime-mc.yaml
+++ b/examples/realtime-mc.yaml
@@ -12,6 +12,7 @@ spec:
       [variables]
       # isolated_cores take a list of ranges; e.g. isolated_cores=2,4-7
       isolated_cores=1
+      #isolate_managed_irq=Y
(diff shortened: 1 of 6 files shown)


jmencak:
Merge after https://github.com/openshift/cluster-node-tuning-operator/pull/123
/hold

examples/realtime-mc.yaml
```diff
@@ -12,6 +12,7 @@ spec:
       [variables]
       # isolated_cores take a list of ranges; e.g. isolated_cores=2,4-7
       isolated_cores=1
+      #isolate_managed_irq=Y
```
yanirq, line 15: so this is an optional override we can use when needed ?  (to override realtime profile)

jmencak: yes, optional and can be used when needed

PR #130, opened 2020-04-30
Switch to ginkgo/gomega for e2e testing.

jmencak:
Switching to ginkgo/gomega for e2e tests.


jmencak:
/cc @yanirq 
This is still WiP, but I'd already appreciate comments whether I'm on the right track.  Thanks!

Makefile
```diff
@@ -70,7 +70,9 @@ $(GOBINDATA_BIN):
 	$(GO) build -o $(GOBINDATA_BIN) ./vendor/github.com/kevinburke/go-bindata/go-bindata
 
 test-e2e: $(BINDATA)
-	KUBERNETES_CONFIG="$(KUBECONFIG)" $(GO) test -v -timeout 20m -tags e2e ./test/e2e/... -run .
+	for d in basic reboots; do \
```
yanirq, line 73: may i recommend a script running the suits recursively and add color setting selectively? 
see https://github.com/openshift-kni/performance-addon-operators/blob/master/hack/run-functests.sh

test/e2e/basic/available.go
```diff
@@ -0,0 +1,50 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	configv1 "github.com/openshift/api/config/v1"
+	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
+	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
+)
+
+var _ = Describe("Node Tuning Operator: availability", func() {
+	It(fmt.Sprintf("clusteroperator/%s available", tunedv1.TunedClusterOperatorResourceName), func() {
```
yanirq, line 20: A good practice would be starting always with Context()  and then nest It() under it

test/e2e/basic/available.go
```diff
@@ -0,0 +1,50 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	configv1 "github.com/openshift/api/config/v1"
+	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
+	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
+)
+
+var _ = Describe("Node Tuning Operator: availability", func() {
```
yanirq, line 19: Suggestion:  change all describe sections according to folder and name , e.g.:
Describe([basic][available]  Node tuning operator availability)

test/e2e/basic/available.go
yanirq, line 20: I'm not sure what is more beneficial, using the constant or just a plain string title. 
One pro for the plain string is it would be easier to find when tests fail.

test/e2e/basic/available.go
```diff
@@ -0,0 +1,50 @@
...
+	It(fmt.Sprintf("clusteroperator/%s available", tunedv1.TunedClusterOperatorResourceName), func() {
+		By(fmt.Sprintf("wait for clusteroperator/%s available", tunedv1.TunedClusterOperatorResourceName))
+		err := wait.PollImmediate(1*time.Second, 5*time.Minute, func() (bool, error) {
+			co, err := cs.ClusterOperators().Get(context.TODO(), tunedv1.TunedClusterOperatorResourceName, metav1.GetOptions{})
+			if err != nil {
+				return false, nil
```
yanirq, line 25: fix to :  return false, err

Makefile
jmencak: I can see some value, however adding a `hack` directory with random scripts is something I managed to avoid so far.  I'd also like to avoid pulling extra dependencies like `github.com/onsi/ginkgo/ginkgo` which are not necessarily needed.  If they really are needed, they'd have to be vendored in as from some build environments you cannot fetch external dependencies.  As for vendoring in, I tend to follow "vendor in only things that are essential".  Can you elaborate whether that dependency is essential?

As for running tests recursively, I wanted to enforce the ordering: basic tests first, advanced (requiring reboots) next.  Also having the flexibility to enable/disable tests seemed handy.

test/e2e/basic/available.go
jmencak: Will do, thank you!

test/e2e/basic/available.go
jmencak: I see what you're saying and it makes sense.  However, if the "default" CR is ever changed you wouldn't need to go through the entire operator code to make appropriate changes.  Just change it in one place.  Also, when the tests fails, it gives you the line and the source file so I don't see locating where the test fails as an issue.  I'd rather leave things as they are.

test/e2e/basic/available.go
jmencak: Will fix and in other places too.  Thank you!

jmencak:
Getting very close to the final state I believe.
/cc @skordas

skordas:
I like it!
/lgtm

jmencak:
/hold
I need to squash first.  Also Yanir will for sure have something to say about this.  Changes expected.

test/e2e/basic/custom_node_labels.go
```diff
@@ -0,0 +1,78 @@
...
+			if node != nil {
+				exec.Command("oc", "label", "node", "--overwrite", node.Name, nodeLabelHugepages+"-")
+			}
+			exec.Command("oc", "delete", "-n", ntoconfig.OperatorNamespace(), "-f", profileHugepages)
+		})
+
+		ginkgo.It(fmt.Sprintf("%s set", sysctlVar), func() {
+			ginkgo.By("getting a list of worker nodes")
+			nodes, err := util.GetNodesByRole(cs, "worker")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
```
yanirq, line 41: is there something we can do with this gomega prefix while keeping the dot imports guidelines ? its a built bulky.

test/e2e/reboots/kernel_parameter_remove.go
```diff
@@ -0,0 +1,157 @@
...
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			ginkgo.By("ensuring the custom worker child profile was set")
+			err = wait.PollImmediate(5*time.Second, 5*time.Minute, func() (bool, error) {
+				cmdlineNew, err = util.GetFileInPod(pod, procCmdline)
+				if err != nil {
+					explain = err.Error()
+					return false, nil
+				}
+				if strings.Contains(cmdlineNew, paramRemove) {
```
yanirq, line 111: In all the contains check, do we need to consider also the order of the parameters ?  There are some use cases i believe like hugepages where the order is important.
If so then I suggest to use regular expressions instead of a string.contains check

test/e2e/reboots/kernel_parameter_remove.go
```diff
@@ -0,0 +1,157 @@
...
+	coreapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
+	nutil "github.com/openshift/cluster-node-tuning-operator/pkg/util"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/util"
+)
+
+// Test the removal of a kernel parameter defined by a parent profile.
+var _ = ginkgo.Describe("[reboots][kernel_parameter_remove] Node Tuning Operator parent profile kernel parameter removal", func() {
```
yanirq, line 21: Maybe i missed that, but why not make this test file more general and include also additions/modifications of a kernel parameter (does'nt has to be in this PR)

test/e2e/basic/custom_node_labels.go
jmencak: If you look upstream k8s or OpenShift tests, none of them use dot imports as dot imports are considered bad practice.  Also, I'm concerned about a possible removal of dot imports in future golang 2.x: https://github.com/golang/go/issues/29326

test/e2e/reboots/kernel_parameter_remove.go
jmencak: Good point, thanks!  Let's address this in a follow-up PR.  The two "reboots" tests are quite similar.  Perhaps they can be optimized/squashed together while reducing the number of reboots and extending the test coverage.

test/e2e/reboots/kernel_parameter_remove.go
jmencak: That's a great point, thank you.  Fixed.

test/e2e/reboots/kernel_parameter_remove.go
jmencak: Added scenario for adding kernel parameters in the child profile.  Marking as resolved for now.  Further improvements will be part of a follow-up PR if you agree.

yanirq:
/lgtm 
The PR seems to be in a good shape now, we can improve/add tests in other PRs if needed.

PR #135, opened 2020-06-11
Support for Operator Force/Managed/Removed/Unmanaged management states.

jmencak:
Changes:
  - support for SLO management states
  - made bootcmdline optional in tuned_types.go (it already was optional in openAPIV3Schema)
  - split `02-crd.yaml` into two files to prepare the ground for fully automated openAPIV3Schema schema generation


pkg/operator/controller.go
```diff
@@ -276,42 +289,66 @@ func (c *Controller) sync(key wqKey) error {
...
 	klog.V(2).Infof("sync(): DaemonSet")
 	err = c.syncDaemonSet(cr)
 	if err != nil {
 		return fmt.Errorf("failed to sync DaemonSet: %v", err)
 	}
 
-	if key.kind == wqKindTuned {
-		// Tuned CR changed, this can affect all profiles, list them and trigger profile updates
-		klog.V(2).Infof("sync(): Tuned %s", key.name)
+	// Tuned CR changed, this can affect all profiles, list them and trigger profile updates
```
sjug, line 316: Why do we not need the tuned kind check anymore here?

pkg/operator/controller.go
```diff
@@ -148,11 +147,50 @@ func (c *Controller) eventProcessor() {
...
+
+	switch cr.Spec.ManagementState {
+	case operatorv1.Force:
+		// Use the same logic as Managed.
+	case operatorv1.Managed, "":
+		// Managed means that the operator is actively managing its resources and trying to keep the component active.
+	case operatorv1.Removed:
+		// Removed means that the operator is actively managing its resources and trying to remove all traces of the component.
+		lastErr = c.removeResources()
+		goto out
```
sjug, line 184: Haven't seen a lot of `goto`, interesting use.

pkg/operator/controller.go
```diff
@@ -346,12 +383,14 @@ func (c *Controller) syncTunedDefault() (*tunedv1.Tuned, error) {
...
 	// Tuned resource found, check whether we need to update it
-	if reflect.DeepEqual(crMf.Spec, cr.Spec) {
+	if reflect.DeepEqual(crMf.Spec.Profile, cr.Spec.Profile) &&
+		reflect.DeepEqual(crMf.Spec.Recommend, cr.Spec.Recommend) {
 		klog.V(2).Infof("syncTunedDefault(): Tuned %s doesn't need updating", crMf.Name)
 		return cr, nil
 	}
 	cr = cr.DeepCopy() // never update the objects from cache
-	cr.Spec = crMf.Spec
+	cr.Spec.Profile = crMf.Spec.Profile
```
sjug, line 392: Are we specifying the fields so we don't copy the `meta`?

pkg/operator/controller.go
```diff
@@ -726,6 +801,59 @@ func (c *Controller) enablePodInformer(enable bool) error {
...
+	var lastErr error
+	dsMf := ntomf.TunedDaemonSet()
+
+	_, err := c.listers.DaemonSets.Get(dsMf.Name)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			lastErr = fmt.Errorf("failed to get DaemonSet %s: %v", dsMf.Name, err)
+		}
+	} else {
+		err = c.clients.Apps.DaemonSets(ntoconfig.OperatorNamespace()).Delete(context.TODO(), dsMf.Name, metav1.DeleteOptions{})
```
sjug, line 814: Do we need unique contexts passed to each delete?

pkg/operator/controller.go
jmencak: So I have no formal proof, but I did go through the code and also tested quite extensively (using the "else" branch -- it was never reached).  So I believe this check was probably some relic of the past code that wasn't cleaned up.  Cleaning up now.

pkg/operator/controller.go
jmencak: I thought you'd mention this :) http://wiki.c2.com/?StructuredProgrammingWithGoToStatements
I believe that `goto`s to quit a function in this manner are quite frequent.  At least in C programs.  If you can think of a better way to eliminate code duplication in this case, I'd be happy to hear!

pkg/operator/controller.go
jmencak: No. You don't want to overwrite the `ManagementState` that might have been changed by the admin.

pkg/operator/controller.go
jmencak: I believe [this requirement](https://github.com/kubernetes/client-go/blob/1e6150831a960cf38075bb8561e01f48837f8d74/kubernetes/typed/extensions/v1beta1/daemonset.go#L156) came with the switch to k8s 1.18.
```
# github.com/openshift/cluster-node-tuning-operator/pkg/operator
pkg/operator/controller.go:814:72: not enough arguments in call to c.clients.Apps.DaemonSets("github.com/openshift/cluster-node-tuning-operator/pkg/config".OperatorNamespace()).Delete
        have (string, "k8s.io/apimachinery/pkg/apis/meta/v1".DeleteOptions)
        want (context.Context, string, "k8s.io/apimachinery/pkg/apis/meta/v1".DeleteOptions)
```

pkg/operator/controller.go
sjug: Nope, not looking to eliminate :smile:

pkg/operator/controller.go
sjug: I was just thinking to have one context variable and just reuse it for each function? `ctx := context.TODO()`?

jmencak: OK, changed, thanks!

PR #138, opened 2020-06-26
Bug 1851214: use lowercase plurals in resource property of relatedObjects

jmencak:
The `resource` property of the `relatedObjects` stanza in the ClusterOperator object needs lowercase plurals.


jmencak:
@sjug, PTAL
thanks!

PR #144, opened 2020-07-21
Add Dockerfile.rhel8

yselkowitz:
A separate file is required, as the changes necessary for building with
RHEL 8 content are not compatible with RHEL 7.

### DIFF
--- a/Dockerfile.rhel8
+++ b/Dockerfile.rhel8
@@ -1,0 +1,43 @@
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.14 AS builder
+WORKDIR /go/src/github.com/openshift/cluster-node-tuning-operator
+COPY . .
+RUN make build
+
+FROM registry.svc.ci.openshift.org/ocp/4.6:base AS tuned
+WORKDIR /root
+COPY assets/tuned /root
+RUN INSTALL_PKGS=" \
+      git rpm-build make desktop-file-utils patch \
+      python3 python3-configobj python3-pyudev \
+      " && \
+    yum install --setopt=tsflags=nodocs -y $INSTALL_PKGS && \
+    cd daemon && \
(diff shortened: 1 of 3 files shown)


Dockerfile.rhel8
```diff
@@ -0,0 +1,43 @@
...
+COPY manifests /manifests
+ENV APP_ROOT=/var/lib/tuned
+ENV PATH=${APP_ROOT}/bin:${PATH}
+ENV HOME=${APP_ROOT}
+WORKDIR ${APP_ROOT}
+COPY --from=builder /go/src/github.com/openshift/cluster-node-tuning-operator/_output/openshift-tuned /usr/bin/
+COPY --from=builder /go/src/github.com/openshift/cluster-node-tuning-operator/assets ${APP_ROOT}
+COPY --from=tuned   /root/rpmbuild/RPMS/noarch /root/rpms
+RUN INSTALL_PKGS=" \
+     hdparm kernel-tools tuna socat \
```
jmencak, line 30: Let's drop `hdparm`, `kernel-tools` and `tuna` and keep only `socat`.  `socat` is the only real dependency for NTO's operand (openshift-tuned).  `tuna` will be pulled in by tuned profiles that need it by `yum --setopt=tsflags=nodocs -y install /root/rpms/*.rpm` and so will `hdparm` and `kernel-tools` on platforms that ship those packages.  On RHEL 7, there's no package `hdparm` and `kernel-tools` for  s390x.  This seems this is no longer the case for RHEL 8, but let's keep Dockerfiles as much close to each other as possible.  Would you be able to adjust Dockerfile and Dockerfile.rhel7 in a similar way?  Thank you!

yselkowitz: In that case, we don't even need a Dockerfile.rhel8.  Modified PR accordingly.

jmencak:
/lgtm
/approve
Thank you!

PR #148, opened 2020-07-29
Tuned daemon fixes.

jmencak:
This PR adds two fixes that resolve rhbz#1746957 and rhbz#1784648 on top of tuned 2.14.  The same fixes that ship with the Tuned daemon in RHEL 8.3.


jmencak:
@sjug PTAL, needs to merge by this Friday. Thanks.

PR #149, opened 2020-08-07
Set maxUnavailable on ds/tuned to 10%

sdodson:
This daemonset isn't critical for ensuring availability so allow up to
10% to be updated at once

On a 250 node cluster we're seeing about 5 pods per minute
upgrading from 4.5.4 to 4.5.5 which isn't horrible, but we can surely
upgrade more than one at a time.


sdodson:
@smarterclayton What'd you say was a reasonable maxUnavailable for daemonsets like these?

sdodson:
/retitle Bug 1867604: Set maxUnavailable on ds/tuned to 10%

PR #152, opened 2020-08-27
Bug 1878602: Support for stall daemon.

jmencak:
This PR adds support for the stalld program to prevent the _starvation_ of
operating system threads in a Linux system.

Changes:
  - support for `[service]` Tuned plugin
  - support the installation of stalld binary and systemd units on hosts fully-managed by the [Machine Config Operator](https://github.com/openshift/machine-config-operator) (MCO).

Example usage:

```
$ oc label node <node-name> node-role.kubernetes.io/worker-rt=

$ oc create -f- <<EOF
kind: Tuned
metadata:
  name: openshift-realtime
spec:
  profile:
  - data: |
      [main]
      summary=Custom OpenShift realtime profile
      include=openshift-node,realtime
      [service]
      service.stalld=start,enable
    name: openshift-realtime

  recommend:
  - machineConfigLabels:
      machineconfiguration.openshift.io/role: "worker-rt"
    priority: 30
    profile: openshift-realtime
EOF

$ oc create -f examples/realtime-mcp.yaml
```

A host-supplied configuration file can be used to override the `stalld` systemd unit.  This file can then be referred to by prefixing the absolute path to the file on the host by the `/host` prefix.  For example:

```
      [service]
      service.stalld=start,enable,file:/host/etc/tuned/openshift-realtime/stalld.conf
```

Note that the file needs to be created on the host by other means, for example, by using a MachineConfig. One of the uses of the drop-in file is to supply environment variables to the stalld unit.  For example, the drop-in file may contain:
```
[Service]
Environment=CLIST=
Environment=AGGR=
Environment=BP="-p 1000000000"
Environment=BR="-r 10000"
Environment=BD="-d 3"
Environment=THRESH="-t 60"
Environment=LOGGING="--log_syslog --log_kmsg"
Environment=FG=--foreground
Environment=PF="--pidfile /run/stalld.pid"
```

### DIFF
--- a/assets/tuned/patches/070-service-plugin.diff
+++ b/assets/tuned/patches/070-service-plugin.diff
@@ -1,0 +1,354 @@
+The service plugin can handle sysvinit, sysv-rc, openrc, systemd.
+Syntax is the following:
+
+[service]
+service.sendmail = start,enable,file:${i:PROFILE_DIR}/tuned-sendmail.conf
+
+'start' means to start the service, 'stop' to stop, 'enable' to enable,
+'disable' to disable, file:FILE is the overlay configuration file
+that will be installed for the service. Multiple directives can be
+separated by comma ',' or semicolon ';'. If the directives are
+conflicting, the last one will be used. The internal variable
+${i:PROFILE_DIR} points to the directory from which the profile is
+loaded.
+
+The service plugin supports configuration overlays only for the systemd,
+for other init systems this directive is ignored. The configuration
+overlay files are copied to the /etc/systemd/system/SERVICE_NAME.service.d/
+directories. Upon unloading if the directory is empty, it's removed.
+
+With systemd, the 'start' directive is implemented by 'restart' in order
+to allow loading of the service configuration file overlay.
+
+With init systems other than systemd, it works only with the current runlevel.
+
+Resolves: rhbz#1869991
+
+See: https://github.com/redhat-performance/tuned/pull/293
+
+diff --git a/tuned/consts.py b/tuned/consts.py
+index 733ad513..876fe69d 100644
+--- a/tuned/consts.py
++++ b/tuned/consts.py
+@@ -49,6 +49,10 @@
+ DEF_CGROUP_MOUNT_POINT = "/sys/fs/cgroup/cpuset"
+ DEF_CGROUP_MODE = 0o770
+ 
++# service plugin configuration
++SERVICE_SYSTEMD_CFG_PATH = "/etc/systemd/system/%s.service.d"
++DEF_SERVICE_CFG_DIR_MODE = 0o755
++
+ # modules plugin configuration
+ MODULES_FILE = "/etc/modprobe.d/tuned.conf"
+ 
+diff --git a/tuned/plugins/plugin_service.py b/tuned/plugins/plugin_service.py
+new file mode 100644
+index 00000000..84a071f7
+--- /dev/null
++++ b/tuned/plugins/plugin_service.py
+@@ -0,0 +1,283 @@
++from . import base
++import collections
++import tuned.consts as consts
++from .decorators import *
++import os
++import re
++import tuned.logs
++from tuned.utils.commands import commands
++
++log = tuned.logs.get()
++cmd = commands()
++
++class Service():
++	def __init__(self, start = None, enable = None, cfg_file = None, runlevel = None):
++		self.enable = enable
++		self.start = start
++		self.cfg_file = cfg_file
++		self.runlevel = runlevel
++
++class InitHandler():
++	def runlevel_get(self):
++		(retcode, out) = cmd.execute(["runlevel"])
++		return out.split()[-1] if retcode == 0 else None
++
++	def daemon_reload(self):
++		cmd.execute(["telinit", "q"])
++
++	def cfg_install(self, name, cfg_file):
++		pass
++
++	def cfg_uninstall(self, name, cfg_file):
++		pass
++
++	def cfg_verify(self, name, cfg_file):
++		return None
++
++# no enable/disable
++class SysVBasicHandler(InitHandler):
++	def start(self, name):
++		cmd.execute(["service", name, "start"])
++
++	def stop(self, name):
++		cmd.execute(["service", name, "stop"])
++
++	def enable(self, name, runlevel):
++		raise NotImplementedError()
++
++	def disable(self, name, runlevel):
++		raise NotImplementedError()
++
++	def is_running(self, name):
++		(retcode, out) = cmd.execute(["service", name, "status"], no_errors = [0])
++		return retcode == 0
++
++	def is_enabled(self, name, runlevel):
++		raise NotImplementedError()
++
++class SysVHandler(SysVBasicHandler):
++	def enable(self, name, runlevel):
++		cmd.execute(["chkconfig", "--level", runlevel, name, "on"])
++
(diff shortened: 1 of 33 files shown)


kpouget (commented):
I didn't test the code, but what I read/understood in the patch seem sensible and consistent, thanks!
I only made a minor style question.

I didn't read in detail the Python code (the service plugin patch), is it new code or is it coming from elsewhere?

pkg/operator/controller.go
```diff
@@ -561,48 +564,84 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
...
+			klog.Infof("created MachineConfig %s with%s", mc.ObjectMeta.Name, logline(haveIgnition, len(bootcmdline) != 0, bootcmdline))
 			return nil
 		}
 		return err
 	}
 
-	if util.StringSlicesEqual(mc.Spec.KernelArguments, kernelArguments) {
+	mcNew := newMachineConfig(name, annotations, labels, kernelArguments, ignFiles, ignUnits)
+
+	ignEqual, err := ignEqual(mc, mcNew)
```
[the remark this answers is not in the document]
jmencak: Yeah, good point.  And even if it was safe, this can probably be very confusing for people to read.  Let me change the variable or function name.

jmencak:
The "Operand cleanup, using long option for stalld." got rid of the limitation in the description.  Removing.  This is a Friday night commit -- it seems to work but more testing and cleanup needed.

examples/realtime-mc.yaml
```diff
@@ -16,6 +16,8 @@ spec:
       not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
       [bootloader]
       cmdline_ocp_realtime=+systemd.cpu_affinity=${not_isolated_cores_expanded}
+      #[service]
+      #service.stalld=start,enable
```
yanirq, line 20: passing variables will be here ? (if not then a reference in the comment will suffice)

yanirq:
Will tests be added in this PR ? (should be)

examples/realtime-mc.yaml
jmencak: Yes, I'm planning to document this.  A "pointer" to an environment file will be passed here.

jmencak:
> Will tests be added in this PR ? (should be)

As time allows one all issues are resolved.  I still see two issues that need to be addressed.

yanirq:
I would add the PR's description to NTO's readme section (or open a new sub folder doc)

jmencak:
@yanirq cannot reproduce the "degraded" MCP state issues I was having last week after several successful tests.  There were some changes in MCO which hopefully fixed those.  I believe I've addressed all your concerns?  Could you please have another look through the code and indicate whether there's something missing?  Thanks!

yanirq:
/lgtm
/hold
looks good overall (unless im missing something) if you feel comfortable release the hold 
I do think we need tests embedded (they usually come with the PR) but because of urgency we can merge then in a very soon to follow PR

jmencak:
> /lgtm
> /hold
> looks good overall (unless im missing something) if you feel comfortable release the hold
> I do think we need tests embedded (they usually come with the PR) but because of urgency we can merge then in a very soon to follow PR

Thanks @yanirq . Which tests do you have in mind?  There is an embedded e2e tests for this new functionality.

jmencak:
@kpouget / @dagrayvid  would you be able to provide a feedback for this PR?  This is close to being complete.  Thank you!

pkg/tuned/host_payload.go
```diff
@@ -0,0 +1,74 @@
...
+func stalldSystemdUnit() ign3types.Unit {
+	unit := `[Unit]
+Description=Stall Monitor
+
+[Service]
+Environment=CLIST=
+Environment=AGGR=
+Environment=BP="-p 1000000000"
+Environment=BR="-r 20000"
+Environment=BD="-d 3"
```
kpouget, line 39: really just a suggestion: long options could be easier to understand, `boost_period` instead of `BP` or `p`

```suggestion
Environment=BP="--boost_period 1000000000"
Environment=BR="--boost_runtime 20000"
Environment=BD="--boost_duration 3"
```

jmencak:
Reiterating
/hold
from Yanir as the commit messages need to be squashed before merge.  We also need another review from PSAP team.

pkg/tuned/tuned.go
```diff
@@ -370,11 +380,17 @@ func (c *Controller) tunedRun() {
...
+			if c.daemon.reloading {
+				c.daemon.reloading = !(strings.Index(l, "static tuning from profile") >= 0 && strings.Index(l, "applied") >= 0)
+				c.daemon.reloaded = !c.daemon.reloading
+			}
+			fmt.Printf("%s\n", l)
 		}
 	}()
 
 	err = c.tunedCmd.Start()
+	c.daemon.reloading = true
```
kpouget, line 393: I'm confused by the lines above, I'm afraid that there might be a race condition, but I have 0 prior knowledge about Goroutines, so take this comment with caution:
* `c.tunedCmd.Start()` unblocks `scanner.Text()`, and the goroutine checks `c.daemon.reloading`
* the main routine sets `	c.daemon.reloading = true`

```suggestion
	c.daemon.reloading = true
	err = c.tunedCmd.Start()
```

kpouget (commented):
looks good to me, except a possible race condition. Please have a look and let me know if you agree, or if I missed something in the synchronization!

assets/tuned/patches/070-service-plugin.diff
```diff
@@ -0,0 +1,354 @@
...
+diff --git a/tuned/consts.py b/tuned/consts.py
+index 733ad513..876fe69d 100644
+--- a/tuned/consts.py
++++ b/tuned/consts.py
+@@ -49,6 +49,10 @@
+ DEF_CGROUP_MOUNT_POINT = "/sys/fs/cgroup/cpuset"
+ DEF_CGROUP_MODE = 0o770
+ 
++# service plugin configuration
++SERVICE_SYSTEMD_CFG_PATH="/etc/systemd/system/%s.service.d"
```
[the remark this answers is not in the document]
jmencak: Thanks.  This is an official upstream tuned patch.  Cannot change it here, but can raise this in upstream tuned.

pkg/tuned/host_payload.go
jmencak: Agreed, however, I tried to diverge from the upstream as little as possible (https://github.com/bristot/stalld/blob/master/redhat/stalld.conf).  Honestly not sure which is better.  Diverge or have long options.

pkg/tuned/tuned.go
jmencak: This is a great catch, thank you!  Will fix.

assets/tuned/patches/070-service-plugin.diff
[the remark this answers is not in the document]
jmencak: Created a PR to get this fixed upstream.

https://github.com/redhat-performance/tuned/pull/295

pkg/tuned/host_payload.go
jmencak: Added comments in the unit to match the upstream `stalld.conf`.

jmencak:
/hold cancel
I believe I've addressed reviewer's comments.  Commits squashed.  If this still looks good, @yanirq feel free to lgtm.

PR #154, opened 2020-09-17
Bug 1879854: Bump vendor dependencies to k8s 1.19.0

jmencak:
Changes:
  - vendor dependencies to k8s 1.19.0
  - latest OpenShift and MCO dependencies
  - switch to golang 1.15
  - switched to controller-gen to generate/update CRDs; there is
    no longer a need to manually edit CRDs after they are generated


kpouget (commented):
Most of the changes look sane and match the comments in the commit message,
just for ` manifests/02-crd-profile.yaml `, `manifests/02-crd-tuned.yaml`, could you confirm that the changes (paragraph filling and order of properties) come from the change of the tool used to generate it?

jmencak:
> Most of the changes look sane and match the comments in the commit message,
> just for `manifests/02-crd-profile.yaml`, `manifests/02-crd-tuned.yaml`, could you confirm that the changes (paragraph filling and order of properties) come from the change of the tool used to generate it?

Thanks for the review, much appreciated!  Yes, they come from the tool.  That was actually the point to have this fully automated.  I've also checked that the yamls semantically match.

jmencak:
/test e2e-aws-upgrade
/test e2e-aws
Fingers crossed.

jmencak:
/test e2e-aws
@kpouget this is the last test to push through the CI.  Have I covered all your concerns?  If so, could I have this PR with a LGTM label?  Thanks!

kpouget:
/lgtm
Thanks Jiri for the PR :)

Makefile
```diff
@@ -75,7 +75,8 @@ pkg/generated: $(API_TYPES)
 	touch $@
 
 crd-schema-gen:
-	$(GO) run ./vendor/github.com/openshift/crd-schema-gen/cmd/crd-schema-gen/ --apis-dir pkg/apis --manifests-dir manifests/
```
sttts, line 78: don't do this. Get the binary release. Never vendor with another set of dependencies which are untested!

manifests/02-crd-tuned.yaml
```diff
@@ -59,78 +69,66 @@ spec:
...
+                              the tuned profile connected by logical AND operator.
                             type: array
-                            items:
-                              # Recursively nested "match" sections are dropped without the following.
-                              x-kubernetes-preserve-unknown-fields: true
-                              # OpenAPI v3 validation schemas containing $ref are not permitted => cannot do recursive validation in 1.14
-                              # https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation
-                              type: object
+                            items: { type: object }
+                            x-kubernetes-preserve-unknown-fields: true
```
sttts, line 112: this looks wrong. Put `x-kubernetes-preserve-unknown-fields: true` into the items schema.

sttts: I bet kube-apiserver prunes the items schema away right now, but leaves the `x-kubernetes-preserve-unknown-fields`, leading oc to fall over. This is a bug. But this bug should never happen if you follow ^^.

Makefile
jmencak: Thanks for the comment @sttts ,  Cannot fetch the binary release in an environment that has no Internet access.  Are you suggesting I fetch the binary release and generate the CRs only in an environment with Internet access or is it ok to vendor in a specific release of "github.com/openshift/crd-schema-gen"?

jmencak:
All checks succeeded.  A way to generate CRD automatically by approved binaries will be done in a follow-up PR.

kpouget:
@sttts's concern with the YAML file has been addressed AFAIU, and his concern with the CRD generator will be addressed in a follow-up PR, so I think all the concern are resolved.

\lgtm

thanks Jiri

kpouget:
@sttts's concern with the YAML file has been addressed AFAIU, and his concern with the CRD generator will be addressed in a follow-up PR, so I think all the concern are resolved.

/lgtm

thanks Jiri

PR #156, opened 2020-09-23
Bug 1881422: Issue a warning when two or more profiles use the same priority.

jmencak:
When two or more profiles have the same priority and they match a node, the
profile selection between these used to be random.  This PR adds some determinism
to the way profile selection is done based on Tuned CR names and also issues a
warning about this to help cluster administrators to discover this problem.

### DIFF
--- a/pkg/operator/profilecalculator.go
+++ b/pkg/operator/profilecalculator.go
@@ -451,18 +451,36 @@ func (pc *ProfileCalculator) tunedsUsePodLabels(tunedSlice []*tunedv1.Tuned) boo
 func tunedRecommend(tunedSlice []*tunedv1.Tuned) []tunedv1.TunedRecommend {
 	var recommendAll []tunedv1.TunedRecommend
 
+	// Tuned profiles should have unique priority across all Tuned CRs and users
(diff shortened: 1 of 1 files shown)


pkg/operator/profilecalculator.go
```diff
@@ -464,6 +464,16 @@ func tunedRecommend(tunedSlice []*tunedv1.Tuned) []tunedv1.TunedRecommend {
 		return recommendAll[i].Priority != nil // undefined priority has the lowest priority
 	})
 
+	for i := 0; i < len(recommendAll)-1; i++ {
```
cynepco3hahue, line 467: Does `recommendAll` sorted by priority?
Like I can imagine the situation when you will have
[profile1, profile2, profile3] and
profile1.priority=30
profile2.priority=50
profile3.priority=30

jmencak: Yep, sorted by priority by the code above.

jmencak:
Thanks for the review @cynepco3hahue .  After talking to Martin S. I've decided to add some determinism to this too.

PR #162, opened 2020-10-07
Update to the latest version of stalld.

jmencak:
Shipping https://gitlab.com/rt-linux-tools/stalld.git
1d3d9329a99b8047ea0886df370de6ddf4c7f5c9


jmencak:
/cc @yanirq 
Feel free to review and /lgtm if this looks good.  Thanks!

MarSik (approved):

pkg/tuned/host_payload.go
```diff
@@ -56,7 +56,7 @@ Environment=BD="-d 3"
 # this value the time the thread must be kept ready but not
 # actually run to decide that the thread is starving
 # ex: THRESH="-t 60"
-Environment=THRESH="-t 60"
+Environment=THRESH="-t 30"
```
MarSik, line 59: I would actually add a comment here to explain the value. @fromanirh can correct me but here is a draft:

The default kernel RCU timeout value is 60 s. All cores in a machine are preempted via NMI (non-maskable interrupt) when a RCU thread is starving that long. That means stalld must react faster.

jmencak: Adjusted.

jmencak:
Trying to remove the stuck hold label.
/hold cancel

pkg/tuned/host_payload.go
fromanirh: @MarSik this is my understanding as well. Reportedly the threshold was also lowered to 20 or even 10s. This brings me to a question/RFE: can we have an option to set this value from a NTO field, letting the user (performance-operator in this case) change this value? Do I need to file an issue for this?

MarSik: I would not expose it via PAO. I think this is so low level that a MCO/Tuned override method is enough. And we can do that already (We can specify a different config/env file in our Tuned profile)

fromanirh: works for me, I was not aware of this override capability. No extra RFE needed then.

fromanirh:
looks good to me

pkg/tuned/host_payload.go
MarSik: https://github.com/openshift-kni/performance-addon-operators/blob/master/docs/configuration_hotfixes.md describes this

PR #168, opened 2020-10-26
Ship the latest version of stalld.

jmencak:
- Shipping https://git.kernel.org/pub/scm/utils/stalld/stalld.git/
    c35b62722ffebb3cf16ad6b470d9275f007f6fa2

### DIFF
--- a/assets/tuned/stalld/Makefile
+++ b/assets/tuned/stalld/Makefile
@@ -7,6 +7,9 @@ CFLAGS	:=	-Wall -O2 -g
 LDFLAGS	:=	-ggdb
 LIBS	:=	 -lpthread
 
+SRC	:=	$(wildcard src/*.c)
+HDR	:=	$(wildcard src/*.h)
+OBJ	:=	$(SRC:.c=.o)
 DIRS	:=	src redhat man
 FILES	:=	Makefile README.md gpl-2.0.txt
 TARBALL	:=	$(NAME)-$(VERSION).tar.xz
@@ -17,11 +20,18 @@ DOCDIR	:=	$(DATADIR)/doc
 MANDIR	:=	$(DATADIR)/man
 LICDIR	:=	$(DATADIR)/licenses
 
-all: src/stalld.o
-	$(CC) -o stalld	 $(LDFLAGS) src/stalld.o $(LIBS)
+.PHONY:	all tests
 
-static: src/stalld.o
-	$(CC) -o stalld-static $(LDFLAGS) --static src/stalld.o $(LIBS)
+all:	stalld
+
+stalld: $(OBJ)
+	$(CC) -o stalld	 $(LDFLAGS) $(OBJ) $(LIBS)
+
+static: $(OBJ)
+	$(CC) -o stalld-static $(LDFLAGS) --static $(OBJ) $(LIBS)
+
+tests:
+	make -C tests
 
 .PHONY: install
 install:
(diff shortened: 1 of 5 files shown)


jmencak:
/hold
until it is confirmed this is the right version to ship

jmencak:
Daniel confirmed this is the right version of `stalld` to use.
/hold cancel
@courtneypacheco  / @sjug  PTAL

assets/tuned/stalld/Makefile
```diff
@@ -17,11 +20,18 @@ DOCDIR	:=	$(DATADIR)/doc
...
+all:	stalld
+
+stalld: $(OBJ)
+	$(CC) -o stalld	 $(LDFLAGS) $(OBJ) $(LIBS)
+
+static: $(OBJ)
+	$(CC) -o stalld-static $(LDFLAGS) --static $(OBJ) $(LIBS)
+
+tests:
+	make -C tests
```
sjug, line 34: We don't have tests as part of this PR so `make tests` wouldn't actually work?

```
$ make tests
make -C tests
make[1]: *** tests: No such file or directory.  Stop.
make: *** [Makefile:34: tests] Error 2
```

assets/tuned/stalld/Makefile
```diff
@@ -33,14 +43,16 @@ install:
 	$(INSTALL) -m 755 -d $(DESTDIR)$(LICDIR)/$(NAME)
 	$(INSTALL) gpl-2.0.txt -m 644 $(DESTDIR)$(LICDIR)/$(NAME)
 
+
```
sjug, line 46: Extra newline?

assets/tuned/stalld/Makefile
```diff
@@ -33,14 +43,16 @@ install:
...
 .PHONY: clean tarball redhat push
 clean:
 	@test ! -f stalld || rm stalld
 	@test ! -f stalld-static || rm stalld-static
 	@test ! -f src/stalld.o || rm src/stalld.o
 	@test ! -f $(TARBALL) || rm -f $(TARBALL)
 	@make -C redhat clean
```
sjug, line 53: Not happy with this redhat line

```
$ make clean
make[1]: *** redhat: No such file or directory.  Stop.
make: *** [Makefile:53: clean] Error 2
```

sjug (commented):
Builds well, some other targets aren't happy.

assets/tuned/stalld/Makefile
jmencak: That's right, tests do not ship with NTO to keep only the bare minimum inside.  Testing of stalld is done in the upstream repo.

assets/tuned/stalld/Makefile
jmencak: Issue/PR against the upstream repo?

assets/tuned/stalld/Makefile
sjug: Ah are you just merging in their artifacts?

assets/tuned/stalld/Makefile
jmencak: Again, NTO ships with the bare minimum to make `stalld` in NTO.

assets/tuned/stalld/Makefile
jmencak: Yes.  The only change I did is to remove the "tests" from the Makefile.  I should've probably changed the three Dockerfile(s) to do "make stalld" instead of just "make".  Tests were not part of old versions of stalld, so I just adjusted the makefile.

sjug: Ack, so we want to keep it the same for ease of patching in the future?

jmencak: We can, so you want me to adjust the three Dockerfiles?  The only reason I didn't do this in the past was adjustments in three places and the fact the target `stalld` was changing a lot.  But I guess this is the last name (starvation_monitor/starved/stalld...).

sdodson:
/retitle Bug 1892457: Ship the latest version of stalld

PR #169, opened 2020-10-27
Set scheduling policy to SCHED_FIFO to stalld and lower threshold.

jmencak:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1887568#c26

### DIFF
--- a/pkg/tuned/host_payload.go
+++ b/pkg/tuned/host_payload.go
@@ -58,7 +58,7 @@ Environment=BD="-d 3"
 # ex: THRESH="-t 60"
 # The default kernel RCU timeout value is 60s.  All available cores are preempted via a NMI when an RCU thread is starving that long.
 # This means stalld must react faster.
-Environment=THRESH="-t 30"
+Environment=THRESH="-t 20"
 
 # Logging options
 #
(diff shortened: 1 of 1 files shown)


pkg/tuned/host_payload.go
```diff
@@ -79,7 +79,7 @@ Environment=FG=--foreground
 # ex: PF=--pidfile /run/stalld.pid
 Environment=PF="--pidfile /run/stalld.pid"
 
-ExecStart=/usr/local/bin/stalld $CLIST $AGGR $BP $BR $BD $THRESH $LOGGING $FG $PF
+ExecStart=chrt -f 10 /usr/local/bin/stalld $CLIST $AGGR $BP $BR $BD $THRESH $LOGGING $FG $PF
```
kpouget, line 82: nit: `chrt` or `/usr/bin/chrt`?

> ExecStart=
> > For each of the specified commands, the first argument must be either an absolute path to an executable or a simple file name without any slashes.

https://www.freedesktop.org/software/systemd/man/systemd.service.html

maybe it used not to be allowed to rely on `$PATH`, or maybe I have something wrong in mind!
but looking at `/usr/lib/systemd/system`, the absolute path seems to be the common usage ... except for systemd itself!

kpouget (commented):
found a small nit, see the comment above; the rest of the patch follows the recommendation form https://bugzilla.redhat.com/show_bug.cgi?id=1887568#c26 to launch `stalld`, so looks good

thanks @jmencak for the PR!

pkg/tuned/host_payload.go
jmencak: Yeah, good point. I was thinking about this when I did the PR.  I decided not to use the absolute path just in case the binary is moved somewhere else to allow for more flexibility.  However, chances it moves are actually pretty slim.  If you think absolute path is better, I don't have a problem with this.

kpouget:
/lgtm

thanks Jiri!

sdodson:
/retitle Bug 1892459: Set scheduling policy to SCHED_FIFO to stalld and lower threshold

sdodson:
/retitle Bug 1892457: Set scheduling policy to SCHED_FIFO to stalld and lower threshold

PR #173, opened 2020-11-03
Change the priorities of example and e2e test profiles.

jmencak:
The default Tuned CR ships with two profiles with priority 30 and 40.  While the application of profiles with the same priority is now deterministic, using the same priority (30) in the examples and e2e tests is not a good idea.  Increase the priority of these profiles by lowering the priority number to 20 (and 10 where needed).

### DIFF
--- a/examples/cpu-partitioning.yaml
+++ b/examples/cpu-partitioning.yaml
@@ -18,5 +18,5 @@ spec:
   recommend:
   - match:
     - label: node-role.kubernetes.io/worker-rt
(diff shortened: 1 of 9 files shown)


ArangoGutierrez (approved):
More info at [custom-tuning-specification](https://docs.openshift.com/container-platform/4.6/nodes/nodes/nodes-node-tuning-operator.html#custom-tuning-specification_nodes-node-tuning-operator)
/lgtm

PR #175, opened 2020-11-10
Bug 1895919: Add a weak dependency on kmod to tuned.

jmencak:
This fixes an issue in the tuned daemon with a missing
dependency on the kmod package.  The `[modules]` plugin uses
modprobe from the kmod package to load kernel modules.

Other changes:
  - Added e2e tests to test this functionality for the loop kernel
    module.
  - Reused PollExecCmdInPod() for GetFileInPod()
  - Removed an unused constant from stalld.go e2e test.

### DIFF
--- a/assets/tuned/patches/090-kmod-weak-dependency.diff
+++ b/assets/tuned/patches/090-kmod-weak-dependency.diff
@@ -1,0 +1,16 @@
+Fixes issue #304 with modprobe being missed by the [modules] plugin.
+
+See: https://github.com/redhat-performance/tuned/pull/305
+
+diff --git a/tuned.spec b/tuned.spec
+index f28e7b58..2a3fc6bf 100644
+--- a/tuned.spec
++++ b/tuned.spec
+@@ -88,6 +88,7 @@ Requires: util-linux, dbus, polkit
+ Recommends: hdparm
+ Recommends: kernel-tools
+ Recommends: python-dmidecode
++Recommends: kmod
+ %endif
+ %if 0%{?rhel} > 7
+ Requires: python3-syspurpose
--- a/test/e2e/basic/modules.go
+++ b/test/e2e/basic/modules.go
@@ -1,0 +1,79 @@
+package e2e
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+
+	coreapi "k8s.io/api/core/v1"
+
+	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
+	util "github.com/openshift/cluster-node-tuning-operator/test/e2e/util"
+)
+
+// Test loading a kernel module by applying a custom profile via node labelling.
+var _ = ginkgo.Describe("[basic][modules] Node Tuning Operator load kernel module", func() {
+	const (
+		profileModules   = "../testing_manifests/tuned_modules_load.yaml"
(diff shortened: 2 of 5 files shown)


kpouget (commented):
thanks for the PR Jiri! 1 line of "patch payload", >80 lines to test it ;-)

Looks good to me overall!

I'm just wondering about the "false negative" of the e2e test, I guess you double checked that the `loop` module is not used by OCP, at least not in the CI environment?

and, does removing the profile means removing the module?

jmencak:
> thanks for the PR Jiri! 1 line of "patch payload", >80 lines to test it ;-)
Thank you for the quick review!  Yeah, exactly a lot more lines to test...

> 
> Looks good to me overall!
> 
> I'm just wondering about the "false negative" of the e2e test, I guess you double checked that the `loop` module is not used by OCP, at least not in the CI environment?
It wasn't used in the CI environment.  I tried to choose a reasonably harmless module that can be removed (unlike fuse) without issues.  If you have a better tip, I'm open to suggestions.

> 
> and, does removing the profile means removing the module?
No, it does not.  While the `tuned` daemon mostly does rollback of the changes this is one of the exceptions I know about.  I believe this is reasonable given the nature of the changes (loading kernel modules).  However, if this feels like a bug, please open an issue in the upstream `tuned` and we can discuss it there.

kpouget:
thanks for the precisions, makes sense.

The `loop` module seems to be a good candidate indeed if it's not used in the CI. I see it's used by 12 "guys" in my workstation, but I don't know who they are :# We'll change it later if ever OCP got to use it!
```
# lsmod | egrep  'Module|loop'
Module                  Size  Used by
loop                   40960  12
```
/lgtm

jmencak:
Changed the module `loop` to `joydev` and Reused `PollExecCmdInPod()` for `GetFileInPod()`.

kpouget (approved):
The core of the patch is the addition of the `kmod` dependency in `tuned.spec`, so that kernel modules can be properly loaded.

the rest of the patch is a `e2e` test loading the `joydev` (a module that should never be used in an OpenShift / RHCOS context) with the NTO,

and a small modification of the `e2e` utility functions: `GetFileInPod(pod *corev1.Pod, file string)` is transformed into a more generic function:
```
PollExecCmdInPod(interval, duration time.Duration, pod *corev1.Pod, args ...string)
```

\lgtm

thanks again Jiri!

kpouget:
> \lgtm

of course I meant 

/lgtm

silly me!

jmencak:
The e2e-aws-operator seems to continuously fail with 
```
request.go:645] Throttling request took 1.073791679s, request: GET:https://api.ci-op-1dy5g69w-0519c.origin-ci-int-aws.dev.rhcloud.com:6443/apis/operators.coreos.com/v1alpha1?timeout=32s\nI1111 06:39:46.372776    7891 request.go:645]
```

/shrug

jmencak:
/hold
This looks as if the stderr and stdout is somehow being mixed in the e2e tests.

jmencak:
There was an issue with e2e tests that has gone unnoticed for several releases with `.CombinedOutput()` being used in e2e tests.  As a result, when the "oc" command sent something on the stderr, some e2e tests could fail.  Sending messages on stderr seems more frequent now, so the issue was noticed and (hopefully) fixed.

/shrug cancel
/hold cancel

jmencak:
@courtneypacheco PTAL, thank you!

courtneypacheco (approved):
/lgtm

PR #183, opened 2020-11-25
E2e test cleanup.

jmencak:
Changes:
  - removing code duplication in `test/e2e/util/util.go`
  - documentation for functions in `test/e2e/util/util.go`
  - more expressive function names where `poll.Wait()` is used
  - more flexibility in terms of specifying the interval and
    timeout for `poll.Wait()`
  - for two e2e tests involving reboots removed a check for the
    original kernel command line parameters as the reboots can now
    cause changes not only to "ostree", but also to "root",
    "rhcos.root", "rw";  this would be difficult to maintain in the
    future
  - adhere more to k8s naming conventions ([Pp]od, ...)

PR #184, opened 2020-11-26
Updating OWNERS file

jmencak:
Disscussed this with @Deepthidharwar

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -4,7 +4,6 @@ approvers:
 reviewers:
   - ArangoGutierrez
   - cep235
-  - Deepthidharwar
   - jmencak
   - sjug
   - zvonkok


jmencak:
Since the recent switch to RHCOS 8.3 (and associated changes with kernel boot parameters after boot), e2e tests will not pass until https://github.com/openshift/cluster-node-tuning-operator/pull/183 is merged.  Putting on hold until then.
/cc @courtneypacheco 
/hold

OWNERS
```diff
@@ -4,7 +4,6 @@ approvers:
 reviewers:
   - ArangoGutierrez
   - cep235
-  - Deepthidharwar
```
ArangoGutierrez, line 7: Would it make sense also to rm from https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-node-tuning-operator/OWNERS#L11 ?

jmencak: Thanks for looking Eduardo.  It would make sense, but if you read the first two lines in that file, it says the file is autogenerated and that we should not edit it.  It is autogenerated presumably from the OWNERS files in this repo.  So I believe once this PR merges, that file gets updated too.

PR #185, opened 2020-12-01
Annotate manifests for single-node-developer cluster profile

cfergeau:
This partially implements phase 1 of https://github.com/openshift/enhancements#482
and does not change behavior. Initially, all cluster-node-tuning-operator
manifests are included in the single-node-developer cluster profile.
Follow-on PRs may exclude any of these that are not needed in the
profile.

### DIFF
--- a/manifests/10-namespace.yaml
+++ b/manifests/10-namespace.yaml
@@ -4,6 +4,7 @@ metadata:
   annotations:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
   labels:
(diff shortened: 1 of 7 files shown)


ArangoGutierrez (approved):
After reading https://github.com/openshift/enhancements/pull/482
/lgtm

jmencak:
Thank you for the PR, @cfergeau .  Only a minor nit.  It would be great if the annotations could be in alphabetical order, but I can clean this up later.

cfergeau:
> Thank you for the PR, @cfergeau . Only a minor nit. It would be great if the annotations could be in alphabetical order, but I can clean this up later.

That's a fair point as I've been paying attention to this in other PRs, I missed that one. This is fixed now.

jmencak:
Thanks again!
/lgtm

PR #188, opened 2020-12-11
Add status conditions and profile applied to Profile(s)

jmencak:
Changes:
  - report Tuned profile currently applied for each of the containerized
    Tuned daemon managed by NTO
  - report two Profile status conditions `Applied` and `Degraded`
    in every Profile indicating whether the Tuned profile was applied and
    whether there were issues during the profile application
  - cleanup of the ClusterOperator settings code; ClusterOperator now also
    reports Reason == ProfileDegraded for the Available condition if any of
    the Tuned Profiles failed to be applied cleanly for any of the  containerized
   Tuned daemons managed by NTO
  - e2e test added to check the status reporting functionality
  - e2e basic/available test enhanced to check for not Degraded condition
  - using `podman build --no-cache` now.  This works around issues such as:
    https://github.com/containers/buildah/issues/2837


jmencak:
/cc @marcel-apf 
Marcel, please review from your perspective whether the newly created statuses fit your needs.  Many thanks!

jmencak:
@MarSik FYI

pkg/apis/tuned/v1/tuned_types.go
```diff
@@ -144,13 +146,65 @@ type ProfileStatus struct {
...
+
+	// reason is the CamelCase reason for the condition's current status.
+	// +optional
+	Reason string `json:"reason,omitempty"`
+
+	// message provides additional information about the current condition.
+	// This is only to be consumed by humans.
+	// +optional
+	Message string `json:"message,omitempty"`
+}
```
marcel-apf, line 190: @jmencak looks good, I think we better merge it and consume it with PAO, and then we can go from there. We will have a better visibility and we can come up with improvement ideas, if needed.

jmencak: Thanks, @marcel-apf .  I'll aim to merge this immediately once 4.8 opens.  @sjug/@dagrayvid , could you please provide a code review?  Thank you!

pkg/tuned/tuned.go
```diff
@@ -109,6 +91,35 @@ type wqKey struct {
 	event string // object event type (add/update/delete) or pass the full object on delete
 }
 
+// Constants
+const (
+	// Constants used for instantiating Profile status conditions
```
dagrayvid, line 96: It might be nice to explain the following three lines in this comment, or at least say that they will be set to 1,2,4 respectively

dagrayvid (commented):
Except for the previous comment, the code changes look good to me.

pkg/tuned/tuned.go
jmencak: A fair point, comments added.  Thank you!

yanirq:
LGTM

jmencak:
/hold
Bot, go easy with the retests, waiting for https://github.com/openshift/cluster-node-tuning-operator/pull/195 to merge first. Then I'll rebase.

pkg/tuned/status.go
```diff
@@ -0,0 +1,127 @@
...
+// old conditions.
+func computeStatusConditions(status Bits, conditions []tunedv1.ProfileStatusCondition) []tunedv1.ProfileStatusCondition {
+	tunedProfileAppliedCondition := tunedv1.ProfileStatusCondition{
+		Type: tunedv1.TunedProfileApplied,
+	}
+	tunedDegradedCondition := tunedv1.ProfileStatusCondition{
+		Type: tunedv1.TunedDegraded,
+	}
+
+	if (status & scApplied) != 0 {
```
yanirq, line 99: where are these coming from  ?  (`status, scApplied, scError, scWarn`)

sjug: https://github.com/openshift/cluster-node-tuning-operator/pull/196/files

jmencak: As pointed out by Sebastian above, I "backported" part of this PR to fix rhbz1919970.

yanirq:
Degrading the cluster operator (NTO operator) if we get even 1 degraded profile status seems a bit harsh. 
This can cause upgrade issues (@jmencak)  but not only. We could have a cluster with multiple nodes having different labels, most of them can run fine with their tuned settings but if we hit one profile in a degraded state this halt NTO completely. 
Another thing is, introducing yet anther degraded state might cause end users to start suffering from it and needing to fix it since it can pose as a blocker issue for them.

Looking from Performance addons operator perspective - I think we should consume the tuned object status anyway if we want a 1:1 report for what might be wrong with performance profiles applied. 
@MarSik  @fromanirh  @cynepco3hahue   agree/disagree ?

jmencak:
> Degrading the cluster operator (NTO operator) if we get even 1 degraded profile status seems a bit harsh.
> This can cause upgrade issues (@jmencak) but not only. We could have a cluster with multiple nodes having different labels, most of them can run fine with their tuned settings but if we hit one profile in a degraded state this halt NTO completely.
> Another thing is, introducing yet anther degraded state might cause end users to start suffering from it and needing to fix it since it can pose as a blocker issue for them.

I completely agree.  If you believe that reporting the "error" status per Profile is enough, I'm more than happy not to touch the `OperatorStatus`.

> Looking from Performance addons operator perspective - I think we should consume the tuned object status anyway if we want a 1:1 report for what might be wrong with performance profiles applied.
> @MarSik @fromanirh @cynepco3hahue agree/disagree ?

yanirq:
> > Degrading the cluster operator (NTO operator) if we get even 1 degraded profile status seems a bit harsh.
> > This can cause upgrade issues (@jmencak) but not only. We could have a cluster with multiple nodes having different labels, most of them can run fine with their tuned settings but if we hit one profile in a degraded state this halt NTO completely.
> > Another thing is, introducing yet anther degraded state might cause end users to start suffering from it and needing to fix it since it can pose as a blocker issue for them.
> 
> I completely agree. If you believe that reporting the "error" status per Profile is enough, I'm more than happy not to touch the `OperatorStatus`.

@jmencak   ack, I think we can remove degration of the cluster operator then.  We can keep status reporting maybe under the operator itself without degrading it.

jmencak:
> @jmencak ack, I think we can remove degration of the cluster operator then. We can keep status reporting maybe under the operator itself without degrading it.

Done.

yanirq:
@jmencak   only thing missing here is the requested bug label

jmencak:
> @jmencak only thing missing here is the requested bug label

The bug label would be needed if we wanted this in 4.7.  For 4.8, the bug label is not necessary.  As soon as 4.8 opens, this will merge.  And it was exactly the plan to have this ready very early on for 4.8.

jmencak:
Let's merge this:
/test all

PR #189, opened 2020-12-14
Bug 1907936: Switch to nto_profile_calculated_total.

jmencak:
Previously, the metric `nto_profile_set_total` was used to indicate the number of times a profile was "set" for a particular node.  This was inaccurate, mostly due to the lack of reporting a profile applied on a given node.  Change to metric name and the code to report the metric accurately.


jmencak:
@skordas  PTAL and let me know if this makes more sense.  Thank you!

kpouget:
the code looks good to me, I don't have comments after looking at the patch, thanks Jiri !

PR #195, opened 2021-01-26
Bug 1920438: Fix openshift-tuned panic on turning debugging on/off.

jmencak:
Fixes openshift-tuned panic on turning debugging on/off introduced by:
https://github.com/openshift/cluster-node-tuning-operator/pull/192


ArangoGutierrez (approved):
After reading the logic behind https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/tuned/tuned.go#L384 this patch makes sense to me
/lgtm

PR #200, opened 2021-02-02
Bug 1924081: Ship the latest tuned-2.15.0

jmencak:
Synchronize with latest upstream tuned release.


assets/tuned/daemon/tuned/plugins/plugin_bootloader.py
```diff
@@ -154,6 +164,25 @@ def _grub2_default_env_patch(self):
...
 
+	def _grub2_default_env_unpatch(self):
+		grub2_default_env = self._cmd.read_file(consts.GRUB2_DEFAULT_ENV_FILE)
+		if len(grub2_default_env) <= 0:
+			log.info("cannot read '%s'" % consts.GRUB2_DEFAULT_ENV_FILE)
+			return False
+
+		write = False
+		if re.search(r"^GRUB_CMDLINE_LINUX_DEFAULT=\"\$\{GRUB_CMDLINE_LINUX_DEFAULT:\+\$GRUB_CMDLINE_LINUX_DEFAULT \}\\\$" +
+			consts.GRUB2_TUNED_VAR + "\"$", grub2_default_env, flags = re.MULTILINE):
```
kpouget, line 175: just a nit, but maybe you can move the regex definition outside of the `if re.search()` call, so that the `if cond` stays in one line, that may make it a bit easier to read (and same at a few other locations in the patch)

assets/tuned/daemon/tuned/plugins/plugin_scheduler.py
```diff
@@ -83,6 +84,18 @@ def __init__(self, monitor_repository, storage_factory, hardware_inventory, devi
...
 
+	def _calc_mmap_pages(self, mmap_pages):
+		if mmap_pages is None:
+			return None
+		try:
+			mp = int(mmap_pages)
+		except ValueError:
+			return 0
+		if mp <= 0:
+			return 0
```
kpouget, line 95: no warning on parsing error?
(just asking, I have no clue where the warnings would end up :#)

kpouget: hum, forget, `0` is used then to trigger the warning!

assets/tuned/daemon/tuned/plugins/plugin_bootloader.py
jmencak: So, NTO takes the tuned daemon from here: https://github.com/redhat-performance/tuned  as is.  If you feel this should be fixed, please create an issue in that project or I can.  Up to you. :)  But NTO ships the tuned daemon as is and only supplies its own patches in the `assets/tuned/patches` directory as needed.

kpouget:
/lgtm

the comment I made above should go to the `tuned` repository, not here :#
(if you want to report it there Jiri, feel free)

PR #201, opened 2021-02-04
Bug 1925072: Updating to the latest stalld v1.7.0.

jmencak:
Changes:
  - updated https://gitlab.com/rt-linux-tools/stalld.git to v1.7.0
    (9ed8ff901c378a1e17041411e78d14b0fa9ed64e)
  - included throttlectl.sh from the repo above to disable/enable
    RT throttling reliably.  The disable of RT throttling was moved
    into the script to run before stalld is started and the enabling
    of it to be run in the script after the stalld service is stopped.

### DIFF
--- a/pkg/tuned/host_payload.go
+++ b/pkg/tuned/host_payload.go
@@ -11,19 +11,16 @@ import (
 	ign3types "github.com/coreos/ignition/v2/config/v3_1/types"
 )
 
-func stalldIgnitionFile() ign3types.File {
-	const stalldPath = "/usr/local/bin/stalld"
-	mode := 0755
-
-	stalldBytes, err := ioutil.ReadFile(stalldPath)
+func ignitionFile(path string, mode int) ign3types.File {
+	stalldBytes, err := ioutil.ReadFile(path)
 	if err != nil {
-		klog.Errorf("failed to read %s: %v", stalldPath, err)
+		klog.Errorf("failed to read %s: %v", path, err)
 		return ign3types.File{}
 	}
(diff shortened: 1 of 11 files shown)


pkg/tuned/host_payload.go
```diff
@@ -94,7 +93,8 @@ func ProvideIgnitionFiles(stalld bool) []ign3types.File {
 	files := []ign3types.File{}
 
 	if stalld {
-		files = append(files, stalldIgnitionFile())
+		files = append(files, ignitionFile("/usr/local/bin/stalld", 0755))
+		files = append(files, ignitionFile("/usr/local/bin/throttlectl.sh", 0755))
```
yanirq, line 97: what will happen in upgrade if the file needs to change/remove ?

yanirq:
any additions might be needed to the testing scheme currently available ?

pkg/tuned/host_payload.go
jmencak: Those files are automatically wiped from the host on a profile change if/when `stalld` is not needed by a selected profile.  During an upgrade/profile change the operator will compare the old/new ignition files and update them accordingly.  I don't believe there is a problem here, is there?

jmencak:
> any additions might be needed to the testing scheme currently available ?

Probably not at the moment.  For 4.8, I'd probably like to add some extra e2e tests.  Do you have any in mind?

pkg/tuned/host_payload.go
yanirq: no, just double checking

jmencak:
In other words, I believe the existing stalld test checking the `pidof` is probably fine at the moment.

PR #202, opened 2021-02-05
Bug 1925072: Remove architecture specific gcc flags.

jmencak:
The `-mtune=generic` is IA32/AMD64/EM64T processor specific.  Also, with `-fcf-protection` GCC fails to build `stalld` on other architectures such as ppc64/s390x.


kpouget:
/lgtm

I wonder why they didn't catch it [upstream]{https://gitlab.com/rt-linux-tools/stalld/-/blob/main/Makefile}, weird!

PR #209, opened 2021-02-11
stalld: Remove another architecture specific gcc flag

yselkowitz:
aarch64 is not a multilib architecture, therefore -m64 is not accepted.
It is already the default on the 64-bit multilib-enabled architectures,
so there is really no need for it anyway.

/cc @jmencak


jmencak (approved):
Thank you for the PR.
/lgtm

PR #211, opened 2021-02-22
Bug 1926903: Keep ignition units in sync with [service] plugin.

jmencak:
Changes:
  - new vendor dependency 'gopkg.in/ini.v1' to parse INI files
  - openshift-tuned now parses Tuned profiles and searches for
    'stalld' enabled/disabled by [service] plugin; systemd
    ignition file units are now created with this enablement
    information
  - removed the patch to disable the tuned [service] enablement
    functionality; the patch is no longer necessary as the
    ignition unit enablement is consistent with the enable/disable
    spec inside the tuned profile
  - fix e2e tests for default_irq_smp_affinity on platforms
    with wider affinity masks than usual (e.g. Xen).


jmencak:
@cynepco3hahue , all tests passed now. This should be now how you proposed the fix originally.

PR #212, opened 2021-02-24
Do not manupulate BLS entries and grub config

jmencak:
NTO currently supports the bootloader plugin only on RHCOS via the
integration with MCO.  Disable patching of BLS entries and grub
configuration files so that error messages from the tuned daemon are
eliminated and their presence not reported further on in k8s objects.


jmencak:
@dagrayvid / @sjug PTAL, thank you.

PR #214, opened 2021-03-03
Add e2e test for adjusting netdev queue count.

jmencak:
Changes:
  - added e2e test to for setting the number of multi-purpose channels
    of network devices
  - be consistent when naming e2e tests and test manifests


test/e2e/basic/netdev_set_channels.go
```diff
@@ -0,0 +1,163 @@
...
+				}
+			}
+
+			if phyDev == "" {
+				util.Logf(fmt.Sprintf("no network devices supporting querying channels found on node %s; skipping test", node.Name))
+				return
+			}
+
+			cmdCombinedChannelsCurrent := []string{"bash", "-c", "ethtool -l " + phyDev + "| sed -n '/Current hardware settings/,/Combined:/{s/^Combined:\\s*//p}'"}
+			cmdCombinedChannelsMax := []string{"bash", "-c", "ethtool -l " + phyDev + "| sed -n '/Pre-set maximums/,/Combined:/{s/^Combined:\\s*//p}'"}
```
kpouget, line 86: one thing I like to do when doing such parsing is to add a comment showing how the text should look like
none of my devices support `ethtool -l`, so I can't double check what's happening here 
something like:
```
     /* sample output of `ethertool -l`:
     ..
     .. only what's relevant ..
     ..
     */
```
(just my 2c)

kpouget: and, in the lines below you do:
```
_, err = util.ExecCmdInPod(pod, "bash", "-c", fmt.Sprintf("ethtool -L %s combined 1", phyDev))
```
which seems a bit easier to read than `"ethtool -l " + phyDev + "...`, perhaps you could unify the string processing, and maybe split the long lines above to help the readability (new line after `-c` and before the `|`)

test/e2e/basic/netdev_set_channels.go
```diff
@@ -0,0 +1,163 @@
...
+			out, err := util.ExecCmdInPod(pod, cmdCombinedChannelsMax...)
+			channelMaximumCombined, err := strconv.Atoi(strings.TrimSpace(out))
+			if err != nil || channelMaximumCombined <= 1 {
+				util.Logf(fmt.Sprintf("network device %s does not seem to support querying channels [%v] on node %s; skipping test: %v",
+					phyDev, channelMaximumCombined, node.Name, err))
+				return
+			}
+
+			out, err = util.ExecCmdInPod(pod, cmdCombinedChannelsCurrent...)
+			channelOrigCombined, err := strconv.Atoi(strings.TrimSpace(out))
```
kpouget, line 98: if `channelOrigCombined == chLen` the should be skipped I guess

kpouget (commented):
thanks for the PR Jiri, I looked at the test and the code makes sense to be, I couldn't find any unexpected

the only thing is that the test may run into false positive if the original channel length is equal to the one you request, so that can be caught already and logged

the other command is bike-shedding about the readability of a few lines, adjust as you prefer!

test/e2e/basic/netdev_set_channels.go
jmencak: I don't think it is needed.  I'm lowering the channel to "1" down below. What is missing though is a check it was really lowered to 1, otherwise one might get a false positive.  Will fix, thanks!

test/e2e/basic/netdev_set_channels.go
jmencak: Agreed, will fix.

jmencak:
> thanks for the PR Jiri, I looked at the test and the code makes sense to be, I couldn't find any unexpected

Thanks for the review, very valuable comments!  Hopefully now all resolved.

test/e2e/basic/netdev_set_channels.go
kpouget: indeed, makes sense:
1. get the original value
2. manually set the current value to 1
3. use NTO to set the value to 2
4. wait for the value to become 2
5. set the original value back

and what's missing is 2.5/ wait for the value to become 1

jmencak: Correct and fixed (2.5) by the latest commit I hope.  You don't really need to "wait" (as in poll) though.  At least I hope these settings (by calling `ethtood` manually) take effect immediately.  If not, the test will fail.

kpouget:
thanks Jiri,
\lgtm

PR #217, opened 2021-03-23
Add "core" e2e tests.

jmencak:
The core e2e tests are checking the ability to retrieve:
  - host and container OS release
  - cluster version

This information is also useful for debugging e2e test failures in cases the developer has no access to the cluster e2e tests are running on.

### DIFF
--- a/test/e2e/core/cluster_version.go
+++ b/test/e2e/core/cluster_version.go
@@ -1,0 +1,45 @@
+package e2e
+
+import (
+	"fmt"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+
+	coreapi "k8s.io/api/core/v1"
+
+	util "github.com/openshift/cluster-node-tuning-operator/test/e2e/util"
+)
+
+// Test the core functionality by reporting host, container OS and cluster version.
+var _ = ginkgo.Describe("[core][cluster_version] Node Tuning Operator host, container OS and cluster version", func() {
+	var (
+		node *coreapi.Node
+	)
+
+	ginkgo.It("host, container OS and cluster version retrievable", func() {
+		ginkgo.By("getting a list of worker nodes")
+		nodes, err := util.GetNodesByRole(cs, "worker")
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		gomega.Expect(len(nodes)).NotTo(gomega.BeZero(), "number of worker nodes is 0")
+
+		node = &nodes[0]
+		ginkgo.By(fmt.Sprintf("getting a Tuned Pod running on node %s", node.Name))
+		pod, err := util.GetTunedForNode(cs, node)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		ginkgo.By(fmt.Sprintf("getting the host OS version on node %s", node.Name))
+		out, err := util.ExecCmdInPod(pod, "cat", "/host/etc/os-release")
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		util.Logf("%s", out)
+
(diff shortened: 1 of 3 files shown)


jmencak:
/hold
until the CI is fixed

jmencak:
@dagrayvid PTAL, thank you!

test/e2e/core/cluster_version.go
```diff
@@ -0,0 +1,40 @@
...
+		pod, err := util.GetTunedForNode(cs, node)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		ginkgo.By(fmt.Sprintf("getting the host OS version on node %s", node.Name))
+		out, err := util.ExecCmdInPod(pod, "cat", "/host/etc/os-release")
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		util.Logf("%s", out)
+
+		ginkgo.By("getting the cluster version")
+		_, _, err = util.ExecAndLogCommand("oc", "get", "clusterversion", "--no-headers")
```
kpouget, line 37: just a nit, why `--no-headers`? 
I think the command only logs this info and it's no parsed, so maybe better keeping the headers to understand what the booleans mean :)

```
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.7.0     True        False         3h21m   Error while reconciling 4.7.0: the cluster operator ingress is degraded
```

jmencak: The reason I dropped it was that it was not aligned nicely. Perhaps devels are really only interested in the real clusterversion, such as 
```
oc get clusterversion -o jsonpath='{.status.desired.version}{"\n"}' version
```
would that make sense?

kpouget: indeed, makes sense

just FYI, I use this command for getting the version, but I don't know if it may give something different than yours:
```
oc version -o json | jq --raw-output '.openshiftVersion'
```
(unfortunately, `jq` is mandatory because `oc version` doesn't support `jsonpath` output :#)

dagrayvid:
> @dagrayvid PTAL, thank you!

Thanks Jiri. Looks like Kevin beat me to it but this looks good to me as well.

jmencak:
/hold 
thanks guys, I'll give you a chance for another review, forgot to add container image info, that will also be useful

jmencak:
/hold cancel
Sorry for the noise, should be complete now.

kpouget:
/lgtm

interesting that you fetch the host OS from `/host/etc/os-release`, I wonder if the GPU Operator could do it as well
(the upgrade is currently broken because the RHEL version is passed via a never-updated environment variable)

jmencak:
> interesting that you fetch the host OS from `/host/etc/os-release`, I wonder if the GPU Operator could do it as well
> (the upgrade is currently broken because the RHEL version is passed via a never-updated environment variable)

The entire host's filesystem is mounted under `/host`, see https://github.com/openshift/cluster-node-tuning-operator/blob/8225b610ff4ee5d0f9ccbdf6d6199939dbc36d29/assets/tuned/manifests/ds-tuned.yaml#L109

dagrayvid:
> > interesting that you fetch the host OS from `/host/etc/os-release`, I wonder if the GPU Operator could do it as well
> > (the upgrade is currently broken because the RHEL version is passed via a never-updated environment variable)
> 
> The entire host's filesystem is mounted under `/host`, see
> 
> https://github.com/openshift/cluster-node-tuning-operator/blob/8225b610ff4ee5d0f9ccbdf6d6199939dbc36d29/assets/tuned/manifests/ds-tuned.yaml#L109

I had this same thought about SRO for a moment. It would make some things easier if we always mounted the host filesystem in our containers ;)

kpouget:
\o/

PR #219, opened 2021-03-25
More precise description of MCP matching.

jmencak:
The documentation was not precise when it came to two or more
MachineConfigPools with nodeSelector matching the same node(s).  This
can happen (e.g. when a node has both master and worker role), but a
single node will always be assigned (a member of) only one or no
MachineConfigPool.


jmencak:
/hold
Use this PR as a test PR to test the current CI changes: https://github.com/openshift/release/pull/17048

jmencak:
/hold cancel
/retest
See if https://github.com/openshift/release/pull/17590 really works.

PR #221, opened 2021-03-31
TuneD: udev: explicitly sort udev properties

jmencak:
Although udev properties are sorted by libudev, according
to libudev/systemd upstream this behavior cannot be relied
upon. Sort them explicitly for consistent behavior.

See upstream TuneD patch:
https://github.com/redhat-performance/tuned/pull/333


yanirq:
LGTM 
(I dont think the upgrade lane failures are related here)

yanirq:
/lgtm
(I dont think the upgrade lane failures are related here)

PR #226, opened 2021-04-27
Bug 1963848: Use OS-shipped stalld vs. the NTO-shipped one.

jmencak:


jmencak:
/hold
Do not merge until the CoreOS shipped `stalld` is at least at version `1.7`. Currently it is at `1.6` and it is broken on CoreOS, therefore the e2e tests will not pass.

kpouget:
seems safe,

just wondering, this :arrow_down:  is a constant, so it cannot change, right? then why guard `if ! useSystemStalld {}` instead of removing it?
```
useSystemStalld = true
```

jmencak:
> seems safe,
> 
> just wondering, this arrow_down is a constant, so it cannot change, right? then why guard `if ! useSystemStalld {}` instead of removing it?
> 
> ```
> useSystemStalld = true
> ```

Thank you for looking Kevin.  My intention was to leave the legacy (shipping custom `stalld` binary and associated units) in place for some time and have a simple toggle to go back to the functioning code just in case an emergency change to a more recent `stalld` not shipped in RHCOS is needed. In that case, you'll just toggle `useSystemStalld = false`. Once we see the code is no longer needed, my intention was to do a full cleanup, not just the section you suggest.  So this is a "mental note" to know what to change. Perhaps I should add a comment above "useSystemStalld = true" to explain this.

kpouget:
yes, make sense to keep the safety net just in case :)

and yes, a comment above the flag could make it clear in the code as well, it doesn't cost much :)

kpouget:
ack, thanks,
now waiting for RHCOS to update `stalld` then!

jmencak:
> ack, thanks,
> now waiting for RHCOS to update `stalld` then!

Exactly, will do upgrade testing and cancel the hold once it lands. Thank you!

jmencak:
CoreOS `48.84.202105211054-0` now ships stalld 1.9.
/hold cancel
/retest

jmencak:
All tests passed, I believe this PR is good to go, Kevin/David.

kpouget:
/lgtm

thanks Jiri :)

PR #229, opened 2021-05-18
Bug 1962219: Switch to client-go leader-with-lease election.

jmencak:
Previously, NTO relied on the operator-sdk (leader-for-life) way of
handling leader election.  The leader-for-life was implemented by
relying on k8s garbage collector mechanism the same way as implemented
by operator-sdk.  Unfortunately, practise shows (rhbz#1916865,
rhbz#1958885) this is not reliable as the ConfigMap locks are sometimes
not garbage collected.

Therefore, switch to leader-with-lease election mechanism as provided by
the client-go package, which seems to be more reliable at this point.

Other changes:
  - fix incorrect error reporting on metrics server shutdown
  - fix default_irq_smp_affinity e2e test for systems with less than 4 CPUs

PR #238, opened 2021-06-21
Bug 1958885: Switch to client-go leader-with-lease election.

jmencak:
This is a backport of https://github.com/openshift/cluster-node-tuning-operator/pull/229 into 4.7.

Previously, NTO relied on the operator-sdk (leader-for-life) way of handling leader election.  The leader-for-life was implemented by relying on k8s garbage collector mechanism the same way as implemented by operator-sdk.  Unfortunately, practise shows (rhbz#1916865, rhbz#1958885, rhbz#1962219) this is not reliable as the ConfigMap locks are sometimes not garbage collected.

Therefore, switch to leader-with-lease election mechanism as provided by the client-go package, which seems to be more reliable at this point.

Other changes:
  - fix incorrect error reporting on metrics server shutdown
  - fix default_irq_smp_affinity e2e test for systems with less then 4 CPUs
  - `go mod tidy`


kpouget:
thanks @jmencak ,
\lgtm

kpouget:
> thanks @jmencak ,
> \lgtm

/lgtm

:|

/joke

PR #239, opened 2021-06-21
Bug 1974277: Fix conditional order for setting net device param.

jmencak:
The conditional for setting a net device parameter in TuneD has a wrong order where it can potentially try to first cast a string to int in a wrong manner and only then check if the value is a specific integer value.

Upstream TuneD fix: https://github.com/redhat-performance/tuned/pull/360

Resolves rhbz#1974277.

### DIFF
--- a/assets/tuned/patches/050-net-device-conditional-order.diff
+++ b/assets/tuned/patches/050-net-device-conditional-order.diff
@@ -1,0 +1,21 @@
+Fix conditional order for setting net device param
+
+The conditional for setting a net device parameter has
+a faulty order where it can potentially try to first cast a string to int
+in a wrong manner and only then check if the value is a specific integer value.
+
+See: https://github.com/redhat-performance/tuned/pull/360
+
+diff --git a/tuned/plugins/plugin_net.py b/tuned/plugins/plugin_net.py
+index 55a3d32b..1f4fabbc 100644
+--- a/tuned/plugins/plugin_net.py
++++ b/tuned/plugins/plugin_net.py
+@@ -397,7 +397,7 @@ def _set_device_parameters(self, context, value, device, sim,
+ 		if dev_params:
+ 			self._check_device_support(context, d, device, dev_params)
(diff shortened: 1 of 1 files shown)


jmencak:
Subject to upstream patch merging: 
https://github.com/redhat-performance/tuned/pull/360
/hold

kpouget:
makes sense to me,
```
(int(dev_params[next(iter(d))]) == 0 or str(dev_params[next(iter(d))]) == 'n/a')
```
the 2nd hand of the `or` can never be true, as `int(...)` will raise an exception if `dev_params[next(iter(d))]` isn't a sequence of digits, eg `n/a`

yanirq:
> makes sense to me,
> 
> ```
> (int(dev_params[next(iter(d))]) == 0 or str(dev_params[next(iter(d))]) == 'n/a')
> ```
> 
> the 2nd hand of the `or` can never be true, as `int(...)` will raise an exception if `dev_params[next(iter(d))]` isn't a sequence of digits, eg `n/a`

yes. this is the reason for the fix. tuned throws an unhandled exception in that case. 
/lgtm

yanirq:
> makes sense to me,
> 
> ```
> (int(dev_params[next(iter(d))]) == 0 or str(dev_params[next(iter(d))]) == 'n/a')
> ```
> 
> the 2nd hand of the `or` can never be true, as `int(...)` will raise an exception if `dev_params[next(iter(d))]` isn't a sequence of digits, eg `n/a`

yes. this is the reason for the fix. tuned throws an unhandled exception in that case. 
/lgtm
/hold - for NTO reviewers

jmencak:
No need for the torture, this patch cannot possibly cause an upgrade failure
/override ci/prow/e2e-upgrade

kpouget:
indeed, makes more sense written like that!

jmencak:
https://github.com/redhat-performance/tuned/pull/360 merged
Squashed commits.
/hold cancel
@yanirq , feel free to add `/lgtm` if this looks good.

PR #252, opened 2021-07-19
IBM Cloud manifest profile patch for operator deployment

csrwng:
Bumps build-machinery-go to latest
Adds patch specific to ibm-cloud-managed profile that generates a manifest that removes node selector for operator deployment
`make verify` verifies that the generated manifest is up to date
`make update` updates the manifest by applying the ibm-cloud-managed patch to the original deployment manifest


jmencak:
Thank you for the PR, @ csrwng .  Could you please explain why we need a separate `50-operator-ibm-cloud-managed.yaml` file and the patching mechanism instead of having a `include.release.openshift.io/ibm-cloud-managed: "true"` annotation to the existing operator manifest?

jmencak:
@csrwng , I think I see the point now, so you cannot use 
```
      nodeSelector:
```
on IBM Cloud for some reason.  How do we make sure  `50-operator.yaml` and `50-operator-ibm-cloud-managed.yaml` are both not rendered on non-IBM Cloud clusters though?

csrwng:
Hi @jmencak this is part of the cluster profiles support in the CVO (https://github.com/openshift/cluster-version-operator/pull/404). At any one time, there is only one cluster profile in effect when the CVO applies manifests to the cluster. The default profile is the `self-managed-high-availability` profile. If a manifest does not include the `include.release.openshift.io/self-managed-high-availability: "true"` annotation, it does not get applied. That means that the new manifest `50-operator-ibm-cloud-managed.yaml` will not get applied in a standard self-hosted OCP.

jmencak:
Thank you for the explanation, Cesar.
/lgtm

jmencak (approved):
/approve

PR #255, opened 2021-07-28
Bug 1986739: gather all Tuned CRs

jmencak:
Fixes gathering all of Tuned CRs with oc adm inspect/must-gather.


kpouget:
makes sense after looking at the BZ, we want to catch all the `tuned` objects
/lgtm

thanks @jmencak for the PR!

ecordell:
[patch-manager] :rocket: Approved for z-stream by score: 0.20

picked

PR #258, opened 2021-08-03
Bug 1985739: Move OpenShift profile to TuneD.

jmencak:
Changes:
  - lift OpenShift profiles from the default Tuned CR to the
    upstream TuneD project; see:
    https://github.com/redhat-performance/tuned/pull/376
  - add a cgroup blacklisting rule to parent openshift
    profile to prevent the [scheduler] plug-in from handling
    containers; see rhbz#1979352
  - make the Profile and Recommend sections optional in the
    Tuned CR
  - rhbz#1774645 has been long fixed in TuneD; remove the
    workaround to keep things clean
  - Makefile: ensure that "make" defaults to the "build"
    target


jmencak:
/uncc @zvonkok 
/cc @yanirq 

Yanir, this will also fix rhbz#1985739 for 4.9.  This is still WiP as I'm hoping to rebase on the latest `TuneD` once https://github.com/redhat-performance/tuned/pull/376 merges.

yanirq:
looks good to me in general,  the only thing missing from pao side is to include openshift-control-plane profile and cleanup the base profile we provide.
Since you removed some code related to the profiles I would like also @Tal-or  to take a look (we are watching them in PAO as well) and see if something could breal

jmencak:
Thank you Yanir.  Looks like [PR376](https://github.com/redhat-performance/tuned/pull/376) will be merged very soon.  @Tal-or , any input from your side?

manifests/20-crd-tuned.yaml
```diff
@@ -35,9 +35,6 @@ spec:
             spec:
               description: 'spec is the specification of the desired behavior of Tuned. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status'
               type: object
-              required:
-                - profile
```
Tal-or, line 39: By having those fields optional it might affect the way PAO present and track the tuned profile's status? https://github.com/openshift-kni/performance-addon-operators/blob/39c74c9b999b57d810934fe0913e8a64eaff162c/controllers/status.go#L240

jmencak: Thank you for looking @Tal-or !  But you're querying [Profile](https://github.com/openshift/cluster-node-tuning-operator/blob/master/manifests/20-crd-profile.yaml) in your code and not [Tuned](https://github.com/openshift/cluster-node-tuning-operator/blob/master/manifests/20-crd-tuned.yaml) that is modified here.  How will this cause problems to PAO?

Tal-or: My mistake. I thought that the profile under Tuned is this [Profile](https://github.com/openshift/cluster-node-tuning-operator/blob/master/manifests/20-crd-profile.yaml)

jmencak:
@dagrayvid if those changes still look good to you, can we have a LGTM? Thanks!

PR #266, opened 2021-08-27
Bug 1998120: Add a cgroup blacklisting rule to parent openshift profile.

jmencak:
Add a cgroup blacklisting rule to parent openshift profile to prevent the `[scheduler]` plug-in from handling containers; see rhbz#1998120.


yanirq:
/lgtm
I would re run and check the flaky tests

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 0.20

adding cherry-pick approved

PR #267, opened 2021-08-30
Bug 1998247: Reload when deps of recommended profile change.

jmencak:
This change fixes operand's behavior during administrator-initiated changes of TuneD profiles that the currently recommended TuneD profile depends on.  Now, changes to any such dependent profile and changes to the recommended TuneD profile will cause TuneD daemon reload.

Resolves rhbz#1998247.

### DIFF
--- a/pkg/tuned/tuned.go
+++ b/pkg/tuned/tuned.go
@@ -48,6 +48,7 @@ const (
 	operandNamespace       = "openshift-cluster-node-tuning-operator"
 	programName            = "openshift-tuned"
 	tunedProfilesDir       = "/etc/tuned"
+	tunedConfFile          = "tuned.conf"
 	tunedActiveProfileFile = tunedProfilesDir + "/active_profile"
 	tunedRecommendDir      = tunedProfilesDir + "/recommend.d"
 	tunedRecommendFile     = tunedRecommendDir + "/50-openshift.conf"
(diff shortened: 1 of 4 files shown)


pkg/tuned/tuned.go
```diff
@@ -313,9 +314,18 @@ func disableSystemTuned() {
...
+		content = []byte{}
+	}
+
+	return profileData == string(content)
+}
+
 // profilesExtract extracts TuneD daemon profiles to the daemon configuration directory.
-// If the data in the to be extracted recommended profile is different than the current
-// recommended profile, the function returns true.
+// If the data in the to be extracted recommended profile or the profiles being included
```
kpouget, line 327: I'm not sure how to read this comment, is it like this?
> If the data in the to-be-extracted-recommended-profile

if yes, then the `-` helps a lot, if no ... the comment should be reformulated :)

jmencak: Your interpretation is correct. I'll add the `-`, thank you!

kpouget:
thanks for the PR @jmencak, I went through the code and I think it all makes sense! the logic was complex to follow (by nature), but I think I managed to follow the thread :)

/lgtm
/hold <-- I let you unhold when ready to merge

jmencak:
/hold cancel
Updated the comment as per reviewer comments.

kpouget:
thanks @jmencak 
/lgtm

PR #274, opened 2021-09-17
WIP: Makefile cleanup, replace yq with yaml-patch from openshift/build-machinery-go

dagrayvid:
Here is a first draft. Replaces crd-schema-gen make target with a call to [add-crd-gen ](https://github.com/openshift/build-machinery-go/blob/4ea9b6da3a37b7ff391a164329b696b91049a1ba/make/targets/openshift/crd-schema-gen.mk#L82) from openshift/build-machinery-go. Renamed the crds to *.crd.yaml as required by add-crd-gen. 

/cc @jmencak 
Let me know if this is what you had in mind.  Not ready to merge yet as I haven't thoroughly tested.


jmencak (commented):
Thank you for the PR, David! I think this is a really good start to clean things up and create a more maintainable `Makefile` that will be familiar to other OpenShift contributors.  A couple of comments/questions though:
  - Is there a reason for the renaming of `20-crd-tuned.yaml`?  If the files must end in `.crd.yaml` for some reason, I wonder whether we should drop the first duplicate "crd" and say something like `20-tuned.crd.yaml` instead.
  - I wonder if it would make sense to move the `20-crd-tuned.crd.yaml-patch` outside of the `/manifests` directory in the `/profile-patches` directory.  Not sure this is possible with current `build-machinery-go` though.
  - I wonder if it makes sense to note somewhere that now to update the CRDs, we need to run `update-codegen-crds`; perhaps in the `Makefile/README.md` or at least the commit message?
  - Now that we are moving more functionality to `build-machinery-go`, the whole `pkg/generated` target is probably a duplication of what `build-machinery-go` already does somewhere.  I wonder if we should keep the target around or just use `build-machinery-go`?
  - `./$(API_TYPES_DIR)` seems a bit strange.  Perhaps do "`API_TYPES_DIR:=./pkg/apis/tuned/v1`" instead?

dagrayvid:
> - Is there a reason for the renaming of `20-crd-tuned.yaml`?  If the files must end in `.crd.yaml` for some reason, I wonder whether we should drop the first duplicate "crd" and say something like `20-tuned.crd.yaml` instead.

Yes, it seems like build-machinery-go expects the CRD manifests to be [named *.crd.yaml](https://github.com/openshift/build-machinery-go/blob/4ea9b6da3a37b7ff391a164329b696b91049a1ba/make/targets/openshift/crd-schema-gen.mk#L60). Agree about removing the duplicate "crd", I thought I already did this! Fixed now.
> - I wonder if it would make sense to move the `20-crd-tuned.crd.yaml-patch` outside of the `/manifests` directory in the `/profile-patches` directory.  Not sure this is possible with current `build-machinery-go` though.

I initially tried this, but patches listed under `/profile-patches/<directory_name>` result in a new .yaml under `/manifests` named by the directory name. Maybe we can change the calls to build-machinery-go so that only the ibm-cloud-managed profile patch is applied this way. Alternatively, maybe we could have a `/patches` directory, with `/patches/profile-patches` that contains patches like the ibm-cloud-managed patch under it (to be used to create these additional files named based on their directory).

> - I wonder if it makes sense to note somewhere that now to update the CRDs, we need to run `update-codegen-crds`; perhaps in the `Makefile/README.md` or at least the commit message?

Noted in the commit message and the Makefile.

> - Now that we are moving more functionality to `build-machinery-go`, the whole `pkg/generated` target is probably a duplication of what `build-machinery-go` already does somewhere.  I wonder if we should keep the target around or just use `build-machinery-go`?

I will have to look into this.

> - `./$(API_TYPES_DIR)` seems a bit strange.  Perhaps do "`API_TYPES_DIR:=./pkg/apis/tuned/v1`" instead?

Agreed, fixed now.

jmencak:
Thank you for the PR, David.  It looks good to me.  Could you please split the commit message into a shorter "subject" and a message body?  For example, something like this might work:

```
Replace yq with yaml-patch in Makefile, from openshift/build-machinery-go

To update CRDs run make update or make update-codegen-crds
```
Thank you!

dagrayvid:
> Could you please split the commit message into a shorter "subject" and a message body? 

Done

PR #278, opened 2021-10-12
Bug 2013321: TuneD: workaround for high CPU utilization of [scheduler] plug-in.

jmencak:
The fix for rhbz#1979352 introduced the [scheduler] plug-in as a
standard part of openshift TuneD profiles.  Unfortunately, the
[scheduler] plug-in can be very CPU intensive, especially on the
OpenShift platform.  The bug for this issue is tracked by rhbz#1921738.
Until this is fixed, work around this problem by adding "runtime=0"
[scheduler] plug-in option.


jmencak:
/uncc @ArangoGutierrez 
/cc @yanirq 
as this is very similar to https://github.com/openshift-kni/performance-addon-operators/pull/579

PR #280, opened 2021-10-13
Bug 2013678: TuneD: workaround for high CPU utilization of [scheduler] plug-in.

jmencak:
The fix for rhbz#1979352 introduced the [scheduler] plug-in as a
standard part of openshift TuneD profiles.  Unfortunately, the
[scheduler] plug-in can be very CPU intensive, especially on the
OpenShift platform.  The bug for this issue is tracked by rhbz#1921738.
Until this is fixed, work around this problem by adding "runtime=0"
[scheduler] plug-in option.

Other changes:
  - Address a race in the stalld e2e test.  The race was sometimes causing
    the TuneD daemon re-enable and start stalld service after it was
    intentionally stopped by the test itself using systemctl.

### DIFF
--- a/test/e2e/reboots/stalld.go
+++ b/test/e2e/reboots/stalld.go
@@ -75,16 +75,19 @@ var _ = ginkgo.Describe("[reboots][stalld] Node Tuning Operator installing syste
 			err = util.WaitForPoolUpdatedMachineCount(cs, "worker-rt", 1)
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
-			ginkgo.By(fmt.Sprintf("waiting for the tuned daemon running on node %s", node.Name))
-			out, err := util.WaitForCmdInPod(pollInterval, waitDuration, pod, "test", "-e", "/run/tuned/tuned.pid")
-			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			waitForTuneD := func() {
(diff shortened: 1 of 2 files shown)


test/e2e/reboots/stalld.go
```diff
@@ -75,16 +75,19 @@ var _ = ginkgo.Describe("[reboots][stalld] Node Tuning Operator installing syste
...
-			ginkgo.By("sleeping 10s to allow for (now unwanted) stalld service starting")
-			time.Sleep(10 * time.Second)
+			waitForTuneD := func() {
+				ginkgo.By(fmt.Sprintf("waiting for the TuneD daemon running on node %s", node.Name))
+				_, err := util.WaitForCmdInPod(pollInterval, waitDuration, pod, "test", "-e", "/run/tuned/tuned.pid")
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+				// 10s is very generous to allow for stalld service starting (if any)
+				ginkgo.By("sleeping 10s to allow for stalld service starting (if any)")
+				time.Sleep(10 * time.Second)
```
kpouget, line 85: I'm not sure to understand the `(if any)` part of the comment, could you explain it a bit?

what I understand from this function is:

1. wait for file `/run/tuned/tuned.pid` to appear, meaning that TuneD started its (early) execution
2. wait 10s for TuneD to start operating properly

PR #288, opened 2021-11-05
openshift-tuned: TuneD built-in function support

jmencak:
Implemented a very basic and (yet) incomplete parser of TuneD built-in functions in the form `${f:function(:argN)*}`.  Supported built-ins are now `${f:virt_check:profile-a:profile-b}` and `${f:exec(:argN)}`.  These are being expanded only in the TuneD "include" statements to discover whether the TuneD process needs to be reloaded due to a custom profile change that uses these built-ins.

Other changes:
  - minor refactoring to lift TuneD profile parser specific code
    from the operand
  - fixed/renamed the target for unit tests, use `make test-unit`

### DIFF
--- a/pkg/tuned/tuned_parser.go
+++ b/pkg/tuned/tuned_parser.go
@@ -1,0 +1,367 @@
+package tuned
+
+import (
+	"fmt"       // Printf()
+	"io"        // io.EOF
+	"io/ioutil" // ioutil.ReadFile()
+	"os/exec"   // os.Exec()
+	"strings"   // strings.Split()
+	"syscall"   // syscall.SIGHUP, ...
+	"time"      // time.Second, ...
+
+	"gopkg.in/ini.v1"
+	"k8s.io/klog/v2"
+)
+
+// getIniFileSectionSlice searches INI file `data` inside [`section`]
+// for key `key`.  It takes the key's value and uses separator
+// `separator` to return a slice of strings.
+func getIniFileSectionSlice(data *string, section, key, separator string) []string {
+	var ret []string
+
+	if data == nil {
+		return ret
+	}
+
+	cfg, err := ini.Load([]byte(*data))
+	if err != nil {
+		// This looks like an invalid INI data or parser error.
+		klog.Errorf("unable to read INI file data: %v", err)
+		return ret
+	}
+
+	if !cfg.Section(section).HasKey(key) {
+		return ret
+	}
+
+	ret = strings.Split(cfg.Section(section).Key(key).String(), separator)
+
+	return ret
+}
+
+// profileHasStalld returns pointer to a boolean value.  The dereferenced
+// pointer value depends on the Tuned [service] plugin enabling/disabling
+// the "stalld" service.  If the "stalld" service "service.stalld" key is
+// not found in Tuned [service] section, nil is returned.
+func profileHasStalld(profile *string) *bool {
+	var ret bool
+
+	stalldServiceVal := getIniFileSectionSlice(profile, "service", "service.stalld", ",")
+
+	if len(stalldServiceVal) == 0 {
+		// There was no service.stalld key in [service] section.
+		return nil
+	}
+
+	for _, v := range stalldServiceVal {
+		if v == "enable" {
+			ret = true
+			return &ret
+		}
+		if v == "disable" {
+			return &ret
+		}
+	}
+
+	// Default to "disable" if "enable" is not present as value by service.stalld key.
+	return &ret
+}
+
+// profileIncludes returns a slice of strings containing TuneD profile names
+// custom (/etc/tuned/<profileName>/) profile 'profileName' includes.
+func profileIncludes(profileName string) []string {
+	profileFile := fmt.Sprintf("%s/%s/%s", tunedProfilesDir, profileName, tunedConfFile)
+
+	content, err := ioutil.ReadFile(profileFile)
+	if err != nil {
+		content = []byte{}
+	}
+
+	s := string(content)
+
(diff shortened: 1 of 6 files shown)


pkg/tuned/tuned_parser.go
```diff
@@ -0,0 +1,354 @@
...
+	return getIniFileSectionSlice(&s, "main", "include", ",")
+}
+
+// profileDepends returns "TuneD profile name"->bool map that custom
+// (/etc/tuned/<profileName>/) profile 'profileName' depends on as keys.
+// The dependency is resolved by finding all the "parent" profiles which are
+// included by using the "include" keyword in the profile's [main] section.
+// Note: no expansion of the TuneD functions into profiles names, such as
+// "f:virt_check" is performed by this function.  Only static parsing of
+// the TuneD INI configuration files is performed.
```
dagrayvid, line 91:
```suggestion
```
This part of the comment is no longer accurate with this PR, right?

dagrayvid:
Thanks @jmencak. I spent some time playing with the test cases in a debugger to understand how the parser works. Other than the problem with the comment I mentioned above, I think this looks good. 

A couple of questions about the parser and built-ins: Do we expect to support more built-ins later? Do we need to support multiple levels of nested functions? If I were writing this I might have tried to take some shortcuts by looking for the substring `${f:exec` or `${f:virt-check`, though this gets messy with nested functions. I haven't fully thought it through, just curious on your reasoning for going with this approach. 

One note: we need to update the ci-config in openshift/release to actually run the unit tests with the new make target. Should we do that before we merge this, or after?

I verified that the new unit tests work, but I did get an error from pkg/manifests/manifests.go... is this user error on my part?
```
pkg/manifests/manifests.go:26:25: undefined: MustAsset
ok  	github.com/openshift/cluster-node-tuning-operator/pkg/clusteroperator	0.004s	coverage: 100.0% of statements
ok  	github.com/openshift/cluster-node-tuning-operator/pkg/tuned	0.016s	coverage: 10.6% of statements
make: *** [Makefile:116: test-unit] Error 1
```

pkg/tuned/tuned_parser.go
jmencak: Correct, good catch, thank you!  Will adjust.

jmencak:
Thank you for looking at this, David, I really appreciate this.

> Do we expect to support more built-ins later?

Possibly, but these are the most important ones in my view and a good starting point.

> Do we need to support multiple levels of nested functions?

Probably not, but it may help with some use cases and results in a more "robust" code.  Also, `TuneD` daemon supports this, hence my reasoning to support this.

> One note: we need to update the ci-config in openshift/release to actually run the unit tests with the new make target. Should we do that before we merge this, or after?

Great thinking.  Definitely and this is my plan.  I was planning to do this after we merge this.

> I verified that the new unit tests work, but I did get an error from pkg/manifests/manifests.go... is this user error on my part?
> 
> ```
> pkg/manifests/manifests.go:26:25: undefined: MustAsset
> ok  	github.com/openshift/cluster-node-tuning-operator/pkg/clusteroperator	0.004s	coverage: 100.0% of statements
> ok  	github.com/openshift/cluster-node-tuning-operator/pkg/tuned	0.016s	coverage: 10.6% of statements
> make: *** [Makefile:116: test-unit] Error 1
> ```

Interesting.  I do not see this in my environment (`go version go1.16.8 linux/amd64`).  The unit tests pass without issues for me.  Will try to look into this.

jmencak:
> ```
> pkg/manifests/manifests.go:26:25: undefined: MustAsset
> ok  	github.com/openshift/cluster-node-tuning-operator/pkg/clusteroperator	0.004s	coverage: 100.0% of statements
> ok  	github.com/openshift/cluster-node-tuning-operator/pkg/tuned	0.016s	coverage: 10.6% of statements
> make: *** [Makefile:116: test-unit] Error 1
> ```

I'm able to reproduce this on clean checkout prior to running "`make`".  I believe this needs to be addressed (at a minimum in the `Makefile`) before we merge his.  Good catch!

jmencak:
`make test-unit` fixed by adding dependency on `$(BINDATA)`.  Leaving `hold` as I've seen some intermittent `unit` test failures that need to be addressed first.

jmencak:
I believe I've fixed the intermittent unit test failures.
/hold cancel

jmencak:
Huh?
```
make: *** [vendor/github.com/openshift/build-machinery-go/make/targets/openshift/../../targets/openshift/controller-gen.mk:30: ensure-controller-gen] Error 22
```
I hope this is a flake.
/retest

jmencak:
Hmm, looked in the live build env.  Looks to me like this is failing:
```
curl -s -f -L https://github.com/openshift/kubernetes-sigs-controller-tools/releases/download/v0.6.0/controller-gen-linux-amd64 -o '_output/tools/bin/controller-gen-v0.6.0'
make: *** [vendor/github.com/openshift/build-machinery-go/make/targets/openshift/../../targets/openshift/controller-gen.mk:30: ensure-controller-gen] Error 22
```
Line 30 of the makefile corresponds to fetching the `controller-gen-v0.6.0`.
But I was able to fetch `controller-gen-v0.6.0` from another running pod directly in the build pipeline without issues.  Continuing digging.  Perhaps really just a flake?

/test verify

jmencak:
Yeah, the curl to get `controller-gen-linux-amd64` is failing for some reason right now..
Let's keep us entertained.
/joke

jmencak:
\o/

dagrayvid:
Looks good. Thanks @jmencak.

/lgtm

PR #291, opened 2021-11-10
go.mod: Bump to go 1.17

jmencak:
Changes:
  - updated go in `go.mod` to 1.17
  - ran `go mod vendor`
  - patched 3 files with `go fmt`  1.17 to pass `make verify`


jmencak:
Also see: https://github.com/openshift/cluster-node-tuning-operator/pull/290

dagrayvid:
Looks good. I did a little googling to figure out if we need both //go:build and //+build in some of those files. Sounds like the //go:build syntax is new for 1.17 and they will be able to coexist for a few releases. 

In order to keep it working on versions <1.17 I suppose we should leave both for now?

jmencak:
@dagrayvid thank you! I was about to comment on your questions, but the last few days were a bit crazy. :)

jmencak:
> Looks good. I did a little googling to figure out if we need both //go:build and //+build in some of those files. Sounds like the //go:build syntax is new for 1.17 and they will be able to coexist for a few releases.
> 
> In order to keep it working on versions <1.17 I suppose we should leave both for now?

Yes, good point.  We will remove `// +build` in the future.  We could probably even do this right now as we'll not compile current NTO with anything lower than `golang 1.17`.  Deferring for a future PR though.

PR #293, opened 2021-12-06
Bug 2029371: controller: update MC after application by TuneD

jmencak:
Tuned Profiles are created by NTO operator and updated by both the
operator and its operand.  The operand's updates are the source of truth
for information such as kernel command-line parameters and presence of
the stalld daemon.  The operator is informed about this via the status
field of the a Tuned Profile that corresponds to a given node.  Operator
updates are the source of truth about which TuneD profile should be
applied.

When a new cluster node is added (for example during a cluster
scale-up), a new Tuned Profile is created.  If the node is quickly added
to a MachinePool for which an NTO-managed MachineConfig already exists,
this addition will cause NTO to overwrite/update the MachineConfig with
data that is incorrectly initialized.  If the operand was too slow to
correct this update (typically during scale-ups when the operand was not
yet running), this would cause unnecessary reboots of nodes sharing the
same MachineConfigPool.

This PR fixes the behaviour above by only letting the operator update
the NTO-managed MachineConfigs once the newly requested Tuned Profile
configuration is successfully applied.

Resolves rhbz#2024682

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -598,12 +598,16 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 	}
 
 	if mcLabels != nil {
-		// The Tuned daemon profile "tunedProfileName" for nodeName matched with MachineConfig
+		// The Tuned daemon profile 'tunedProfileName' for nodeName matched with MachineConfig
 		// labels set for additional machine configuration.  Sync the operator-created
 		// MachineConfig for MachineConfigPools 'pools'.
-		err := c.syncMachineConfig(getMachineConfigNameForPools(pools), mcLabels, profile.Status.Bootcmdline, profile.Status.Stalld)
-		if err != nil {
-			return fmt.Errorf("failed to update Profile %s: %v", profile.Name, err)
+		if profile.Status.TunedProfile == tunedProfileName && profileApplied(profile) {
+			// Synchronize MachineConfig only once the (calculated) TuneD profile 'tunedProfileName'
+			// has been successfully applied.
+			err := c.syncMachineConfig(getMachineConfigNameForPools(pools), mcLabels, profile.Status.Bootcmdline, profile.Status.Stalld)
+			if err != nil {
+				return fmt.Errorf("failed to update Profile %s: %v", profile.Name, err)
+			}
 		}
 	}
 	if profile.Spec.Config.TunedProfile == tunedProfileName &&
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -81,7 +81,22 @@ func (c *Controller) getOrCreateOperatorStatus() (*configv1.ClusterOperator, err
 	return co, nil
 }
 
-// profileDegraded returns true if Tuned Profiles 'profile' has not been applied
+// profileApplied returns true if Tuned Profile 'profile' has been applied.
+func profileApplied(profile *tunedv1.Profile) bool {
+	if profile == nil || profile.Spec.Config.TunedProfile != profile.Status.TunedProfile {
+		return false
+	}
+
+	for _, sc := range profile.Status.Conditions {
+		if sc.Type == tunedv1.TunedProfileApplied && sc.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+
+	return false
+}
+
+// profileDegraded returns true if Tuned Profile 'profile' has not been applied
 // or applied with errors (Degraded).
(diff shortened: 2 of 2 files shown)


pkg/operator/controller.go
```diff
@@ -601,9 +601,13 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 		// The Tuned daemon profile "tunedProfileName" for nodeName matched with MachineConfig
 		// labels set for additional machine configuration.  Sync the operator-created
 		// MachineConfig for MachineConfigPools 'pools'.
-		err := c.syncMachineConfig(getMachineConfigNameForPools(pools), mcLabels, profile.Status.Bootcmdline, profile.Status.Stalld)
-		if err != nil {
-			return fmt.Errorf("failed to update Profile %s: %v", profile.Name, err)
+		if profile.Status.TunedProfile == tunedProfileName && profileApplied(profile) {
+			// Synchronize MachineConfig only once we the (calculated) TuneD profile tunedProfileName
+			// has been successfully applied.
```
dagrayvid, line 606:
```suggestion
			// Synchronize MachineConfig only once the (calculated) TuneD profile tunedProfileName
			// has been successfully applied.
```
Typo?

pkg/operator/controller.go
```diff
@@ -601,9 +601,13 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 		// The Tuned daemon profile "tunedProfileName" for nodeName matched with MachineConfig
 		// labels set for additional machine configuration.  Sync the operator-created
 		// MachineConfig for MachineConfigPools 'pools'.
-		err := c.syncMachineConfig(getMachineConfigNameForPools(pools), mcLabels, profile.Status.Bootcmdline, profile.Status.Stalld)
-		if err != nil {
-			return fmt.Errorf("failed to update Profile %s: %v", profile.Name, err)
+		if profile.Status.TunedProfile == tunedProfileName && profileApplied(profile) {
```
dagrayvid, line 604: If the condition is false, should we requeue / return err? Or will the profile status be updated by the operand once the profile is applied, triggering operator reconciliation?

jmencak: The latter. the profile status be updated by the operand once the profile is applied, triggering operator reconciliation.

pkg/operator/status.go
```diff
@@ -81,6 +81,21 @@ func (c *Controller) getOrCreateOperatorStatus() (*configv1.ClusterOperator, err
 	return co, nil
 }
 
+// profileApplied returns true if Tuned Profiles 'profile' has been applied.
```
dagrayvid, line 84:
```suggestion
// profileApplied returns true if Tuned Profile "profile" has been applied.
```
Sorry I didn't catch this in the first review. After this, the PR looks good to me.

jmencak: Thank you.  Fixed in other place too.  Using `'` to quote variable names elsewhere in function comments (e.g. `mc.go`, `profilecalculator.go`), so switching to that for consistency.

dagrayvid:
/hold

@jmencak We should take a look at the e2e-aws-operator failure... it didn't seem like a flake to me.

jmencak:
> /hold
> 
> @jmencak Can you take a look at the e2e-aws-operator failure and unhold if you are ready to merge? It didn't seem like a CI flake

Thank you for pointing out the `e2e` test failure to me.  This one is new to me.  Now we seem to have two "intermittently" failing 2 e2e tests I know of:
- `basic/sysctl_d_override.go`
- `reboots/stalld.go`

At this point it is not clear to me why they sometimes fail (I have a hunch about the second one).  We'll address this in future PRs.  However, I'm 100% sure this PR's code has nothing to do with the failure you pointed out (`basic/sysctl_d_override.go`) as this test has nothing in common with the MachineConfig matching functionality we are adjusting here.

As the retest succeeded, cancelling hold and adding e2e test review on our TODO list.

/hold cancel

jmencak:
> At this point it is not clear to me why they sometimes fail (I have a hunch about the second one). We'll address this in future PRs. However, I'm 100% sure this PR's code has nothing to do with the failure you pointed out (`basic/sysctl_d_override.go`) as this test has nothing in common with the MachineConfig matching functionality we are adjusting here.
> 
> As the retest succeeded, cancelling hold and adding e2e test review on our TODO list.

Jira story to address the intermittent e2e failures: https://issues.redhat.com/browse/PSAP-576

PR #296, opened 2021-12-09
controller: ignore non-linux hosts

jmencak:
Controller has a sync loop that creates Tuned Profiles for every node on
the cluster.  As TuneD supports only Linux hosts, there is no point in
creating Profiles for hosts other than Linux.


dagrayvid:
This seems fine to me.
/lgtm

PR #297, opened 2021-12-09
Bug 2026387: Handle certificate rotation in pkg/metrics/server.go

dagrayvid:
This  PR is intended to solve RHBZ#2026387. 

With this change, a file watcher is added to the metrics server using the fsnotify package, to watch for changes to the TLS certificate for the metrics endpoint. I took some inspiration from https://github.com/openshift/cluster-ingress-operator/pull/334/. Instead of restarting the operator when the certificate changes as is done in that PR, the server will be shut down and restarted.

These changes have been tested by editing the node-tuning-operator-tls secret (comments can be added to the certificate / key by putting them outside of the `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----`   lines) and by deleting the secret altogether which results in a new key pair being created.

After some review, I will do more testing using the same NTP process described in the BZ to "expire" the certificate.

@jmencak PTAL

Signed-off-by: David Gray <dagray@redhat.com>

### DIFF
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -1,25 +1,32 @@
 package metrics
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
+	"os"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"gopkg.in/fsnotify.v1"
 
 	"k8s.io/klog/v2"
 )
 
 const (
-	tlsCRT = "/etc/secrets/tls.crt"
-	tlsKey = "/etc/secrets/tls.key"
+	tlsSecretDir = "/etc/secrets"
+	tlsCert      = tlsSecretDir + "/tls.crt"
+	tlsKey       = tlsSecretDir + "/tls.key"
 )
 
-// RunServer starts the metrics server.
-func RunServer(port int, stopCh <-chan struct{}) {
+func buildServer(port int) *http.Server {
 	if port <= 0 {
 		klog.Error("invalid port for metric server")
-		return
+		return nil
 	}
 
 	handler := promhttp.HandlerFor(
@@ -37,44 +44,136 @@ func RunServer(port int, stopCh <-chan struct{}) {
 		Handler: router,
 	}
 
-	go func() {
-		if err := srv.ListenAndServeTLS(tlsCRT, tlsKey); err != nil {
-			klog.Errorf("error starting metrics server: %v", err)
+	return srv
+}
+
+func startServer(srv *http.Server) {
+	klog.Infof("starting metrics server")
+	if err := srv.ListenAndServeTLS(tlsCert, tlsKey); err != nil && err != http.ErrServerClosed {
+		klog.Errorf("error from metrics server: %v", err)
+	}
+}
+
+func stopServer(srv *http.Server) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	klog.Infof("stopping metrics server")
+	if err := srv.Shutdown(ctx); err != nil {
+		klog.Errorf("error or timeout stopping metrics listener: %v", err)
+	}
+}
+
+// RunServer starts the server, and watches the tlsCert and tlsKey for certificate changes.
+func RunServer(port int, stopCh <-chan struct{}) {
+	srv := buildServer(port)
+	if srv == nil {
+		return
+	}
+
+	go startServer(srv)
+
+	// Set up and start the file watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if watcher == nil || err != nil {
+		klog.Errorf("failed to create file watcher, cert/key rotation will be disabled  %v", err)
+	} else {
+		defer watcher.Close()
+		if err := watcher.Add(tlsSecretDir); err != nil {
+			klog.Errorf("failed to add %v to watcher, cert/key rotation will be disabled: %v", tlsSecretDir, err)
 		}
-	}()
-	<-stopCh
-	if err := srv.Shutdown(context.Background()); err != nil && err != http.ErrServerClosed {
-		klog.Errorf("error stopping metrics listener: %v", err)
+	}
+
+	origCertChecksum := checksumFile(tlsCert)
+	origKeyChecksum := checksumFile(tlsKey)
+
+	for {
+		select {
+		case <-stopCh:
+			stopServer(srv)
+			return
+		case event := <-watcher.Events:
+			klog.V(2).Infof("event from filewatcher on file: %v, event: %v", event.Name, event.Op)
+
+			if event.Op == fsnotify.Chmod || event.Op == fsnotify.Remove {
+				continue
+			}
+
+			if certsChanged(origCertChecksum, origKeyChecksum) {
+				// Update file checksums with latest files.
+				origCertChecksum = checksumFile(tlsCert)
+				origKeyChecksum = checksumFile(tlsKey)
+
+				// restart server
+				klog.Infof("restarting metrics server to rotate certificates")
+				stopServer(srv)
+				srv = buildServer(port)
+				go startServer(srv)
+			}
+		case err = <-watcher.Errors:
+			klog.Warningf("error from metrics server certificate file watcher: %v", err)
+		}
 	}
 }
 
-// PodLabelsUsed indicates whether the deprecated Pod label matching functionality
-// is turned on.
-func PodLabelsUsed(enable bool) {
-	if enable {
-		podLabelsUsed.Set(1)
-		return
+// Determine if the certificates have changed and need to be updated.
+// Returns true if both files have changed AND neither is an empty file.
+func certsChanged(origCertChecksum []byte, origKeyChecksum []byte) bool {
+	// Check if both files exist.
+	certNotEmpty, err := fileExistsAndNotEmpty(tlsCert)
+	if err != nil {
+		klog.Warningf("error checking if changed TLS cert file empty/exists: %v", err)
+		return false
 	}
-	podLabelsUsed.Set(0)
-}
-
-// ProfileCalculated keeps track of the number of times a given Tuned profile
-// resource was calculated for node 'nodeName'.
-func ProfileCalculated(nodeName, profileName string) {
-	profileCalculated.With(map[string]string{"node": nodeName, "profile": profileName}).Inc()
-}
-
-// RegisterVersion exposes the Operator build version.
-func RegisterVersion(version string) {
-	buildInfo.WithLabelValues(version).Set(1)
-}
-
-// Degraded sets the metric that indicates whether the operator is in degraded
-// mode or not.
-func Degraded(deg bool) {
-	if deg {
-		degradedState.Set(1)
-		return
+	keyNotEmpty, err := fileExistsAndNotEmpty(tlsKey)
+	if err != nil {
+		klog.Warningf("error checking if changed TLS key file empty/exists: %v", err)
+		return false
 	}
-	degradedState.Set(0)
+
+	if !certNotEmpty || !keyNotEmpty {
+		// One of the files is missing despite some file event.
+		klog.V(1).Infof("certificate or key is missing or empty, certificates will not be rotated")
+		return false
+	}
+	currentCertChecksum := checksumFile(tlsCert)
+	currentKeyChecksum := checksumFile(tlsKey)
+
+	klog.V(2).Infof("certificate checksums before: %x, %x. checksums after: %x, %x", origCertChecksum, origKeyChecksum, currentCertChecksum, currentKeyChecksum)
+	// Check if the non-empty certificate/key files have actually changed.
+	if !bytes.Equal(origCertChecksum, currentCertChecksum) && !bytes.Equal(origKeyChecksum, currentKeyChecksum) {
+		klog.Infof("cert and key changed, need to restart the server.")
+		return true
+	}
+
+	return false
+}
+
+// Compute the sha256 checksum for file 'fName'.
+func checksumFile(fName string) []byte {
+	file, err := os.Open(fName)
+	if err != nil {
+		klog.Infof("failed to open file %v for checksum: %v", fName, err)
+	}
+	defer file.Close()
+
+	hash := sha256.New()
+
+	if _, err = io.Copy(hash, file); err != nil {
+		klog.Infof("failed to compute checksum for file %v: %v", fName, err)
+	}
+
+	return hash.Sum(nil)
+}
+
+// Check if a file exists and has file.Size() not equal to 0.
+// Returns any error returned by os.Stat other than os.ErrNotExist.
+func fileExistsAndNotEmpty(fName string) (bool, error) {
+	if fi, err := os.Stat(fName); err == nil {
+		return (fi.Size() != 0), nil
+	} else if errors.Is(err, os.ErrNotExist) {
+		return false, nil
+	} else {
+		// Some other error, file may not exist.
+		return false, err
+	}
 }
--- a/test/e2e/basic/metrics_cert_rotation.go
+++ b/test/e2e/basic/metrics_cert_rotation.go
@@ -1,0 +1,110 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
+	util "github.com/openshift/cluster-node-tuning-operator/test/e2e/util"
+)
+
+var _ = ginkgo.Describe("[basic][metrics] Node Tuning Operator certificate rotation", func() {
+	ginkgo.Context("TLS certificate rotation", func() {
+		ginkgo.It("delete certificate Secret and check that the server restarts with latest certificate", func() {
+			const (
+				pollInterval = 5 * time.Second
+				waitDuration = 5 * time.Minute
+			)
+
+			// Delete the existing secret to force certificate rotation.
+			ginkgo.By("deleting node-tuning-operator-tls Secret to trigger certificate rotation")
+			err := rotateTLSCertSecret()
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			// Delete the existing secret again to force certificate rotation.
+			// This is done twice to test whether the file watcher in the metrics server
+			// will continue to watch the new certificates for changes after a rotation.
+			ginkgo.By("deleting node-tuning-operator-tls Secret a second time")
+			err = rotateTLSCertSecret()
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			ginkgo.By("getting a list of worker nodes")
+			nodes, err := util.GetNodesByRole(cs, "worker")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			gomega.Expect(len(nodes)).NotTo(gomega.BeZero(), "number of worker nodes is 0")
+
+			node := &nodes[0]
+			ginkgo.By(fmt.Sprintf("getting a TuneD Pod running on node %s", node.Name))
+			tunedPod, err := util.GetTunedForNode(cs, node)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			ginkgo.By("getting cluster-node-tuning-operator Pod")
+			operatorPod, err := util.GetNodeTuningOperatorPod(cs)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			ginkgo.By("checking if server TLS certificate matches TLS certificate in Secret")
+			err = wait.PollImmediate(pollInterval, waitDuration, func() (bool, error) {
+				tlsSecret, err := cs.Secrets(ntoconfig.OperatorNamespace()).Get(context.TODO(), "node-tuning-operator-tls", metav1.GetOptions{})
+				if err != nil {
+					util.Logf("error getting secret/node-tuning-operator-tls. May not exist yet. Err: %v", err)
+					return false, nil
+				}
+				secretCertContents := string(tlsSecret.Data["tls.crt"])
+
+				operatorPodIP := operatorPod.Status.PodIP
+				opensslCmd := "/host/usr/bin/openssl s_client -connect " + operatorPodIP + ":60000 2>/dev/null </dev/null | sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p'"
+
+				serverCertContents, err := util.ExecCmdInPod(tunedPod, "/bin/bash", "-c", opensslCmd)
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+				if len(serverCertContents) > 0 && strings.Contains(secretCertContents, serverCertContents) {
+					return true, nil
+				}
+				return false, nil
+			})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		})
+	})
+})
+
+// Deletes the secret/node-tuning-operator-tls and
+// waits up to 2 minutes for it to be recreated with the new certificate.
+func rotateTLSCertSecret() error {
+	// Get original Secret.
+	tlsSecret, err := cs.Secrets(ntoconfig.OperatorNamespace()).Get(context.TODO(), "node-tuning-operator-tls", metav1.GetOptions{})
+	if err != nil {
+		util.Logf("Error getting secret/node-tuning-operator-tls. May not exist yet. Err: %v", err)
+		return err
+	}
+	origCertContents := string(tlsSecret.Data["tls.crt"])
+
+	_, _, err = util.ExecAndLogCommand("oc", "delete", "-n", ntoconfig.OperatorNamespace(), "secret/node-tuning-operator-tls")
+	if err != nil {
+		util.Logf("Error deleting secret/node-tuning-operator/tls")
+		return err
+	}
+
+	// Wait for new certificate to be injected into the Secret.
+	err = wait.PollImmediate(2*time.Second, 2*time.Minute, func() (bool, error) {
+		tlsSecret, err = cs.Secrets(ntoconfig.OperatorNamespace()).Get(context.TODO(), "node-tuning-operator-tls", metav1.GetOptions{})
+		if err != nil {
+			util.Logf("Error getting secret/node-tuning-operator-tls. May not exist yet. Err: %v", err)
+			return false, nil
+		}
+		latestCertContents := string(tlsSecret.Data["tls.crt"])
+		if len(latestCertContents) > 0 && latestCertContents != origCertContents {
+			//Secret has been updated.
+			return true, nil
+		}
+		return false, nil
+	})
+	return err
+}
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -64,6 +64,26 @@ func GetTunedForNode(cs *framework.ClientSet, node *corev1.Node) (*corev1.Pod, e
 	return &podList.Items[0], nil
 }
 
+// GetNodeTuningOperator returns the node tuning operator Pod.
+// If more than one operator Pod is running will return the first Pod found.
+func GetNodeTuningOperatorPod(cs *framework.ClientSet) (*corev1.Pod, error) {
+	listOptions := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{"name": "cluster-node-tuning-operator"}).String(),
+	}
+
+	podList, err := cs.Pods(ntoconfig.OperatorNamespace()).List(context.TODO(), listOptions)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't list potential NTO operator Pods: %v", err)
+	}
+
+	if len(podList.Items) == 0 {
+		return nil, fmt.Errorf("failed to find the cluster-node-tuning-operator Pods")
+	}
+
+	// Return the first operator pod if multiple are running
+	return &podList.Items[0], nil
+}
+
 // execCommand executes command 'name' with arguments 'args' and optionally
 // ('log') logs the output.  Returns captured standard output, standard error
 // and the error returned.
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -52,3 +52,34 @@ func init() {
 		degradedState,
 	)
 }
+
+// PodLabelsUsed indicates whether the deprecated Pod label matching functionality
+// is turned on.
+func PodLabelsUsed(enable bool) {
+	if enable {
+		podLabelsUsed.Set(1)
+		return
+	}
+	podLabelsUsed.Set(0)
+}
+
+// ProfileCalculated keeps track of the number of times a given Tuned profile
+// resource was calculated for node 'nodeName'.
+func ProfileCalculated(nodeName, profileName string) {
+	profileCalculated.With(map[string]string{"node": nodeName, "profile": profileName}).Inc()
+}
+
+// RegisterVersion exposes the Operator build version.
+func RegisterVersion(version string) {
+	buildInfo.WithLabelValues(version).Set(1)
+}
+
+// Degraded sets the metric that indicates whether the operator is in degraded
+// mode or not.
+func Degraded(deg bool) {
+	if deg {
+		degradedState.Set(1)
+		return
+	}
+	degradedState.Set(0)
+}


pkg/metrics/server.go
```diff
@@ -1,25 +1,32 @@
...
 
 	"k8s.io/klog/v2"
 )
 
 const (
-	tlsCRT = "/etc/secrets/tls.crt"
-	tlsKey = "/etc/secrets/tls.key"
+	certsDir = "/etc/secrets"
+	tlsCRT   = certsDir + "/tls.crt"
+	tlsKey   = certsDir + "/tls.key"
```
fcami, line 23: Could you use filepath.Join instead ( https://pkg.go.dev/path/filepath#Join )?

pkg/metrics/server.go
```diff
@@ -37,14 +44,77 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+	}
+}
+
+func stopServer(srv *http.Server) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if err := srv.Shutdown(ctx); err != nil {
+		klog.Errorf("error or timeout stopping metrics listener: %v", err)
+	} else {
+		// TODO delete debugging lines
```
fcami, line 63: I think that comment is obsolete.

pkg/metrics/server.go
```diff
@@ -78,3 +148,66 @@ func Degraded(deg bool) {
...
+
+//Compute the md5 checksum for a file.
+func checksumFile(fName string) []byte {
+	file, err := os.Open(fName)
+	if err != nil {
+		klog.Infof("failed to open file %v for checksum: %v", fName, err)
+	}
+	defer file.Close()
+
+	hash := md5.New()
```
fcami, line 195: Could you use something "better" than MD5, as we're trying to minimizeMD5 usage in RHEL? sha256 for instance.

pkg/metrics/server.go
```diff
@@ -1,25 +1,32 @@
...
+	"gopkg.in/fsnotify.v1"
 
 	"k8s.io/klog/v2"
 )
 
 const (
-	tlsCRT = "/etc/secrets/tls.crt"
-	tlsKey = "/etc/secrets/tls.key"
+	certsDir = "/etc/secrets"
+	tlsCRT   = certsDir + "/tls.crt"
```
fcami, line 22: Minor nitpick: tlsCert seems more readable, and more aligned with tlsKey, than tlsCRT.

fcami (changes requested):
The change LGTM, 
The one concern I have is the behavior of the fsnotify watcher for the /etc/secrets directory. Is the secrets volume unmounted/mounted again on certificate renewal or are the certificate and key files deleted and then recreated inside the same volume? The umount/mount operation might throw the watcher off.
Is there a way to add a CI test for this, deleting the secrets twice?

pkg/metrics/server.go
jmencak: I think this line and the line below is the debugging that should be removed.  Anyways, still WiP and will be cleaned I believe.

jmencak (commented):
Thank you for the PR, David, much appreciated.  Overall this looks sane to me.  I didn't test the code, but I agree with François that a CI test for this would be useful.  Left some comments for inspiration/cleanup.

pkg/metrics/server.go
dagrayvid: Agreed, I had the same thought myself.

pkg/metrics/server.go
dagrayvid: I had this thought, however we can't use a function call to set a constant.

pkg/metrics/server.go
```diff
@@ -37,14 +44,77 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+	origCertChecksum := checksumFile(tlsCRT)
+	origKeyChecksum := checksumFile(tlsKey)
+
+	for {
+		select {
+		case <-stopCh:
+			stopServer(srv)
+		case event, ok := <-watcher.Events:
+			if !ok {
+				klog.Errorf("file watcher events channel closed")
```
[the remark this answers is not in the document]
dagrayvid: I think this should never happen. My first thought is to just `continue` the loop, which will mean certificate rotation is disabled (no more events from watcher channel). 

Alternatively we could kill the whole operator in the case of unexpected failures like this. Any guidance you can provide based on how other parts of the operator handle similar situations?

pkg/metrics/server.go
```diff
@@ -78,3 +148,66 @@ func Degraded(deg bool) {
...
+
+	hash := md5.New()
+
+	if _, err = io.Copy(hash, file); err != nil {
+		klog.Infof("failed to compute checksum for file %v: %v", fName, err)
+	}
+
+	return hash.Sum(nil)
+}
+
```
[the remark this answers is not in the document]
dagrayvid: Added, thanks.

pkg/metrics/server.go
dagrayvid: I had sha256 originally. I (perhaps naively) switched to md5 for the slight performance improvement because I figured security was not a concern here. I switched it back as performance really shouldn't be a problem here, and it is the current standard.

pkg/metrics/server.go
dagrayvid: Done

pkg/metrics/server.go
```diff
@@ -78,3 +148,66 @@ func Degraded(deg bool) {
...
+			klog.Infof("cert and key changed, we should restart the server.")
+			return true
+		}
+	}
+
+	klog.Infof("At least one of the cert / key files has not changed")
+	return false
+}
+
+//Compute the md5 checksum for a file.
```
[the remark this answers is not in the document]
dagrayvid: Fixed, thanks.

pkg/metrics/server.go
```diff
@@ -78,3 +148,66 @@ func Degraded(deg bool) {
 	}
 	degradedState.Set(0)
 }
+
+// Determine if the certificates have changed and need to be updated
```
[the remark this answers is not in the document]
dagrayvid: Fixed these, thanks.

pkg/metrics/server.go
```diff
@@ -78,3 +148,66 @@ func Degraded(deg bool) {
...
+		klog.Infof("certificate or key is missing or empty. Don't rotate.")
+		return false
+	} else {
+		currentCertMD5 := checksumFile(tlsCRT)
+		currentKeyMD5 := checksumFile(tlsKey)
+
+		klog.Infof("checksums before: %x, %x. checksums after: %x, %x", origCertMD5, origKeyMD5, currentCertMD5, currentKeyMD5)
+		// Check if the non-empty certificate / key files have actually changed
+		if !bytes.Equal(origCertMD5, currentCertMD5) && !bytes.Equal(origKeyMD5, currentKeyMD5) {
+			klog.Infof("cert and key changed, we should restart the server.")
```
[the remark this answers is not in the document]
dagrayvid: Fixed that, thanks... I think I wrote this before I had implemented the restart ;)

pkg/metrics/server.go
```diff
@@ -78,3 +148,66 @@ func Degraded(deg bool) {
...
+	if err != nil {
+		klog.Warningf("error checking if changed TLS key file empty / exists: %v", err)
+		return false
+	}
+
+	if !certNotEmpty || !keyNotEmpty {
+		// One of the files is missing despite some file event
+		klog.Infof("certificate or key is missing or empty. Don't rotate.")
+		return false
+	} else {
```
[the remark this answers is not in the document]
dagrayvid: Good catch, remove this nesting makes the function a lot cleaner.

pkg/metrics/server.go
```diff
@@ -78,3 +148,66 @@ func Degraded(deg bool) {
...
+
+		klog.Infof("checksums before: %x, %x. checksums after: %x, %x", origCertMD5, origKeyMD5, currentCertMD5, currentKeyMD5)
+		// Check if the non-empty certificate / key files have actually changed
+		if !bytes.Equal(origCertMD5, currentCertMD5) && !bytes.Equal(origKeyMD5, currentKeyMD5) {
+			klog.Infof("cert and key changed, we should restart the server.")
+			return true
+		}
+	}
+
+	klog.Infof("At least one of the cert / key files has not changed")
```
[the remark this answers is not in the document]
dagrayvid: This was really just here for debugging purposes, I removed this one and changed a few other logs to V(1) or V(2).

pkg/metrics/server.go
```diff
@@ -37,14 +44,77 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+	go startServer(srv, stopCh)
+
+	// Set up and start the file watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		klog.Errorf("failed to create watcher: %v", err)
+	}
+	defer watcher.Close()
+	if err := watcher.Add(certsDir); err != nil {
+		klog.Errorf("failed to add %v to watcher: %v", certsDir, err)
```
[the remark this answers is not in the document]
dagrayvid: Good idea. I added this.

pkg/metrics/server.go
```diff
@@ -37,14 +44,77 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+	} else {
+		klog.Infof("watching: %v", certsDir)
```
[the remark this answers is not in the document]
dagrayvid: Really just there for debugging. I removed this one.

pkg/metrics/server.go
```diff
@@ -37,14 +44,77 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+	for {
+		select {
+		case <-stopCh:
+			stopServer(srv)
+		case event, ok := <-watcher.Events:
+			if !ok {
+				klog.Errorf("file watcher events channel closed")
+			}
+
+			klog.Infof("file: %v, event: %v", event.Name, event.Op)
```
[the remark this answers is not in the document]
dagrayvid: Good catch. I made this V(2) as it may be useful for future debugging. I also made it so that we will skip the checksum step if the file event is chmod or remove. The other event types are create, write, and rename.

pkg/metrics/server.go
```diff
@@ -37,14 +44,77 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+				go startServer(srv, stopCh)
+			}
+		case err = <-watcher.Errors:
+			klog.Warningf("error from metrics server certificate file watcher: %v", err)
 		}
-	}()
-	<-stopCh
-	if err := srv.Shutdown(context.Background()); err != nil && err != http.ErrServerClosed {
-		klog.Errorf("error stopping metrics listener: %v", err)
+
```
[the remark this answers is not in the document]
dagrayvid: Fixed.

dagrayvid:
Added a new revision with many of the comments resolved. I added it in a new commit just to make things clear, but I can squash the commits once we are done iterating.

I decided to move functions PodLabelsUsed,  ProfileCalculated, RegisterVersion and Degraded to metrics.go rather than server.go to keep server.go specific to running the server. To me this is a more logical separation, but open to other ideas.

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
-func PodLabelsUsed(enable bool) {
-	if enable {
-		podLabelsUsed.Set(1)
+func stopServer(srv *http.Server) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if err := srv.Shutdown(ctx); err != nil {
+		klog.Errorf("error or timeout stopping metrics listener: %v", err)
+	}
+	klog.V(2).Infof("stopping metrics server without error")
```
jmencak, line 63: stopping or stopped?  This could be my English though, so you're in a better position to asses which is better.

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+
+	origCertChecksum := checksumFile(tlsCert)
+	origKeyChecksum := checksumFile(tlsKey)
+
+	for {
+		select {
+		case <-stopCh:
+			stopServer(srv)
+		case event, ok := <-watcher.Events:
+			if !ok {
```
jmencak, line 93: I wonder if we even need this check.  I have been happily runing with the detection of channel closing in `tuned` for years.  Also, my advice of add "cert/key rotation will be disabled" is probably not a good one looking more closely.  My appologies.  This might be just a temporary blip and the rotation code might be completely unaffected.  `watcher.Add()` are fatal though in my view and not checking for `nil` will bring down the whole operator so we need to be extra careful there.

jmencak (commented):
Thank you for the changes, David.  Steadily making improvements.  My main concern now is error checking and handling `SIGTERM` and shutdown gracefully.  This needs to be tested.

pkg/metrics/server.go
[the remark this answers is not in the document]
jmencak: > I think this should never happen. My first thought is to just `continue` the loop, which will mean certificate rotation is disabled (no more events from watcher channel).

As mentioned elsewhere, you're probably right the likelihood of this happening is very low.  The rest is dependent on the underlying `fsnotify` implementation which I have not studied... 

> 
> Alternatively we could kill the whole operator in the case of unexpected failures like this. Any guidance you can provide based on how other parts of the operator handle similar situations?

Something like `klog.Fatal()` (and let the operator be restarted) is certainly an option, but probably not here.  In the operator, we don't have any "retry" loops when things go really bad.  We do have one in the operand though.  Saying that, it is probably pointless to have one in the operator unless you just want to make a local one but I do not believe this is needed.

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+		klog.Errorf("failed to add %v to watcher, cert/key rotation will be disabled: %v", certsDir, err)
+	}
+
+	origCertChecksum := checksumFile(tlsCert)
+	origKeyChecksum := checksumFile(tlsKey)
+
+	for {
+		select {
+		case <-stopCh:
+			stopServer(srv)
```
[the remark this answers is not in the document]
jmencak: After a little testing, this should not block `SIGTERM` even if you do not return.  However, at exit you'll have running goroutines you don't need to have.  You can check this with `runtime.NumGoroutine()` and I still think we should `return` here (perhaps for future cleanup at exit and avoid the leak).

pkg/metrics/server.go
dagrayvid: You're definitely right this is awkward as written... It was fixed on my local copy :)

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+		case event, ok := <-watcher.Events:
+			if !ok {
+				klog.Errorf("file watcher events channel closed, cert/key rotation will be disabled")
+				continue
+			}
+
+			klog.V(2).Infof("event from filewatcher on file: %v, event: %v", event.Name, event.Op)
+
+			if event.Op != fsnotify.Chmod && event.Op != fsnotify.Remove &&
```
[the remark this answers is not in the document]
dagrayvid: Agreed it is more readable that way. I had this originally but switched it to save on lines... but on second thought, I think overall the function reads better with it.

pkg/metrics/server.go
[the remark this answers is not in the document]
dagrayvid: Yes, agreed.

pkg/metrics/server.go
```diff
@@ -1,25 +1,32 @@
...
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"gopkg.in/fsnotify.v1"
 
 	"k8s.io/klog/v2"
 )
 
 const (
-	tlsCRT = "/etc/secrets/tls.crt"
-	tlsKey = "/etc/secrets/tls.key"
+	certsDir = "/etc/secrets"
```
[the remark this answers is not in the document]
dagrayvid: I changed it to tlsSecretDir.

pkg/metrics/server.go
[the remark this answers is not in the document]
dagrayvid: This should now be fixed.

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
-			klog.Errorf("error starting metrics server: %v", err)
-		}
-	}()
-	<-stopCh
-	if err := srv.Shutdown(context.Background()); err != nil && err != http.ErrServerClosed {
-		klog.Errorf("error stopping metrics listener: %v", err)
+	return srv
+}
+
+func startServer(srv *http.Server, stopCh <-chan struct{}) {
```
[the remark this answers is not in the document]
dagrayvid: Good catch. This was leftover from my original draft of splitting the runServer function into smaller functions. The stopCh doesn't need to be here as the server will be shutdown if there is a message in stopCh.

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+
+	go startServer(srv, stopCh)
+
+	// Set up and start the file watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		klog.Errorf("failed to create watcher: %v", err)
+	}
+	defer watcher.Close()
+	if err := watcher.Add(certsDir); err != nil {
```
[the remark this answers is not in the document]
dagrayvid: This should now be fixed. If the watcher is nil certificate rotation will be "disabled" but the server will still run.

pkg/metrics/server.go
dagrayvid: I removed this check. If somehow `event` were set to nil here, I believe the worst case is that we would check if the files really changed.

dagrayvid:
> Is there a way to add a CI test for this, deleting the secrets twice?
@jmencak @fcami, I spent some time looking into options for testing certificate rotation as part of the e2e test. Originally, I was thinking of using a test pod to curl from the metrics endpoint on the `node-tuning-operator` service, but was unable to get this working using the trusted-ca bundle.... maybe I was doing it wrong :\ 

However, in order to test the full NTO metrics "stack", the idea I'm currently leaning towards is:

1. delete the tls Secret (this will result in [a new certificate and key being put into the secret](https://docs.openshift.com/container-platform/4.9/security/certificates/service-serving-certificate.html#manually-rotate-service-ca_service-serving-certificate)). 
2. wait ~1 minute
3. until success or timeout, query prometheus for an nto metric from the last ~30 seconds and make sure the result is not empty. e.g.:
```
oc exec -n openshift-monitoring -c prometheus prometheus-k8s-0 -- \
curl -s "http://localhost:9090/api/v1/query?query=nto_degraded_info\
&start=$(date -d "30 seconds ago" -u +%FT%T)&end=$(date -u +%FT%T)&step=15s" \
| jq -r
```

If certificate rotation is failing, metrics won't be reported and the result will be something like
```
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": []
  }
}
```
If metrics are being scraped from NTO successfully, `data.result` will be nonempty

WDYT? Any better ideas? Maybe I'm missing a simpler way to query prometheus from the e2e test?

fcami:
> > Is there a way to add a CI test for this, deleting the secrets twice?
> 
> @jmencak @fcami, I spent some time looking into options for testing certificate rotation as part of the e2e test. Originally, I was thinking of using a test pod to curl from the metrics endpoint on the `node-tuning-operator` service, but was unable to get this working using the trusted-ca bundle.... maybe I was doing it wrong :\
> 
> However, in order to test the full NTO metrics "stack", the idea I'm currently leaning towards is:
> 
> 1. delete the tls Secret (this will result in [a new certificate and key being put into the secret](https://docs.openshift.com/container-platform/4.9/security/certificates/service-serving-certificate.html#manually-rotate-service-ca_service-serving-certificate)).
> 2. wait ~1 minute
> 3. until success or timeout, query prometheus for an nto metric from the last ~30 seconds and make sure the result is not empty. e.g.:
> 
> ```
> oc exec -n openshift-monitoring -c prometheus prometheus-k8s-0 -- \
> curl -s "http://localhost:9090/api/v1/query?query=nto_degraded_info\
> &start=$(date -d "30 seconds ago" -u +%FT%T)&end=$(date -u +%FT%T)&step=15s" \
> | jq -r
> ```
> 
> If certificate rotation is failing, metrics won't be reported and the result will be something like
> 
> ```
> {
>   "status": "success",
>   "data": {
>     "resultType": "vector",
>     "result": []
>   }
> }
> ```
> 
> If metrics are being scraped from NTO successfully, `data.result` will be nonempty
> 
> WDYT? Any better ideas? Maybe I'm missing a simpler way to query prometheus from the e2e test?

Say we delete the secrets twice, we have cert1, cert2 and cert3.
The problem I see is that since cert1 is still valid and not part of any revocation list, the endpoint will continue working in this test even if the operator failed to load cert2 then cert3 and is still using cert1.
Prometheus will still be able to scrape data, so the test is invalid.

The only way to make sure the proper certificate is used seems to be comparing the content of the certificate in the secret with what's exposed / loaded by the operator.

dagrayvid:
> The problem I see is that since cert1 is still valid and not part of any revocation list, the endpoint will continue working in this test even if the operator failed to load cert2 then cert3 and is still using cert1.
Prometheus will still be able to scrape data, so the test is invalid.

You're absolutely right, sorry for my misunderstanding.

> The only way to make sure the proper certificate is used seems to be comparing the content of the certificate in the secret with what's exposed / loaded by the operator.

One way to do this is to get the server certificate using the openssl tool:
`openssl s_client -connect node-tuning-operator:60000`. 

Then compare the certificate with the one from the secret.

dagrayvid:
@jmencak @fcami, I added an e2e test for certificate rotation. It behaves as expected from my testing. 

I'm not thrilled with the complexity of using Go to exec an oc debug pod in the container from the openshift/tools imagestream, to run an openssl command to get the certs from the server... if you have a more elegant solution I'm open to ideas.

I also am not yet sure why @fcami suggested we should delete the secret twice. From my testing it has the exact same effect as deleting it once.

dagrayvid:
/retest 

The failed e2e-aws-operator test was due to infrastructure issues.

jmencak:
> @jmencak @fcami, I added an e2e test for certificate rotation. It behaves as expected from my testing.
> 
> I'm not thrilled with the complexity of using Go to exec an oc debug pod, using the container from the openshift/tools imagestream, to run an openssl command to get the certs from the server... if you have a more elegant solution I'm open to ideas.

Neither am I "thrilled" about that.  Without digging too much (yet), you might be able to use the `openssl` command on the `/host` from a `TuneD` container.  This works for me on my cluster, but had to use the service IP directly as it wouldn't resolve:
```
/host/usr/bin/openssl s_client -connect 10.130.0.65:60000
```
alternativel, you might be able to use the `crypto/tls` package and the raw structures it provides (again untested):
https://stackoverflow.com/questions/31751764/get-remote-ssl-certificate-in-golang

dagrayvid:
Thanks for the ideas Jiri!

> you might be able to use the crypto/tls package and the raw structures it provides (again untested):
https://stackoverflow.com/questions/31751764/get-remote-ssl-certificate-in-golang

I was originally thinking of doing something like this, but the metrics service will only be available from within the cluster, so this go code would need to run in some pod, rather than as part of the e2e test, which makes it a bit uglier IMO.

> you might be able to use the openssl command on the /host from a TuneD container. This works for me on my cluster, but had to use the service IP directly as it wouldn't resolve... Not sure what the best way to resolve the host in the e2e test is, but perhaps just resolving it in the operator pod and then using it in the openshift-tuned pod to issue the openssl command on the host might be the easiest way to solve this?

I think the service IP is just the NTO Pod IP, which we can get by inspecting the NTO Pod object. I will try this.

dagrayvid:
@jmencak I added a new version of the e2e test which uses the TuneD pod as discussed.

jmencak:
> @jmencak I added a new version of the e2e test which uses the TuneD pod as discussed.

Thank you David, will have another look at this PR soon, likely tomorrow morning.

test/e2e/basic/metrics_cert_rotation.go
```diff
@@ -49,15 +60,14 @@ var _ = ginkgo.Describe("[basic][metrics] Node Tuning Operator certificate rotat
...
 				secretCertContents := string(tlsSecret.Data["tls.crt"])
 
-				// Check server certificate using openssl tool.
-				stdout, stderr, err := util.ExecAndLogCommand("/bin/bash", "-c", "oc debug istag/tools:latest --quiet -n openshift -- openssl s_client -connect node-tuning-operator.openshift-cluster-node-tuning-operator.svc:60000 </dev/null |  sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p'")
-				if err != nil {
-					util.Logf("Error getting TLS certificate from metrics server. Stderr: %v, \n err: %v", stderr.String(), err)
-					return false, nil
-				}
-				serverCertContents := stdout.String()
+				ginkgo.By(fmt.Sprintf("Getting certificate from NTO metrics server"))
```
jmencak, line 63: Nit, consistency: `s/Getting/getting/`

test/e2e/basic/metrics_cert_rotation.go
```diff
@@ -49,15 +60,14 @@ var _ = ginkgo.Describe("[basic][metrics] Node Tuning Operator certificate rotat
...
+				operatorPodIP := operatorPod.Status.PodIP
+				opensslCmd := "/host/usr/bin/openssl s_client -connect " + operatorPodIP + ":60000 2>/dev/null </dev/null |  sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p'"
```
jmencak, line 65: `s/|  sed/| sed/` # i.e. remove one space.

test/e2e/util/util.go
```diff
@@ -64,6 +64,26 @@ func GetTunedForNode(cs *framework.ClientSet, node *corev1.Node) (*corev1.Pod, e
...
+	podList, err := cs.Pods(ntoconfig.OperatorNamespace()).List(context.TODO(), listOptions)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't list potential NTO operator pods: %v", err)
+	}
+
+	if len(podList.Items) != 1 {
+		if len(podList.Items) == 0 {
+			return nil, fmt.Errorf("failed to find the cluster-node-tunig-operator pods")
+		}
+		return nil, fmt.Errorf("too many (%d) cluster-node-tunig-operator pods", len(podList.Items))
```
jmencak, line 82: Do we need to do this?  Say we have two operators running and one waiting to be elected as the leader.  Is this what are you worried about?  Wouldn't it be better to return just the first one?  For what we currently need this would be likely be better and wouldn't fail an e2e test if two operators were running at the same time.

test/e2e/util/util.go
```diff
@@ -64,6 +64,26 @@ func GetTunedForNode(cs *framework.ClientSet, node *corev1.Node) (*corev1.Pod, e
 	return &podList.Items[0], nil
 }
 
+// GetTunedForNode returns a Pod that runs on a given node.
```
jmencak, line 67: Comment doesn't match what the function does.

test/e2e/util/util.go
```diff
@@ -64,6 +64,26 @@ func GetTunedForNode(cs *framework.ClientSet, node *corev1.Node) (*corev1.Pod, e
...
+func GetNodeTuningOperatorPod(cs *framework.ClientSet) (*corev1.Pod, error) {
+	listOptions := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{"name": "cluster-node-tuning-operator"}).String(),
+	}
+
+	podList, err := cs.Pods(ntoconfig.OperatorNamespace()).List(context.TODO(), listOptions)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't list potential NTO operator pods: %v", err)
```
jmencak, line 75: I *think* it should be `s/pods/Pods/`.  At least that what I use in the logs elsewhere.  Similarly below.

test/e2e/basic/metrics_cert_rotation.go
dagrayvid: Thanks, fixed this and others.

test/e2e/basic/metrics_cert_rotation.go
dagrayvid: Thanks, fixed.

test/e2e/util/util.go
dagrayvid: Yes, this is what I was trying to handle. Taking the first Pod is probably better... though if there are multiple running pods I'm not sure which one the metrics service might point to.

test/e2e/util/util.go
dagrayvid: Fixed, sorry about that.

test/e2e/util/util.go
dagrayvid: Fixed in a few spots, thanks.

test/e2e/util/util.go
jmencak: Aaah, good point.  I should probably leave the reviews for the morning. :)

test/e2e/util/util.go
```diff
@@ -64,6 +64,26 @@ func GetTunedForNode(cs *framework.ClientSet, node *corev1.Node) (*corev1.Pod, e
 	return &podList.Items[0], nil
 }
 
+// GetNodeTuningOperator returns the node tuning operator Pod,
+// or the first one found if multiple are running
```
jmencak, line 68: Nit: `s/are running/are running./`

dagrayvid: Fixed this, thanks.

dagrayvid:
@jmencak, I tried it with two operator pods running, and the openssl command will work regardless of which NTO pod it is connecting to (as long as they both rotate their certificates). AFAICT, this test is valid even if we don't know which NTO pod we will get, so I removed the check that the pod list is of length 1.

test/e2e/basic/metrics_cert_rotation.go
```diff
@@ -0,0 +1,77 @@
...
+var _ = ginkgo.Describe("[basic][metrics] Node Tuning Operator certificate rotation", func() {
+
+	ginkgo.Context("TLS certificate rotation", func() {
+		ginkgo.It("delete certificate Secret and check that the server restarts with latest certificate", func() {
+			const (
+				pollInterval = 5 * time.Second
+				waitDuration = 5 * time.Minute
+			)
+
+			// Delete the existing secret to force certificate rotation
```
jmencak, line 27: Nit: `s/rotation/rotation./`

test/e2e/basic/metrics_cert_rotation.go
```diff
@@ -0,0 +1,77 @@
...
+			_, _, err := util.ExecAndLogCommand("oc", "delete", "-n", ntoconfig.OperatorNamespace(), "secret/node-tuning-operator-tls")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			ginkgo.By("sleeping 5s to wait for Secret to be refreshed")
+			time.Sleep(5 * time.Second)
+
+			// Delete the secret again to force certificate rotation
```
jmencak, line 33: I didn't quite get why we're doing this twice.  Is this based on what @fcami suggested?  Can we have an explanation?

test/e2e/basic/metrics_cert_rotation.go
```diff
@@ -0,0 +1,77 @@
...
+			const (
+				pollInterval = 5 * time.Second
+				waitDuration = 5 * time.Minute
+			)
+
+			// Delete the existing secret to force certificate rotation
+			_, _, err := util.ExecAndLogCommand("oc", "delete", "-n", ntoconfig.OperatorNamespace(), "secret/node-tuning-operator-tls")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			ginkgo.By("sleeping 5s to wait for Secret to be refreshed")
+			time.Sleep(5 * time.Second)
```
jmencak, line 31: Can we be reasonably sure the secrets are refreshed within in 5 seconds?  Is it possible to do this in a cleaner way without the sleep?  Ignore if it is too hard to implement.

jmencak (commented):
Thank you David.  This looks very close to being ready to be merged.  Only a few nits and questions.  I'd also appreciated @fcami input.

test/e2e/basic/metrics_cert_rotation.go
fcami: If the fsnotify watches a file that's deleted, it will report the delete event, which might trigger reloading the certificate properly. However, rotating the certificate a second time will not work since the fsnotify does not watch the new file.
In short, testing that renewing the cert twice results in the last certificate being loaded protects against a refactoring that would simplify the fsnotify logic too much.
Does that help @jmencak ?

jmencak: It does, thank you.  I need to look into this more.  In any case, an explanation like this should go in the comments.

test/e2e/basic/metrics_cert_rotation.go
dagrayvid: I rewrote this part into a function that will delete it, then check every 2 seconds for up to 2 minutes whether it is recreated. It seems to usually be back by the second check. I think this is "better" so we don't have to rely on it being refreshed within 5 seconds, but it will also not wait as long most of the time.

(Will update PR once I address the error handling in the server restart)

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+			return
+		case event := <-watcher.Events:
+			klog.V(2).Infof("event from filewatcher on file: %v, event: %v", event.Name, event.Op)
+
+			if event.Op == fsnotify.Chmod || event.Op == fsnotify.Remove {
+				continue
+			}
+
+			if certsChanged(origCertChecksum, origKeyChecksum) {
+
```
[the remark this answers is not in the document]
dagrayvid: Fixed, thanks. It seems I tend to subconsciously add extra whitespace in my code.

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
 }
 
-// ProfileCalculated keeps track of the number of times a given Tuned profile
-// resource was calculated for node 'nodeName'.
-func ProfileCalculated(nodeName, profileName string) {
-	profileCalculated.With(map[string]string{"node": nodeName, "profile": profileName}).Inc()
+// Determine if the certificates have changed and need to be updated.
+// Returns true if both files have changed AND neither is an empty file.
+func certsChanged(origCertChecksum []byte, origKeyChecksum []byte) bool {
+	// Check if both files exist
```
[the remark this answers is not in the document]
dagrayvid: Fixed

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+		return true
+	}
+
+	return false
 }
 
-// RegisterVersion exposes the Operator build version.
-func RegisterVersion(version string) {
-	buildInfo.WithLabelValues(version).Set(1)
+// Compute the sha256 checksum for a file.
```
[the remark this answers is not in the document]
dagrayvid: Fixed

test/e2e/basic/metrics_cert_rotation.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
+	util "github.com/openshift/cluster-node-tuning-operator/test/e2e/util"
+)
+
+var _ = ginkgo.Describe("[basic][metrics] Node Tuning Operator certificate rotation", func() {
+
```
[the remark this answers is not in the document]
dagrayvid: No reason. Fixed.

test/e2e/basic/metrics_cert_rotation.go
dagrayvid: Fixed.

test/e2e/basic/metrics_cert_rotation.go
dagrayvid: I added the explanation to the second deletion.

test/e2e/basic/metrics_cert_rotation.go
dagrayvid: Pushed the change to this part.

dagrayvid:
I squashed the commits into one. Retitling to include BZ.

/retitle Bug 2026387: Handle certificate rotation in pkg/metrics/server.go

pkg/metrics/server.go
```diff
@@ -37,44 +44,137 @@ func RunServer(port int, stopCh <-chan struct{}) {
...
+
+			if certsChanged(origCertChecksum, origKeyChecksum) {
+
+				// Update file checksums with latest files.
+				origCertChecksum = checksumFile(tlsCert)
+				origKeyChecksum = checksumFile(tlsKey)
+
+				// restart server
+				klog.Infof("restarting metrics server to rotate certificates")
+				stopServer(srv)
```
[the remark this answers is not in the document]
dagrayvid: As discussed on slack, the alternatives don't seem much better so we will leave it for now.

dagrayvid:
/hold

I want to do some final testing, but ready for /lgtm

jmencak:
As mentioned over chat earlier, the last nit I see is `s/pod if multiple are running/pod if multiple are running./` but I can live with that or can be changed for consistency later.  Thank you again for the PR, it 
/lgtm
now.

dagrayvid:
Thanks @jmencak.

/unhold

PR #289, opened 2021-11-09
WIP: Refactor controller runtime

cynepco3hahue:

PR #302, opened 2021-12-15
Refactor NTO to use controller runtime lib

yanirq:
Refactor cluster node tuning operator to use controller runtime library ([release 0.11](https://github.com/kubernetes-sigs/controller-runtime/tree/release-0.11)).
The functionality is internal only and replaces the direct application of a controller with the controller runtime scheme.

The new controller(s) structure:
- 2 Controllers under one manager : CVO controller and Tuned controller using controller runtime library.
- Metrics server added as runnable to the controller runtime manager. 
- Tuned daemon controller left untouched. 

Current implementation that might be subject to change depending on the operator performance (or reviews): 
- Tuned controller does not keep internal maps for tracking node names and pods. It watches pod and node label changes and in the reconcile logic will list all nodes before syncing tuned profiles.
- Pod labels watch will run always but will never act on changes if no tuned CR has pod type.

Pending tasks checklist: 
- [ ] Ensure metric reporting works as expected (not only on e2e)
- [ ] e2e and unit test coverage should be added (either in this PR or the following)
- [ ]  Unused methods will be removed once the PR is in an acceptable review state. 
- [x] Add profile deletion when a node is deleted
- [x] Test managed state
- [ ] Test operator's performance vs previous implementation
- [ ] Test all status reporting scenarios
- [x] Client-go leader election changes applied correctly

This is also a preliminary work to set up the stage for moving [Performance addons operator ](https://github.com/openshift-kni/performance-addon-operators)under NTO as documented here: https://github.com/openshift/enhancements/pull/867


jmencak:
Congrats on the `e2e-aws-operator ` tests passing @yanirq , We will start a review this week.
/cc @dagrayvid

yanirq:
> Congrats on the `e2e-aws-operator ` tests passing @yanirq , We will start a review this week. /cc @dagrayvid

All the previous PRs are now closed. 
There is a checklist still to be completed but reviews should definitely start asap. 
You will notice that there are some` //TODO `sections with alternatives to the current implementation in some sections.

/cc @cynepco3hahue

Dockerfile.rhel8
```diff
@@ -20,7 +20,7 @@ RUN INSTALL_PKGS=" \
 
 FROM registry.ci.openshift.org/ocp/4.10:base
 COPY --from=builder /go/src/github.com/openshift/cluster-node-tuning-operator/_output/cluster-node-tuning-operator /usr/bin/
-COPY manifests/*.yaml manifests/image-references /manifests
+COPY manifests/*.yaml manifests/image-references /manifests/
```
dagrayvid, line 23: I'm glad to see I'm not the only one who hit this :)

pkg/operator/controller.go
```diff
@@ -1,481 +1,302 @@
...
-				c.workqueue.Forget(obj)
-				klog.Errorf("expected wqKey in workqueue but got %#v", obj)
-				return
-			}
+	// Fetch the Tuned instance
+	tunedInstance := &tunedv1.Tuned{}
+	err := r.Get(ctx, req.NamespacedName, tunedInstance)
+	if err != nil {
+		// for deletion events perform all the needed syncs as a regular event.
+		if !errors.IsNotFound(err) {
```
cynepco3hahue, line 208: if err equals to `IsNotFound` we should exit the reconcile loop early

pkg/operator/controller.go
```diff
@@ -1,481 +1,302 @@
...
-		err = c.syncProfile(cr, key.name)
-		if err != nil {
-			return fmt.Errorf("failed to sync Profile %s: %v", key.name, err)
-		}
-		return nil
-
-	default:
-	}
-
 	// Tuned CR changed and the operator components need to be managed.
```
cynepco3hahue, line 252: probably we should update the comment to reflect what really happens

pkg/operator/controller.go
```diff
@@ -785,341 +572,90 @@ func (c *Controller) getMachineConfigNamesForTuned() (map[string]bool, error) {
...
-}
-
-func (c *Controller) removeResources() error {
+func (r *TunedReconciler) removeResources(ctx context.Context) error {
 	var lastErr error
+
 	dsMf := ntomf.TunedDaemonSet()
-	ctx := context.TODO()
+	ds := &appsv1.DaemonSet{}
+	key := types.NamespacedName{
```
cynepco3hahue, line 609: The deletion removes the DS but the operator will re-create it.

cynepco3hahue (commented):
So quick initial review.

cynepco3hahue:
The upgrade lane is failing with the event related to tuning operator, I can see that CO status does not container all relevant information
```
"conditions": [
                    {
                        "lastTransitionTime": "2022-01-11T11:45:37Z",
                        "message": "Cluster has deployed \"4.10.0-0.ci.test-2022-01-11-101808-ci-op-wm3j5379-latest\"",
                        "reason": "AsExpected",
                        "status": "True",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": "2022-01-11T11:45:37Z",
                        "message": "Cluster version is \"4.10.0-0.ci.test-2022-01-11-101808-ci-op-wm3j5379-latest\"",
                        "reason": "AsExpected",
                        "status": "False",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2022-01-11T11:45:37Z",
                        "message": "DaemonSet \"tuned\" available",
                        "reason": "AsExpected",
                        "status": "False",
                        "type": "Degraded"
                    }
                ],
```
it does not include upgradeable status.

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -1,64 +1,149 @@
...
-		go metrics.RunServer(metrics.MetricsPort, stopCh)
-		metrics.RegisterVersion(version.Version)
-		controller, err := operator.NewController()
+		// we have two namespaces that we need to watch
+		// 1. NTO namespace - for NTO resources
+		// 2. None namespace - for cluster wide resources
+		ntoNamespace := config.OperatorNamespace()
+		namespaces := []string{
+			ntoNamespace,
+			metav1.NamespaceNone,
```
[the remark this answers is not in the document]
yanirq: I understand that its because  `MultiNamespacedCacheBuilder` by default creates a global cache for cluster scoped resource. But we also have a need to watch pods in all namespaces when tuned CRs are pod typed driven. 
In that case would you also recommend to remove the `metav1.NamespaceNone` from the cache builder ?

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -1,64 +1,149 @@
...
 			os.Exit(0)
 		}
 
-		go metrics.RunServer(metrics.MetricsPort, stopCh)
-		metrics.RegisterVersion(version.Version)
-		controller, err := operator.NewController()
+		// we have two namespaces that we need to watch
+		// 1. NTO namespace - for NTO resources
+		// 2. None namespace - for cluster wide resources
+		ntoNamespace := config.OperatorNamespace()
```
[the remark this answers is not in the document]
yanirq: At this point the metrics configuration is under the server.go class which is added as a runnable to the manager.

pkg/metrics/server.go
```diff
@@ -63,23 +66,27 @@ func stopServer(srv *http.Server) {
 	}
 }
 
+func (Server) Start(ctx context.Context) error {
+	return RunServer(MetricsPort, ctx)
+}
+
 // RunServer starts the server, and watches the tlsCert and tlsKey for certificate changes.
-func RunServer(port int, stopCh <-chan struct{}) {
+func RunServer(port int, ctx context.Context) error {
```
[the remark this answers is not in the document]
yanirq: The question is if the metrics server provided by the controller runtime will have all the functionality encompassed in the server we have in hand here. Tracking the history in NTO I saw added functionality to the server in hand.
We need to understand if the server here can be replaced completely with the controller runtime one or the diff is too big.
@jmencak  @dagrayvid   this might need to be revisited in this PR or we can refactor the server move in another PR.
At this point of time the metrics reporting is not fully tested (e2e current coverage might not be enough)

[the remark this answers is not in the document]
cynepco3hahue: I use the same metrics, but drop it under the different endpoint, see - https://github.com/openshift/cluster-node-tuning-operator/pull/289/files#diff-b1da0f702ee090192a7506701ce502f780265a647ee3a7b29ac0ce37d2fea1aaR116

cmd/cluster-node-tuning-operator/main.go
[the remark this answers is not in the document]
cynepco3hahue: mm good point, let leave it for now

pkg/operator/controller.go
```diff
@@ -1,481 +1,302 @@
...
-		return nil, err
+			return !reflect.DeepEqual(profileOld.Status.Conditions, profileNew.Status.Conditions)
+		},
 	}
 
-	// ClusterOperator
-	controller.clients.ConfigV1Client, err = configv1client.NewForConfig(controller.kubeconfig)
-	if err != nil {
-		return nil, err
+	podPredicates := predicate.Funcs{
```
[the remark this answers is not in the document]
yanirq: these are not the same perdicates. But I do notice I have a bug in the node watch that uses this perdicat. ill remove it

pkg/operator/controller.go
```diff
@@ -1,481 +1,302 @@
...
-	controller.clients.MC, err = mcfgclientset.NewForConfig(controller.kubeconfig)
+	err := ctrl.NewControllerManagedBy(mgr).
+		For(&tunedv1.Tuned{}).
+		Owns(&tunedv1.Profile{}, builder.WithPredicates(profilePredicates)).
+		Watches(
+			&source.Kind{Type: &corev1.Node{}},
+			handler.EnqueueRequestsFromMapFunc(r.defaultTunedRequests),
+			//TODO - consider using builder.OnlyMetadata instead these perdicats
+			builder.WithPredicates(podPredicates, predicate.Or(
+				predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{}))).
```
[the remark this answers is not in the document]
yanirq: its usually used in conjunction with the labelsChanged predicate but this is probably redundant in our case. I will clean it up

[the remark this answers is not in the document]
yanirq: ill also need to change Or to And here

pkg/operator/controller.go
```diff
@@ -1,481 +1,302 @@
...
+		Watches(
+			&source.Kind{Type: &corev1.Pod{}},
+			handler.EnqueueRequestsFromMapFunc(r.podLabelsToTuned),
+			builder.WithPredicates(podPredicates,
+				predicate.Or(
+					predicate.Funcs{UpdateFunc: func(e event.UpdateEvent) bool {
+						return r.state.enablePodEvents
+					}},
+					predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{}))).
```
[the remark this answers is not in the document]
yanirq: same as above

pkg/operator/controller.go
```diff
@@ -1,481 +1,302 @@
...
+		Watches(
+			&source.Kind{Type: &mcov1.MachineConfigPool{}},
+			handler.EnqueueRequestsFromMapFunc(r.defaultTunedRequests)).
```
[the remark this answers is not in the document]
yanirq: This is a good point and brings me to a question :
@jmencak  If an MCP status has changed and now degraded for example , will we want the another recommend profile to kick in? (not an existing functionality at the moment) 
If the answer is no then we can filter out status update events.  
If there are more filtering events that come to mind feel free to share them here.

pkg/operator/controller.go
yanirq: Since we are using a tree based structure for tuned -> profiles, when a tuned CR is deleted we need to go through the whole sync to update the relevant artifacts.  you will notice that the sync work is mostly done against the default tuned CR and not the reconciled CR itself (it is just the trigger) 
We are Listing all the tuned CRs to get the recommended profiles down the pipeline here .
I'm not entirely happy this behavior and open for suggestions.

pkg/operator/controller.go
yanirq: I would just remove it in our case

pkg/operator/controller.go
yanirq: good point.  this is handling for the ManagementState : Removed  but the CVO reconciler will just recreate it.  Probably this logic should be applied and synced on the CVO controller .

pkg/operator/controller.go
```diff
@@ -1,481 +1,302 @@
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return r.state.enablePodEvents
+		},
 	}
 
-	// Tuned
-	controller.clients.Tuned, err = tunedset.NewForConfig(controller.kubeconfig)
-	if err != nil {
-		return nil, err
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &corev1.Pod{}, "spec.nodeName", func(o client.Object) []string {
```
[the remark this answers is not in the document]
yanirq: added a comment

jmencak (commented):
Thank you for the PR, Yanir.  On the whole, I see changes for the better in store for us.  To review this PR in one go fully without melting one's brain is quite impossible, so here is at least a start.  I'll have another look later -- I didn't go over `controller.go/reconciler.go` at all.

Important
  - Upgrades don't work for me.  After upgrading the daemonset is stuck in its old version.


Nits/consistency
  - For error reporting (including logging), please use lowercase letters.  E.g.:
    klog.Infof("succeeded to 
    not
    klog.Infof("Succeeded to ...
    unless the message starts with something that needs an uppercase letter such as "TuneD/Tuned" or similar.

  - Comments starting on a new line prior a code block that are beginning a to form a sentence
    should start with a capital letter and end by a dot.  Example:
    // we have two namespaces that we need to watch
    vs.
    // We have two namespaces that we need to watch:

  - We are trying to end a longer comment that is a "sentence" by a dot.  Example: "# Needed by new leases mechanism.".
    Can we please follow this convention in the code?

pkg/operator/controller.go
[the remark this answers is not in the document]
jmencak: > If an MCP status has changed and now degraded for example , will we want the another recommend profile to kick in? 

That sounds like overengineering to me, so I'd say the answer is no.

pkg/metrics/server.go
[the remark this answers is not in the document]
dagrayvid: It seems most operators which use controller-runtime use the kube-rbac-proxy sidecar container to protect the metrics endpoint rather than handling the TLS within the operator. It looks like kube-rbac-proxy does [handle certificate changes](https://github.com/brancz/kube-rbac-proxy/pull/47/files).

I think the cluster-ingress-operator is a good example. The manifests ([servicemonitor](https://github.com/openshift/cluster-ingress-operator/blob/master/manifests/0000_90_ingress-operator_02_servicemonitor.yaml), [deployment](https://github.com/openshift/cluster-ingress-operator/blob/0d5f49acfb118d7a6c553709b730ab1e7b2a6f6c/manifests/02-deployment.yaml#L79-L98), [service](https://github.com/openshift/cluster-ingress-operator/blob/master/manifests/01-service.yaml) look how I would expect. 

I don't fully understand the reasoning for the way the controller-runtime prometheus registry is [combined with the ingress-operator's](https://github.com/openshift/cluster-ingress-operator/blob/0d5f49acfb118d7a6c553709b730ab1e7b2a6f6c/pkg/operator/metrics.go#L13). I was hoping it would be as simple as what is described here https://book.kubebuilder.io/reference/metrics.html#publishing-additional-metrics, but the comment in the ingress operator makes it sound like it might need to be a bit more complicated.

manifests/30-monitoring.yaml
```diff
@@ -85,7 +85,7 @@ spec:
   - targetPort: 60000
     interval: 60s
     scheme: https
-    path: /metrics
+    path: /nto-metrics
```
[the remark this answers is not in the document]
yanirq: this is a relic from an attempt to use the controller runtime  for metrics.  Will revert.

yanirq:
> Does the controller-runtime have the functionality to reload after certificate updates? That was the key change recently done by David. If so, we should probably switch to keep things simple/managed by controller-runtime, but I'd say not necessarily in this PR.

It seems that it contains support for cert updates : https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/internal/certwatcher  but if the current implementation works with the runnable server then we can leave it to another PR indeed

pkg/operator/controller.go
yanirq: @cynepco3hahue   the CVO reconciler logic is now changed to first act on managed state.
It will first inspect the ManagementState. In case the operator is under Unmanaged or Removed: update the status and take no further actions.
see https://github.com/openshift/cluster-node-tuning-operator/pull/302/commits/81de5a6f332cc0eed5e106187dfebf9510a499dd

pkg/operator/controller.go
```diff
@@ -1,484 +1,331 @@
...
+		profile := &tunedv1.Profile{}
+		key := types.NamespacedName{
+			Namespace: ntoconfig.OperatorNamespace(),
+			Name:      node.Name,
+		}
+		if err := r.Get(context.TODO(), key, profile); err != nil {
+			klog.Errorf("could not get profile for node %s deletion event %v", node.Name, err)
+			return nil
+		}
+		if err := r.Delete(context.TODO(), profile); err != nil {
```
cynepco3hahue, line 142: I remember we discussed it, WDT about moving deletion logic to reconcile loop?
I think one of the reasons why you should not delete the object here is because some of the commands can fail and you do not have control to re-enter the reconcile loop.

pkg/operator/controller.go
```diff
@@ -1,484 +1,331 @@
...
+			builder.WithPredicates(predicate.LabelChangedPredicate{})).
+		Watches(
+			&source.Kind{Type: &corev1.Pod{}},
+			handler.EnqueueRequestsFromMapFunc(r.podLabelsToTuned),
+			builder.WithPredicates(podPredicates,
+				predicate.And(
+					predicate.Funcs{UpdateFunc: func(e event.UpdateEvent) bool {
+						return r.state.enablePodEvents
+					}}, predicate.LabelChangedPredicate{}))).
+		Watches(
```
cynepco3hahue, line 119: What about using predicates for MachineConfig pools, I think we are interested in some specific changes under the MCP.

pkg/operator/controller.go
cynepco3hahue: ok works for me

pkg/operator/controller.go
```diff
@@ -1,484 +1,331 @@
...
-				return fmt.Errorf("failed to get Tuned %s: %v", tunedv1.TunedDefaultResourceName, err)
-			}
-		}
+	if err != nil {
+		return fmt.Errorf("failed to get default Tuned CR: %v", err)
 	}
-	// We have the default Tuned custom resource (cr)
 
-	switch cr.Spec.ManagementState {
+	switch defaultTuned.Spec.ManagementState {
```
cynepco3hahue, line 230: @jmencak Do you know if someone is really using management state? I think most of OpenShift controllers do not use it.

pkg/metrics/server.go
cynepco3hahue, line 74: Can you please specify where are we calling this method?

pkg/operator/controller.go
yanirq: will be added in the next push

pkg/operator/controller.go
yanirq: will be fixed in the next push

pkg/metrics/server.go
yanirq: called here:
https://github.com/openshift/cluster-node-tuning-operator/blob/d534c86bfb48e2245dd4556e2bb2c1b2aa1119ff/pkg/metrics/server.go#L69
As a runnable here: 
https://github.com/openshift/cluster-node-tuning-operator/blob/d534c86bfb48e2245dd4556e2bb2c1b2aa1119ff/cmd/cluster-node-tuning-operator/main.go#L123
Started here:
https://github.com/openshift/cluster-node-tuning-operator/blob/d534c86bfb48e2245dd4556e2bb2c1b2aa1119ff/cmd/cluster-node-tuning-operator/main.go#L140

cynepco3hahue: oh, it's a nice one.

pkg/operator/controller.go
jmencak: > @jmencak Do you know if someone is really using management state? I think most of OpenShift controllers do not use it.

Not sure what you're asking.  Customers using this functionality in NTO or other operators apart from NTO?  If the former, then no, I do not know.  If the latter, yes, there are OpenShift operators that implement this (dns-operator, cluster-image-registry-operator, cluster-kube-apiserver-operator, cluster-network-operator, ...).  If they do not, then they probably do not comply.

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+		klog.Errorf("failed to get Default Tuned %s: %v", crMf.Name, err)
+		return err
+	}
+	// Tuned resource found, check whether we need to update it
+	if reflect.DeepEqual(crMf.Spec.Profile, cr.Spec.Profile) &&
+		reflect.DeepEqual(crMf.Spec.Recommend, cr.Spec.Recommend) {
+		klog.V(2).Infof("syncTunedDefault(): Tuned %s doesn't need updating", crMf.Name)
+		return nil
+	}
+	cr = cr.DeepCopy() // never update the objects from cache
```
jmencak, line 392: Is this still needed or does the controller-runtime client handle this by not returning objects from cache?

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+				"The operator is removed",
+			); err != nil {
+				klog.Errorf("failed to update the cluster operator %q status conditions", instance.Name)
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{}, nil
+		}
+	}
+
+	klog.Infof("reconciling ClusterOperator %q", instance.Name)
```
jmencak, line 231: Can we drop verbosity here?  This message is spamming the operator logs.

pkg/operator/controller.go
```diff
@@ -1,484 +1,312 @@
...
+	}})
+	return requests
+}
+
+// Reconcile reads that state of the cluster for a Tuned object and makes changes based on the state read and what is in the Tuned.Spec .
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *TunedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	klog.Infof("reconciling Tuned")
```
jmencak, line 178: Can we drop verbosity here?  This message is spamming the operator logs.

jmencak (commented):
I had a quick look at the `reconciller.go`.  Looks ok, but I'll need to go through it a bit more carefully.  I've also tested the Managed/Unmanaged/Removed functionality.  Upgrades by updating the operator manifest now also work, so thank you for the changes so far.  Still haven't looked at the `controller.go` rewrite at all.  Will try to ASAP.

jmencak:
I did some preliminary "performance" testing of the current implementation (`b922bdda`) and the old one (`3c5760e9`).  Overall, the numbers look already quite good for the current implementation:

Current implementation, NTO idle (checking for utime(14)/stime(15) in `/proc/<operator-pid>/stat`).  Lower (delta) is better.
```
2022-24-01 12:23:19: 492 242
2022-24-01 12:24:19: 495 244
2022-24-01 12:25:19: 501 247
2022-24-01 12:26:19: 502 249
2022-24-01 12:27:19: 506 250
2022-24-01 12:28:19: 508 251
2022-24-01 12:29:19: 512 254
2022-24-01 12:30:19: 515 256
2022-24-01 12:31:19: 518 258
2022-24-01 12:32:19: 520 259
2022-24-01 12:33:19: 524 260
2022-24-01 12:34:19: 525 262
2022-24-01 12:35:19: 530 265
2022-24-01 12:36:19: 532 267
2022-24-01 12:37:19: 536 268
2022-24-01 12:38:19: 538 269
2022-24-01 12:39:19: 542 271
2022-24-01 12:40:19: 545 273
2022-24-01 12:41:19: 549 274
2022-24-01 12:42:19: 551 276
2022-24-01 12:43:19: 554 277
----------------------------
delta:     20 min  :  62  35
```
During creating 1000 pods and using a pod-label functionality.
```
2022-24-01 12:55:13: 691 309
2022-24-01 12:56:13: 837 329
2022-24-01 12:57:13: 996 346
2022-24-01 12:58:13: 1125 361
=============================
delta:     20 min  :  434 52
```

Old implementation

```
Idle CPU (20 minute run):
~~~~~~~~~~~~~~~~~~~~~~~~~
2022-24-01 13:09:24: 63 12
2022-24-01 13:10:24: 65 13
2022-24-01 13:11:24: 72 15
2022-24-01 13:12:24: 74 16
2022-24-01 13:13:24: 79 19
2022-24-01 13:14:24: 80 19
2022-24-01 13:15:24: 85 21
2022-24-01 13:16:24: 87 21
2022-24-01 13:17:24: 89 22
2022-24-01 13:18:24: 92 23
2022-24-01 13:19:24: 95 23
2022-24-01 13:20:24: 99 24
2022-24-01 13:21:24: 103 25
2022-24-01 13:22:24: 106 26
2022-24-01 13:23:24: 108 27
2022-24-01 13:24:24: 111 27
2022-24-01 13:25:24: 114 28
2022-24-01 13:26:24: 116 28
2022-24-01 13:27:24: 118 29
2022-24-01 13:28:24: 119 29
2022-24-01 13:29:25: 123 30
----------------------------
delta:     20 min  :  60 18
```
During creating 1000 pods and using a pod-label functionality.

```
2022-24-01 13:35:25: 518 87
2022-24-01 13:36:25: 772 115
2022-24-01 13:37:25: 1024 145
2022-24-01 13:38:25: 1289 177
=============================
delta:     20 min  :  771 90
```

So the old implementation seems better at "idle CPU utilization" when the operator "does nothing", while it performs worse with the TuneD pod matching functionality turned on.  I wonder if we can make some improvements to the current implementation wrt. `stime` which seems to be performing %-wise quite a bit worse than the old implementation.  Overall, however, I believe the results already look quite good.

Wrt. memory use, I haven't seen any major difference.

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+const ClusterOperatorName = "node-tuning"
+
+// Reconciler reconciles a Tuned object and creates the tuned daemon set.
+type Reconciler struct {
+	Client    client.Client
+	Scheme    *runtime.Scheme
+	Namespace string
+}
+
+// SetupWithManager creates a new PerformanceProfile Controller and adds it to the Manager.
```
dagrayvid, line 45: If I understand correctly, this  SetupWithManager creates a new Tuned Controller and adds it to the manager. 
```suggestion
// SetupWithManager creates a new Tuned Controller and adds it to the Manager.
```

jmencak (commented):
I quickly went through the new `controller.go` code.  I pointed out only nits so far.  I still need to go through it again (thoroughly) to understand the new semantics a bit more.  Thank you for the changes in the old code that help consistency.

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+						Namespace: ntoconfig.OperatorNamespace(),
+					}},
+				}
+			}),
+			builder.WithPredicates(profilePredicates)).
+		Complete(r)
+	if err != nil {
+		return err
+	}
+	//TODO - not sure this is the right place
```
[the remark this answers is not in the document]
dagrayvid: Well, it definitely works as is, but I'm not sure it's logical. Why not just keep it in main.go?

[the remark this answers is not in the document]
yanirq: Because we need the server to first start and then register the version. 
This is not possible in the current implementation since after starting the manager :  
https://github.com/openshift/cluster-node-tuning-operator/blob/b922bddae5c88ba489020ba7f6b235fc13d00ec3/cmd/cluster-node-tuning-operator/main.go#L140
We cannot add additional functionality right after that command.

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,402 @@
...
+	return nil
+}
+
+func validateUpdateEvent(e *event.UpdateEvent) bool {
+	if e.ObjectOld == nil {
+		klog.Error("update event has no old runtime object to update")
+		return false
+	}
+	if e.ObjectNew == nil {
+		klog.Error("update event has no new runtime object for update")
```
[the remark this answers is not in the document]
yanirq: No special reason. 
Changing both to this format : 
update event has no new/old runtime object

pkg/operator/reconciler.go
yanirq: The responsibilities of the controllers are commented above each `Reconcile` func.
This is the template controller runtime had in the past but we can change it if we wish.

pkg/operator/reconciler.go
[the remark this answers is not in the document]
dagrayvid: @yanirq That is what I thought initially, but in this current location in reconciler.go `RegisterVersion(version.Version)` is being called before mgr.Start, so it would be equivalent to put it in main.go (before mgr.Start). 

I don't think RegisterVersion needs to be called after mgr.Start because it the metrics will be stored in the local prometheus registry until the server comes up, at which point they will be scraped by OCP monitoring. I did a few tests to confirm this.

pkg/operator/reconciler.go
yanirq: We still need this for handling mutated objects for updates (@cynepco3hahue keep me honest here)
Maybe we should change the name of the recipient to  mutatedCr / mutatedDefaultTuned

pkg/operator/reconciler.go
dagrayvid: I think for this PR, we should at least correct the reference to `PerformanceProfile` in this comment.

yanirq: already done in the last commit https://github.com/openshift/cluster-node-tuning-operator/pull/302/commits/329389d756c0e6317ce4d13401fde7c25a98a914

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -1,64 +1,149 @@
+/*
```
jmencak, line 1: Let's remove the license text in this file too please.  I believe the `LICENSE` file is enough.

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -1,64 +1,149 @@
...
 	"github.com/openshift/cluster-node-tuning-operator/pkg/tuned"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/util"
 	"github.com/openshift/cluster-node-tuning-operator/version"
 )
 
+const (
+	operandFilename  = "openshift-tuned"
+	operatorFilename = "cluster-node-tuning-operator"
+	metricsHost      = "0.0.0.0"
+	leaderElectionID = "node-tuning-operator-lock"
```
jmencak, line 51: Duplicate constant. Already defined in `pkg/config/config.go`.

pkg/operator/status.go
```diff
@@ -323,3 +322,105 @@ func conditionTrue(conditions []configv1.ClusterOperatorStatusCondition, condTyp
...
+				Type:               c.Type,
+				Status:             configv1.ConditionTrue,
+				LastTransitionTime: now,
+				Reason:             reason,
+				Message:            message,
+			}
+			continue
+		}
+
+		// Once we update the available condition we also should update the cluster operator to be upgradeable except the unmanaged state.
```
jmencak, line 348: Do you mean:
`s/be upgradeable except the unmanaged state/be upgradeable unless the operator is in Unmanaged state/`
?  In any case, I still think we probably do not want to set the `OperatorUpgradeable` condition.

jmencak (commented):
I had another look at the PR.  Mostly nits, but one thing I'm really uneasy about is the new `updateClusterOperatorStatus()` code and associated functions.  It seems unnecessarily complicated and also seems to change the transition time for all conditions irrespective whether the state flips or not.. Why do we need the new functions? /cc @dagrayvid

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -1,64 +1,149 @@
...
+			Namespace:               ntoNamespace,
+		})
+
 		if err != nil {
-			klog.Fatal(err)
+			klog.Exit(err)
 		}
 
-		controller.BecomeLeader(stopCh)
+		mgr.Add(metrics.Server{})
```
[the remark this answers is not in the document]
yanirq: The class names where kept to have some sort of a diff (at least the controller)
But now I intend to change the names to:
reconciler.go -> cvo_controller.go
controller.go -> tuned_controller.go
The struct names will be changed accordingly to be more clear.

More detailed comments will be added in main to explain that, but in general
cvo_controller - reconciles requests for CVO objects. It owns (default) tuned and deamonset it creates.
It also watches for all other tuned objects and profiles in order to update the status. 

tuned_controller - reconciles requests for tuned objects and performs the logic for syncing profiles/nodes/rendered tuned in correlation with machine configs. 

I agree that there would be some more cleanups to make the code more clear and move/remove some code sections from one controller or the other.
I do think that having 2 reconciles / controllers should be the form of implementation here but I'm open for changes
(cc @cynepco3hahue)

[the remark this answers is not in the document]
jmencak: > The class names were kept to have some sort of a diff (at least the controller) but now I intend to change the names to: reconciler.go -> cvo_controller.go controller.go -> tuned_controller.go The struct names will be changed accordingly to be more clear.

It is still not clear to me why we need two controllers, but I'm not against it as it might aid modularity and a clear separation of tasks.  But the "tasks" need to be clear, agreed and make sense.  If we really do need two (or even more controllers), then perhaps having a `controller` prefix to separate them would be better.  E.g. `controller_whatever.go`.  But `controller_cvo.go` still seems like a strange name and misleading.  We need to come up with a better name if we do split it.  I do not have any better suggestions at this point.

pkg/operator/controller.go
```diff
@@ -3,482 +3,293 @@ package operator
...
+		Owns(&tunedv1.Profile{}, builder.WithPredicates(profilePredicates)).
+		Watches(
+			&source.Kind{Type: &corev1.Node{}},
+			handler.EnqueueRequestsFromMapFunc(r.defaultTunedRequests),
+			//TODO - consider using builder.OnlyMetadata instead these perdicats
+			builder.WithPredicates(predicate.LabelChangedPredicate{})).
+		Watches(
+			&source.Kind{Type: &corev1.Pod{}},
+			handler.EnqueueRequestsFromMapFunc(r.podLabelsToTuned),
+			builder.WithPredicates(podPredicates,
```
jmencak, line 109: I did a little bit of testing when creating 1000 pods with the operator config when pod-matching functionality is not in use and the impact on the CPU is noticeable.  This is a clear regression over the old implementation and is likely to be picked up when the Perf team does large-scale OCP testing.  Is there another way to implement this in controller-runtime to make sure we do not regress at all?

yanirq: I was looking into https://github.com/kubernetes-sigs/controller-runtime/issues/246  which will allow us to add a watch for pods dynamically but I still need to figure out how to remove the watch (if this is even possible).

cmd/cluster-node-tuning-operator/main.go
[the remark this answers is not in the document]
cynepco3hahue: In general, the assumption of the controller is that the controller should work only on one type of resource. Like you said it is better from a modularity and readability perspective. So the idea is to have:
1. operator controller to deploy the default set of resources(default tuned and default daemon set)
2. controller to handle the behavior around the tuned resources.

dagrayvid:
@yanirq @jmencak,

I did a bit more testing similar to what Jiri already shared.  I tested 3 cases for the old code and the new code: 1. fully idle 2. creating ~1000 pods in the background without any custom profile, 3. creating ~1000 pods in the background with a Profile that matches to those pods. Here are the results:

1. Fully idle:
```
Old implementation:                 Controller-runtime rewrite:
2022-04-02 17:39:25: 101 18         2022-04-02 16:36:52: 103 20
2022-04-02 17:40:25: 103 19         2022-04-02 16:37:52: 106 22
2022-04-02 17:41:25: 105 20         2022-04-02 16:38:52: 110 24
2022-04-02 17:42:25: 105 20         2022-04-02 16:39:52: 114 25
delta:               4   2                               11  5
```

2. Creating pods with no label matching used. Should be pretty idle…
```
Old implementation:                 Controller-runtime rewrite:
2022-04-02 17:46:33: 112 22         2022-04-02 16:48:01: 153 44
2022-04-02 17:47:33: 113 22         2022-04-02 16:49:01: 226 55
2022-04-02 17:48:33: 115 23         2022-04-02 16:50:01: 299 67
2022-04-02 17:49:33: 117 23         2022-04-02 16:51:01: 365 76
delta:               5   1                               212 32
```

```
Old implementation:                 Controller-runtime rewrite:
2022-04-02 17:53:18: 169  33        2022-04-02 17:04:48: 104  21
2022-04-02 17:54:18: 539  56        2022-04-02 17:05:48: 1122 85
2022-04-02 17:55:18: 969  72        2022-04-02 17:06:48: 2959 189
2022-04-02 17:56:18: 1488 86        2022-04-02 17:07:48: 5740 299
delta:               1319 53                             5636 278
```

This confirms that the controller-runtime implementation is doing some work when pods are created even when the pod label matching functionality is unused. It also shows that when the pod label matching functionality is used, the new implementation is using many more CPU cycles than the old implementation.

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -2,63 +2,133 @@ package main
...
+
+		restConfig := ctrl.GetConfigOrDie()
+		le := util.GetLeaderElectionConfig(restConfig, enableLeaderElection)
+		mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+			NewCache:                cache.MultiNamespacedCacheBuilder(namespaces),
+			Scheme:                  scheme,
+			LeaderElection:          true,
+			LeaderElectionID:        config.OperatorLockName,
+			LeaderElectionNamespace: ntoNamespace,
+			LeaseDuration:           &le.LeaseDuration.Duration,
```
camilamacedo86, line 96: Why are you defining a lease duration?

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -2,63 +2,133 @@ package main
...
+		ntoNamespace := config.OperatorNamespace()
+		namespaces := []string{
+			ntoNamespace,
+			metav1.NamespaceNone,
+		}
+
+		restConfig := ctrl.GetConfigOrDie()
+		le := util.GetLeaderElectionConfig(restConfig, enableLeaderElection)
+		mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+			NewCache:                cache.MultiNamespacedCacheBuilder(namespaces),
```
camilamacedo86, line 91: Why would you like to use this option?
How many ns would you like to inform here?

See that if you try cache all -1 or 2 namespaces on the cluster then, you will probably check performance issues.
See: 

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/multi_namespace_cache.go#L40-L46

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -2,63 +2,133 @@ package main
...
+			Scheme:                  scheme,
+			LeaderElection:          true,
+			LeaderElectionID:        config.OperatorLockName,
+			LeaderElectionNamespace: ntoNamespace,
+			LeaseDuration:           &le.LeaseDuration.Duration,
+			RetryPeriod:             &le.RetryPeriod.Duration,
+			RenewDeadline:           &le.RenewDeadline.Duration,
+			Namespace:               ntoNamespace,
```
camilamacedo86, line 99: You are passing an `NS and NewCache: cache.MultiNamespacedCacheBuilder(namespaces)`
Note that you need to have permissions ( scope ) to read/update/delete resources and in this way cache them where the operator will be installed. Then, you can:

a) [Watching/Catching resources in a set of Namespaces](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#watching-resources-in-a-set-of-namespaces)
It is possible to use [MultiNamespacedCacheBuilder](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder) from [Options](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#Options) to watch and manage resources in a set of Namespaces.

OR 

b) Watching/Catching resources in a single Namespace (where the. operator will be installed) by using the Namespace option. See: 

> // Namespace if specified restricts the manager's cache to watch objects in
> // the desired namespace Defaults to all namespaces
> //
> // Note: If a namespace is specified, controllers can still Watch for a
> // cluster-scoped resource (e.g Node).  For namespaced resources the cache
> // will only hold objects from the desired namespace.
> Namespace [string](https://pkg.go.dev/builtin#string)
 
Ref: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.1/pkg/manager#Manager

OR 

c) Do not add both which means grant cluster-scope permission for the project. Your operator will be watching/catching the whole cluster 

Be aware that how much more resources/namespaces do you catching/watching more resources you will consume.

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -2,63 +2,133 @@ package main
...
+		})
```
camilamacedo86, line 100: By default SDK/Kubebuilder scaffold the projects as cluster-scope, see: https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/main.go#L68-L79

But you can change the. scope: https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/

manifests/40-rbac.yaml
```diff
@@ -34,10 +34,10 @@ rules:
 - apiGroups: ["security.openshift.io"]
```
camilamacedo86, line 34: by adopting SDK/Kubebuilder you will be able to work with makers, e.g: https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/controllers/memcached_controller.go#L44-L48

Then, when you run `make generate` the RBAC will be generated at the[config](https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v3/memcached-operator/config)/rbac/ dir: https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v3/memcached-operator/config/rbac

Also, you can use `make bundle` and have the whole OLM bundle generated by you with all your kustomize configs, see: https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v3/memcached-operator/bundle

That is very helpful to work with the releases and provide the solutions for OLM.
You can also add customizations on in your base CSV: [operator-sdk](https://github.com/operator-framework/operator-sdk)/[testdata](https://github.com/operator-framework/operator-sdk/tree/master/testdata)/[go](https://github.com/operator-framework/operator-sdk/tree/master/testdata/go)/[v3](https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v3)/[memcached-operator](https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v3/memcached-operator)/[config](https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v3/memcached-operator/config)/[manifests](https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v3/memcached-operator/config/manifests)/bases/

pkg/metrics/server.go
camilamacedo86: You can pass the metrics to the manager: https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/main.go#L70

By using SDK/Kubebuilder scaffolds you won that by free as well. See: https://book.kubebuilder.io/reference/metrics.html

pkg/operator/mc.go
```diff
@@ -1,13 +1,15 @@
 package operator
```
camilamacedo86, line 1: What means mc.go?
What is its purpose?

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,246 @@
...
+		CreateFunc: func(e event.CreateEvent) bool {
+			if e.Object == nil {
+				klog.Error("create event has no runtime object to create")
+				return false
+			}
+
+			return e.Object.GetName() == tunedv1.TunedClusterOperatorResourceName
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if !validateUpdateEvent(&e) {
```
camilamacedo86, line 52: Are you doing it to address the scenarios that the reconciliation fails because the resource changed on the cluster? If yes, I'd suggest using the client and fetching the resource that you want to change/update before calling the update. Then, if it fails return err in the reconciliation to ensure that it will be executed again.

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,246 @@
...
+				return false
+			}
+
+			return e.ObjectNew.GetName() == tunedv1.TunedClusterOperatorResourceName &&
+				e.ObjectOld.GetName() == tunedv1.TunedClusterOperatorResourceName
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			if e.Object == nil {
```
camilamacedo86, line 60: Are you here trying to avoid the error when the CR was deleted? If yes, 
It also is not necessary. I'd suggest you replace it by fetching the CR in the beginner of the reconcile and if it no be found then return nill to stop the reconciliation such as: https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/controllers/memcached_controller.go#L62-L76

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,246 @@
...
+	Namespace string
+}
+
+// SetupWithManager creates a new ClusterOperator Controller and adds it to the Manager.
+// The Manager will set fields on the Controller and Start it when the Manager is Started.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// We want to initiate reconcile loop only on changes under the NTO cluster operator.
+	cvoPredicates := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			if e.Object == nil {
```
camilamacedo86, line 44: Same of. https://github.com/openshift/cluster-node-tuning-operator/pull/302/files#r815030772

pkg/operator/reconciler.go
```diff
@@ -0,0 +1,246 @@
...
+
+		return !equality.Semantic.DeepEqual(dsOld.Status, dsNew.Status) ||
+			!equality.Semantic.DeepEqual(dsOld.Spec, dsNew.Spec)
+	}}
+
+	err := ctrl.NewControllerManagedBy(mgr).
+		For(&configv1.ClusterOperator{}, builder.WithPredicates(cvoPredicates)).
+		Owns(&appsv1.DaemonSet{}, builder.WithPredicates(dsUpdatePredicate)).
+		Owns(&tunedv1.Tuned{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Complete(r)
```
camilamacedo86, line 85: The predicates are useful when we want to filter the watching, e.g.:

I want to watch all Pods created/owned by this controller but I just want to re-trigger the reconciliation when they are updated if they have the label = n. It helps improve the performance since is possible create conditionals.

pkg/util/leaderelection.go
```diff
@@ -10,22 +10,26 @@ import (
 	"k8s.io/klog"
 )
 
-func GetLeaderElectionConfig(ctx context.Context, restcfg *rest.Config) configv1.LeaderElection {
+// GetLeaderElectionConfig returns leader election configs defaults based on the cluster topology
+func GetLeaderElectionConfig(restcfg *rest.Config, enabled bool) configv1.LeaderElection {
```
camilamacedo86, line 14: Why do you need this ultis for leader election? 
By default, you have the C+R implementation for leader election : https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/leaderelection/leader_election.go

What is the extra requirement on top of that? 

Also, you might give a look in the doc https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#leader-election which has compressive info about it.

go.mod
```diff
@@ -24,10 +24,12 @@ require (
 	k8s.io/klog v1.0.0
 	k8s.io/klog/v2 v2.30.0
 	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
+	sigs.k8s.io/controller-runtime v0.11.0
```
camilamacedo86, line 27: Also, see; wdyt about check it with the bug fix: https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.11.1

go.mod
```diff
@@ -116,5 +122,8 @@ replace (
 	k8s.io/kubernetes => k8s.io/kubernetes v1.23.0
 	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.23.0
 	k8s.io/metrics => k8s.io/metrics v0.23.0
+	k8s.io/mount-utils => k8s.io/mount-utils v0.23.0
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.23.0
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.11.0
+	sigs.k8s.io/controller-tools => sigs.k8s.io/controller-tools v0.7.0
```
camilamacedo86, line 128: why do you need these replaces? 
Why do you import sigs.k8s.io/controller-tools?

cmd/cluster-node-tuning-operator/main.go
yanirq: this was to keep the original behavior before the refactor

cmd/cluster-node-tuning-operator/main.go
yanirq: the main reason here was to watch cluster resources such as nodes and pods and have a distinct namespace for NTO to be used for filtering

yanirq:
@camilamacedo86   Thank you for the extensive review, it is super informative and helpful (also for deeper understanding of controller runtime "under the hood").
This PR will probably be closed soon since we switched to an alternative solution but it will be kept in case we do choose to use this refactor in the future.
Some of the information is insightful also for the current implementation so thanks again !  I will answer the comments here that could be addressed.

camilamacedo86:
Hi @yanirq,

https://github.com/openshift/cluster-node-tuning-operator/pull/302#issuecomment-1053623069
Thank you for your reply. I hope that it can help out.
 
Feel free to reach out if you need.

yanirq:
This PR will not be the format we will be using.
The solution is to copy over the PAO code and add it to the manager as a runnable instead of refactoring whole of NTO to use controller runtime.  
The PR can be reopened or use as a reference if needed.

PR #303, opened 2021-12-16
TuneD: switch to FDP TuneD delivery

jmencak:
Switch to Fast Data Path (FDP) method of delivering TuneD.  Keep the old
method of delivery in place for emergency rollback and update the
version of TuneD to match the version current in the FDP channel.


jmencak:
A few notes to reviewers:
- This should merge this week.
- `tuned-profiles-openshift, tuned-profiles-postgresql and tuned-profiles-spectrumscale` profiles are not shipped via CDN yet, therefore I decided to ship all profiles locally.  Once this is fixed, we'll switch to shipping all the profiles from CDN too.
- This might be a potentially breaking change for non `x86_64` architectures.  The odds are that they will not be, but we will not know until this is merged and the `Dockerfile.rhel8` reconciled with the downstream CI.  I'll keep an eye on this and take appropriate actions if/when things break.

jmencak:
I think it is unlikely we're getting those missing 3 packages in CDN by the end of this week, therefore:
/retitle TuneD: switch to FDP TuneD delivery

dagrayvid:
Thanks @jmencak for the PR. The changes make sense (after discussing on slack).

/lgtm

PR #304, opened 2021-12-16
TuneD profiles: switch to FDP TuneD delivery completely

jmencak:
All TuneD profiles packages are now being correctly shipped via CDN --
use them.

### DIFF
--- a/Dockerfile.rhel8
+++ b/Dockerfile.rhel8
@@ -14,7 +14,7 @@ RUN INSTALL_PKGS=" \
     LC_COLLATE=C cat ../patches/*.diff | patch -Np1 && \
     dnf build-dep tuned.spec -y && \
     make rpm PYTHON=/usr/bin/python3 && \
-    rm -rf /root/rpmbuild/RPMS/noarch/{tuned-2.*,tuned-gtk*,tuned-utils*,tuned-profiles-compat*} && \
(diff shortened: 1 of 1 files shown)


jmencak:
/assign @dagrayvid 
One last PR to review before you go on leave, David, please :)

Dockerfile.rhel8
```diff
@@ -14,7 +14,7 @@ RUN INSTALL_PKGS=" \
     LC_COLLATE=C cat ../patches/*.diff | patch -Np1 && \
     dnf build-dep tuned.spec -y && \
     make rpm PYTHON=/usr/bin/python3 && \
-    rm -rf /root/rpmbuild/RPMS/noarch/{tuned-2.*,tuned-gtk*,tuned-utils*,tuned-profiles-compat*} && \
+    rm -rf /root/rpmbuild/RPMS/noarch/{tuned-gtk*,tuned-utils*,tuned-profiles-compat*} && \
```
dagrayvid, line 17: Why do we need this line if we are removing line 39 below, where we would be installing these?

jmencak: The short answer is we do not need it.  Longer answer: this whole stage will be removed in the future once FDP proves itself.  There is also somewhere in the CI that I believe expects 3 stages, so I didn't want to risk any breakage at this point.  Need to look into this, but I'm afraid I'm out of time at this point.

dagrayvid:
Thanks @jmencak for clarifying my question.

I successfully built the operator using the FDP repo and my credentials, thanks for the instructions.

/lgtm

PR #307, opened 2022-01-04
Bug 2037036: grub2-editenv: no errors when using [bootloader] plug-in

jmencak:
Workaround for `TuneD` running in the container not to emit
```
ERROR    tuned.utils.commands: Executing grub2-editenv error: [Errno 2] No such file or directory: 'grub2-editenv': 'grub2-editenv'
```
message while keeping the `[bootloader]` plug-in functional.

PR #308, opened 2022-01-04
Bug 2036303: Prevent unnecessary ksm* error log messages.

jmencak:
Adding upstream TuneD patch:
https://github.com/redhat-performance/tuned/pull/331

The patch checks the presence of `ksm` and `ksmtuned` units before (un)masking them.

The change prevents messages in the logs such as:
```
2021-02-24 14:46:52,399 ERROR    tuned.plugins.plugin_script: script '/usr/lib/tuned/cpu-partitioning/script.sh' error output: 'Unit ksm.service does not exist, proceeding anyway.
```

which caused per-node Profiles go into Degraded state.

Resolves: rhbz#2036303

### DIFF
--- a/assets/tuned/patches/050-no-ksm-errors.diff
+++ b/assets/tuned/patches/050-no-ksm-errors.diff
@@ -1,0 +1,29 @@
+Check the presence of `ksm` and `ksmtuned` units before (un)masking them.
+
+Prevents messages in the logs such as:
+2021-02-24 14:46:52,399 ERROR    tuned.plugins.plugin_script: script '/usr/lib/tuned/cpu-partitioning/script.sh' error output: 'Unit ksm.service does not exist, proceeding anyway.
+
+See: https://github.com/redhat-performance/tuned/pull/331
+
+diff --git a/profiles/functions b/profiles/functions
+index 0e1836b6..636995c6 100644
+--- a/profiles/functions
++++ b/profiles/functions
+@@ -559,6 +559,8 @@ disable_ksm()
+ 		if ! touch $KSM_MASK_FILE; then
+ 			die "failed to create $KSM_MASK_FILE"
+ 		fi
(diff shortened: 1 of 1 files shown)


dagrayvid:
@jmencak, why isn't the bootloader workaround from #307 needed for 4.9?

jmencak:
> @jmencak, why isn't the bootloader workaround from #307 needed for 4.9?

Good question.  Because we have a custom patch for it in 4.9: https://github.com/openshift/cluster-node-tuning-operator/blob/release-4.9/assets/tuned/patches/030-no-bootloader-errors.diff which isn't in `TuneD` upstream and 4.10 now ship a completely unmodified `TuneD` version.

dagrayvid:
Thanks for clarifying @jmencak. Makes sense.

/lgtm

jmencak:
**Backport risk assessment**

- [ ] If the Bugzilla associated with the PR has the "FastFix" keyword, the subjective assessment on the issue has already been done and a customer is impacted. These PRs should be prioritized for merge.
  - not applicable
- [ ] For branches that are in the Maintenance lifecycle phase:
  - not applicable
- [x] The bug has significant impact either through severity, reduction in supportability, or number of users affected.
  - yes, cluster administrators using PAO will otherwise be alarmed by the Profiles being in Degraded state
- [x] The severity field of the bug must be set to accurately reflect criticality.
  - bugzilla/severity-high
- [x] The PR is merged in the next newer release branch and the bug for that newer release is VERIFIED by QE.
  - yes
- [x] The PR’s description is well formed with user-focused release notes that state the bug number, impact, cause, and resolution. Where appropriate, it should also contain information about how a user can identify whether a particular cluster is affected.
  - yes

/label backport-risk-assessed

jmencak:
@wabouhamad , can we get `cherry-pick-approved` label?  Thank you!

PR #314, opened 2022-02-06
Vendor and run performance-addon-operator as part of the controller-runtime NTO manager

cynepco3hahue:


yanirq:
@dagrayvid  we can start testing the performance of this combination of PAO and NTO with this PR (based on #316)
The manifests in this folder should be applied manually (including the new additions).

manifests/40-webhook-configuration.yaml
```diff
@@ -0,0 +1,55 @@
+---
```
dagrayvid, line 1: We should probably increase the prefix number for this and the following files by 10. Or perhaps use `45-webhook-configuration.yaml` if changing the name of the other manifests could cause problems?

cynepco3hahue: yee, I want to avoid renaming, 45 will work for me

dagrayvid:
This PR performs very similarly to the current implementation of NTO. As we expected, the addition of the PAO control loop has no impact when there are no PerformanceProfiles to reconcile.

1. Fully idle:
```
Old implementation:                 pr-314:
2022-04-02 17:39:25: 101 18         2022-14-02 11:27:42: 45 9
2022-04-02 17:40:25: 103 19         2022-14-02 11:28:42: 46 10
2022-04-02 17:41:25: 105 20         2022-14-02 11:29:42: 48 10
2022-04-02 17:42:25: 105 20         2022-14-02 11:30:42: 49 12
delta:               4   2                               4  3
```

2. Creating pods with no label matching used.
```
Old implementation:                 pr-314:
2022-04-02 17:46:33: 112 22         2022-14-02 11:32:54: 52 13
2022-04-02 17:47:33: 113 22         2022-14-02 11:33:54: 53 13
2022-04-02 17:48:33: 115 23         2022-14-02 11:34:54: 54 14
2022-04-02 17:49:33: 117 23         2022-14-02 11:35:54: 56 14
delta:               5   1                               4 1
```

```
Old implementation:                 pr-314:
2022-04-02 17:53:18: 169  33        2022-14-02 11:43:02: 130 28
2022-04-02 17:54:18: 539  56        2022-14-02 11:44:02: 501 51
2022-04-02 17:55:18: 969  72        2022-14-02 11:45:02: 981 71
2022-04-02 17:56:18: 1488 86        2022-14-02 11:46:02: 1547 86
delta:               1319 53                             1417 58
```

With this, I am comfortable merging #316. @jmencak, WDYT?

jmencak:
> With this, I am comfortable merging #316. @jmencak, WDYT?

This looks good.  Do we have any data wrt. memory usage?

jmencak:
> > This looks good. Do we have any data wrt. memory usage?
> 
> I haven't looked at memory usage yet. I can probably gather some data on this today if needed. Is memory usage as much a concern as CPU usage for core OCP operators?

Not as much as CPU usage, but it would be nice to know.

dagrayvid:
I ran the first and third test cases to get an idea of the memory usage before / after these changes.

It's probably easiest to just share it with a graph of memory usage over time. There is a line for fully idle, and then a line for memory over time while creating 1000 pods using pod-label matching, and then deleting them to see if the usage goes back down to where it was before creating any pods.

Screen Shot 2022-02-14 at 5 06 04 PM

One note: for the PR-314 results, you may be wondering why the idle memory usage is lower than the memory usage at the start of the pod creation case. I created the custom profile before starting the second test case, without creating any matching pods. This slight jump in memory usage after creating the custom profile was only seen in the new implementation. I tested it twice to double check that it wasn't due to some human error.

dagrayvid:
I ran the first and third test cases to get an idea of the memory usage before / after these changes.

It's probably easiest to just share it with a graph of memory usage over time. There is a line for fully idle, and then a line for memory over time while creating 1000 pods using pod-label matching, and then deleting them to see if the usage goes back down to where it was before creating any pods.

Screen Shot 2022-02-14 at 6 16 48 PM

One note: In case you wonder why the idle memory usage is lower than the memory usage at the start of the pod creation case. I created the custom profile before starting the second test case, without creating any matching pods. This jump was seen in both the new and old implementations (corrected from previous comment)

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -100,12 +105,37 @@ func main() {
...
+
+		// configure webhook server
+		webHookServer := mgr.GetWebhookServer()
+		webHookServer.Port = webhookPort
+		webHookServer.CertDir = webhookCertDir
+		webHookServer.CertName = webhookCertName
+		webHookServer.KeyName = webhookKeyName
+
+		if err = (&performancev2.PerformanceProfile{}).SetupWebhookWithManager(mgr); err != nil {
+			klog.Exitf("unable to create PerformanceProfile v2 webhook : %v", err)
```
jmencak, line 136: Nit, as noted above.

manifests/40-rbac.yaml
```diff
@@ -69,6 +69,12 @@ rules:
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
   verbs: ["create","get","update","patch"]
+- apiGroups: ["node.k8s.io"]
```
jmencak, line 72: In the past, I was asked by ProdSec to review RBAC.  After the review, I added a few comments here.  Can we also do this for PAO?  Especially here for `runtimeclasses`.  `performance.openshift.io` should be clear to all.

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -100,12 +105,37 @@ func main() {
 
 		controller, err := operator.NewController()
 		if err != nil {
-			klog.Fatal(err)
+			klog.Fatalf("failed to create new controller %v", err)
```
jmencak, line 108: Consistency.  `s/controller %v/controller: %v/` ?

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -100,12 +105,37 @@ func main() {
...
+		}
+
+		if err := mgr.Add(controller); err != nil {
```
jmencak, line 111: Thank you for doing this, we missed this in the earlier review. :/

jmencak (commented):
Thank you Artyom!  Untested (yet), but the changes look good to me.  Only nits right now.

yanirq:
@cynepco3hahue  in the current vendoring form PAO still uses `openshift-performance-addon-operator` namesspace:

https://github.com/openshift-kni/performance-addon-operators/blob/c61353605e60f48ac75432a37489d278d32eb4ce/controllers/performanceprofile_controller.go#L227

If we will keep it that way we also need to add the namespace to the manifests.

cmd/cluster-node-tuning-operator/main.go
cynepco3hahue: Done

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -100,12 +105,37 @@ func main() {
...
-		mgr.Add(controller)
-		mgr.Add(metrics.Server{})
 		metrics.RegisterVersion(version.Version)
 
+		if err = (&controllers.PerformanceProfileReconciler{
+			Client:   mgr.GetClient(),
+			Scheme:   mgr.GetScheme(),
+			Recorder: mgr.GetEventRecorderFor("performance-profile-controller"),
+		}).SetupWithManager(mgr); err != nil {
+			klog.Exitf("unable to create PerformanceProfile controller : %v", err)
```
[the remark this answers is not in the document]
cynepco3hahue: typo, done :)

cmd/cluster-node-tuning-operator/main.go
cynepco3hahue: Done

manifests/40-rbac.yaml
cynepco3hahue: Done

cynepco3hahue:
> @cynepco3hahue in the current vendoring PAO still uses `openshift-performance-addon-operator` namespace:
> 
> https://github.com/openshift-kni/performance-addon-operators/blob/c61353605e60f48ac75432a37489d278d32eb4ce/controllers/performanceprofile_controller.go#L227
> 
> If we will keep it that way we also need to add the namespace to the manifests.

It should not affect us, it is used only to generate correct manifests automatically. We are adding manifests manually so all kube builder thing is not relevant for us.

cynepco3hahue:
not relevant anymore

PR #316, opened 2022-02-08
wrap NTO controller with controller runtime lib

yanirq:
Refactor cluster node tuning operator to use controller runtime library ([release 0.11](https://github.com/kubernetes-sigs/controller-runtime/tree/release-0.11)).

The functionality is internal only and replaces the direct application of a controller with the controller runtime scheme.
This is a refactor to the main invocation of cluster node tuning controller.
The controller and the metrics server are wrapped and started by the controller runtime library.

This is also a preliminary work to set up the stage for moving [Performance addons operator ](https://github.com/openshift-kni/performance-addon-operators)under NTO as documented here: https://github.com/openshift/enhancements/pull/867


yanirq:
@jmencak  @dagrayvid 
This PR is an alternative to replace a full refactor of NTO's controller presented here:  https://github.com/openshift/cluster-node-tuning-operator/pull/302 
It keeps most of the original code and will save a lot of regressions if proven as a valid solution. 
The performance should be tested as well but since the main functionality is kept then the results should not differ from the current implementation of NTO.

dagrayvid:
As expected this PR seems to perform about the same as the current implementation:

1. Fully idle:
```
Old implementation:                 Controller-runtime wrapper:
2022-04-02 17:39:25: 101 18         2022-08-02 16:52:44: 54 12
2022-04-02 17:40:25: 103 19         2022-08-02 16:53:44: 56 13
2022-04-02 17:41:25: 105 20         2022-08-02 16:54:44: 57 13
2022-04-02 17:42:25: 105 20         2022-08-02 16:55:44: 60 15
delta:               4   2                               6  3
```

2. Creating pods with no label matching used. 
```
Old implementation:                 Controller-runtime wrapper:
2022-04-02 17:46:33: 112 22         2022-08-02 17:00:22: 69 18
2022-04-02 17:47:33: 113 22         2022-08-02 17:01:22: 71 18
2022-04-02 17:48:33: 115 23         2022-08-02 17:02:22: 73 19
2022-04-02 17:49:33: 117 23         2022-08-02 17:03:22: 75 19
delta:               5   1                               6 1
```

```
Old implementation:                 Controller-runtime wrapper:
2022-04-02 17:53:18: 169  33        2022-08-02 17:09:37: 129 27
2022-04-02 17:54:18: 539  56        2022-08-02 17:10:37: 587 48
2022-04-02 17:55:18: 969  72        2022-08-02 17:11:37: 1101 65
2022-04-02 17:56:18: 1488 86        2022-08-02 17:12:37: 1696 80
delta:               1319 53                             1567 53
```

pkg/operator/controller.go
```diff
@@ -1045,97 +1042,20 @@ func (c *Controller) run(stopCh <-chan struct{}) {
...
-func (c *Controller) BecomeLeader(stopCh <-chan struct{}) {
-	// Become the leader before proceeding.
-	klog.Info("trying to become a leader")
-
-	// Check the presence of the legacy node-tuning-operator-lock ConfigMap.
-	// If the owner references exist in the ConfigMap, we have an old garbage
-	// collection based locking scheme and we cannot continue until the ConfigMap
-	// is GC or deleted manually.  If the owner references do not exist, just go
-	// ahead with the new client-go leader-election code.
-loop:
```
yanirq, line 1069: @jmencak   In case the move to the controller runtime wrapper does not handle this GC, will we still need to keep this cleanup for legacy NTO lock ?  
What is the upgrade path for NTO ?  can we skip major versions and miss this cleanup ?

jmencak: So this cleanup code was added in 4.8.  In my opinion we no longer need to keep it since any of the 4.8, 4.9, 4.10 versions will take care of this cleanup.  In other words, we don's support direct upgrades from say 4.7->4.11.

pkg/util/leaderelection.go
```diff
@@ -10,22 +10,26 @@ import (
...
+		configv1.LeaderElection{
+			Disable: !enabled,
+		},
 		"", "",
 	)
 
-	if infra, err := clusterstatus.GetClusterInfraStatus(ctx, restcfg); err == nil && infra != nil {
-		if infra.ControlPlaneTopology == configv1.SingleReplicaTopologyMode {
-			return leaderelection.LeaderElectionSNOConfig(defaultLeaderElection)
+	if enabled {
```
jmencak, line 25: I wonder if we can reverse the logic here and say something like `if !enabled return xyz`.  This will handle the "edge" cases sooner and also keep the (future) code below simpler and nesting down.  This, in my opinion, is a better coding practice.

jmencak (commented):
Thank you for the PR, @yanirq !  I had a quick look at this and it looks mostly good to me.  Only nits so far, but I will take another look.  Also, @dagrayvid , can you take a look?

dagrayvid:
The changes look good to me after my first look through.

@yanirq @jmencak is the long term plan to maintain the NTO controller as a client-go based controller like this? Or do we intend to rewrite it using the controller-runtime eventually, assuming we can find a way to do so without a significant performance regression?

yanirq:
> The changes look good to me after my first look through.
> 
> @yanirq @jmencak is the long term plan to maintain the NTO controller as a client-go based controller like this? Or do we intend to rewrite it using the controller-runtime eventually, assuming we can find a way to do so without a significant performance regression?

The main goal here is to have NTO wrapped by controller runtime in order for us to add [PAO](https://github.com/openshift-kni/performance-addon-operators), which is already written completely with controller-runtime, under this repository (see https://github.com/openshift/enhancements/pull/867).
When approaching this at first we thought about having a main reconcile loop (controller runtime implemented) that would take care of both NTO and PAO.
In retrospective https://github.com/openshift/cluster-node-tuning-operator/pull/302  showed that a full conversion of NTO controller to use controller runtime will need a deeper design change so it would be efficient and and written according to [best practices](https://sdk.operatorframework.io/docs/best-practices/managing-resources/).

With the changes introduced in this PR, the next step would be to add PAO controller with controller runtime manager:
 https://github.com/openshift/cluster-node-tuning-operator/blob/86248b9245e69177fdb53e3a5f39d1e57dd1faaa/cmd/cluster-node-tuning-operator/main.go#L85

yanirq:
> So now that we have two options, I wonder how they'd compare performance-wise with PAO already merged in as a controller. Have you tried that, @yanirq? What's the expectation here? At the moment, this PR is the "winner", but will it still be a "winner" when running idle with PAO merged in?

The approach of adding PAO would be similar in both PRs. We will add it as a separate controller (with its reconciler) under the manger.
The reconciler ideally should handle only related CRs (Performance Profiles in our case and watch for its created artifacts)

yanirq:
> > So now that we have two options, I wonder how they'd compare performance-wise with PAO already merged in as a controller. Have you tried that, @yanirq? What's the expectation here? At the moment, this PR is the "winner", but will it still be a "winner" when running idle with PAO merged in?
> 
> I had the same thought. I can probably test this using #314.

@cynepco3hahue  should have a more updated PR for that (at least a WIP one)

dagrayvid:
> > I had the same thought. I can probably test this using #314.
> 
> @cynepco3hahue should have a more updated PR for that (at least a WIP one)

I will wait for an update to PR#314 before testing. Although I don't expect any surprises, I would like to do this test before merging this PR.

jmencak:
Thank you for the changes, @yanirq .  This looks good to me now.  Can we squash the commits?  @dagrayvid , unless you have objections from your side, I think this is ready to be merged.

yanirq:
> Thank you for the changes, @yanirq . This looks good to me now. Can we squash the commits? @dagrayvid , unless you have objections from your side, I think this is ready to be merged.

@jmencak  commits are already squashed

jmencak:
/lgtm
/approve
/hold
David, please cancel the hold if you have no objections.

PR #321, opened 2022-03-14
Address e2e test races

jmencak:
The NTO e2e test suite runs as a series of short e2e tests starting
immediately once the previous test completes.  Some e2e tests rely on
tuning set by the default worker `TuneD` profile -- `openshift-node`.  Most
e2e test will complete without doing a full profile rollback (and/or)
application of the default worker `TuneD` profile.  If an e2e test relies
on this profile to be applied, block the execution of the test until it
applied.

Other changes:
  - Make sure `tuned_errors_and_recovery.go` e2e test can be run several
    times in a sequence on the same cluster.
  - Some formatting nits.

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -24,6 +24,11 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/test/framework"
 )
 
+const (
+	// The default worker profile.  See: assets/tuned/manifests/default-cr-tuned.yaml
+	DefaultWorkerProfile = "openshift-node"
+)
+
 // Logf formats using the default formats for its operands and writes to
 // ginkgo.GinkgoWriter and a newline is appended.  It returns the number of
 // bytes written and any write error encountered.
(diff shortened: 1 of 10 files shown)


jmencak:
This is the last PR I'm aware of we want to get in for 4.11 from NTO side, @yanirq 

This should not affect the NTO/PAO merge, just NTO e2e test fixes.

test/e2e/util/util.go
```diff
@@ -280,6 +285,42 @@ func WaitForClusterOperatorConditionReason(cs *framework.ClientSet, interval, du
 	return nil
 }
 
+// WaitForProfileConditionStatus blocks until Profile with name `profile`
+// has applied TuneD profile `profileExpect` with its `conditionType` equal to
+// the value of 'conditionStatus'.
```
dagrayvid, line 290: I think this comment suggests that the function blocks until the TuneD profile is applied, but that isn't necessarily true. It blocks until the Profile object status condition `conditionType` matches  `conditionStatus` (regardless of whether the TuneD profile is applied or not, right? 

If you agree, maybe we can reword it to something like:
```suggestion
// WaitForProfileConditionStatus blocks until Profile with name `profile`
// is reporting its `conditionType` with the value of 'conditionStatus', 
// for the TuneD profile `profileExpect`.
```

jmencak: I agree, your comment is a lot more precise than the original.  Removed the trailing space and using it.  Thank you!

jmencak:
Upgrades failing.  MCO with master pool degraded, no chance this PR caused this
/retest

dagrayvid:
/lgtm

Potential CI flakes need to be investigated, but seem unrelated to this PR.

PR #322, opened 2022-03-15
Performance addon operator code base move to NTO

yanirq:
This PR includes : 
- PAO code base copy to NTO .  See snapshot here: https://github.com/openshift-kni/performance-addon-operators/pull/862
- PAO code invocation and manifests (taken from the PAO vendoring PR: https://github.com/openshift/cluster-node-tuning-operator/pull/314 

Additional notes:
- First 2 commits are to preserve git values and enable a reasonable method for cherry-picks
- PAO code moved to NTO current paths and structure
- Most of the vendoring added in one additional commit


jmencak (commented):
The structure overall looks reasonable to me.  A couple of notes though:
- Can the `/testdata` be placed somewhere under `/test` to keep the root a bit more tidy?
- How will `/config/pao/rbac/role_binding.yaml` and `/config/pao/rbac/role.yaml` get instantiated?  Should we place them in `/manifests` instead to leave this up to CVO?
- Why is `/assets/pao/configs` plural and `/config` singlular?  I.e. consistency.
- I like the addition of `/docs`.  We should shorten the NTO `/README.md` and move most of the docs from the `/README.md` to that folder alongside PAO.  Not necessarily as part of this PR.

yanirq:
> The structure overall looks reasonable to me. A couple of notes though:

>      Can the `/testdata` be placed somewhere under `/test` to keep the root a bit more tidy?
Yes, I will work on that on a cleanup separate commit (maybe after some more comments)

>     * How will `/config/pao/rbac/role_binding.yaml` and `/config/pao/rbac/role.yaml` get instantiated?  Should we place them in `/manifests` instead to leave this up to CVO?
Actually I might be able to remove that folder completely , our test data depends on that and there is still work to be done in the code generators (not in this PR at this point of time)
Next commit should have the relevant manifests under manifests folder.

>     * Why is `/assets/pao/configs` plural and `/config` singlular?  I.e. consistency.
ack
>     * I like the addition of `/docs`.  We should shorten the NTO `/README.md` and move most of the docs from the `/README.md` to that folder alongside PAO.  Not necessarily as part of this PR.
I will need to really cleanup the docs here . The content should be adjusted + the links.
We can either have it in this PR and fix in a follow up or leave it to a different PR.

Makefile
```diff
@@ -49,21 +51,24 @@ clone-tuned:
...
+build: $(BINDATA) generate-deepcopy pkg/generated
 	$(GO_BUILD_RECIPE)
 	ln -sf $(PACKAGE_BIN) $(OUT_DIR)/openshift-tuned
 
 $(BINDATA): $(GOBINDATA_BIN) $(ASSETS)
 	$(GOBINDATA_BIN) -mode 420 -modtime 1 -pkg manifests -o $(BINDATA) assets/...
 	gofmt -s -w $(BINDATA)
 
-pkg/generated: $(API_TYPES)
+generate-deepcopy:
```
jmencak, line 62: This change basically makes the reason for using `Makefile` useless.  Please keep the dependencies.

Makefile
```diff
@@ -141,3 +147,29 @@ $(call add-crd-gen,tuned,$(API_TYPES_DIR),./manifests,./manifests)
 $(call add-profile-manifests,manifests,./profile-patches,./manifests)
 
 .PHONY: all build deepcopy crd-schema-gen test-e2e verify verify-gofmt clean local-image local-image-push
+
+# Performance addons operations
```
jmencak, line 151: Nit: Above you use PAO as a comment, why use "Performance addons" here and a new line?

Makefile
yanirq: why is that ? this will re generated the zz deep copy files in case the api/types.go for the types we have are changed.   Is this handled somewhere else?
notice that `pkg/generated `was moved to line 71

yanirq: ok.  I now see I removed $(API_TYPES) by accident in the `pkg/generated:` section

cynepco3hahue:
Why do we need to copy `config/pao` directory? It is used mostly to generate CSV.

yanirq:
> Some initial review.
> 
>     1. Why do we need to copy `config/pao` directory? It is used mostly to generate CSV.
There is some code still using it, examining how to remove it completely atm

>     2. Probably the doc should now have a more clear hierarchy, same regarding examples.
Docs will be removed from this PR and will be handled in a separate one. 
>     3. We should update NTO README with the relevant PAO section.
Same comment as above.
>     4. Probably we do not need `pkg/pao/utils/csvtools`.
There is a section that still uses a tool there. I will see if it can be removed
>     5. Probably the pao directory hierarchy can be more clear like if it is placed under e2e we do not really need `functests` 
I think the current hierarchy can be kept. Mainly if considering cherry picks and familiarity

Makefile
```diff
@@ -33,6 +32,10 @@ ORG=openshift
 TAG=$(shell git rev-parse --abbrev-ref HEAD)
 IMAGE=$(REGISTRY)/$(ORG)/origin-cluster-node-tuning-operator:$(TAG)
 
+# PAO variables
+CLUSTER ?= "ci"
+PAO_CRD_APIS :=$(addprefix ./pkg/apis/pao/,v2 v1 v1alpha1)
```
dagrayvid, line 37: Maybe we should use `API_TYPES_DIR` here?
```suggestion
PAO_CRD_APIS :=$(addprefix ./$(API_TYPES_DIR)/pao/,v2 v1 v1alpha1)
```

dagrayvid:
Overall, I think the structure of the PR looks good. 

@jmencak @yanirq do you think we should repeat the performance tests that we ran on some of the previous PRs for CPU / memory utilization? I'm guessing without creating any PerformanceProfiles to reconcile, this PR should not have much impact.

dagrayvid:
> > @jmencak @yanirq do you think we should repeat the performance tests that we ran on some of the previous PRs for CPU / memory utilization? I'm guessing without creating any PerformanceProfiles to reconcile, this PR should not have much impact.
> 
> In my view this is not necessary, but if you feel it should be done no-one will stop you.

Agreed. I will not run these tests.

fromanirh:
I think we should get rid of `csvtools`. You can just inline the `MarhsallObject` method in the `cmd/performance-profile-creator/cmd/root.go` file (PAO paths), in a separate commit

fromanirh:
the overall direction looks good to me and after a first pass the most import bits of PAO seem in the right places.
I'll have another pass later to look at the PAO side of things.

cynepco3hahue:
The PR looks good to me in the light that we will have additional PRs for cleanups and documentation updates.
Mayne some additional nil is to add a separate OWNER file for all packages that contain PAO related stuff, to allow us to lgtm and reduce the PAO review load from NTO folks.

fromanirh:
> The PR looks good to me in the light that we will have additional PRs for cleanups and documentation updates. Mayne some additional nil is to add a separate OWNER file for all packages that contain PAO related stuff, to allow us to lgtm and reduce the PAO review load from NTO folks.

having specific OWNERS for PAO stuff is a great idea indeed

jmencak (commented):
Thank you for the PR, Yanir.  It looks good to me.  Only nits.  They should be very easy to address.
Once addressed, please squash the commits and add a good title and summary
description to the commit log.

Things that I believe should be addressed (now or in the future):

- Inconsistent script perms.  I believe they should all be 755:
```
assets/pao/scripts/hugepages-allocation.sh 755
assets/pao/scripts/low-latency-hooks.sh 644
assets/pao/scripts/set-rps-mask.sh 644
```
- In a future PR, we should probably move the NTO examples to `examples/tuned`.
This will need adjusting e2e tests and docs elsewhere.

_Disclaimer: I did not review the `pkg/pao` and `test/e2e/pao` content._

yanirq:
@jmencak  commits are now squashed except the copy and git mv commits that should help support cherry picking.

> Inconsistent script perms. I believe they should all be 755:
assets/pao/scripts/hugepages-allocation.sh 755
assets/pao/scripts/low-latency-hooks.sh 644
assets/pao/scripts/set-rps-mask.sh 644

I would keep the perms as they are since they are defined the same way in PAO.
These scripts are injected to machine configs and I'm pretty sure there was a reason for these specific permissions.
We can review this again and adjust in a different PR if need be.

yanirq:
>  The PR looks good to me in the light that we will have additional PRs for cleanups and documentation updates. Mayne some additional nil is to add a separate OWNER file for all packages that contain PAO related stuff, to allow us to lgtm and reduce the PAO review load from NTO folks.
> 
> having specific OWNERS for PAO stuff is a great idea indeed

We could do that in a different PR and there are logical folders to place PAO specific owners files but there are some locations that the 2 worlds are merged and will also require our attention.
As the time will pass I can also see more infusion here so maybe its worth considering adding to the main OWNERS file all of the PAO maintainers.

yanirq:
/hold
tide check is missing for some reason. holding in case this needs to be resolved first and not to allow auto merge.

jmencak:
Last nits:

What's the official name of PAO?  If it really is Performance Addon Operator as the original repo sugests, then I suggest to adjust the title of this PR and drop the plural (Addon/s/).  The same for the git log message please.  If you decide to keep things clean and do this, I'd also suggest to improve the git log formatting/message too a bit.  At least this:
```
s/by controller/  by controller/
s/kubernetes 1.23/kubernetes 1.23.3/  # it already was at 1.23.0
```

jmencak:
Thank you for the PR.
/approve
@dagrayvid please LGTM if this looks good to you.

jmencak:
Looking at the `#forum-dp-platform` the CI issue is still not resolved/

yanirq:
@jmencak  @dagrayvid  can we lgtm this PR now all tests have passed ?

jmencak:
Looks like the bot didn't see David's
/lgtm

PR #323, opened 2022-03-24
Fix e2e tests for aarch64 and SNO

jmencak:
This PR fixes operator e2e tests for for aarch64 and clusters with
worker nodes that also have a master role, such as SNO.

Other changes:
  - Fix for `sysctl_d_override.go` rollback (node/pod was always nil
    and rollback in case of a test failure was not performed).
  - Consistency changes across the tests.

### DIFF
--- a/examples/hugepages.yaml
+++ b/examples/hugepages.yaml
@@ -14,7 +14,7 @@ spec:
       # Comments that do not start at the beginning of a line as the TuneD
       # documentation states are currently (2021-09-15) a grey area and likely
       [sysctl]
-      vm.nr_hugepages=16	# a bad idea as rhbz#2004508 shows.
+      vm.nr_hugepages=1		# a bad idea as rhbz#2004508 shows.
     name: openshift-hugepages
   recommend:
   - match:
--- a/test/e2e/basic/custom_node_labels.go
+++ b/test/e2e/basic/custom_node_labels.go
@@ -52,8 +52,8 @@ var _ = ginkgo.Describe("[basic][custom_node_labels] Node Tuning Operator custom
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 			// Expect the default worker node profile applied prior to getting any current values.
-			ginkgo.By(fmt.Sprintf("waiting for TuneD profile %s on node %s", util.DefaultWorkerProfile, node.Name))
-			err = util.WaitForProfileConditionStatus(cs, pollInterval, waitDuration, node.Name, util.DefaultWorkerProfile, tunedv1.TunedProfileApplied, coreapi.ConditionTrue)
+			ginkgo.By(fmt.Sprintf("waiting for TuneD profile %s on node %s", util.GetDefaultWorkerProfile(node), node.Name))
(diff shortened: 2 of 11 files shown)


kpouget:
thanks @jmencak, the PR looks good
could you explain quickly the rational behind the changes?
I see
1. vm.nr_hugepages=16  -> 1. Not enough memory for 16 huge pages ........ ?
2. `joydev` module replaced by `m4`, I assume `joydev` isn't available on some platform, but `m4` is
3. master/infra nodes get `DefaultMasterProfile = "openshift-control-plane"`, makes sense
4. I don't understand what this refers to (in the code):
> Fix for sysctl_d_override.go rollback (node/pod was always nil and rollback in case of a test failure was not performed).

jmencak:
> thanks @jmencak, the PR looks good could you explain quickly the rational behind the changes? I see

Thank you for looking and commenting, Kevin!

>     1. vm.nr_hugepages=16  -> 1. Not enough memory for 16 huge pages ........ ?

This is not necessarily architecture specific.  However, I had issues allocating the 16 huge pages on standard AWS Graviton2 instances, vs. no issues on `x86_64`.  So, this is an attempt at lowering the barrier for passing the test.  If I see failures even with this in the future, I'll consider removing this test completely from the e2e test suite.

>     2. `joydev` module replaced by `m4`, I assume `joydev` isn't available on some platform, but `m4` is

Exactly.

>     4. I don't understand what this refers to (in the code):
> 
> 
> > Fix for sysctl_d_override.go rollback (node/pod was always nil and rollback in case of a test failure was not performed).

```
pod, err = 
```
vs
```
pod, err :=
```

kpouget:
I still don't understand 4, but we can discuss it tomorrow. (I thrust you on that, just curious to understand what's happening)

/lgtm
/hold
^^^ unhold when you want to merge

kpouget:
> > I still don't understand 4, but we can discuss it tomorrow. (I thrust you on that, just curious to understand what's happening)
> 
> `node` and `pod` variables are declared _and_ assigned `nil` value [here](https://github.com/openshift/cluster-node-tuning-operator/blob/03838478d355fca77c3e6be30a0ecab95bbed5c8/test/e2e/basic/sysctl_d_override.go#L30) (declared+assigned==defined). Now, [node here](https://github.com/openshift/cluster-node-tuning-operator/blob/03838478d355fca77c3e6be30a0ecab95bbed5c8/test/e2e/basic/sysctl_d_override.go#L59) and [pod here](https://github.com/openshift/cluster-node-tuning-operator/blob/03838478d355fca77c3e6be30a0ecab95bbed5c8/test/e2e/basic/sysctl_d_override.go#L61) unfortunately defines new variables in the scope of the function, therefore not using/masking the variables outside the function. I.e. `node` and `pod` were always `nil` until this change. Makes sense now?
> 

hum, got it, with `node := ...` you were masking the var defined at the beginning of the `Context` function, and thus `if node != nil: util.ExecAndLogCommand...` was never executed, because `node == nil` was true any time

tricky to spot a single-char `:` bug, good catch :)

PR #324, opened 2022-03-24
Performance profile render command integration

MarSik:
The original PAO code supports an offline render mode where it reads the PerformanceProfile manifest from a file and generates all the outputs as files. This is useful for Day 0 and GitOps based installations.

This PR is work in progress and depends on https://github.com/openshift/cluster-node-tuning-operator/pull/322

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,12 @@ test-e2e:
 	  KUBERNETES_CONFIG="$(KUBECONFIG)" $(GO) test -v -timeout 40m ./test/e2e/$$d -ginkgo.v -ginkgo.noColor -ginkgo.failFast || exit; \
 	done
 
+.PHONY: test-e2e-local
+test-e2e-local:
+	for d in performanceprofile/functests-render-command/1_render_command; do \
+	  $(GO) test -v -timeout 40m ./test/e2e/$$d -ginkgo.v -ginkgo.noColor -ginkgo.failFast || exit; \
+	done
+
 verify:	verify-gofmt
 
 verify-gofmt:
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -14,6 +14,8 @@ import (
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	olmv1 "github.com/operator-framework/api/pkg/operators/v1"
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -26,6 +28,7 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/pkg/config"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/metrics"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/operator"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/cmd/render"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/signals"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/tuned"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/util"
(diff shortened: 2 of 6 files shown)


MarSik:
@jmencak Hi, this PR introduces cobra to the main.go file. PAO has the dependency anyway, but I changed the way flags are parsed. Would this approach be acceptable?

The render command is then executed as

podman run -it cluster-node-tuning-operator:4.11 render --help

When no subcommand is used, NTO starts as usual.

jmencak (commented):
Thank you for the PR.  I think this approach is acceptable even though I'm far from thrilled by the cobra usage.  But as you say, we already have the dependency.  The PR needs cleaning though:
- Can we have back the help for both `cluster-node-tuning-operator` and `openshift-tuned` that would show it can be used with the `-v` options?
- `-v` options should work!
- Also, some comments about the usage for `openshift-tuned` would be welcome as you already added them for `cluster-node-tuning-operator`

test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml
```diff
@@ -61,7 +61,8 @@ spec:
   - machineConfigLabels:
       machineconfiguration.openshift.io/role: worker-cnf
     operand:
-      debug: false
+      tunedConfig:
+        reapply_sysctl: null
```
jmencak, line 65: `reapply_sysctl` is a Boolean, why does it have `null` value?

jmencak (changes requested):
Thank you for the PR and the changes.  While the e2e tests pass, the `-v=<number>` option for the operator itself doesn't work.  This needs to be fixed prior the merge.

```
$ ./cluster-node-tuning-operator -v=2
Error: invalid argument "2" for "-v, --version" flag: strconv.ParseBool: parsing "2": invalid syntax
Usage:
  cluster-node-tuning-operator [flags]
  cluster-node-tuning-operator [command]

Available Commands:
  completion  generate the autocompletion script for the specified shell
  help        Help about any command
  render      Render performance-addon-operator manifests

Flags:
      --enable-leader-election   Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager. (default true)
  -h, --help                     help for cluster-node-tuning-operator
  -v, --version                  Show program version and exit.

Use "cluster-node-tuning-operator [command] --help" for more information about a command.
```

```

File: test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml (hunk 1 above)
Comment (MarSik) [reply test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml:65]: You tell me :) This is the value the test is seeing, I only updated the expected value for it to pass as PAO does not care about this part.

File: cmd/cluster-node-tuning-operator/main.go [hunk 2]
```diff
@@ -63,97 +65,125 @@ func printVersion() {
...
+
+func operatorRun() {
+	klog.InitFlags(nil)
+	printVersion()
+
+	if showVersionAndExit {
+		return
+	}
+
+	klog.Infof("Starting the Cluster node tuning operator.")
```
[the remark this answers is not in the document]
MarSik: It was a debug helper, because the operator was failing to start and I could not figure out where. I can remove it.

test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml
jmencak: I might tell you, but I have no idea about the test you're talking about. :)  I guess it is `null` because `tunedConfig` was not specified at all?  Good to know it was manually updated by you.

cmd/cluster-node-tuning-operator/main.go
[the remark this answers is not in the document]
jmencak: I'll leave the decision up to you.  Either remove or change the log message.  Whichever you feel is better.

Makefile
```diff
@@ -97,6 +97,12 @@ test-e2e:
 	  KUBERNETES_CONFIG="$(KUBECONFIG)" $(GO) test -v -timeout 40m ./test/e2e/$$d -ginkgo.v -ginkgo.noColor -ginkgo.failFast || exit; \
 	done
 
+.PHONY: test-e2e-local
+test-e2e-local:
```
yanirq, line 101: Should we make the make target more specific as in https://github.com/openshift/cluster-node-tuning-operator/blob/131604110a23144fa0ccafc7be5a204a4cb6e4b5/Makefile#L202 ?

yanirq:
@MarSik  please make sure to also update this section:
https://github.com/openshift/cluster-node-tuning-operator/blob/master/docs/performanceprofile/performance_addon.md#render-mode

Makefile
MarSik: I was wondering if we can have more tests that are e2e in nature, but do not require a running cluster.. but the name is not important obviously.

test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml
MarSik: This is the expected value of Tuned when PerfProfile is processed by the render stage. If you check couple of lines above this, we mostly care about the generated tuned profile. The rest is just what NTO structures serialize to by default.

docs/performanceprofile/performance_addon.md
```diff
@@ -68,12 +68,12 @@ export ASSET_OUTPUT_DIR=<output path for the rendered manifests>
 
 Build and invoke the binary
 ```
-build/_output/bin/performance-addon-operators render (FIXME)
+_output/cluster-node-tuning-operator render (FIXME)
```
yanirq, line 71: we can remove the (FIXME) part, I placed it specifically for this PR

jmencak (commented):
The `-v` options seem to work now.  Once the nits:

- `(FIXME)`
- `s/tuned/TuneD/`

are addressed and the commits squashed, this is good to go from my side.

Makefile
yanirq: if we are  planing to add a render lane in openshift/release then it should correspond to the make target

yanirq: @MarSik  we currently have https://github.com/openshift/cluster-node-tuning-operator/blob/master/hack/run-render-command-functests.sh under this repo.  If this target comes to replace the invocation of this script [like we did in PAO ](https://github.com/openshift-kni/performance-addon-operators/blob/1df302f67a2d8338af090a5084ccaef6e902b283/Makefile#L280) then we can remove the script (although the script contains some additional decorations)

MarSik:
@yanirq @jmencak It seems all is ready for the final check now.

yanirq:
looks good to me. 
I will add a CI lane for "e2e-local"  under openshift/release.
@jmencak  feel free to give the final ack

PR #328, opened 2022-03-28
hack: fix installing the ginkgo

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -6,8 +6,9 @@ LATENCY_TEST_RUN=${LATENCY_TEST_RUN:-"false"}
 which ginkgo
 if [ $? -ne 0 ]; then
 	echo "Downloading ginkgo tool"
-	#go install github.com/onsi/ginkgo/ginkgo
-  go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+	# drop -mod=vendor flags, otherwise the installation will fail
+	# because of the package can not be installed under the vendor directory
+    GOFLAGS='' go install github.com/onsi/ginkgo/ginkgo@v1.16.5
 fi
 
 NO_COLOR=""
--- a/hack/run-perf-profile-creator-functests.sh
+++ b/hack/run-perf-profile-creator-functests.sh
@@ -4,9 +4,10 @@ GINKGO_SUITS=${GINKGO_SUITS:-"test/e2e/pao/functests-performance-profile-creator
 
 which ginkgo
 if [ $? -ne 0 ]; then
-	echo "Downloading ginkgo tool"
-	#go install github.com/onsi/ginkgo/ginkgo
-  go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+    echo "Downloading ginkgo tool"
+	# drop -mod=vendor flags, otherwise the installation will fail
+	# because of the package can not be installed under the vendor directory
+    GOFLAGS='' go install github.com/onsi/ginkgo/ginkgo@v1.16.5
 fi
 
 NO_COLOR=""
--- a/hack/run-render-command-functests.sh
+++ b/hack/run-render-command-functests.sh
@@ -4,9 +4,10 @@ GINKGO_SUITS=${GINKGO_SUITS:-"test/e2e/pao/functests-render-command"}
 
 which ginkgo
 if [ $? -ne 0 ]; then
-	echo "Downloading ginkgo tool"
-	#go install github.com/onsi/ginkgo/ginkgo
-  go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+    echo "Downloading ginkgo tool"
+	# drop -mod=vendor flags, otherwise the installation will fail
+	# because of the package can not be installed under the vendor directory
+    GOFLAGS='' go install github.com/onsi/ginkgo/ginkgo@v1.16.5
 fi
 
 NO_COLOR=""
--- a/manifests/45-webhook-configuration.yaml
+++ b/manifests/45-webhook-configuration.yaml
@@ -26,6 +26,8 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
     service.beta.openshift.io/inject-cabundle: "true"
   name: performance-addon-operator
 webhooks:


jmencak:
I don't think this will work in environments that are cut off from Internet access like the ART pipeline.  The question is whether we need this.  I specifically wanted to avoid installing `ginkgo` in the past to keep things a bit more robust with less cruft, but perhaps it is time for a change.  Thoughts?

cynepco3hahue:
> I don't think this will work in environments that are cut off from Internet access like the ART pipeline. The question is whether we need this. I specifically wanted to avoid installing `ginkgo` in the past to keep things a bit more robust with less cruft, but perhaps it is time for a change. Thoughts?

@jmencak Thanks for the comment. Does the ART run e2e tests?

jmencak:
> > I don't think this will work in environments that are cut off from Internet access like the ART pipeline. The question is whether we need this. I specifically wanted to avoid installing `ginkgo` in the past to keep things a bit more robust with less cruft, but perhaps it is time for a change. Thoughts?
> 
> @jmencak Thanks for the comment. Does the ART run e2e tests?

It does not to my knowledge and we are running the nightly tests separately on the nightlies instead.  So as I say, I'm open to change and perhaps the worry/current config is unfounded.

jmencak:
> > I don't think this will work in environments that are cut off from Internet access like the ART pipeline. The question is whether we need this. I specifically wanted to avoid installing `ginkgo` in the past to keep things a bit more robust with less cruft, but perhaps it is time for a change. Thoughts?
> 
> In general, we can do the same as we are doing with other tools, and build it from the vendor directory.

It is an option, but what you're doing is probably simpler and will keep the `Makefile` cleaner.  Let's wait for other thoughts. :)  I'm not saying what you do in the PR is a no-go.

jmencak:
If we do decide to build `gingko` from `vendor`, then we should also probably change the `Makefile` to make full use of it when running NTO e2e tests.

hack/run-functests.sh
```diff
@@ -7,7 +7,7 @@ which ginkgo
 if [ $? -ne 0 ]; then
 	echo "Downloading ginkgo tool"
 	#go install github.com/onsi/ginkgo/ginkgo
-  go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+  GOFLAGS='' go install github.com/onsi/ginkgo/ginkgo@v1.16.5
```
fromanirh, line 10: could you please add a comment/note in the commit message about what was the problem and why this fix resolves the problem?

fromanirh:
do we need ginkgo-the-runner or can we just use the ginkgo packages?

cynepco3hahue:
> do we need ginkgo-the-runner or can we just use the ginkgo packages?

We need the runner.

fromanirh:
> > do we need ginkgo-the-runner or can we just use the ginkgo packages?
> 
> We need the runner.

ok, last question: how costly is to avoid using the runner?

cynepco3hahue:
Ok the CI lane looks better now, it still has failures but at least it succeeded to run
For future investigation:
```
[Fail] [rfe_id:27368][performance] Validation webhook with API version v1alpha1 profile [It] should reject the creation of the profile with overlapping CPUs 
/go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/pao/functests/1_performance/performance.go:840
[Fail] [rfe_id:27368][performance] Validation webhook with API version v1alpha1 profile [It] should reject the creation of the profile with overlapping CPUs 
/go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/pao/functests/1_performance/performance.go:841
```

cynepco3hahue:
> ok, last question: how costly is to avoid using the runner?

@fromanirh Just to be sure, by this do you mean how costly it is to create a test binary that we can run without ginkgo?

hack/run-functests.sh
cynepco3hahue: Sure will add the comment.

yanirq:
@jmencak  @cynepco3hahue   either approach is fine. 
I suggest having a fix that will release other PRs if one approach is more costly and will take more time (which we can do in another PR)

fromanirh:
I'm not against the current approach btw

cynepco3hahue:
> I mean creating a ginkgo entry point like this: https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/e2e_serial_test.go which runs all the tests but does not need the `ginkgo` runner; we can just use `go test`.

I think in this case we probably can not use ginkgo flags, like fail fast. We can consider using it in the future, but first, we should be sure that all our tests are stable for 100%.

fromanirh:
> > I mean creating a ginkgo entry point like this: https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/e2e_serial_test.go which runs all the tests but does not need the `ginkgo` runner; we can just use `go test`.
> 
> I think in this case we probably can not use ginkgo flags, like fail fast. We can consider using it in the future, but first, we should be sure that all our tests are stable for 100%.

we can, but I agree let's get out of the woods first and if this is the quickest way, let's go with it

jmencak:
OK, let's wait for `ci/prow/e2e-gcp-pao` to pass.  Then we can squash, merge and cleanup (hopefully...) later.

fromanirh:
> > > ok, last question: how costly is to avoid using the runner?
> > 
> > 
> > @fromanirh Just to be sure, by this do you mean how costly it is to create a test binary that we can run without ginkgo?
> 
> I mean creating a ginkgo entry point like this: https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/e2e_serial_test.go which runs all the tests but does not need the `ginkgo` runner; we can just use `go test`.

*NON BLOCKING* PoC here: https://github.com/openshift/cluster-node-tuning-operator/pull/329 let's see how it goes

fromanirh:
> > > > ok, last question: how costly is to avoid using the runner?
> > > 
> > > 
> > > @fromanirh Just to be sure, by this do you mean how costly it is to create a test binary that we can run without ginkgo?
> > 
> > 
> > I mean creating a ginkgo entry point like this: https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/e2e_serial_test.go which runs all the tests but does not need the `ginkgo` runner; we can just use `go test`.
> 
> _NON BLOCKING_ PoC here: #329 let's see how it goes

the initial experiments demonstrate this is doable BUT having a 1:1 replacement is not so easy. We would possibly need some changes in how we run the suites, so by all means let's unblock quickly the e2e testing with this approach while #329 matures (or a better alternative emerges)

jmencak:
Great to see the `ci/prow/e2e-gcp-pao` job succeeded.  Can we squash the commits?  Then I'm happy to approve.

cynepco3hahue:
> Great to see the `ci/prow/e2e-gcp-pao` job succeeded. Can we squash the commits? Then I'm happy to approve.

Done.

jmencak:
Thank you for a clean commit message.
/approve
/lgtm

cynepco3hahue:
Fixed under different PR.

PR #331, opened 2022-03-29
Remove pao namespace from watched namespaces

yanirq:
Performance addons operator namespace removed from the controller runtime manager's namespace watch list to decrease the number of watches.

The only code aligned to this fix is the PAO OLM removal which is a one time operation that will not use the namespace above explicitly.


jmencak:
For reference, this should mitigate the issues mentioned [here](https://github.com/openshift/cluster-node-tuning-operator/pull/330#issue-1184664789).

PR #336, opened 2022-04-05
Fix PAO OLM operator removal

yanirq:
Perfromance addons operator removal will be done by searching for PAO deployment name that will  always euqal `performance-operator` . It will remove the csv and its matching subscription.

This fix will also ensure that all versions lower than 4.11
will remove the PAO.

unit test added.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile_controller.go
+++ b/pkg/performanceprofile/controller/performanceprofile_controller.go
@@ -30,7 +30,6 @@ import (
 	profileutil "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
-	olmv1 "github.com/operator-framework/api/pkg/operators/v1"
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -156,12 +155,11 @@ func (r *PerformanceProfileReconciler) removeOLMOperator() error {
 		subscription := &subscriptions.Items[i]
 		if subscription.Name == "performance-addon-operator" {
```
cynepco3hahue, line 156: IIRC we should not assume the name of the subscription to be `performance-addon-operator`

yanirq: @camilamacedo86  perhaps you can advise here.
In case we want to remove a 4.10 optional operator, following the [CLI steps](https://docs.openshift.com/container-platform/4.10/operators/admin/olm-deleting-operators-from-cluster.html) translated into code as suggested in this PR, what would be the cretin field of the subscription we can track and delete for our operator - `performance-addon-operator`.
In case  `Subscription.Name` field can be something other than `performance-addon-operator` , is there a field that will always contain the constant for `performance-addon-operator` for all types of deployments ? (CLI/UI/Manual)

cynepco3hahue: Probably you can find if the `subscription.Status.CurrentCSV` includes `performance-addon-operator`.

yanirq: I have changed the removal mechanism to get the csv first by the url

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -174,27 +154,45 @@ func (r *PerformanceProfileReconciler) removeOLMOperator() error {
 
 	for i := range csvs.Items {
 		csv := &csvs.Items[i]
-		if csv.Name == paoCSV {
-			klog.Infof("Removing performance-addon-operator CSV %s", paoCSV)
-			if err := r.Delete(context.TODO(), csv); err != nil {
-				return err
+		deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
```
cynepco3hahue, line 157: can we just check the deployment name? It should be hardcoded under the CSV, so it should be good enough for us.

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -174,27 +154,45 @@ func (r *PerformanceProfileReconciler) removeOLMOperator() error {
...
+			for _, deployment := range deploymentSpecs {
+				if &deployment.Spec.Template.Spec != nil {
+					containers := deployment.Spec.Template.Spec.Containers
+					if containers != nil {
+						for _, container := range containers {
+							// "performance-addon" will always be a substring of performance-addon-operator URLs.
+							if strings.Contains(container.Image, "performance-addon") {
+								deletedCSVs = append(deletedCSVs, csv.Name)
+								klog.Infof("Removing performance-addon-operator related CSV %s", csv.Name)
+								if err := r.Delete(context.TODO(), csv); err != nil {
```
cynepco3hahue, line 168: I am unsure if it is a good idea to delete the CSV before the subscription(OLM can re-create it)

yanirq: you have a good point here.

pkg/performanceprofile/controller/performanceprofile_controller.go
yanirq: will look into that

PR #337, opened 2022-04-07
Update NTO-generated annotations in MCs

jmencak:
When updating NTO-generated MachineConfigs, update `tuned.openshift.io/generated-by-controller-version` annotation with the NTO version performing the update.


dagrayvid:
Thanks Jiri, the code is easy to understand and seems correct.

/lgtm

PR #339, opened 2022-04-11
Implement a new workload hints functionality

cynepco3hahue:
The new API should make it easier to configure nodes for specific types of workloads via the performance profile.

### How does it look?
```
spec:
  workloadHints:
    realTime: true
...
spec:
  workloadHints:
    highPowerConsumption: false
```

### How does it work?

The tuned profile has:

Have a separate set of kernel arguments for each workload hint.
Set sysctl, cpu and scheduler section parameters according to
workload hints.


pkg/apis/performanceprofile/v1/performanceprofile_types.go
```diff
@@ -78,7 +81,7 @@ type CPUSet string
 // CPU defines a set of CPU related features.
 type CPU struct {
 	// Reserved defines a set of CPUs that will not be used for any container workloads initiated by kubelet.
-	Reserved *CPUSet `json:"reserved,omitempty"`
+	Reserved *CPUSet `json:"reserved"`
```
fromanirh, line 84: this is probably correct but should not be done in this PR. Or if it is, is not clear why and should be done on a separate commit with proper rationale

fromanirh (commented):
LGTM - one pretty minor comment inside
I'll need to review again the hints implementation proper, but the logic per se seems fine

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -121,12 +129,24 @@ avc_cache_threshold=8192
...
 initrd_dst_img=
 initrd_add_dir=
+
 # overrides cpu-partitioning cmdline
-cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} intel_pstate=disable nosoftlockup
+cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} systemd.cpu_affinity=${not_isolated_cores_expanded} intel_iommu=on iommu=pt
+
 {{if .StaticIsolation}}
-cmdline_realtime=+tsc=nowatchdog intel_iommu=on iommu=pt isolcpus=domain,managed_irq,${isolated_cores} systemd.cpu_affinity=${not_isolated_cores_expanded}
+cmdline_isolation=+isolcpus=domain,managed_irq,${isolated_cores}
```
fromanirh, line 137: the only other thing I'm not comfortable with in this PR is that in this commit we are both changing the profiles proper and the way we create them. I would be happier if, if possible, we can have 1. a commit which takes the current profiles unchanged and just changes the way we combine the existing data, using the workload hints. 2. a commit which changes the profiles per se, but NOT the way we combine the data. IOW, I'd love to split the responsabilities in this commit.
This is NOT BLOCKING, depending on what the other reviewers prefer.

pkg/apis/performanceprofile/v1/performanceprofile_types.go
cynepco3hahue: done

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -166,6 +168,21 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 		}
 	}
 
+	// TODO: I think we need to parse additional kernel arguments and in case when it intersect with workload hints
+	// kernel arguments we should or override it or drop an error
+	// set default values for workload hints
+	templateArgs[templateRealTime] = "true"
+	templateArgs[templateHighPowerConsumption] = "true"
```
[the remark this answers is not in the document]
cynepco3hahue: done

assets/performanceprofile/tuned/openshift-node-performance
cynepco3hahue: done

fromanirh (approved):
/lgtm
/hold

PR #340, opened 2022-04-11
Fix performance profile deepcopy generation

cynepco3hahue:
Move deepcopy generation to the separate make target, because
once the `pkg/generated` folder exists and is updated, the target will
continue to return `make: 'pkg/generated' is up to date` also when
deepcopy files require updates.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>


Makefile
```diff
@@ -52,21 +52,23 @@ clone-tuned:
...
-pkg/generated: $(API_TYPES)
+deepcopy:
 	$(GO) run k8s.io/code-generator/cmd/deepcopy-gen \
 	  --input-dirs $(PACKAGE)/$(API_TYPES_DIR)/tuned/v1,$(PACKAGE)/$(API_TYPES_DIR)/performanceprofile/v1alpha1,$(PACKAGE)/$(API_TYPES_DIR)/performanceprofile/v1,$(PACKAGE)/$(API_TYPES_DIR)/performanceprofile/v2 \
 	  -O $(API_ZZ_GENERATED) \
 	  --go-header-file $(API_GO_HEADER_FILE) \
 	  --bounding-dirs $(PACKAGE)/$(API_TYPES_DIR) \
 	  --output-base tmp
+
+pkg/generated: $(API_TYPES)
```
fromanirh, line 71: I believe is a bit better if we depend on `deepcopy` here instead of in `build`. IOW:
```
build: $(BINDATA) deepcopy pkg/generated build-performance-profile-creator
...

pkg/generated: $(API_TYPES) deepcopy
```

fromanirh (commented):
/lgtm
/hold
hold to give time to other reviewer and because of a inline non-blocking comment

Makefile
cynepco3hahue: I am unsure because deepcopy files are placed under pkg/apis/../.. directories, so it does not really relate to pkg/generated
but if you insist I can move it.

fromanirh: so let's wait for another review to lift the `hold`

yanirq:
In general looks good to me as well but since we are touching the main Makefile lets add @jmencak  for a final review

Makefile
jmencak: This change means that every time I make a simple change to NTO code and run `make`, `deepcopy-gen` will run, unnecessarily generating things that do not need to be generated and slowing the compilation down.  Can we not do this please? :)

cynepco3hahue: @jmencak Hm but I got another error, that during the change of the API it did not generate the updated deepcopy files. Any ideas on how to have the best of two worlds?

jmencak:
I guess I'm not understanding the problem here.  If I change the `*_types.go` files and type `make`, the `pkg/generated` target is executed and the `zz_generated.deepcopy.go` files updated.  If not, `pkg/generated` target is not executed.  Exactly what I expect. What am I missing?

Makefile
```diff
@@ -21,6 +21,7 @@ TUNED_DIR:=daemon
 API_TYPES_DIR:=pkg/apis
 API_TYPES:=$(shell find $(API_TYPES_DIR) -name \*_types.go)
 API_ZZ_GENERATED:=zz_generated.deepcopy
+API_ZZ_GENERATED_FILES:=$(shell find $(API_TYPES_DIR) -name \$(API_ZZ_GENERATED)*)
```
jmencak, line 24: Could you provide an reproducer for me to try and see this is really needed?

jmencak (commented):
Thank you for the PR, Artyom.  This looks mostly good to me, but I'm unconfortable adding extra complexity to the `Makefile` unless it is really needed.  Do you have a reproducer (set of steps) I can try to see and understand why this change is needed?

Makefile
cynepco3hahue: The reproducer will be:
1. Run `make pkg/generated`
2. Edit manually `pkg/apis/performanceprofile/v2/zz_generated.deepcopy.go`
3. Re-run `make pkg/generated`

Result:
It does not restore the `pkg/apis/performanceprofile/v2/zz_generated.deepcopy.go` to the correct format

In general, it is nothing that can not be fixed via `git` commands, but from generation tools, I am expecting that it will restore the content of generated files to have correct content, similar to files under `pkg/generated`.

jmencak: Would a simple change like
```
API_TYPES:=$(shell find $(API_TYPES_DIR) -name \*.go)
```
do what you want without adding extra complexity to the `Makefile`?  Perhaps adjusting the name of `API_TYPES` to something that more reflects what those sources are?

jmencak:
Thank you for the changes, Artyom.
/lgtm
/approve

pkg/apis/performanceprofile/v1/doc.go
```diff
@@ -0,0 +1,4 @@
+// +k8s:deepcopy-gen=package
+
+// Package v1 is the v2 version of the API.
+package v1
```
yanirq, line 4: Package v1 is the v2 version of the API -> Package v1 is the `v1` version of the API  ?

cynepco3hahue: thanks, done

PR #341, opened 2022-04-12
Bug 2074062: Remove TuneD profiles no longer used

jmencak:
TuneD profiles are extracted into all operands on every Tuned CR change. After their deletionin the CR, they were not removed from the operand's filesystems. In the past, it was not necessary to remove these profiles.  This changed with the introduction of conditional profile loading into TuneD.

As NTO's Cloud-specific profiles rely on conditional profile loading, this change makes sure all TuneD profiles the current recommended profile depends on are removed from /etc/tuned/<profile>/ once the same TuneD <profile> is no longer defined in any of Tuned CRs.

Resolves rhbz#2074062.


pkg/tuned/controller.go
```diff
@@ -411,11 +418,46 @@ func profilesExtract(profiles []tunedv1.TunedProfile) (bool, error) {
...
+// - Error if any or nil.
+func profilesSync(profiles []tunedv1.TunedProfile) (bool, error) {
+	change, extracted, recommendedProfileDeps, err := profilesExtract(profiles)
+	if err != nil {
+		return change, err
+	}
+
+	// Deal with TuneD profiles absent from Tuned CRs, but still present in /etc/tuned/<profile>/ the recommended profile depends on.
+	for profile := range recommendedProfileDeps {
+		if !extracted[profile] {
```
dagrayvid, line 448: Thanks @jmencak. Overall the code makes sense, just one idea for improvement:

I find the name of `extracted` a bit confusing here because the "problem" is that the profile IS extracted (or was), and because it was extracted, we need to delete it. However, we are checking if it is `!extracted` and if `!extracted`, we delete the corresponding profile dir.

I think renaming the variable to something like `newlyExtracted` or `extractedNew` might clarify things.

Similarly the name `recommendedProfileDeps` sounds like the list of profiles that we would need to keep in /etc/tuned/<profiles>, but it refers to the current (soon to be previous?) recommended-profile-dependencies. Maybe there is a better name we can come up with? I'm probably okay with keeping this as-is, if we make it clear that `extracted` is referring to the latest (future) needed dependencies.

Hope that makes sense, open to discussing further.

jmencak: Thank you for the suggestions, David.  I've replaced `s/extracted/extractedNew/` .  As for `recommendedProfileDeps` I do not have a better name. :/ It is one of the hardest problems in computer science. :)

dagrayvid:
Thanks for the change.

/lgtm

PR #342, opened 2022-04-12
Bug 2073938: Replace RuntimeClass version from v1beta1 to v1 

yanirq:
RuntimeClass in the node.k8s.io/v1beta1 API version
will [no longer be served in v1.25](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#runtimeclass-v125), therefore all occurrences in code containing v1beta1 are replaced with v1.
Kubernetes already supports the seamless transition for existing runtimeclasses.node.k8s.io v1beta1 to v1.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go
@@ -10,7 +10,7 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/tuned"
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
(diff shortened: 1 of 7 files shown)


jmencak:
This is a straingforward `s///` and this change looks good to me.  Even the commit message looks mostly ok, so thank you for that.  However, could we please keep the first letter of the commit message uppercase?  I.e. `s/replace/Replace/`.  Thank you!

cynepco3hahue:
@yanirq The change lgtm, can you please verify if it works correctly during the upgrade?

yanirq:
> @yanirq The change lgtm, can you please verify if it works correctly during the upgrade?

@cynepco3hahue I tested a 4.11 cluster with a runtime object defined as v1beta1 which immediately got replaced with v1.
After that I have deployed NTO with the current content, a performance profile with the relevant machine config pool to go with it without any issues.

jlojosnegros (approved):
/lgtm

PR #343, opened 2022-04-13
  Add build and ppc test invocation under one target

yanirq:
Add build and ppc test targets under `test-e2e-local`
that already contains the render test so there will be one main
target for local e2e tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ test-e2e:
 	done
 
 .PHONY: test-e2e-local
-test-e2e-local:
+test-e2e-local: build performance-profile-creator-tests
 	for d in performanceprofile/functests-render-command/1_render_command; do \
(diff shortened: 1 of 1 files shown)


yanirq:
This PR will correspond with https://github.com/openshift/release/pull/27797

jmencak:
The title seems a bit misleading to me.  You're just adding dependencies to a target, right?

yanirq:
> The title seems a bit misleading to me. You're just adding dependencies to a target, right?

Right, but making the test-e2e-local the main target to be triggered by the CI lane, also for future non cluster e2e tests.
At the moment there are only 2.

yanirq:
> The title seems a bit misleading to me. You're just adding dependencies to a target, right?

@jmencak  commit title changed

jmencak:
Thank you for the change, Yanir.
/lgtm

PR #345, opened 2022-04-18
Add performanceProfile API offline cores

marioferh:
## Description

Provide the ability through the performance addon operator to offline a set of unused cpus to reduce the overall power consumption of the cluster.

## Type of change

- [x] Add PerformanceProfile API 
- [ ] Add e2e test
- [ ] Add must gather data collect

## Testing steps
Launch test with PerformanceProfile:

  cpu:
    isolated: "1-2"
    reserved: "0"
    offlined: "4"

check if offlined cpu are offline:
sh-4.4# cat /sys/devices/system/cpu/offline 
4

### DIFF
--- a/assets/performanceprofile/scripts/set-cpus-offline.sh
+++ b/assets/performanceprofile/scripts/set-cpus-offline.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+
+for cpu in ${OFFLINE_CPUS//,/ };
+  do
+    online_cpu_file="/sys/devices/system/cpu/cpu$cpu/online"
+    if [ ! -f "${online_cpu_file}" ]; then
+      echo "ERROR: ${online_cpu_file} does not exist, abort script execution"
+      exit 1
+    fi
+  done
+
+echo "All cpus offlined exists, set them offline"
+
+for cpu in ${OFFLINE_CPUS//,/ };
+  do
+    online_cpu_file="/sys/devices/system/cpu/cpu$cpu/online"
+    echo 0 > "${online_cpu_file}"
+    echo "offline cpu num $cpu"
+  done
+
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
@@ -127,7 +127,12 @@ func (r *PerformanceProfile) validateCPUs() field.ErrorList {
 		}
 
 		if r.Spec.CPU.Isolated != nil && r.Spec.CPU.Reserved != nil {
-			cpuLists, err := components.NewCPULists(string(*r.Spec.CPU.Reserved), string(*r.Spec.CPU.Isolated))
+			var offlined string
+			if r.Spec.CPU.Offlined != nil {
+				offlined = string(*r.Spec.CPU.Offlined)
+			}
+
+			cpuLists, err := components.NewCPULists(string(*r.Spec.CPU.Reserved), string(*r.Spec.CPU.Isolated), offlined)
 			if err != nil {
 				allErrs = append(allErrs, field.InternalError(field.NewPath("spec.cpu"), err))
 			}
@@ -141,8 +146,14 @@ func (r *PerformanceProfile) validateCPUs() field.ErrorList {
 					allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu.isolated"), r.Spec.CPU.Isolated, "isolated CPUs can not be empty"))
 				}
 
-				if overlap := cpuLists.Intersect(); len(overlap) != 0 {
+				if overlap := components.Intersect(cpuLists.GetIsolated(), cpuLists.GetReserved()); len(overlap) != 0 {
 					allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu"), r.Spec.CPU, fmt.Sprintf("reserved and isolated cpus overlap: %v", overlap)))
+				}
+			}
+
+			if r.Spec.CPU.Offlined != nil {
+				if overlap := components.Intersect(cpuLists.GetReserved(), cpuLists.GetOfflined()); len(overlap) != 0 {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu"), r.Spec.CPU, fmt.Sprintf("reserved and offlined cpus overlap: %v", overlap)))
 				}
 			}
 		}
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
@@ -20,9 +20,11 @@ const (
 	// HugePagesCount defines the huge page count used for tests
 	HugePagesCount = 4
 	// IsolatedCPUs defines the isolated CPU set used for tests
-	IsolatedCPUs = CPUSet("4-7")
+	IsolatedCPUs = CPUSet("4-6")
 	// ReservedCPUs defines the reserved CPU set used for tests
 	ReservedCPUs = CPUSet("0-3")
+	// ReservedCPUs defines the reserved CPU set used for tests
+	OfflinedCPUs = CPUSet("7")
 	// SingleNUMAPolicy defines the topologyManager policy used for tests
 	SingleNUMAPolicy = "single-numa-node"
 
@@ -48,6 +50,7 @@ func NewPerformanceProfile(name string) *PerformanceProfile {
 	size := HugePageSize1G
 	isolatedCPUs := IsolatedCPUs
 	reservedCPUs := ReservedCPUs
+	offlinedCPUs := OfflinedCPUs
 	numaPolicy := SingleNUMAPolicy
 
 	netDeviceName := NetDeviceName
@@ -64,6 +67,7 @@ func NewPerformanceProfile(name string) *PerformanceProfile {
 			CPU: &CPU{
 				Isolated: &isolatedCPUs,
 				Reserved: &reservedCPUs,
+				Offlined: &offlinedCPUs,
 			},
 			HugePages: &HugePages{
 				DefaultHugePagesSize: &size,
@@ -140,9 +144,11 @@ var _ = Describe("PerformanceProfile", func() {
 
 		It("should reject cpus allocation with no reserved CPUs", func() {
 			reservedCPUs := CPUSet("")
-			isolatedCPUs := CPUSet("0-7")
-			profile.Spec.CPU.Isolated = &isolatedCPUs
+			isolatedCPUs := CPUSet("0-6")
+			offlinedCPUs := CPUSet("7")
 			profile.Spec.CPU.Reserved = &reservedCPUs
+			profile.Spec.CPU.Isolated = &isolatedCPUs
+			profile.Spec.CPU.Offlined = &offlinedCPUs
 			errors := profile.validateCPUs()
 			Expect(errors[0].Error()).To(ContainSubstring("reserved CPUs can not be empty"))
 		})
@@ -157,7 +163,17 @@ var _ = Describe("PerformanceProfile", func() {
 			Expect(errors[0].Error()).To(ContainSubstring("isolated CPUs can not be empty"))
 		})
 
-		It("should reject cpus allocation with overlapping sets", func() {
+		It("should allow cpus allocation with no offlined CPUs", func() {
+			cpusIsolaled := CPUSet("0")
+			cpusReserved := CPUSet("1")
+			profile.Spec.CPU.Isolated = &cpusIsolaled
+			profile.Spec.CPU.Reserved = &cpusReserved
+			profile.Spec.CPU.Offlined = nil
+			errors := profile.validateCPUs()
+			Expect(errors).To(BeEmpty())
+		})
+
+		It("should reject cpus allocation with overlapping sets between reserved and isolated", func() {
 			reservedCPUs := CPUSet("0-7")
 			isolatedCPUs := CPUSet("0-15")
 			profile.Spec.CPU.Reserved = &reservedCPUs
@@ -165,6 +181,18 @@ var _ = Describe("PerformanceProfile", func() {
 			errors := profile.validateCPUs()
 			Expect(errors).NotTo(BeEmpty(), "should have validation error when reserved and isolation CPUs have overlap")
 			Expect(errors[0].Error()).To(ContainSubstring("reserved and isolated cpus overlap"))
+		})
+
+		It("should reject cpus allocation with overlapping sets between reserved and offlined", func() {
+			reservedCPUs := CPUSet("0-7")
+			isolatedCPUs := CPUSet("8-11")
+			offlinedCPUs := CPUSet("0,12-15")
+			profile.Spec.CPU.Reserved = &reservedCPUs
+			profile.Spec.CPU.Isolated = &isolatedCPUs
+			profile.Spec.CPU.Offlined = &offlinedCPUs
+			errors := profile.validateCPUs()
+			Expect(errors).NotTo(BeEmpty(), "should have validation error when reserved and isolation CPUs have overlap")
+			Expect(errors[0].Error()).To(ContainSubstring("reserved and offlined cpus overlap"))
 		})
 	})
 
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -15,6 +15,7 @@ import (
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
 	"k8s.io/utils/pointer"
 
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
@@ -48,6 +49,7 @@ const (
 	udevRpsRules       = "99-netdev-rps.rules"
 	// scripts
 	hugepagesAllocation = "hugepages-allocation"
+	setCPUsOffline      = "set-cpus-offline"
 	ociHooks            = "low-latency-hooks"
 	setRPSMask          = "set-rps-mask"
 )
@@ -76,6 +78,7 @@ const (
 	environmentHugepagesSize  = "HUGEPAGES_SIZE"
 	environmentHugepagesCount = "HUGEPAGES_COUNT"
 	environmentNUMANode       = "NUMA_NODE"
+	environmentOfflineCpus    = "OFFLINE_CPUS"
 )
 
 const (
@@ -139,7 +142,7 @@ func getIgnitionConfig(profile *performancev2.PerformanceProfile) (*igntypes.Con
 
 	// add script files under the node /usr/local/bin directory
 	mode := 0700
-	for _, script := range []string{hugepagesAllocation, ociHooks, setRPSMask} {
+	for _, script := range []string{hugepagesAllocation, ociHooks, setRPSMask, setCPUsOffline} {
 		dst := getBashScriptPath(script)
 		content, err := assets.Scripts.ReadFile(fmt.Sprintf("scripts/%s.sh", script))
 		if err != nil {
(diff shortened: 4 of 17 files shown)


assets/performanceprofile/scripts/add-offline-cpus.sh
```diff
@@ -0,0 +1,18 @@
...
+for cpu in $cpus;
+  do
+    online_cpu_file="/sys/devices/system/cpu/cpu$cpu/online"
+    if [ ! -f "${online_cpu_file}" ]; then
+      echo "ERROR: ${online_cpu_file} does not exist"
+      exit 1
+    fi
+    echo "$disable_cpu" > "${online_cpu_file}"
+    echo "offline cpu num $cpu"
+  done
```
yanirq, line 18: add new line

manifests/50-operator.yaml
```diff
@@ -40,7 +40,7 @@ spec:
         tolerationSeconds: 120 # Evict pods within 2 mins.
       containers:
         - name: cluster-node-tuning-operator
-          image: registry.ci.openshift.org/openshift/origin-v4.0:cluster-node-tuning-operator
+          image: quay.io/mariofer/cluster-node-tuning-operator:4.11_offline
```
yanirq, line 43: make sure to revert this

manifests/50-operator.yaml
```diff
@@ -59,7 +59,7 @@ spec:
           - name: RESYNC_PERIOD
             value: "600"
           - name: CLUSTER_NODE_TUNED_IMAGE
-            value: registry.ci.openshift.org/openshift/origin-v4.0:cluster-node-tuning-operator
+            value: quay.io/mariofer/cluster-node-tuning-operator:4.11_offline
```
yanirq, line 62: same

hack/clean-deploy.sh
```diff
@@ -59,3 +39,9 @@ do
     $OC_TOOL delete mcp $mcp
 done
 
+oc delete subscription performance-addon-operator -n openshift-performance-addon-operator
```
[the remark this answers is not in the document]
marioferh: right

assets/performanceprofile/scripts/add-offline-cpus.sh
```diff
@@ -0,0 +1,19 @@
...
+  do
+    online_cpu_file="/sys/devices/system/cpu/cpu$cpu/online"
+    if [ ! -f "${online_cpu_file}" ]; then
+      echo "ERROR: ${online_cpu_file} does not exist"
+      exit 1
+    fi
+    echo "$disable_cpu" > "${online_cpu_file}"
+    echo "offline cpu num $cpu"
+  done 
+
```
yanirq, line 19: redundant line at EOF

yanirq:
[manifests/50-operator-ibm-cloud-managed.yaml](https://github.com/openshift/cluster-node-tuning-operator/pull/345/files#diff-cca9add0dc7e5f27efb2a7027d6f8531814119ca08eea003627715bbfa85eac7)  is wrongly deleted in this PR.
I suggest using pin point commits with only the relevant files and/or use stashing before pushing changes.

assets/performanceprofile/scripts/add-offline-cpus.sh
marioferh: done

assets/performanceprofile/scripts/add-offline-cpus.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+
+disable_cpu="0"
+
+cpus="$(echo "${OFFLINE_CPUS}" | grep -o -E '[0-9]+')"
```
fromanirh, line 7: please consider using the shell builtins here: https://stackoverflow.com/questions/27702452/loop-through-a-comma-separated-shell-variable
```
for i in ${OFFLINE_CPUS//,/ }; do echo "$i"; done
```

assets/performanceprofile/scripts/add-offline-cpus.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+
+disable_cpu="0"
```
fromanirh, line 5: I think "0" and "1" are idiomatic in bash, so this is a bit of overdoing

assets/performanceprofile/scripts/add-offline-cpus.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/bash
```
fromanirh, line 1: more than _ADD_ offline cpus, this is more like _SET_ offline cpus (or even better set cpus offline)

fromanirh (commented):
possible improvements inline

jmencak:
@marioferh  is this going to work with core ranges, such as `4-7` as `isolated` and `reserved` do, or do I need to use `4,5,6,7` instead?

assets/performanceprofile/scripts/add-offline-cpus.sh
marioferh: done

assets/performanceprofile/scripts/add-offline-cpus.sh
marioferh: done

assets/performanceprofile/scripts/add-offline-cpus.sh
marioferh: done

marioferh:
> 

sure, it woks in the same way that isolated and reserverd, with cpusets.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
```diff
@@ -33,6 +33,27 @@ const hugepagesAllocationService = `
...
+      - contents: |
+          [Unit]
+          Description=OfflineCpus - offlined cpus: 6,7
+          Before=kubelet.service
+
+          [Service]
+          Environment=OFFLINE_CPUS=6,7
+          Type=oneshot
+          RemainAfterExit=true
+          ExecStart=/usr/local/bin/add-offline-cpus.sh
```
fromanirh, line 46: need updates (add -> set)
probably also in other places

marioferh: done

jmencak:
> > @marioferh is this going to work with core ranges, such as `4-7` as `isolated` and `reserved` do, or do I need to use `4,5,6,7` instead?
> 
> sure, it woks in the same way that isolated and reserverd, with cpusets.

I see, thank you.  Missed the `cpuset.Parse()`.

fromanirh:
/test e2e-no-cluster
it seems more like a infra issue. Let's retry to see if it's a glitch or not.

fromanirh:
ok, so it's not a gltich, but seems weird indeed. Let's start checking the config of the prow lane

marioferh:
> ok, so it's not a gltich, but seems weird indeed. Let's start checking the config of the prow lane

@fromanirh https://github.com/openshift/cluster-node-tuning-operator/pull/351 merged and rebased. Now it should work.

fromanirh (commented):
the logic here seems OK. Do we need validation here? ATM the user is free to offline any amount of CPUs, and any of them, from the reserved pool, from the isolated pool, anything.
Perhaps we should reject to offline CPUs in the reserved pool, and we shoudl also maybe enforce that at least one (physical) isolated CPU should be online any given time.
Before to merge, I think we need to add the first validation to avoid to involontarily kill the machine

fromanirh:
the test failure is legit:
```
 Summarizing 1 Failure:
[Fail] render command e2e test With a single performance-profile [It] Gets cli args and produces the expected components to output directory 
/go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests-render-command/1_render_command/render_test.go:100 
```

yanirq:
> the test failure is legit:
> 
> ```
>  Summarizing 1 Failure:
> [Fail] render command e2e test With a single performance-profile [It] Gets cli args and produces the expected components to output directory 
> /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests-render-command/1_render_command/render_test.go:100 
> ```

I agree,  since you are changing API and injecting new content to machineconfig you need to see that the expected rendered files being compared in the tests will match your changes

marioferh:
> the logic here seems OK. Do we need validation here? ATM the user is free to offline any amount of CPUs, and any of them, from the reserved pool, from the isolated pool, anything. Perhaps we should reject to offline CPUs in the reserved pool, and we shoudl also maybe enforce that at least one (physical) isolated CPU should be online any given time. Before to merge, I think we need to add the first validation to avoid to involontarily kill the machine

> Perhaps we should reject to offline CPUs in the reserved pool,
it makes no sense right? reserved isolated and offlined CPUS are added in performanceProfile, it cannot set offline a reserved one.

I can add a validation to not overlap offlined with reserved and isolated as now there is a validation to not overlap reserved and isolated.

fromanirh:
/retest
infra issue - cluster failed to go up

yanirq:
There is a real failure on the e2e-no-cluster lane:
`rendered manual_machineconfig.yaml file is not identical to its reference file`

marioferh:
> There is a real failure on the e2e-no-cluster lane: `rendered manual_machineconfig.yaml file is not identical to its reference file`

oh... two days ago was passing :S

pkg/performanceprofile/controller/performanceprofile/components/utils.go
```diff
@@ -103,9 +116,11 @@ func NewCPULists(reservedList, isolatedList string) (*CPULists, error) {
 	if err != nil {
 		return nil, err
 	}
+	offlined, _ := cpuset.Parse(offlinedList)
```
fromanirh, line 119: why are we ignoring the error here?

pkg/performanceprofile/controller/performanceprofile/components/utils.go
```diff
@@ -79,11 +79,12 @@ func CPUListToMaskList(cpulist string) (hexMask string, err error) {
 type CPULists struct {
 	reserved cpuset.CPUSet
 	isolated cpuset.CPUSet
+	offlined cpuset.CPUSet
 }
 
 // Intersect returns cpu ids found in both the provided cpuLists, if any
-func (c *CPULists) Intersect() []int {
-	commonSet := c.reserved.Intersection(c.isolated)
+func (c *CPULists) Intersect(firstSet cpuset.CPUSet, secondSet cpuset.CPUSet) []int {
```
fromanirh, line 86: minor/not blocking: this method doesn't use fields of `CPULists` anymore so it should be a free function.
And maybe, being it a two-liner, could be inlined in the calling sites.

pkg/performanceprofile/controller/performanceprofile/components/utils.go
fromanirh: This also triggers a bell. In a future PR, we should make sure we get and handle the parsing errors for the cpu sets above.

marioferh: missed. fix.

pkg/performanceprofile/controller/performanceprofile/components/utils.go
marioferh: Fix. Thanks again

pkg/apis/performanceprofile/v1/performanceprofile_types.go
```diff
@@ -98,6 +98,9 @@ type CPU struct {
 	// Defaults to "true"
 	// +optional
 	BalanceIsolated *bool `json:"balanceIsolated,omitempty"`
+	// Offline defines a set of CPUs that will be unused and set offline
+	// +optional
+	Offlined *CPUSet `json:"offlined"`
```
yanirq, line 103: Why are we adding the offlined field to v1 ?   do we need to support both v1 and v2 with new additions ? 
(cc @fromanirh)

fromanirh: right, good catch. We need to add it to `v2` only actually

yanirq:
I think this PR will also require us to add a new e2e test (could be under https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/1_performance/performance.go or a new class under that folder)
Where we take the existing profile that gets created by the e2e tests (or used in discovery mode), add an offline field with values (with probably amending the isolated and reserved), check values propagated to the machine and revert the profile back.

yanirq:
> > There is a real failure on the e2e-no-cluster lane: `rendered manual_machineconfig.yaml file is not identical to its reference file`
> 
> oh... two days ago was passing :S

It probably changed if you changed the content of the machine config, or more specifically - the set_offline_cpus script

pkg/apis/performanceprofile/v1/performanceprofile_types.go
marioferh: done

pkg/performanceprofile/controller/performanceprofile/components/utils.go
fromanirh: It seems to me we are still ignoring the error? should be
```
offlined, err := cpuset.Parse(offlinedList)
if err != nil {
   ...
```

fromanirh (commented):
almost there! now that only the v2 API is fixed, once the CI is happy we can have a final pass to make sure all loose ends are tied and then merge.
The e2e tests can be added on a followup PR.

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -127,7 +127,7 @@ func (r *PerformanceProfile) validateCPUs() field.ErrorList {
 		}
 
 		if r.Spec.CPU.Isolated != nil && r.Spec.CPU.Reserved != nil {
-			cpuLists, err := components.NewCPULists(string(*r.Spec.CPU.Reserved), string(*r.Spec.CPU.Isolated))
+			cpuLists, err := components.NewCPULists(string(*r.Spec.CPU.Reserved), string(*r.Spec.CPU.Isolated), string(*r.Spec.CPU.Offlined))
```
yanirq, line 130: What will happen if `r.Spec.CPU.Offlined` is `nil` ? I think you will need a more selective check and list creation here.

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -141,10 +141,16 @@ func (r *PerformanceProfile) validateCPUs() field.ErrorList {
 					allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu.isolated"), r.Spec.CPU.Isolated, "isolated CPUs can not be empty"))
 				}
 
-				if overlap := cpuLists.Intersect(); len(overlap) != 0 {
+				if overlap := components.Intersect(cpuLists.GetIsolated(), cpuLists.GetReserved()); len(overlap) != 0 {
```
yanirq, line 144: We need to make sure this is documented d/s as well

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -352,3 +396,11 @@ func renderCrioConfigSnippet(profile *performancev2.PerformanceProfile, src stri
...
 	return crioConfig.Bytes(), nil
 }
+
+func listToString(cpus []int) string {
+	items := make([]string, len(cpus))
+	for idx, cpu := range cpus {
+		items[idx] = strconv.FormatInt(int64(cpu), 10)
+	}
+	return strings.Join(items, ",")
+}
```
yanirq, line 406: If you keep the function then please move it to one of the util folder such as `pkg/performanceprofile/utils` or `pkg/util`

pkg/performanceprofile/controller/performanceprofile/components/utils.go
```diff
@@ -142,3 +152,11 @@ func CPUMaskToCPUSet(cpuMask string) (cpuset.CPUSet, error) {
 
 	return builder.Result(), nil
 }
+
+func ListToString(cpus []int) string {
```
yanirq, line 156: Function title is a bit general for int list to string.
For general knowledge , there are some one liners that could do the same such as:
`strings.Trim(strings.Join(strings.Fields(fmt.Sprint(int_slice)), ","), "[]")`

fromanirh: I asked about avoiding a similar oneliner, and I still think we should avoid them in general

yanirq: note taken

marioferh: this was the original one
```
func stringSplit(cpuListSlice []int) string {
	return strings.Trim(strings.Join(strings.Fields(fmt.Sprint(cpuListSlice)), ","), "[]")
}
```
Changed due a comment request

fromanirh: yes, comment was from me. Let's please keep it in the current form (and place?)

marioferh: I've mode to utils that make more sense for me, its an auxiliary func, right?

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
marioferh: right, refactoring

marioferh: add a fix. Not sure if it is the most clean one but... Suggestions are welcome

marioferh:
> I think this PR will also require us to add a new e2e test (could be under https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/1_performance/performance.go or a new class under that folder) where we take the existing profile that gets created by the e2e tests (or used in discovery mode), add an offline field with values (with probably amending the isolated and reserved), check values propagated to the machine and revert the profile back.

Good idea, test ongoing.

yanirq:
/lgtm
lets get @fromanirh  final review as well and we can release the hold once approved.

fromanirh (commented):
non-blocking comments which can and should be fixed in a later PR

fromanirh:
/approve
let's see if I can

fromanirh:
@marioferh once this merges please add links to the followup PRs

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -166,6 +182,18 @@ var _ = Describe("PerformanceProfile", func() {
 			Expect(errors).NotTo(BeEmpty(), "should have validation error when reserved and isolation CPUs have overlap")
 			Expect(errors[0].Error()).To(ContainSubstring("reserved and isolated cpus overlap"))
 		})
+
+		It("should reject cpus allocation with overlapping sets between reserved and offlined", func() {
```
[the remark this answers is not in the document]
marioferh: the old test is still present. line 176, this one is an addition

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
```diff
@@ -73,5 +94,15 @@ var _ = Describe("Machine Config", func() {
...
 
+		It("should add systemd unit to offlineCPUs", func() {
+			Expect(manifest).To(ContainSubstring(offlineCPUS))
+		})
+	})
+
+	Context("check listToString ", func() {
+		It("should create string from CPUSet", func() {
+			res := components.ListToString(CPUs)
+			Expect(res).To(ContainSubstring(CPUstring))
```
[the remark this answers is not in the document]
marioferh: rigth

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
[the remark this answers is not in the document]
fromanirh: right, it was just renamed. My bad, and less work for the future

assets/performanceprofile/scripts/set-cpus-offline.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+
+for cpu in ${OFFLINE_CPUS//,/ };
+  do
+    online_cpu_file="/sys/devices/system/cpu/cpu$cpu/online"
+    if [ ! -f "${online_cpu_file}" ]; then
+      echo "ERROR: ${online_cpu_file} does not exist"
+      exit 1
```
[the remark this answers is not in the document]
marioferh: done

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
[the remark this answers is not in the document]
marioferh: done

fromanirh (commented):
/lgtm

PR #346, opened 2022-04-19
Move PAO OLM removal to main

yanirq:
This will remove the olm artifacts before the PAO
controller starts instead of doing it in reconciliation.
Removal of subscription and csvs logic fixed.


cmd/cluster-node-tuning-operator/main.go
```diff
@@ -124,6 +126,10 @@ func operatorRun() {
 		klog.Exit(err)
 	}
 
+	if err := removePerformanceOLMOperator(restConfig); err != nil {
```
fromanirh, line 129: this make intuitive sense to me, but maybe elaborate a bit more in the commit message about why better to do this in main?

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -44,6 +45,24 @@ var _ = Describe("[performance][config] Performance configuration", func() {
 		RunningOnSingleNode = isSNO
 	})
 
+	It("should remove OLM artifacts for performance-addon-operator", func() {
```
fromanirh, line 48: I'm trying to think if we should have e2e tests for both the scenarios:

1. <= 4.10 PAO installed
2. no PAO installed at all
and I'm inclined to think yes.

However we can start with integration tests to cover all the flows, like we had previously

yanirq: This test case is pretty general but it will cover the basics of 1 and 2. 
But now thinking of it the test might need to be fixed to get matching subscriptions first according to the new code.

cmd/cluster-node-tuning-operator/main.go
yanirq: Added some more info on the commit message

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -177,6 +183,65 @@ func tunedOperandRun() {
 	tuned.Run(stopCh, &boolVersion, version.Version)
 }
 
+// Uninstall PAO OLM operator.
+// This should apply only from version 4.11
```
jmencak, line 187: While this comment is clear to us now, it may not be to someone else in the future.  Perhaps saying something about PAO shipping a core operator from 4.11 and no longer as OLM operator and that's why we are removing it?  I wonder if we should also add something like "Remove this code in OCP version 4.XY" if we know what XY should be, or at least "Remove this code in the future.".

jmencak (commented):
Thank you for the explanation in the commit log why the uninstall code should go in `main.go`. If/When you adjust the commit log, I suggest also `s/csvs/CSVs/`.  Leaving the correctness of the uninstallation code for LGTM to someone from the PAO team.

cmd/cluster-node-tuning-operator/main.go
yanirq: Note taken.  We can add  "Remove this code in OCP version 4.XY" if we can tell for sure that the upgrade path will not come from 4.10 (and below) to the version in hand (be it 4.11, 4.12 , ...)

cmd/cluster-node-tuning-operator/main.go
fromanirh: was this resolved?

yanirq: If the current (amended) [commit message](https://github.com/openshift/cluster-node-tuning-operator/pull/346/commits/3130030d8fa1871d419f668b17fce42cac455305) is still not satisfactory let me know.

test/e2e/performanceprofile/functests/0_config/config.go
fromanirh: ok, I guess we can revisit later if needed

cmd/cluster-node-tuning-operator/main.go
yanirq: comment updated.

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -177,6 +183,67 @@ func tunedOperandRun() {
 	tuned.Run(stopCh, &boolVersion, version.Version)
 }
 
+// Uninstall PAO OLM operator since PAO is shipped
+// as a core operator from 4.11.
+// This is relevant for any upgrade path of an Openshift cluster
```
jmencak, line 188: `s/Openshift/OpenShift/`

jmencak (commented):
Only nit.  Also, if we could fix the `s/olm/OLM/` in the commit log it would be great.

jmencak:
My concerns were addressed. Thank you for the changes, Yanir.  Let's have a LGTM from PAO team.

jmencak:
Whoops:
```
'_output/tools/bin/controller-gen-v0.6.0' schemapatch:manifests="./manifests" paths="./pkg/apis/tuned/v1" output:dir="/tmp/tmp.0Bortnd4Ml"
Error: unable to parse option "output:dir=/tmp/tmp.0Bortnd4Ml": [binary literal has no digits (at <input>:1:10)]
Usage:
  controller-gen [flags]
```

yanirq:
> Whoops:
> 
> ```
> '_output/tools/bin/controller-gen-v0.6.0' schemapatch:manifests="./manifests" paths="./pkg/apis/tuned/v1" output:dir="/tmp/tmp.0Bortnd4Ml"
> Error: unable to parse option "output:dir=/tmp/tmp.0Bortnd4Ml": [binary literal has no digits (at <input>:1:10)]
> Usage:
>   controller-gen [flags]
> ```

@jmencak  rebase was needed.  now complete.

yanirq:
@fromanirh  please take a final look and approval from our side

cmd/cluster-node-tuning-operator/main.go
fromanirh: the description of the change is still a bit too terse. Is not obvious what is the benefit of doing this like this, outside of the reconcile loop. Now that everyone has the context is pretty clear, but without some records this knowledge is going to fade away in few months.
The extra description in the commit message doesn't have to be an essay. Few lines describing the why and recording context are sufficient (the "how" is evident from the change itself).

yanirq: ack.  commit message has been modified.

fromanirh:
/lgtm
thanks for the update

PR #351, opened 2022-04-24
Fix local e2e test target

yanirq:
Change test-e2e-local to use only local directives
instead of the whole build target.


marioferh (approved):
/lgtm

PR #354, opened 2022-04-26
Bug 2087159: Modify PPC to generate new "offlined" parameter in Performance Profile

jlojosnegros:
A new parameter is going to be added to Performance Profile  by #345. 
This PR aims to modify the Performance Profile Creator to be able to generate it.

To calculate offlined cpus we take into account the following input parameters:
- `offlined-cpu-count` : the number of cpus the user wants to set offline.
- `disable-ht`: Marks if the user want to disable logical processor siblings to disable hyperthreading
- `power-consumption-mode`: values: “default”, “low-latency”, “ultra-low-latency”

if `disable-ht` is true sibling logical processors will not be considered in the calculation of any of the CPUSets.
Unless `power-consumption-mode` is equal to `ultra-low-latency` ( which is considered a high-consumption scenario), we will first try to look for a complete socket to set offline, that is a socket where all the logical processors are still not in any set.
If we still have not set offline enough logical processors to fulfill the `offlined-cpu-count` we will try to set offline as many logical processor siblings as possible
If we still have not set offline enough logical processors to fulfill the `offlined-cpu-count` we will try to set offline any logical processor that is not already in a set.

Depends on #345

### DIFF
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -87,6 +87,7 @@ var (
 type ProfileData struct {
 	isolatedCPUs             string
 	reservedCPUs             string
+	offlinedCPUs             string
 	nodeSelector             *metav1.LabelSelector
 	mcpSelector              map[string]string
 	performanceProfileName   string
@@ -178,6 +179,7 @@ func NewRootCommand() *cobra.Command {
 	}
 
 	root.PersistentFlags().IntVar(&pcArgs.ReservedCPUCount, "reserved-cpu-count", 0, "Number of reserved CPUs (required)")
+	root.PersistentFlags().IntVar(&pcArgs.OfflinedCPUCount, "offlined-cpu-count", 0, "Number of offlined CPUs")
 	root.PersistentFlags().BoolVar(&pcArgs.SplitReservedCPUsAcrossNUMA, "split-reserved-cpus-across-numa", false, "Split the Reserved CPUs across NUMA nodes")
 	root.PersistentFlags().StringVar(&pcArgs.MCPName, "mcp-name", "", "MCP name corresponding to the target machines (required)")
 	root.PersistentFlags().BoolVar(&pcArgs.DisableHT, "disable-ht", false, "Disable Hyperthreading")
@@ -356,6 +358,10 @@ func getDataFromFlags(cmd *cobra.Command) (ProfileCreatorArgs, error) {
 	if err != nil {
 		return creatorArgs, fmt.Errorf("failed to parse reserved-cpu-count flag: %v", err)
 	}
+	offlinedCPUCount, err := strconv.Atoi(cmd.Flag("offlined-cpu-count").Value.String())
+	if err != nil {
+		return creatorArgs, fmt.Errorf("failed to parse offlined-cpu-count flag: %v", err)
+	}
 	splitReservedCPUsAcrossNUMA, err := strconv.ParseBool(cmd.Flag("split-reserved-cpus-across-numa").Value.String())
 	if err != nil {
 		return creatorArgs, fmt.Errorf("failed to parse split-reserved-cpus-across-numa flag: %v", err)
@@ -394,6 +400,7 @@ func getDataFromFlags(cmd *cobra.Command) (ProfileCreatorArgs, error) {
 		MustGatherDirPath:           mustGatherDirPath,
 		ProfileName:                 profileName,
 		ReservedCPUCount:            reservedCPUCount,
+		OfflinedCPUCount:            offlinedCPUCount,
 		SplitReservedCPUsAcrossNUMA: splitReservedCPUsAcrossNUMA,
 		MCPName:                     mcpName,
 		TMPolicy:                    tmPolicy,
@@ -456,7 +463,11 @@ func getProfileData(args ProfileCreatorArgs, cluster ClusterData) (*ProfileData,
 	// same from hardware topology point of view
 
 	nodeHandle := cluster[mcp][0]
-	reservedCPUs, isolatedCPUs, err := nodeHandle.GetReservedAndIsolatedCPUs(args.ReservedCPUCount, args.SplitReservedCPUsAcrossNUMA, args.DisableHT)
+	systemInfo, err := nodeHandle.GatherSystemInfo()
+	if err != nil {
+		return nil, fmt.Errorf("failed to compute get system information: %v", err)
+	}
+	reservedCPUs, isolatedCPUs, offlinedCPUs, err := profilecreator.CalculateCPUSets(systemInfo, args.ReservedCPUCount, args.OfflinedCPUCount, args.SplitReservedCPUsAcrossNUMA, args.DisableHT, args.PowerConsumptionMode == ultraLowLatency)
 	if err != nil {
 		return nil, fmt.Errorf("failed to compute the reserved and isolated CPUs: %v", err)
 	}
@@ -465,6 +476,7 @@ func getProfileData(args ProfileCreatorArgs, cluster ClusterData) (*ProfileData,
 	kernelArgs := profilecreator.GetAdditionalKernelArgs(args.DisableHT)
 	profileData := &ProfileData{
 		reservedCPUs:           reservedCPUs.String(),
+		offlinedCPUs:           offlinedCPUs.String(),
 		isolatedCPUs:           isolatedCPUs.String(),
 		nodeSelector:           mcp.Spec.NodeSelector,
 		mcpSelector:            mcpSelector,
@@ -473,6 +485,7 @@ func getProfileData(args ProfileCreatorArgs, cluster ClusterData) (*ProfileData,
 		rtKernel:               args.RTKernel,
 		additionalKernelArgs:   kernelArgs,
 		userLevelNetworking:    args.UserLevelNetworking,
+		disableHT:              args.DisableHT,
 	}
 
 	// setting workload hints
@@ -517,6 +530,7 @@ type ProfileCreatorArgs struct {
 	MustGatherDirPath           string `json:"must-gather-dir-path"`
 	ProfileName                 string `json:"profile-name"`
 	ReservedCPUCount            int    `json:"reserved-cpu-count"`
+	OfflinedCPUCount            int    `json:"offlined-cpu-count"`
 	SplitReservedCPUsAcrossNUMA bool   `json:"split-reserved-cpus-across-numa"`
 	DisableHT                   bool   `json:"disable-ht"`
 	RTKernel                    bool   `json:"rt-kernel"`
@@ -528,6 +542,7 @@ type ProfileCreatorArgs struct {
 
 func createProfile(profileData ProfileData) error {
 	reserved := performancev2.CPUSet(profileData.reservedCPUs)
+
 	isolated := performancev2.CPUSet(profileData.isolatedCPUs)
 	// TODO: Get the name from MCP if not specified in the command line arguments
 	profile := &performancev2.PerformanceProfile{
@@ -552,6 +567,11 @@ func createProfile(profileData ProfileData) error {
 				TopologyPolicy: &profileData.topologyPolicy,
 			},
 		},
+	}
+
+	if len(profileData.offlinedCPUs) > 0 {
+		offlined := performancev2.CPUSet(profileData.offlinedCPUs)
+		profile.Spec.CPU.Offlined = &offlined
 	}
 
 	if len(profileData.additionalKernelArgs) > 0 {
--- a/pkg/performanceprofile/profilecreator/profilecreator.go
+++ b/pkg/performanceprofile/profilecreator/profilecreator.go
@@ -61,6 +61,19 @@ const (
 	allCores = -1
 )
 
+var (
+	// This filter is used to avoid offlining the first logical processor of each core.
+	// LogicalProcessors is a slice of integers representing the logical processor IDs assigned to
+	// a processing unit for a core. GHW API gurantees that the logicalProcessors correspond
+	// to hyperthread pairs and in the code below we select only the first hyperthread (id=0)
+	// of the available logical processors.
+	// Please refer to https://www.kernel.org/doc/Documentation/x86/topology.txt for more information on
+	// x86 hardware topology. This document clarifies the main aspects of x86 topology modelling and
+	// representation in the linux kernel and explains why we select id=0 for obtaining the first
+	// hyperthread (logical core).
+	filterFirstLogicalProcessorInCore = func(index, lpID int) bool { return index != 0 }
+)
+
 func getMustGatherFullPathsWithFilter(mustGatherPath string, suffix string, filter string) (string, error) {
 	var paths []string
 
@@ -227,6 +240,31 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
 	return ghw.CPU(ghwHandler.snapShotOptions)
 }
 
+func (ghwHandler GHWHandler) SortedCPU() (*cpu.Info, error) {
+	cpuInfo, err := ghw.CPU(ghwHandler.snapShotOptions)
+	if err != nil {
+		return nil, fmt.Errorf("can't obtain cpuInfo info from GHW snapshot: %v", err)
+	}
+
+	sort.Slice(cpuInfo.Processors, func(x, y int) bool {
+		return cpuInfo.Processors[x].ID < cpuInfo.Processors[y].ID
+	})
+
+	for _, processor := range cpuInfo.Processors {
+		for _, core := range processor.Cores {
+			sort.Slice(core.LogicalProcessors, func(x, y int) bool {
+				return core.LogicalProcessors[x] < core.LogicalProcessors[y]
+			})
+		}
+
+		sort.Slice(processor.Cores, func(x, y int) bool {
+			return processor.Cores[x].ID < processor.Cores[y].ID
+		})
+	}
+
+	return cpuInfo, nil
+}
+
 // SortedTopology returns a TopologyInfo struct that contains information about the Topology sorted by numa ids and cpu ids on the host system
 func (ghwHandler GHWHandler) SortedTopology() (*topology.Info, error) {
 	topologyInfo, err := ghw.Topology(ghwHandler.snapShotOptions)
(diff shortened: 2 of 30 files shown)


pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -289,30 +289,108 @@ func topologyHTDisabled(info *topology.Info) *topology.Info {
...
-		return cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("please specify the reserved CPU count in the range [1,%d]", cpuInfo.TotalThreads-1)
+		return cpuset.CPUSet{}, cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("please specify the reserved CPU count in the range [1,%d]", cpuInfo.TotalThreads-1)
+	}
+
+	if offlinedCPUCount >= int(cpuInfo.TotalThreads) {
+		return cpuset.CPUSet{}, cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("please specify the offlined CPU count in the range [0,%d]", cpuInfo.TotalThreads-1)
 	}
+
+	if reservedCPUCount+offlinedCPUCount >= int(cpuInfo.TotalThreads) {
+		return cpuset.CPUSet{}, cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("please specify the reserved plus offlined CPU count in the range [1,%d]", cpuInfo.TotalThreads-1)
```
fromanirh, line 310: shouldn't this range be `[0, len(cpuInfo.TotalThreads)-reservedCPUCount-1]` ?

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -289,30 +289,108 @@ func topologyHTDisabled(info *topology.Info) *topology.Info {
...
 	}
+
 	htEnabled, err := ghwHandler.IsHyperthreadingEnabled()
 	if err != nil {
-		return cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("can't determine if Hyperthreading is enabled or not: %v", err)
+		return cpuset.CPUSet{}, cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("can't determine if Hyperthreading is enabled or not: %v", err)
+	}
+
+	//Calculate offlined cpus
+	offlined, err := ghwHandler.GetOfflinedCPUs(cpuInfo, topologyInfo, offlinedCPUCount, disableHTFlag, htEnabled)
```
fromanirh, line 325: so offlined takes precedence over reserved and isolated when computing cpu set partitioning, right?
Would it be simpler (and safer) if we compute first the reserved set, then the offlined set and then we trivially (it seems trivial at least) compute the isolated set as the remaining CPU set?

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -289,30 +289,108 @@ func topologyHTDisabled(info *topology.Info) *topology.Info {
...
+
+	if lpOfflined < offlinedCPUCount {
+		log.Warnf("could not offline enough logical processors (required:%d, offlined:%d)", offlinedCPUCount, lpOfflined)
 	}
+	return offlinedCPUs.Result(), nil
+}
+
+func (ghwHandler GHWHandler) GetReservedCPUs(offlinedCPUs cpuset.CPUSet, topologyInfo *topology.Info, reservedCPUCount int, splitReservedCPUsAcrossNUMA bool, disableHTFlag bool, htEnabled bool) (cpuset.CPUSet, *topology.Info, error) {
+
+	// TODO maybe it is not a good idea to modify topoly information inside this function
```
fromanirh, line 386: Altering the topology data provided by the GHW is not wrong per se, but seems a smell. However it's possible that avoiding so would require significant refactoring (I can't say, but can't rule out either)

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: yeah you are right

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: I have nothing against doing it as you propose but I do not understand why it would be simpler and safer to calculate first reserved. 

I thought of calculating offlined first because:
- if user do not want to disable HT and offlinedCount is enough to disable a complete socket we can do it to save some power ( as far as I understood @marioferh  measurements that's the only scenario where some power saving occurs)
- otherwise I will try to offline the sibling threads to help with performance.

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: Yeah, that's true. I think we could revisit this after this feature is done.

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -346,15 +438,23 @@ func newCPUAccumulator() *cpuAccumulator {
...
+func (ca *cpuAccumulator) AddCoresWithFilter(max int, cores []*cpu.ProcessorCore, filterLogicalProcessor func(int, int) bool) int {
+	initialCount := ca.count
 	for _, processorCore := range cores {
-		for _, core := range processorCore.LogicalProcessors {
+		for index, logicalProcessorId := range processorCore.LogicalProcessors {
 			if ca.count < max || max == allCores {
-				ca.builder.Add(core)
-				ca.count++
+				if filterLogicalProcessor(index, logicalProcessorId) {
+					ca.builder.Add(logicalProcessorId)
```
jlojosnegros, line 451: I think there could be an error here with cpuAccumulator.
If `logicalProcessorId` is already present in the builder it will be updated but the number of elements will not be increased therefore `ca.count` will have a wrong value.

If `build.Add` could return the number of elements added this would help ... otherwise I think ca.count should be calculated as `builder.Size` to avoid the problem.
@fromanirh, @Tal-or @swatisehgal  what do you think?

pkg/performanceprofile/profilecreator/profilecreator.go
fromanirh: well, the offline amount is anyway bound between 1 and (total - reserved), so it seems nicer to let this limitation be visible in the error message that, unless I'm missing something, can otherwise be misleading.
And yes, we totally need to have at least 1 reserved core

pkg/performanceprofile/profilecreator/profilecreator.go
fromanirh: the reasoning is that reserved is compulsory, and the reserved set is the most important set because it's needed for the node to be functional. IOW, without proper setting reserved set, we can kill the node, which is basically among the worst possible outcomes.

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -289,30 +304,105 @@ func topologyHTDisabled(info *topology.Info) *topology.Info {
...
+}
+
+func (ghwHandler GHWHandler) GetOfflinedCPUs(cpuInfo *cpu.Info, topologyInfo *topology.Info, offlinedCPUCount int, disableHTFlag bool, htEnabled bool) (cpuset.CPUSet, error) {
+	offlinedCPUs := newCPUAccumulator()
+	lpOfflined := 0
+
+	//currently HT is enabled on the system and the user wants to disable HT
+	if htEnabled && disableHTFlag {
+		for _, node := range topologyInfo.Nodes {
+			lpOfflined += offlinedCPUs.AddCoresWithFilter(offlinedCPUCount, node.Cores, filterFirstLogicalProcessorInCore)
```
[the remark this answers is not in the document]
jlojosnegros: Well I do not know If this is an error perse ... I mean I have chosen not to keep offlining cpus as if `disable-ht` parameter has "precedence" over the number of cpus to disable, but I do not know if that assumption is correct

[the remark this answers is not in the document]
fromanirh: uhm, let's just clarify somewhere (commit message, code comments, design doc, you name it) how we want the options to interact and let's be consistent.
I don't actually remember how PPC treats HT disabling.

jlojosnegros:
just rebased over master

pkg/performanceprofile/profilecreator/profilecreator.go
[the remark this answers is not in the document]
jlojosnegros: Ok I have tried to clarify both the parameters and the decision flow in the design document. 
https://docs.google.com/document/d/1ZYs2CD9dfdXECzubpnpdkZXQtjajvhdKJ_jb28h_V4c/edit?pli=1#heading=h.5h27mxaz7he0

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: As this is an offline tool and I do not think we expect cpuAccumulator to have thousands of elements, could we go with using ca.builder.Size() to update ca.count in the meantime?

fromanirh: I guess we can. I was (and still am) thinking more about code clarity (`cpuset.Builder` reuse is effectively doing more harm than good now) rather than performance.

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: ok, let me then change the order in which they are calculated.

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: Make a change proposal here : https://github.com/openshift/cluster-node-tuning-operator/pull/354/commits/954adafefd73bcae947dcaaeea1c745d6a82687e

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -334,31 +334,53 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
+		done:  false,
 	}
 }
 
 // AddCores adds logical cores from the slice of *cpu.ProcessorCore to a CPUset till the cpuset size is equal to the max value specified
 // In case the max is specified as allCores, all the cores from the slice of *cpu.ProcessorCore are added to the CPUSet
-func (ca *cpuAccumulator) AddCores(max int, cores []*cpu.ProcessorCore) {
+func (ca *cpuAccumulator) AddCores(max int, cores []*cpu.ProcessorCore) int {
+	if ca.done {
+		return -1
```
fromanirh, line 491: let's return the error condition out of band:
```
func (ca *cpuAccumulator) AddCores(max int, cores []*cpu.ProcessorCore) (int, error) {
	if ca.done {
		return -1, fmt.Errorf("CPU accumulator finalized")
        }
        ...
}
```

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -334,31 +334,53 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
-				ca.count++
+				_, found := ca.elems[logicalProcessorId]
+				ca.elems[logicalProcessorId] = struct{}{}
+				if !found {
+					ca.count++
+				}
 			}
 		}
 	}
+	return ca.count - initialCount
```
fromanirh, line 507: this is covered by existing tests, right?

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -334,31 +334,53 @@ func (ghwHandler GHWHandler) GetReservedAndIsolatedCPUs(reservedCPUCount int, sp
...
+}
+
+func (ca *cpuAccumulator) mapKeys() []int {
+	keys := make([]int, len(ca.elems))
+	i := 0
+	for k := range ca.elems {
+		keys[i] = k
```
fromanirh, line 514: maybe just use append? also, we can probably inline this function in `Result`
```
func (ca *cpuAccumulator) Result() cpuset.CPUSet {
        var keys []int
        for k := range ca.elems {
             keys = append(keys, k)
        }
	return cpuset.NewCPUSet(keys...)
}
```

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -465,6 +472,7 @@ func getProfileData(args ProfileCreatorArgs, cluster ClusterData) (*ProfileData,
 	kernelArgs := profilecreator.GetAdditionalKernelArgs(args.DisableHT)
 	profileData := &ProfileData{
 		reservedCPUs:           reservedCPUs.String(),
+		offlinedCPUs:           "", //todo assing value when calculated
```
fromanirh: OK, this (and the comment above) are fixed in the next commit. In the future please consider fixing them in the same commit

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -289,30 +304,121 @@ func topologyHTDisabled(info *topology.Info) *topology.Info {
...
 	cpuInfo, err := ghwHandler.CPU()
 	if err != nil {
-		return cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("can't obtain CPU info from GHW snapshot: %v", err)
+		return cpuset.CPUSet{}, cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("can't obtain CPU info from GHW snapshot: %v", err)
 	}
 
+	// Check limits are in range
 	if reservedCPUCount <= 0 || reservedCPUCount >= int(cpuInfo.TotalThreads) {
-		return cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("please specify the reserved CPU count in the range [1,%d]", cpuInfo.TotalThreads-1)
+		return cpuset.CPUSet{}, cpuset.CPUSet{}, cpuset.CPUSet{}, fmt.Errorf("please specify the reserved CPU count in the range [1,%d]", cpuInfo.TotalThreads-1)
```
fromanirh, line 342: this should be addressed in a later PR, however is too early in the flow to target the user in the error message ("please specify..."). At this layer we should just report the error, and let another layer treat with the user interface

pkg/performanceprofile/profilecreator/profilecreator.go
fromanirh: typo: `topoly`

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -242,6 +242,31 @@ func (ghwHandler GHWHandler) CPU() (*cpu.Info, error) {
 	return ghw.CPU(ghwHandler.snapShotOptions)
 }
 
+func (ghwHandler GHWHandler) SortedCPU() (*cpu.Info, error) {
```
fromanirh, line 245: this feature should be proposed to GHW - we should file an issue at least.
Additionally, I cant' help but keep wondering if instead of mangling `cpu.Info` we shoudl just compute here the representation we need, which seems a simple `[]int`.
IOW, instead of

1. computing a new `cpuInfo` with the sorted values
2. iterate over the mangled `cpuInfo` to extract the cpuid slice
3. use the cpuid slice

just compute the cpuid slice here and use it directly

fromanirh (commented):
partial review, need to get to the allocation logic proper

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: Sorry but I thin I do not fully understand your comment. 

we need all the socket -> core -> logical processor structure if we want to use that information to selectively disable some processors.

fromanirh: right, we need to keep some hierarchy. I was focusing a too narrow use case.

fromanirh: so ok, also considering we have already `SortedTopology` let's keep this as is and let's address possible cleanups/ghw extensions later

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: nope but I have just added some new unit test to cover it.

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: done

cmd/performance-profile-creator/cmd/root.go
jlojosnegros: ok

test/e2e/performanceprofile/functests-performance-profile-creator/1_performance-profile_creator/ppc.go
```diff
@@ -168,6 +172,43 @@ var _ = Describe("[rfe_id:OCP-38968][ppc] Performance Profile Creator", func() {
 		})
 	})
 
+	It("[test_id:XXXX1] Verify PPC fails when offlined cpu count is more than available cpus", func() {
```
jlojosnegros, line 175: There is no "test_id" for these tests.
They are really simple tests, but I have left this here just to raise the question: Should I ask for a test_id or just ignore the tag?

fromanirh: let's start ignoring the tag, we can add it later

test/e2e/performanceprofile/functests-performance-profile-creator/1_performance-profile_creator/ppc.go
```diff
@@ -168,6 +172,43 @@ var _ = Describe("[rfe_id:OCP-38968][ppc] Performance Profile Creator", func() {
...
+		_, errData, _ := testutils.ExecAndLogCommandWithStderr(ppcPath, cmdArgs...)
+		ppcErrorString := errorStringParser(errData)
+		Expect(ppcErrorString).To(ContainSubstring("please specify the offlined CPU count in the range"))
+	})
+
+	It("[test_id:XXXX3] Verify PPC fails when the sum of offlined cpu count and reserved cpu count is more than available cpus", func() {
+		ppcArgs := []string{
+			fmt.Sprintf("--offlined-cpu-count=%d", 50),
+			fmt.Sprintf("--reserved-cpu-count=%d", 50),
+			"--rt-kernel=false",
```
fromanirh, line 203: do we want e2e tests for:

- offline and reserved sets overlapping
- all the above with rt-kernel=true?

pkg/performanceprofile/profilecreator/profilecreator_test.go
```diff
@@ -786,19 +786,50 @@ var _ = Describe("PerformanceProfileCreator: Test Helper cpuAccumulator", func()
...
 			cores := acc.Result().ToSlice()
 			Expect(cores).Should(Equal([]int{0, 1, 2}))
 		})
 
+		It("should not count elements already in the accumulator", func() {
+			acc := newCPUAccumulator()
+			node := topology1.Nodes[0]
+			n1, err := acc.AddCores(allCores, node.Cores)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(n1).To(BeNumerically("==", 4))
```
fromanirh, line 1120: or `To(Equal(4))` ?

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -289,30 +304,147 @@ func topologyHTDisabled(info *topology.Info) *topology.Info {
...
+
+func (ghwHandler GHWHandler) GetOfflinedCPUs(alreadyUsedCPUs cpuset.CPUSet, cpuInfo *cpu.Info, topologyInfo *topology.Info, offlinedCPUCount int, disableHTFlag bool, htEnabled bool) (cpuset.CPUSet, error) {
+	offlinedCPUs := newCPUAccumulator()
+	lpOfflined := 0
+
+	//currently HT is enabled on the system and the user wants to disable HT
+	if htEnabled && disableHTFlag {
+		for _, node := range topologyInfo.Nodes {
+			acc, err := offlinedCPUs.AddCoresWithFilter(offlinedCPUCount, node.Cores, func(index, lpID int) bool {
+				return !alreadyUsedCPUs.Contains(lpID) && filterFirstLogicalProcessorInCore(index, lpID)
```
fromanirh, line 406: do we have a e2e test which checks the CPUs picked by this logic and ensures we offlined only thread sibilings?

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -289,30 +304,147 @@ func topologyHTDisabled(info *topology.Info) *topology.Info {
...
+				if err != nil {
+					return cpuset.CPUSet{}, err
+				}
+				lpOfflined += acc
+			}
+		}
+	} else {
+		for _, processor := range cpuInfo.Processors {
+			if processor.NumThreads <= uint32(offlinedCPUCount-lpOfflined) {
+				// We can offline a complete CPU socket
```
fromanirh, line 427: same, do we have a e2e test which ensures that we picked all the cpus pertaining a socket? e.g. nothing more? (and probably nothing less?)

pkg/performanceprofile/profilecreator/profilecreator_test.go
```diff
@@ -404,6 +404,287 @@ var _ = Describe("PerformanceProfileCreator: Populating Reserved and Isolated CP
...
+			mustGatherDirAbsolutePath, err = filepath.Abs(mustGatherDirPath)
+			Expect(err).ToNot(HaveOccurred())
+			handle, err = NewGHWHandler(mustGatherDirAbsolutePath, node)
+			Expect(err).ToNot(HaveOccurred())
+			reservedCPUSet, isolatedCPUSet, offlinedCPUSet, err := handle.CalculateCPUSets(reservedCPUCount, offlinedCPUCount, splitReservedCPUsAcrossNUMA, disableHT)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(offlinedCPUSet.Intersection(reservedCPUSet).IsEmpty()).To(BeTrue())
+			Expect(offlinedCPUSet.Intersection(isolatedCPUSet).IsEmpty()).To(BeTrue())
+			Expect(offlinedCPUSet.Size()).To(BeNumerically("==", offlinedCPUCount))
```
fromanirh, line 460: we should probalby add a check that the union of the returned set is the total set of the CPUs in the system

fromanirh (commented):
/lgtm
/hold
Nice work! not blocking comments, let's see what other reviewers think

fromanirh:
/approve
we surely want this change

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: Yes and no .... 
There is a test addressing this case but to disable a complete socket is almost impossible right now. 
As reservedCPUs are calculated first and we do not have socket layout into account to do it, it is almost certain that at least one cpu of each socket would be reserved.

pkg/performanceprofile/profilecreator/profilecreator.go
fromanirh: ok, but we don't actually check in the system that cpuID=X is or is not sibling of cpuID=Y, right? meaning, we hardcoded the correct setting in the test, computed manually. Is not verified querying some data, is it?
Doing like this is OK, we can extend later.

pkg/performanceprofile/profilecreator/profilecreator.go
fromanirh: anyway, we can hunt and use more system topologies if this helps us

marioferh (approved):
Tested and from the feature point of view seems right. We can discuss about the criteria to choose offlined cpus but for this PR is more than enough. Thanks @jlojosnegros

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: Nope, we do not query any data in the tests, just compute the correct result manually.

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: > But we can reserve only cpus from socket 0 and offline all socket 1 in theory, can't we?

Surely we could do it, but right now we are not doing it at all. We do not consult CPU info to choose reserved CPUs, just TopologyInfo so we just choose the cpus in the same NUMA zone or balance them between the NUMA zones but we do not really know if they belong to the same or different sockets.

fromanirh: ok, so we need to first clarify/unify our own terminology and then iterate from there. Even with multi-NUMA single-socket CPUs offlining an entire NUMA zone can yield significant improvements.

pkg/performanceprofile/profilecreator/profilecreator_test.go
jlojosnegros: I have been trying to add this but reserved + offlined + isolated = total is not always true. 

When `disable-ht` is `true` and reserved are calculated topology is modified deleting all the "sibling" logical processors so they could not be eligible when isolated is calculated. 

This modified topology is used to calculate isolated but not for the offflined because in fact in this case we really want to offline the sibling logical processors.

TL;DR
So, if `disable-ht` is `true` then `reserved + offlined + isolated = total` will be `false` **unless** offlined-cpu-count >= number of logical processor siblings

fromanirh: That's a great testcase to add to document this behaviour :) This kind of knowledge should be explicitely recorded, better in tests (+ explanation comments where helpful)

jlojosnegros: done

fromanirh:
/hold cancel
lgtm is gone, no point in holding anymore

test/e2e/performanceprofile/functests-performance-profile-creator/1_performance-profile_creator/ppc.go
jlojosnegros: done

test/e2e/performanceprofile/functests-performance-profile-creator/1_performance-profile_creator/ppc.go
jlojosnegros: I can add them but `rt-kernel` is not checked to calculate any of the cpuSets.

fromanirh: ok, then no point. We're good as we are now.

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -61,6 +61,21 @@ const (
 	allCores = -1
 )
 
+var (
+	noFilerLogicalProcessors = func(int, int) bool { return true }
```
yanirq, line 65: s/noFilerLogicalProcessors/noFil**t**erLogicalProcessors

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -61,6 +61,21 @@ const (
...
+
+	// This filter is used to avoid offlining the first logical processor of each core.
+	// LogicalProcessors is a slice of ints representing the logical processor IDs assigned to
```
yanirq, line 68: s/ints/integers

pkg/performanceprofile/profilecreator/profilecreator.go
yanirq: other suggestion : change var name to unfilteredLogicalProcessors / allLogicalProcessors / LogicalProcessors

pkg/performanceprofile/profilecreator/profilecreator.go
yanirq: this is an old behavior we already had.  we should indeed scan and sweep these messages into a different layer.

test/e2e/performanceprofile/testdata/ppc-expected-profiles/profile1a.json
```diff
@@ -0,0 +1,8 @@
+{
+  "must-gather-dir-path": "must-gather.bare-metal",
+  "mcp-name": "worker-cnf",
+  "reserved-cpu-count": 4,
+  "offlined-cpu-count": 4,
+  "rt-kernel": true,
+  "power-consumption-mode": "low-latency"
+}
```
yanirq, line 8: missing new line

test/e2e/performanceprofile/testdata/ppc-expected-profiles/profile1b.json
```diff
@@ -0,0 +1,8 @@
+{
+  "must-gather-dir-path": "must-gather.bare-metal",
+  "mcp-name": "worker-cnf",
+  "reserved-cpu-count": 4,
+  "offlined-cpu-count": 40,
+  "rt-kernel": true,
+  "power-consumption-mode": "low-latency"
+}
```
yanirq, line 8: missing new line

test/e2e/performanceprofile/testdata/ppc-expected-profiles/profile1a.json
yanirq: this is repetitive in all the json files added here.

test/e2e/performanceprofile/testdata/ppc-expected-profiles/profile1a.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: performance.openshift.io/v2
+kind: PerformanceProfile
+metadata:
+  name: performance
+spec:
+  cpu:
+    isolated: 1,3-39,41,43,45,47,49-51,53-79
+    reserved: 0,2,40,42
+    offlined: 44,46,48,52
```
yanirq, line 9: @jlojosnegros   in all the expected profile nder this test data , do we have the `offlined` always set with values ? 
If the answer is yes then I would add at least one test profile without that field (since its optional) to test the default behavior.

test/e2e/performanceprofile/functests-performance-profile-creator/1_performance-profile_creator/ppc.go
jlojosnegros: ok. the parameter is irrelevant but it is mandatory.

pkg/performanceprofile/profilecreator/profilecreator.go
jlojosnegros: Maybe would be better to scan and sweep in a different PR, is that ok ?

jlojosnegros:
Minor changes and rebase over last master to avoid some CI problems.

jlojosnegros:
rebasing over last master

PR #359, opened 2022-05-04
Add e2e tests for PAO mustgather

mrniranjan:

### DIFF
--- a/test/e2e/performanceprofile/functests/6_mustgather_testing/mustgather.go
+++ b/test/e2e/performanceprofile/functests/6_mustgather_testing/mustgather.go
@@ -1,0 +1,120 @@
+package pao_mustgather
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
+	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var mustGatherPath = os.Getenv("MUSTGATHER_DIR")
+var mustGatherFailed bool
+var mustGatherContentDir string
+
+var _ = Describe("[rfe_id: 50649] Performance Addon Operator Must Gather", func() {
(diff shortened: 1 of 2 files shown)


ffromani (commented):
looks like a nice start. We should probably squash all these commits together.
Do we want to run `must-gather` in the testsuite or we want somehow depend on the must-gather output already be available? (I'd prefer the former, at least optionally)

mrniranjan:
@fromanirh i want to run it as part of downstream test suite.

fromanirh:
> @fromanirh i want to run it as part of downstream test suite.

ok, but in general this won't prevent us to possibly call must-gather in the testsuite, will it?

mrniranjan:
> > @fromanirh i want to run it as part of downstream test suite.
> 
> ok, but in general this won't prevent us to possibly call must-gather in the testsuite, will it?

No, Should i extend this test suite to call  mustgather from a specified image and then run the tests ?

test/e2e/performanceprofile/functests/6_mustgather_testing/mustgather.go
```diff
@@ -0,0 +1,120 @@
...
+			workerRTNodes, err = nodes.GetByLabels(testutils.NodeSelectorLabels)
+			Expect(err).ToNot(HaveOccurred())
+			workerRTNodes, err = nodes.MatchingOptionalSelector(workerRTNodes)
+			Expect(err).ToNot(HaveOccurred())
+			for _, node := range workerRTNodes {
+				nodeSpecificFiles := []string{
+					fmt.Sprintf("nodes/%s/%s_logs_kubelet.gz", node.Name, node.Name),
+					fmt.Sprintf("nodes/%s/lscpu", node.Name),
+					fmt.Sprintf("nodes/%s/lspci", node.Name),
+					fmt.Sprintf("nodes/%s/proc_cmdline", node.Name),
```
marioferh, line 88: maybe add here "sys/devices/system/cpu/offline" to confirm last must gather commit is working. Then I am creating a e2e to deploy performance profile with offlined cpus. Maybe you can use that one to check is the offline file has the right content.

fromanirh:
> > > @fromanirh i want to run it as part of downstream test suite.
> > 
> > 
> > ok, but in general this won't prevent us to possibly call must-gather in the testsuite, will it?
> 
> No, Should i extend this test suite to call mustgather from a specified image and then run the tests ?

I think we should, but not in this PR. We can start simple.

mrniranjan:
@fromanirh @marioferh I have squashed all the commits to one commit. Regarding offline cpus checking in mustgather. I will send a separate PR. Since the files  are in sysinfo.tgz. I need to figure out how to untar it in a test case.

fromanirh:
> @fromanirh @marioferh I have squashed all the commits to one commit. Regarding offline cpus checking in mustgather. I will send a separate PR. Since the files are in sysinfo.tgz. I need to figure out how to untar it in a test case.

We do something very similar in ghw, this can be a good example: https://github.com/jaypipes/ghw/blob/main/pkg/snapshot/unpack.go#L65

test/e2e/performanceprofile/functests/6_mustgather_testing/mustgather.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	testutils.BeforeAll(func() {
+		if mustGatherFailed {
+			Skip("No mustgather directory provided, check MUSTGATHER_DIR environment variable")
+		} else {
+			mustgatherPathContent, err := ioutil.ReadDir(mustGatherPath)
+			Expect(err).To(BeNil(), "failed to read the Mustgather Directory %s: %v", mustGatherPath, err)
+			for _, file := range mustgatherPathContent {
+				if strings.Contains(file.Name(), "registry") {
```
marioferh, line 45: why registry? the idea is take sysinfo.tgz, extract it and validate files.

PR #360, opened 2022-05-05
Add e2e offlined cpus

marioferh:
## Description

Add e2e test to check cpus are set offline after perfromanceProfile change and revert back to default status

## Type of change

- [x ] Add e2e test

## Testing steps
- Launch config default PerformanceProfile
- Change PefromanceProfile adding offlined cpus
- Check cpus in perfromance profile are offline in selected nodes
- Revert back to default config without offlined cpus

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -73,11 +73,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			Skip("Discovery mode enabled, performance profile not found")
 		}
 
-		workerRTNodes, err = nodes.GetByLabels(testutils.NodeSelectorLabels)
-		Expect(err).ToNot(HaveOccurred())
-		workerRTNodes, err = nodes.MatchingOptionalSelector(workerRTNodes)
-		Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("error looking for the optional selector: %v", err))
-		Expect(workerRTNodes).ToNot(BeEmpty(), "cannot find RT enabled worker nodes")
+		workerRTNodes = getUpdatedNodes()
 		profile, err = profiles.GetByNodeLabels(nodeLabel)
 		Expect(err).ToNot(HaveOccurred())
 		performanceMCP, err = mcps.GetByProfile(profile)
@@ -643,6 +639,74 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
 		})
 	})
+
+	Context("Offlined CPU API", func() {
+		BeforeEach(func() {
+			//Saving the old performance profile
+			initialProfile = profile.DeepCopy()
+		})
+
+		It("[disruptive] should set offline cpus after deploy PAO", func() {
+			for _, node := range workerRTNodes {
+				onlineCPUCount, err := nodes.ExecCommandOnNode([]string{"nproc", "--all"}, &node)
+				Expect(err).ToNot(HaveOccurred())
+				onlineCPUInt, err := strconv.Atoi(onlineCPUCount)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(onlineCPUInt).Should(BeNumerically(">=", 3))
+			}
+
+			// Create new performance with offlined
+			reserved := performancev2.CPUSet("0")
+			isolated := performancev2.CPUSet("1")
+			offlined := performancev2.CPUSet("2-3")
+			profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Enable UserLevelNetworking and add Devices in Profile")
+			profile.Spec.CPU = &performancev2.CPU{
+				Reserved: &reserved,
+				Isolated: &isolated,
+				Offlined: &offlined,
+			}
+			By("Updating the performance profile")
+			profiles.UpdateWithRetry(profile)
+
+			By("Applying changes in performance profile and waiting until mcp will start updating")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+
+			By("Waiting for MCP being updated")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+
+			workerRTNodes = getUpdatedNodes()
+			//Check offlined cpus are setting correctly
+			for _, node := range workerRTNodes {
+				offlinedOutput, err := nodes.ExecCommandOnNode([]string{"cat", "/sys/devices/system/cpu/offline"}, &node)
+				Expect(err).ToNot(HaveOccurred())
+				offlinedCPUSet, err := cpuset.Parse(offlinedOutput)
+				offlinedCPUSetProfile, err := cpuset.Parse(string(offlined))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(offlinedCPUSet.Equals(offlinedCPUSetProfile))
+			}
+		})
+
+		AfterEach(func() {
+			By("Reverting the Profile")
+			spec, err := json.Marshal(initialProfile.Spec)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(testclient.Client.Patch(context.TODO(), profile,
+				client.RawPatch(
+					types.JSONPatchType,
+					[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+				),
+			)).ToNot(HaveOccurred())
+
+			By("Applying changes in performance profile and waiting until mcp will start updating")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+
+			By("Waiting when mcp finishes updates")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+		})
+	})
 })
 
 func hugepagesPathForNode(nodeID, sizeINMb int) string {
@@ -668,4 +732,13 @@ func countHugepagesOnNode(node *corev1.Node, sizeInMb int) (int, error) {
 		count += t
 	}
 	return count, nil
+}
+
+func getUpdatedNodes() []corev1.Node {
+	workerRTNodes, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
+	workerRTNodes, err = nodes.MatchingOptionalSelector(workerRTNodes)
+	Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("error looking for the optional selector: %v", err))
+	Expect(workerRTNodes).ToNot(BeEmpty(), "cannot find RT enabled worker nodes")
+	return workerRTNodes
 }


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -1018,6 +1019,58 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			}
+			By("Updating the performance profile")
+			profiles.UpdateWithRetry(profile)
+
+			By("Waiting for MCP being updated")
+			mcps.WaitForCondition("worker-cnf", mcv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+
+			//Check offlined cpus are setting correctly
+			for _, node := range workerRTNodes {
+				Eventually(func() bool {
```
marioferh, line 1055: This check when mcp reboot the node. Maybe it's not neccesary

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -1018,6 +1019,58 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			profile.Spec.CPU = &performancev2.CPU{
+				Reserved: &reserved,
+				Isolated: &isolated,
+				Offlined: &offlined,
+			}
+			By("Updating the performance profile")
+			profiles.UpdateWithRetry(profile)
+
+			By("Waiting for MCP being updated")
+			mcps.WaitForCondition("worker-cnf", mcv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
```
marioferh, line 1051: Sometimes reaches tiemout. Second reboot takes more time than first one and go beyond 10 minutes of timeout. Maybe extend timeout of MCP check?

mrniranjan: @marioferh Can you move this test case to functests/2_performance_update/updating_profile.go.  We generally keep all the performance profile update tests in this file.

marioferh: @mrniranjan thanks, in fact in updating_profile.go there are more similar test

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -1018,6 +1019,52 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+
+	Context("Offlined CPU API", func() {
+		It("should set offline cpus after deploy PAO", func() {
+			By("Backing up the profile")
+			initialProfile := profile.DeepCopy()
+
+			// Create new performance with offlined
+			reserved := performancev2.CPUSet("0")
+			isolated := performancev2.CPUSet("1")
+			offlined := performancev2.CPUSet("2-3")
```
fromanirh, line 1031: we assume the node has >= 4 cpus, ideally we should check that's the case, perhaps skipping if the node has less than X cpus - to make sure we offline only a tiny fraction of them

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -1018,6 +1019,52 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+			By("Waiting for MCP being updated")
+			mcps.WaitForCondition("worker-cnf", mcv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+
+			//Check offlined cpus are setting correctly
+			for _, node := range workerRTNodes {
+				Eventually(func() bool {
+					offlinedOutput, err := nodes.ExecCommandOnMachineConfigDaemon(&node, []string{"cat", "/sys/devices/system/cpu/offline"})
+					Expect(err).ToNot(HaveOccurred())
+					fmt.Printf("mario %s", offlinedOutput)
+					return strings.Contains(string(offlinedOutput), string(offlined))
```
fromanirh, line 1053: we should parse the `offlinedOutput` and make sure it is the same cpuset we asked to offline. String comparison is a bit too brutal

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -1018,6 +1019,52 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+
+			By("Reverting the Profile")
+			spec, err := json.Marshal(initialProfile.Spec)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(testclient.Client.Patch(context.TODO(), profile,
+				client.RawPatch(
+					types.JSONPatchType,
+					[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+				),
+			)).ToNot(HaveOccurred())
```
fromanirh, line 1065: we need to wait for the revert to happen before to call the test done, and we need to make sure we revert is attempted anyway even if the test exits early

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -1018,6 +1019,52 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			})
 		})
 	})
+
+	Context("Offlined CPU API", func() {
+		It("should set offline cpus after deploy PAO", func() {
```
fromanirh, line 1024: let's add a label to make it possible to skip (or to focus on) this very potentially disruptive test, something like` [disruptive]` or something else

fromanirh (commented):
some of the comments potentially already addressed, let's doublecheck

test/e2e/performanceprofile/functests/1_performance/performance.go
marioferh: done

test/e2e/performanceprofile/functests/1_performance/performance.go
fromanirh: because there are multiple different representation of the same set, and we can't (nor should) the representation reported by the kernel

marioferh:
How can I prevent this error?
`   Offlined CPU API
  /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1023
    should set offline cpus after deploy PAO [It]
    /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1024
    Unexpected error:
              
        <*errors.errorString | 0xc0009e3940>: {
            s: "failed to run command [cat /sys/devices/system/cpu/offline]: output ; error ",
        }
        failed to run command [cat /sys/devices/system/cpu/offline]: output ; error 
    occurred
`

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -643,6 +644,71 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+					offlinedOutput, err := nodes.ExecCommandOnNode([]string{"cat", "/sys/devices/system/cpu/offline"}, &node)
+					Expect(err).ToNot(HaveOccurred())
+					offlinedCPUSet, err := cpuset.Parse(offlinedOutput)
+					offlinedCPUSetProfile, err := cpuset.Parse(string(offlined))
+					Expect(err).ToNot(HaveOccurred())
+					return offlinedCPUSet.Equals(offlinedCPUSetProfile)
+				}, cluster.ComputeTestTimeout(15*time.Minute, false), 1*time.Minute).Should(BeTrue(), "could not get correct offline cpus")
+			}
+
+			By("Reverting the Profile")
```
fromanirh, line 695: this should probably moved (in some form) in a `AfterEach` or `defer` block (I'm fine with either), to make sure it runs even if an `Expect` in lines 683-693 fails. We totally don't want to pollute the cluster state

marioferh: So right. 
Add beforeEach and afterEach to make it clearer and less buggy

fromanirh: thanks, looks nice

marioferh:
> How can I prevent this error?
> 
> ```
>    Offlined CPU API
>   /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1023
>     should set offline cpus after deploy PAO [It]
>     /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1024
>     Unexpected error:
>               
>         <*errors.errorString | 0xc0009e3940>: {
>             s: "failed to run command [cat /sys/devices/system/cpu/offline]: output ; error ",
>         }
>         failed to run command [cat /sys/devices/system/cpu/offline]: output ; error 
>     occurred
> ```

@fromanirh any idea?

fromanirh:
> > How can I prevent this error?
> > ```
> >    Offlined CPU API
> >   /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1023
> >     should set offline cpus after deploy PAO [It]
> >     /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1024
> >     Unexpected error:
> >               
> >         <*errors.errorString | 0xc0009e3940>: {
> >             s: "failed to run command [cat /sys/devices/system/cpu/offline]: output ; error ",
> >         }
> >         failed to run command [cat /sys/devices/system/cpu/offline]: output ; error 
> >     occurred
> > ```
> 
> Still failing for this. @fromanirh any idea?

could be a bunch of things. Does it fail all the time? On all the nodes?

marioferh:
> > > How can I prevent this error?
> > > ```
> > >    Offlined CPU API
> > >   /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1023
> > >     should set offline cpus after deploy PAO [It]
> > >     /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1024
> > >     Unexpected error:
> > >               
> > >         <*errors.errorString | 0xc0009e3940>: {
> > >             s: "failed to run command [cat /sys/devices/system/cpu/offline]: output ; error ",
> > >         }
> > >         failed to run command [cat /sys/devices/system/cpu/offline]: output ; error 
> > >     occurred
> > > ```
> > 
> > 
> > Still failing for this. @fromanirh any idea?
> 
> could be a bunch of things. Does it fail all the time? On all the nodes?

in local is working, it fails when it launches the cmd and the cluster is down due to reboot

fromanirh:
> > > > How can I prevent this error?
> > > > ```
> > > >    Offlined CPU API
> > > >   /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1023
> > > >     should set offline cpus after deploy PAO [It]
> > > >     /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:1024
> > > >     Unexpected error:
> > > >               
> > > >         <*errors.errorString | 0xc0009e3940>: {
> > > >             s: "failed to run command [cat /sys/devices/system/cpu/offline]: output ; error ",
> > > >         }
> > > >         failed to run command [cat /sys/devices/system/cpu/offline]: output ; error 
> > > >     occurred
> > > > ```
> > > 
> > > 
> > > Still failing for this. @fromanirh any idea?
> > 
> > 
> > could be a bunch of things. Does it fail all the time? On all the nodes?
> 
> in local is working, it fails when it launches the cmd and the cluster is down due to reboot

ok, then are waiting correctly for the node to be ready before to try to exec the command?

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -643,6 +644,75 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+			profiles.UpdateWithRetry(profile)
+
+			By("Applying changes in performance profile and waiting until mcp will start updating")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+
+			By("Waiting for MCP being updated")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+
+			//Check offlined cpus are setting correctly
+			for _, node := range workerRTNodes {
```
marioferh, line 686: (continue the discussion here for cleanliness).

Working in local.

Failing when node reboots but should be do the check only after waiting for MCP being updated, then node should be finish the reboot and be available, right?

fromanirh: yes, provided we wait for the right conditions (let's check)

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -643,6 +644,75 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+				Offlined: &offlined,
+			}
+			By("Updating the performance profile")
+			profiles.UpdateWithRetry(profile)
+
+			By("Applying changes in performance profile and waiting until mcp will start updating")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+
+			By("Waiting for MCP being updated")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
```
fromanirh, line 683: could be worth to wait for all the relevant nodes to be `Ready` here

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -643,6 +644,75 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+					types.JSONPatchType,
+					[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+				),
+			)).ToNot(HaveOccurred())
+
+			By("Applying changes in performance profile and waiting until mcp will start updating")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+
+			By("Waiting when mcp finishes updates")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
```
fromanirh, line 713: could be worth to wait for all the relevant nodes to be `Ready` here

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -643,6 +644,82 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+			By("Applying changes in performance profile and waiting until mcp will start updating")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+
+			By("Waiting for MCP being updated")
+			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+
+			//Check offlined cpus are setting correctly
+			for _, node := range workerRTNodes {
+				Eventually(func() bool {
+					// Check if node is Ready
```
fromanirh, line 688: need to get an update node object here

marioferh: how?

fromanirh: https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/utils/nodes/nodes.go

marioferh: Ok, deleted eventually and add the updated nodes, if "Waiting for MCP being updated" waits correctly, node should be ready and its not worth to check if its actually ready, right?

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -643,6 +639,74 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
 		})
 	})
+
+	Context("Offlined CPU API", func() {
+		BeforeEach(func() {
+			By("Saving the old performance profile")
```
fromanirh, line 645: this step is cheap and can't fail (bar disasters) so not really worth a `By` comment, but no big deal

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -643,6 +639,74 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+		AfterEach(func() {
+			By("Reverting the Profile")
+			spec, err := json.Marshal(initialProfile.Spec)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(testclient.Client.Patch(context.TODO(), profile,
+				client.RawPatch(
+					types.JSONPatchType,
+					[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+				),
+			)).ToNot(HaveOccurred())
```
fromanirh, line 701: this should be enclosed in a `Eventually` or a retry loop of sorts. The revert must really be careful about not leaving the cluster in a corrupted state

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -669,3 +733,12 @@ func countHugepagesOnNode(node *corev1.Node, sizeInMb int) (int, error) {
 	}
 	return count, nil
 }
+
+func getUpdatedNodes() []corev1.Node {
```
fromanirh, line 737: what's the benefit of helper functions used once? I'd inline back this code in the calling site if we don't plan to reuse it (and sometimes even if we plan to reuse it, depending on the code flow)

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
fromanirh: so, are we sure this is not needed? what provides this guarantee? (applies also above)

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
marioferh: copied from here: https://github.com/openshift/cluster-node-tuning-operator/blob/0686c0a24caffb32567dbb516d0b847cc613b933/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go#L628
should I also change it there?

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
marioferh: its called twice, extracted from here:
https://github.com/openshift/cluster-node-tuning-operator/blob/0686c0a24caffb32567dbb516d0b847cc613b933/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go#L76

fromanirh: ok

marioferh: "[fromanirh](https://github.com/fromanirh) [8 minutes ago](https://github.com/openshift/cluster-node-tuning-operator/pull/360#discussion_r869123175)

so, are we sure this is not needed? what provides this guarantee? (applies also above)"

I've answered before. If MCP is in stated updated, the node is ready. There are similar snippets in other tests.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
fromanirh: no, should be fixed on a separate PR

yanirq: I think we switched in the past to use patch instead of Update with retry on purpose : https://github.com/openshift-kni/performance-addon-operators/commit/a646abfa26253a6cef1c44285ae67e0f8d2a8c37

fromanirh: Yes, nice catch. Then ok, if Patch avoids the conflict then no need to wrap in a loop.

fromanirh:
LGTM
the test logic is OK for me. Leaving the cluster in a corrupted state because of a leaking test makes failures very hard to debug and troubleshoot because the failure can happen in totally unrelated places down the line, in the tests which happen to run later.
This is especially true for operations so invasive as disabling CPUs.
We took steps to prevent leaking tests, we will need to watch the next runs to debug and fix any failures.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
marioferh: It's a good idea to add the updated node object event if its not necessary (in this case) because the ExecCommandOnNode, only needs node.name and that doesn't change

`// GetMachineConfigDaemonByNode returns the machine-config-daemon pod that runs on the specified node
func GetMachineConfigDaemonByNode(node *corev1.Node) (*corev1.Pod, error) {
	listOptions := &client.ListOptions{
		Namespace:     testutils.NamespaceMachineConfigOperator,
		FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": node.Name}),
		LabelSelector: labels.SelectorFromSet(labels.Set{"k8s-app": "machine-config-daemon"}),
	}

	mcds := &corev1.PodList{}
	if err := testclient.Client.List(context.TODO(), mcds, listOptions); err != nil {
		return nil, err
	}

	if len(mcds.Items) < 1 {
		return nil, fmt.Errorf("failed to get machine-config-daemon pod for the node %q", node.Name)
	}
	return &mcds.Items[0], nil
}
`

Just to show that it was not the root issue of the bug

PR #363, opened 2022-05-12
Bug 2089414: PAO Add validation check between isolated and offlined CPUs

marioferh:
Isolated and offlined can not share any CPUs.
If there is an overlap, it fails.


jlojosnegros (approved):
/lgtm

PR #367, opened 2022-05-30
Bug 2083428: crio: Use absolute path for binaries in crio hook script

yanirq:
Fixes Bug 2083428: low-latency-hooks.sh script requires ip binary in container image

The binaries in the script used by cri-o prestart hook are fixed to use absolute path since 
$PATH is not guaranteed to contain all the needed paths for binaries invocations in prestart phase.


jmencak:
Thank you for the changes, they
/lgtm

jmencak (approved):

assets/performanceprofile/scripts/low-latency-hooks.sh
```diff
@@ -1,35 +1,38 @@
 #!/usr/bin/env bash
 
+BIN_PATH="/usr/bin"
+SBIN_PATH="/usr/sbin"
```
fromanirh, line 4: I'd rather use variables to encode the paths of all the binaries we depend on:
```
JQ="/usr/bin/jq"
IP="/usr/sbin/ip"
...
```
The extra benefit is we have all the deps explicit.
We should do this for all the commands we use which are not shell builtins
Even better, we should replace these scripts with real binaries in the future.

PR #368, opened 2022-06-03
e2e: performance profile: Latency tests stabilization

shajmakh:
- e2e: latency: fail the test if CPUs number is less than 1
- images: change ImagePullPolicy to default
- e2e: pp-latency: verify tool args
- e2e: pp-latency: add the runtime to stabilize the pod-phase condition

Please find each commit for more details

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -548,13 +548,13 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			err := testclient.Client.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
 
-			err = pods.WaitForPredicate(testpod, 10*time.Minute, func(pod *corev1.Pod) (bool, error) {
+			currentPod, err := pods.WaitForPredicate(testpod, 10*time.Minute, func(pod *corev1.Pod) (bool, error) {
 				if pod.Status.Phase != corev1.PodPending {
 					return true, nil
 				}
 				return false, nil
 			})
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred(), "expected the pod to keep pending, but its current phase is %s", currentPod.Status.Phase)
 
 			updatedPod := &corev1.Pod{}
 			err = testclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(testpod), updatedPod)
--- a/test/e2e/performanceprofile/functests/4_latency/latency.go
+++ b/test/e2e/performanceprofile/functests/4_latency/latency.go
@@ -313,7 +313,7 @@ func getLatencyTestCpus() (int, error) {
 		if err != nil {
 			return val, fmt.Errorf("the environment variable LATENCY_TEST_CPUS has incorrect value %q, it must be a positive integer with maximum value of %d: %w", latencyTestCpusEnv, math.MaxInt32, err)
 		}
-		if val < 0 || val > math.MaxInt32 {
+		if val <= 0 || val > math.MaxInt32 {
 			return val, fmt.Errorf("the environment variable LATENCY_TEST_CPUS has an invalid number %q, it must be a positive integer with maximum value of %d", latencyTestCpusEnv, math.MaxInt32)
 		}
 		return val, nil
--- a/test/e2e/performanceprofile/functests/4_latency/latency.go
+++ b/test/e2e/performanceprofile/functests/4_latency/latency.go
@@ -454,12 +454,17 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 	Expect(err).ToNot(HaveOccurred())
 
 	By("Waiting two minutes to download the latencyTest image")
-	err = pods.WaitForPhase(testPod, corev1.PodRunning, 2*time.Minute)
+	currentPod, err := pods.WaitForPredicate(testPod, 2*time.Minute, func(pod *corev1.Pod) (bool, error) {
+		if pod.Status.Phase == corev1.PodRunning {
+			return true, nil
+		}
+		return false, nil
+	})
 	if err != nil {
 		testlog.Error(err)
 		logEventsForPod(testPod)
 	}
-	Expect(err).ToNot(HaveOccurred())
+	Expect(err).ToNot(HaveOccurred(), "expected the pod to reach running phase, its current phase is %s", currentPod.Status.Phase)
 
 	if runtime, _ := strconv.Atoi(latencyTestRuntime); runtime > 1 {
 		By("Checking actual CPUs number for the running pod")
@@ -476,8 +481,9 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 	if err != nil {
 		testlog.Error(err)
 		logEventsForPod(testPod)
+		testlog.Info(getLogFile(node, logName))
 	}
-	Expect(err).ToNot(HaveOccurred(), getLogFile(node, logName))
+	Expect(err).ToNot(HaveOccurred())
 }
 
 func extractLatencyValues(logName string, exp string, node *corev1.Node) string {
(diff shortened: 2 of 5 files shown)


test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -207,31 +207,31 @@ func clearEnv() {
...
 	//testCpus: for tests that expect a success output message, note that an even CPU number is needed, otherwise the test would fail with SMTAlignmentError
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "6", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "1", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "60", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
+	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "30", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
```
Tal-or, line 224: If it applies to multiple cases I would set this value as a const.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -455,7 +456,9 @@ func createLatencyTestPod(testPod *corev1.Pod, node *corev1.Node, logName string
 		testlog.Error(err)
 		logEventsForPod(testPod)
 	}
-	Expect(err).ToNot(HaveOccurred())
+	updatedPod := &corev1.Pod{}
+	testclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(testPod), updatedPod)
+	Expect(err).ToNot(HaveOccurred(), "Expected the pod to reach running phase but its current phase is %s", updatedPod.Status.Phase)
```
fromanirh, line 461: this test looks wrong, don't we need to check the `updatedPod` phase?

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -207,31 +207,33 @@ func clearEnv() {
...
+	//1. to let the tools have their time to measure latency properly 2. have time to check that the pod phase turned running and not completed immediately
 	//testCpus: for tests that expect a success output message, note that an even CPU number is needed, otherwise the test would fail with SMTAlignmentError
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "2", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "0", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, testCpus: "6", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "1", testRun: "true", testRuntime: "10", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "60", testRun: "true", testRuntime: "2", testMaxLatency: guaranteedLatency, outputMsgs: []string{success}, toolToTest: toolToTest})
+	successRuntime := "30"
```
fromanirh, line 224: do we want to use the same value for all the tests? previously it was not the case.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -121,17 +122,27 @@ var _ = table.DescribeTable("Test latency measurement tools tests", func(testGro
...
+					testlog.Error("the tool to test was not set")
 				}
-				if test.toolToTest == cyclictest {
-					commandRegex = fmt.Sprintf("running the cyclictest command with arguments .*-D %s", test.testRuntime)
-				}
-				if test.toolToTest == hwlatdetect {
-					commandRegex = fmt.Sprintf("running the hwlatdetect command with arguments .*--duration %s", test.testRuntime)
+
+				for _, argument := range passedArgs {
+					Expect(strings.Contains(string(output), argument)).To(BeTrue(), "The tool command didn't pass the argument %q", argument)
```
fromanirh, line 144: please check the gomega matchers `ContainElement` and `ContainSubstring`

marioferh:
Trying to fix e2e-upgrade, if fails again, just retest, it fails 1 of 3 times more or less.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
shajmakh: yes, this is part of stabling the tests, the value is used for tests that expect a success message hence the tool is executed, it can be different from one test to another but we want to retain the runtime relevant to manage other checks in the main test like verifying the pod gets to running phase. so just for simplicity I used the same value for these tests.

test/e2e/performanceprofile/functests/4_latency/latency.go
shajmakh: the status is checked here:
https://github.com/openshift/cluster-node-tuning-operator/pull/368/files#diff-41952f42e8241558ea4f4d9cea270beab57c936f488bc6ff41984c20546278a6L453
maybe the name of the variable is misleading here, as it is not updated pod in terms of specs but of status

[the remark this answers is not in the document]
shajmakh: perhaps it's mistake calling the variable updatedpod, this variable gets the exact same pod with no changes in the object except for the status. so it should be called something like "currentPod".I saw the need to know what is the actual pod's status that didn't match `corev1.PodRunning` here:
https://github.com/openshift/cluster-node-tuning-operator/pull/368/files#diff-41952f42e8241558ea4f4d9cea270beab57c936f488bc6ff41984c20546278a6L453

shajmakh: I see, ok I'll make that a separate klog error

fromanirh: ok, my too terse statement may have been misleading, sorry about that. The problem here is intermixing the check to `err` and the message regarding the updated pod status. it should be probably be better fixed in `pods.WaitForPhase`.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
shajmakh: That sounds a good suggestion but it seems that ContainElement matcher expects array/ slice/map:
"Actual must be an array, slice or map"
https://onsi.github.io/gomega/#containelementselement-interface

fromanirh: fair enough, feel free to ignore

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -548,13 +548,13 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", func() {
 			err := testclient.Client.Create(context.TODO(), testpod)
```
fromanirh, line 548: nit: generally avoid Uppercase Letters In The Commit Message - with the possible exceptions of acronyms

fromanirh (commented):
/lgtm

fromanirh:
/approve
of course

fromanirh:
/hold
please unhold when ready (e.g. WIP in the title)

shajmakh:
/hold cancel
Thank you folks for the valuable reviews!

PR #370, opened 2022-06-06
PAO: fix performance update flaky test

marioferh:


marioferh (commented):
flaky test cannot be reproduced...

PR #372, opened 2022-06-09
Bug 2051443: latency-tests: oslat: increase minimal number of CPUs

Tal-or:
Oslat should exclude the sibling cpu of the main thread, hence we need one additional isolated CPU for making the test meaningful.
If there are not enough isolated CPUs, the test will skip.


dgonyier:
Not officially a reviewer, but LGTM 😄

Tal-or:
Should be handled on the runner side: https://github.com/openshift-kni/cnf-features-deploy/pull/1143

PR #373, opened 2022-06-13
Rename Performance Addon Controller to Performance Profile Controller

marioferh:
Rename old names
Clean lower case and uniform all names


docs/performanceprofile/api-versions.md
```diff
@@ -2,18 +2,18 @@
 
 ## Supported API Versions
 
-The Performance Addon controller supports *v2*, *v1* and *v1alpha1* [Performance Profile](performance_profile.md) versions.
+The Performance Controller Operator supports *v2*, *v1* and *v1alpha1* [Performance Profile](performance_profile.md) versions.
```
fromanirh, line 5: this is not a operator no more, so we should write "Performance Profile Controller" or something along these lines (here and below)

marioferh: done

fromanirh:
/lgtm
/hold
generally speaking this seems right, it's 4.12+ material, right?
hold to let other reviewers chime in

marioferh:
> /lgtm /hold generally speaking this seems right, it's 4.12+ material, right? hold to let other reviewers chime in

Well the change has been doing in 4.11. It's not super important but I don't have a strong opinion about it.

fromanirh:
It's a weird timing for this change to land in 4.11, but we can re-evaluate

marioferh:
> It's a weird timing for this change to land in 4.11, but we can re-evaluate

yep then better for 4.12

fromanirh:
/approve
/hold cancel
for 4.12/main branch only

PR #385, opened 2022-07-08
Switch to CentOS Stream for upstream Containerfile

jmencak:
CentOS 8 is dead, switch to CentOS Stream 8.


jmencak:
Based on our Slack conversation.
/cc @yanirq 
/cc @bartwensley 

Using `Dockerfile` (vs. `Dockerfine.rhel8`) makes building NTO without access to TuneD FDP CDN repos easier.

PR #388, opened 2022-07-15
WIP Bug 2102287: WorkloadHint realtime is not applied correctly on Performance Profile Controller

marioferh:


marioferh (commented):
/hold
Re order commits after rebases.
Wait for e2e test.

PR #395, opened 2022-07-20
deps: pin pcidb to 1.0.0

fromanirh:
Up until version 1.0.0 excluded, pcidb tried to download from network
the pcidb if that was missing, which was the case for container images
which did not install, directly or indirectly, the `hwdata` package.

The conditions to trigger a network download are actually non trivial -
the code needed to require PCI listing functionality, which are unlikely
here.

Nevertheless, the version 1.0.0 of `pcidb` switched the default to a
saner "do not attempt network fetch unless explicitely asked for" while
being otherwise backward compatible (API-wise), so let's consume this
dependency and let's get rid of this potential issue entirely.


fromanirh:
ok, CI is having hiccups so we have time to decide anyway.
For extra context: here's the 0.9.0 version of `ghw` bumping the dep to `pcidb` 1.0.0: https://github.com/jaypipes/ghw/blob/v0.9.0/go.mod note all previous versions depends on `pcidb` 0.6.0. 
pcidb changes: https://github.com/jaypipes/pcidb/releases/tag/v1.0.0

fromanirh:
trivial rebase

jmencak:
still
/lgtm

MarSik (approved):

fromanirh:
thanks @jmencak ! we are discussing the approach with @MarSik

fromanirh:
obsolete, we will re-evaluate later

PR #396, opened 2022-07-20
Bug 2105123: tuned: disable irqbalance

fromanirh:
The tuned irqbalance plugin clears the irqbalance banned CPUs list when tuned starts. The list is then managed dynamically by the runtime handlers.

On node restart, the tuned pod can be started AFTER the workload pods (kubelet nor kubernetes offers ordering guarantees when recovering the node state); clearing the banned CPUs list while pods are running and compromising the IRQ isolation guarantees. Same holds true if the NTO pod restarts for whatever reason.

The fix is twofold:
1. disable the irqbalance plugin, so tuned doesn't clear the irqbalance banned list anymore.
2. depend on crio to reset the banned CPU list earlier in the boot

in #413 we add e2e tests to verify the aforementioned steps.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2105123

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -38,8 +38,10 @@ transparent_hugepages=never
 
 {{if not .GloballyDisableIrqLoadBalancing}}
 [irqbalance]
-#> Override the value set by cpu-partitioning with an empty one
-banned_cpus=""
+# Disable the plugin entirely, which was enabled by the parent profile `cpu-partitioning`.
+# It can be racy if TuneD restarts for whatever reason.
+#> cpu-partitioning
+enabled=false
 {{end}}
 
 [scheduler]
--- a/pkg/performanceprofile/controller/performanceprofile/components/utils.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/utils.go
@@ -140,7 +140,7 @@ func CPUMaskToCPUSet(cpuMask string) (cpuset.CPUSet, error) {
 		}
 		mask, err := strconv.ParseUint(chunk, 16, bitsInWord)
 		if err != nil {
-			return cpuset.NewCPUSet(), fmt.Errorf("failed to parse the CPU mask %q: %v", cpuMask, err)
+			return cpuset.NewCPUSet(), fmt.Errorf("failed to parse the CPU mask %q (chunk %q): %v", cpuMask, chunk, err)
 		}
 		for j := 0; j < bitsInWord; j++ {
 			if mask&1 == 1 {
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -1,0 +1,247 @@
+package __performance
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/tuned"
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/mcps"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/util"
+	"github.com/openshift/cluster-node-tuning-operator/test/framework"
+)
+
+var (
+	cs = framework.NewClientSet()
+)
+
+var _ = Describe("[performance] Checking IRQBalance settings", func() {
+	var workerRTNodes []corev1.Node
+	var targetNode *corev1.Node
+	var profile *performancev2.PerformanceProfile
+	var performanceMCP string
+	var err error
+
+	BeforeEach(func() {
+		if discovery.Enabled() && testutils.ProfileNotFound {
+			Skip("Discovery mode enabled, performance profile not found")
+		}
+
+		workerRTNodes, err = nodes.GetByLabels(testutils.NodeSelectorLabels)
+		Expect(err).ToNot(HaveOccurred())
+		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+		Expect(err).ToNot(HaveOccurred())
+		performanceMCP, err = mcps.GetByProfile(profile)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Verify that worker and performance MCP have updated state equals to true
+		for _, mcpName := range []string{testutils.RoleWorker, performanceMCP} {
+			mcps.WaitForCondition(mcpName, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+		}
+
+		nodeIdx := pickNodeIdx(workerRTNodes)
+		targetNode = &workerRTNodes[nodeIdx]
+		By(fmt.Sprintf("verifying worker node %q", targetNode.Name))
+	})
+
+	Context("Verify irqbalance configuration handling", func() {
+
+		It("Should not overwrite the banned CPU set on tuned restart", func() {
+			if profile.Status.RuntimeClass == nil {
+				Skip("runtime class not generated")
+			}
+
+			if tuned.IsIRQBalancingGloballyDisabled(profile) {
+				Skip("this test needs dynamic IRQ balancing")
+			}
+
+			Expect(targetNode).ToNot(BeNil(), "missing target node")
+
+			irqAffBegin, err := getIrqDefaultSMPAffinity(targetNode)
+			Expect(err).ToNot(HaveOccurred(), "failed to extract the default IRQ affinity from node %q", targetNode.Name)
+			testlog.Infof("IRQ Default affinity on %q when test begins: {%s}", targetNode.Name, irqAffBegin)
+
+			bannedCPUs, err := getIrqBalanceBannedCPUs(targetNode)
+			Expect(err).ToNot(HaveOccurred(), "failed to extract the banned CPUs from node %q", targetNode.Name)
+			testlog.Infof("banned CPUs on %q when test begins: {%s}", targetNode.Name, bannedCPUs.String())
+
+			smpAffinitySet, err := nodes.GetDefaultSmpAffinitySet(targetNode)
+			Expect(err).ToNot(HaveOccurred(), "failed to get default smp affinity")
+
+			onlineCPUsSet, err := nodes.GetOnlineCPUsSet(targetNode)
+			Expect(err).ToNot(HaveOccurred(), "failed to get Online CPUs list")
+
+			// expect no irqbalance run in the system already, AKA start from pristine conditions.
+			// This is not an hard requirement, just the easier state to manage and check
+			Expect(smpAffinitySet.Equals(onlineCPUsSet)).To(BeTrue(), "found default_smp_affinity %v, expected %v - IRQBalance already run?", smpAffinitySet, onlineCPUsSet)
+
+			cpuRequest := 2 // minimum amount to be reasonably sure we're SMT-aligned
+			annotations := map[string]string{
+				"irq-load-balancing.crio.io": "disable",
+				"cpu-quota.crio.io":          "disable",
+			}
+			testpod := getTestPodWithProfileAndAnnotations(profile, annotations, cpuRequest)
+			testpod.Spec.NodeName = targetNode.Name
+
+			data, _ := json.Marshal(testpod)
+			testlog.Infof("using testpod:\n%s", string(data))
+
+			err = testclient.Client.Create(context.TODO(), testpod)
+			Expect(err).ToNot(HaveOccurred())
+			defer func() {
+				if testpod != nil {
+					testlog.Infof("deleting pod %q", testpod.Name)
+					deleteTestPod(testpod)
+				}
+				bannedCPUs, err := getIrqBalanceBannedCPUs(targetNode)
+				Expect(err).ToNot(HaveOccurred(), "failed to extract the banned CPUs from node %q", targetNode.Name)
+
+				testlog.Infof("banned CPUs on %q when test ends: {%s}", targetNode.Name, bannedCPUs.String())
+
+				irqAffBegin, err := getIrqDefaultSMPAffinity(targetNode)
+				Expect(err).ToNot(HaveOccurred(), "failed to extract the default IRQ affinity from node %q", targetNode.Name)
+
+				testlog.Infof("IRQ Default affinity on %q when test ends: {%s}", targetNode.Name, irqAffBegin)
+			}()
+
+			err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
+			logEventsForPod(testpod)
+			Expect(err).ToNot(HaveOccurred())
+
+			// now we have something in the IRQBalance cpu list. Let's make sure the restart doesn't overwrite this data.
+			postCreateBannedCPUs, err := getIrqBalanceBannedCPUs(targetNode)
+			Expect(err).ToNot(HaveOccurred(), "failed to extract the banned CPUs from node %q", targetNode.Name)
+			testlog.Infof("banned CPUs on %q just before the tuned restart: {%s}", targetNode.Name, postCreateBannedCPUs.String())
+
+			Expect(postCreateBannedCPUs.IsEmpty()).To(BeFalse(), "banned CPUs %v should not be empty on node %q", postCreateBannedCPUs, targetNode.Name)
+
+			By(fmt.Sprintf("getting a TuneD Pod running on node %s", targetNode.Name))
+			pod, err := util.GetTunedForNode(cs, targetNode)
+			Expect(err).NotTo(HaveOccurred())
+
+			By(fmt.Sprintf("causing a restart of the tuned pod (deleting the pod) on %s", targetNode.Name))
+			_, _, err = util.ExecAndLogCommand("oc", "delete", "pod", "--wait=true", "-n", pod.Namespace, pod.Name)
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func() error {
+				By(fmt.Sprintf("getting again a TuneD Pod running on node %s", targetNode.Name))
+				pod, err = util.GetTunedForNode(cs, targetNode)
+				if err != nil {
+					return err
+				}
+
+				By(fmt.Sprintf("waiting for the TuneD daemon running on node %s", targetNode.Name))
+				_, err = util.WaitForCmdInPod(5*time.Second, 5*time.Minute, pod, "test", "-e", "/run/tuned/tuned.pid")
+				return err
+			}).WithTimeout(5 * time.Minute).WithPolling(10 * time.Second).ShouldNot(HaveOccurred())
+
+			By(fmt.Sprintf("re-verifying worker node %q after TuneD restart", targetNode.Name))
+			postRestartBannedCPUs, err := getIrqBalanceBannedCPUs(targetNode)
+			Expect(err).ToNot(HaveOccurred(), "failed to extract the banned CPUs from node %q", targetNode.Name)
+			testlog.Infof("banned CPUs on %q after the tuned restart: {%s}", targetNode.Name, postRestartBannedCPUs.String())
+
+			Expect(postRestartBannedCPUs.ToSlice()).To(Equal(postCreateBannedCPUs.ToSlice()), "banned CPUs changed post tuned restart on node %q", postRestartBannedCPUs.ToSlice(), targetNode.Name)
+		})
+	})
+})
+
+// nodes.BannedCPUs fails (!!!) if the current banned list is empty because, deep down, ExecCommandOnNode expects non-empty stdout.
+// In turn, we do this to at least have a chance to detect failed commands vs failed to execute commands (we had this issue in
+// not-so-distant past, legit command output lost somewhere in the communication). Fixing ExecCommandOnNode isn't trivial and
+// require close attention. For the time being we reimplement a form of nodes.BannedCPUs which can handle empty ban list.
+func getIrqBalanceBannedCPUs(node *corev1.Node) (cpuset.CPUSet, error) {
+	cmd := []string{"cat", "/rootfs/etc/sysconfig/irqbalance"}
+	conf, err := nodes.ExecCommandOnNode(cmd, node)
+	if err != nil {
+		return cpuset.NewCPUSet(), err
+	}
+
+	keyValue := findIrqBalanceBannedCPUsVarFromConf(conf)
+	if len(keyValue) == 0 {
+		// can happen: everything commented out (default if no tuning ever)
+		testlog.Warningf("cannot find the CPU ban list in the configuration (\n%s)\n", conf)
+		return cpuset.NewCPUSet(), nil
+	}
+
+	testlog.Infof("banned CPUs setting: %q", keyValue)
+
+	items := strings.FieldsFunc(keyValue, func(c rune) bool {
+		return c == '='
+	})
+	if len(items) != 2 {
+		return cpuset.NewCPUSet(), fmt.Errorf("malformed CPU ban list in the configuration")
+	}
+
+	bannedCPUs := unquote(strings.TrimSpace(items[1]))
+	testlog.Infof("banned CPUs: %q", bannedCPUs)
+
+	banned, err := components.CPUMaskToCPUSet(bannedCPUs)
+	if err != nil {
+		return cpuset.NewCPUSet(), fmt.Errorf("failed to parse the banned CPUs: %v", err)
+	}
+
+	return banned, nil
+}
+
+func getIrqDefaultSMPAffinity(node *corev1.Node) (string, error) {
+	cmd := []string{"cat", "/rootfs/proc/irq/default_smp_affinity"}
+	return nodes.ExecCommandOnNode(cmd, node)
+}
+
+func findIrqBalanceBannedCPUsVarFromConf(conf string) string {
+	scanner := bufio.NewScanner(strings.NewReader(conf))
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if strings.HasPrefix(line, "#") {
+			continue
+		}
+		if !strings.HasPrefix(line, "IRQBALANCE_BANNED_CPUS") {
+			continue
+		}
+		return line
+	}
+	return ""
+}
+
+func pickNodeIdx(nodes []corev1.Node) int {
+	name, ok := os.LookupEnv("E2E_PAO_TARGET_NODE")
+	if !ok {
+		return 0 // "random" default
+	}
+	for idx := range nodes {
+		if nodes[idx].Name == name {
+			testlog.Infof("node %q found among candidates, picking", name)
+			return idx
+		}
+	}
+	testlog.Infof("node %q not found among candidates, fall back to random one", name)
+	return 0 // "safe" default
+}
+
+func unquote(s string) string {
+	q := "\""
+	s = strings.TrimPrefix(s, q)
+	s = strings.TrimSuffix(s, q)
+	return s
+}
--- a/test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml
+++ b/test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml
@@ -22,8 +22,9 @@ spec:
       Different values will override the original values in parent profiles.\n\n[variables]\n#>
       isolated_cores take a list of ranges; e.g. isolated_cores=2,4-7\n\nisolated_cores=1\n\n\nnot_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}\n\n[cpu]\n#>
       latency-performance\n#> (override)\nforce_latency=cstate.id:1|3\ngovernor=performance\nenergy_perf_bias=performance\nmin_perf_pct=100\n\n\n[service]\nservice.stalld=start,enable\n\n\n[vm]\n#>
-      network-latency\ntransparent_hugepages=never\n\n\n[irqbalance]\n#> Override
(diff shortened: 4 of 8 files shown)


assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
```diff
@@ -0,0 +1,12 @@
...
+set -euo pipefail
+
+# const
+SED="/bin/sed"
+# tunable - overridable for testing purposes
+IRQBALANCE_CONF="${1:-/etc/sysconfig/irqbalance}"
+
+[ ! -f ${IRQBALANCE_CONF} ] && exit 0
+
+${SED} -i '/^IRQBALANCE_BANNED_CPUS=/d' ${IRQBALANCE_CONF} || exit 0
```
jmencak, line 11: It probably is not really needed, but wouldn't the following be a bit more robust?
`'/^\s*IRQBALANCE_BANNED_CPUS\b/d'`

assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# const
+SED="/bin/sed"
```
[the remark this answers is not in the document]
fromanirh: nothing but laziness on author side :) will fix!

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -38,8 +38,8 @@ transparent_hugepages=never
 
 {{if not .GloballyDisableIrqLoadBalancing}}
 [irqbalance]
-#> Override the value set by cpu-partitioning with an empty one
-banned_cpus=""
+#> Disable the plugin entirely. It can be racy if NTO restarts for whatever reason. We depend now on a systemd unit.
+enabled=false
```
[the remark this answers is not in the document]
fromanirh: thanks, this helps a lot!

pkg/performanceprofile/controller/performanceprofile/components/assets_scripts_test.go
```diff
@@ -0,0 +1,219 @@
...
+			data, err := os.ReadFile(fName)
+			Expect(err).ToNot(HaveOccurred())
+
+			bannedCPUs, ok := extractBannedCPUs(string(data))
+			Expect(ok).To(BeTrue())
+			Expect(bannedCPUs).To(BeEmpty())
+		})
+
+		It("should clear existing ban list", func() {
+			// the actual ban list doesn't make sense, we just need a value
```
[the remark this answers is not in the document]
fromanirh: yes indeed, I'll fix in the next upload

pkg/performanceprofile/controller/performanceprofile/components/assets_scripts_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+			_, err = cmd.Output()
+			Expect(err).ToNot(HaveOccurred())
+
+			data, err := os.ReadFile(fName)
+			Expect(err).ToNot(HaveOccurred())
+
+			bannedCPUs, ok := extractBannedCPUs(string(data))
+			Expect(ok).To(BeTrue())
+			Expect(bannedCPUs).To(Equal(bannedCPUs))
```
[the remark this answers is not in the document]
fromanirh: oops, this is nothing more than a misstep on my side. This ofc doesn't make sense. Will fix and re-upload

fromanirh:
thanks @jmencak for the early review. I want to add at least a e2e test to cover the "tuned restart while pods running" scenario and then I'll lift the WIP.

fromanirh:
content wise ready. Now testing and checking the e2e test.

fromanirh:
the PR is ready for review. I'm happy enough with the e2e test, but suggestions very welcome.

test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml
```diff
@@ -22,8 +22,9 @@ spec:
       Different values will override the original values in parent profiles.\n\n[variables]\n#>
       isolated_cores take a list of ranges; e.g. isolated_cores=2,4-7\n\nisolated_cores=1\n\n\nnot_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}\n\n[cpu]\n#>
       latency-performance\n#> (override)\nforce_latency=cstate.id:1|3\ngovernor=performance\nenergy_perf_bias=performance\nmin_perf_pct=100\n\n\n[service]\nservice.stalld=start,enable\n\n\n[vm]\n#>
-      network-latency\ntransparent_hugepages=never\n\n\n[irqbalance]\n#> Override
-      the value set by cpu-partitioning with an empty one\nbanned_cpus=\"\"\n\n\n[scheduler]\nruntime=0\ngroup.ksoftirqd=0:f:11:*:ksoftirqd.*\ngroup.rcuc=0:f:11:*:rcuc.*\nsched_min_granularity_ns=10000000\nsched_migration_cost_ns=5000000\nnuma_balancing=0\n\nsched_rt_runtime_us=-1\n\n\ndefault_irq_smp_affinity
+      network-latency\ntransparent_hugepages=never\n\n\n[irqbalance]\n#> Disable the
+      plugin entirely. It can be racy if NTO restarts for whatever reason. We depend
```
jmencak, line 26: nit: `s/if NTO restarts/if TuneD/ restarts/` should be more precise.  NTO (operator) restarting will have no effect on this.

pkg/performanceprofile/controller/performanceprofile/components/utils.go
```diff
@@ -140,7 +140,7 @@ func CPUMaskToCPUSet(cpuMask string) (cpuset.CPUSet, error) {
 		}
 		mask, err := strconv.ParseUint(chunk, 16, bitsInWord)
 		if err != nil {
-			return cpuset.NewCPUSet(), fmt.Errorf("failed to parse the CPU mask %q: %v", cpuMask, err)
+			return cpuset.NewCPUSet(), fmt.Errorf("failed to parse the CPU mask %q (chuink %q): %v", cpuMask, chunk, err)
```
jmencak, line 143: `s/chuink/chunk/`

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -38,8 +38,8 @@ transparent_hugepages=never
 
 {{if not .GloballyDisableIrqLoadBalancing}}
 [irqbalance]
-#> Override the value set by cpu-partitioning with an empty one
-banned_cpus=""
+#> Disable the plugin entirely. It can be racy if NTO restarts for whatever reason. We depend now on a systemd unit.
```
jmencak, line 41: Same as above NTO restart have nothing to do with this.  `TuneD` restarts do.

jmencak (commented):
Thank you for the PR and the changes so far, Francesco!  This looks good to me with the exceptions of some nits, but it might be useful if Yanir/Martin could have a look also from PAO side.

test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml
fromanirh: right, fixed. Thanks!

assets/performanceprofile/tuned/openshift-node-performance
fromanirh: fixed, thanks.

fromanirh:
/hold
I was made aware of https://github.com/cri-o/cri-o/blob/v1.24.1/internal/runtimehandlerhooks/high_performance_hooks.go#L360 (which I skimmed over before) - so the new unit may be unnecessary (not harmful, but redundant). I'll issue a reduced PR with the e2e test to test if this crio functionality is enough

fromanirh:
The orig_cpu_banned list is created the first time the server starts ever - likely when the node is provisioned and crio first start, thus without any tuning, and never deleted or updated.

No known services delete or update this file besides crio; in turn crio only creates it once at startup, and only if the file does not exist.

With the new PR deployed, we expect the crio code to do nothing (see https://github.com/cri-o/cri-o/blob/a3bbde8a77c323aa6a485da9a9046299155c6016/internal/runtimehandlerhooks/high_performance_hooks.go#L401 )

Without the PR deployed, crio should reset the affinity mask as the unit deployed in the PR does, but by side effect.
Crio would restore the ban list to a known value, which happen to be the empty list because of the sum of side effects.

In both cases, the behaviour of crio depends on the assumption that the orig ban list file will hold the empty cpu list.

To make the behavior explicit and predictable, a crio patch would be needed to either
Optionally disable (PAO will configure as disabled) the cpu ban list code, relying on the explicit unit to disable it
Possibly, read the restore file from custom location (controlled and managed by PAO). This will avoid the new unit payload

Alternatively, PAO can override the restore file with a known value. We will need a unit file to do so.

bartwensley:
> Hence, let's just keep with this PR, but with extra care to make sure the `/etc/sysconfig/orig_irq_banned_cpus` contains the values which will make the restore flow do nothing :)

IMO this is introducing an unnecessary dependency between NTO and cri-o. As your analysis shows, the existing cri-o code will restore the original banned CPU list whenever the server restarts, which should be sufficient. I thought that was by design when this was added to cri-o (see https://github.com/cri-o/cri-o/pull/4441).

Just my opinion, but I think there is value in simplicity and adding this extra systemd unit does add some (IMO unnecessary) complexity. One potential issue I see (maybe I'm wrong) is that this new code (clear-irqbalance-banned-cpus.sh) assumes the cpu list is at /etc/sysconfig/irqbalance - but I think the location is configurable in cri-o and according to a comment on the PR, the location is different for ubuntu (for example): https://github.com/cri-o/cri-o/pull/4441#discussion_r545234016

fromanirh:
> > Hence, let's just keep with this PR, but with extra care to make sure the `/etc/sysconfig/orig_irq_banned_cpus` contains the values which will make the restore flow do nothing :)
> 
> IMO this is introducing an unnecessary dependency between NTO and cri-o. As your analysis shows, the existing cri-o code will restore the original banned CPU list whenever the server restarts, which should be sufficient. I thought that was by design when this was added to cri-o (see [cri-o/cri-o#4441](https://github.com/cri-o/cri-o/pull/4441)).
> 
> Just my opinion, but I think there is value in simplicity and adding this extra systemd unit does add some (IMO unnecessary) complexity. One potential issue I see (maybe I'm wrong) is that this new code (clear-irqbalance-banned-cpus.sh) assumes the cpu list is at /etc/sysconfig/irqbalance - but I think the location is configurable in cri-o and according to a comment on the PR, the location is different for ubuntu (for example): [cri-o/cri-o#4441 (comment)](https://github.com/cri-o/cri-o/pull/4441#discussion_r545234016)

Without the unit, what in the system guarantees crio will reset the irqbalance config to the empty value?
I cannot prove the current flow resets the mask deliberately to nil value (no affinity) and not as combination of factors which can be broken any moment.

I'm more than happy to simplify and remove the unit, granted that we have guarantees from the system (which the unit is providing, that's the only reason to have it), and that guarantees are easily backportable.

A crio change, which I want to do anyway, is the best, but would make the backport harder.

fromanirh:
An alternative could be to overwrite `/etc/sysconfig/orig_irq_banned_cpus` when we deploy the performance profile.
Doesn't feel too nice, because crio owns that file - it's private data.

The change I'm making is indeed tampering the same private data, but only because crio unconditionally runs the restore - the unit wants to won the irqbalance config restore, so this is temporary until crio gains the option (in all the supported versions) to skip the restore stage.

The core issue here is the crio restore flow doesn't completely fit the goal here. It would be fine if we can control (or inject) the value that crio would restore on startup. *that* would be nicer, simpler, would provide guarantees and will make the unit fully redundant.
But it will require a (backportable) crio change. 

Tampering the crio private data (/etc/sysconfig/orig_irq_banned_cpus) is evil as long term solution, is less evil as mid term solution until we can disable the restore flow or control the value to be injected.

fromanirh:
crio improvements: https://github.com/cri-o/cri-o/pull/6105

bartwensley:
If I understand, the current cri-o design is to store the banned cpu mask settings when crio first starts and then restore these every time the host restarts. In what scenario doesn't that work for us?

I'd be in favor of delivering the NTO change to disable the tuned irqbalance plugin on its own. That should be safe and can easily be propped to earlier releases.

If you want to enhance the crio handling of the banned cpu mask, it seems that could be done separately and wouldn't be required in earlier releases.

fromanirh:
> If I understand, the current cri-o design is to store the banned cpu mask settings when crio first starts and then restore these every time the host restarts. In what scenario doesn't that work for us?

It seems the case, by side effect. Meaning the intent of the change and the required guarantees will be implicit.

> I'd be in favor of delivering the NTO change to disable the tuned irqbalance plugin on its own. That should be safe and can easily be propped to earlier releases.

What makes unsafe to backport the extra systemd unit?

fromanirh:
> Tampering the crio private data (`/etc/sysconfig/orig_irq_banned_cpus`) is evil as long term solution, is less evil as mid term solution until we can disable the restore flow or control the value to be injected.

But still, this is a evil. Relaying on crio picking the right values to restore is fragile. It's _likely_ to be right most of time, but no guarantees. Simply put, no piece of the stack even tries to provide them. Doing it explicitely requires tampering the cri-o private data. This is actually a good argument to avoid the systemd unit. Need to re-evaluate.

assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+set -euo pipefail
+set -x
+
+# const
+SED="/usr/bin/sed"
+# tunable - overridable for testing purposes
+IRQBALANCE_CONF="${1:-/etc/sysconfig/irqbalance}"
```
bartwensley, line 8: This assumes the cpu list is at /etc/sysconfig/irqbalance - but I think the location is configurable in cri-o and according to a comment on a PR, the location is different for ubuntu (for example): https://github.com/cri-o/cri-o/pull/4441#discussion_r545234016

bartwensley:
> > If I understand, the current cri-o design is to store the banned cpu mask settings when crio first starts and then restore these every time the host restarts. In what scenario doesn't that work for us?
> 
> It seems the case, by side effect. Meaning the intent of the change and the required guarantees will be implicit.

You still haven't given an example where the existing cri-o code does the wrong thing. IMO you are adding complexity to solve a problem that doesn't exist.

> > I'd be in favor of delivering the NTO change to disable the tuned irqbalance plugin on its own. That should be safe and can easily be propped to earlier releases.
> 
> What makes unsafe to backport the extra systemd unit?

I'm saying that disabling the irqbalance plugin (on its own) is a much smaller and safer change than introducing an additional systemd unit.

fromanirh:
> > > If I understand, the current cri-o design is to store the banned cpu mask settings when crio first starts and then restore these every time the host restarts. In what scenario doesn't that work for us?
> > 
> > 
> > It seems the case, by side effect. Meaning the intent of the change and the required guarantees will be implicit.
> 
> You still haven't given an example where the existing cri-o code does the wrong thing.

Because the existing crio code won't do the right provided the cpu mask is saved correctly, which we don't control. It is probably saved correctly, but we can't enforce (we can only check ex-poste) nor we can declare the desired state.

> IMO you are adding complexity to solve a problem that doesn't exist.

The current solution is indeed sub-optimal. This is not under discussion, I can't argue. The optimal solution would be either depend on crio, or disable the crio code. We can't do either at this point in time. The pure crio approach is _likely_ to be correct, but no guarantees. It depends on the content the cpu mask snapshot, which we don't control.

> > What makes unsafe to backport the extra systemd unit?
> 
> I'm saying that disabling the irqbalance plugin (on its own) is a much smaller and safer change than introducing an additional systemd unit.

There's a tradeoff here for sure, and the unit approach is not risk free. Depending on a snapshot we don't predict or control of the system state doesn't seem ideal either.

assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
fromanirh: true, but on RHCOS/RHEL this is where the config file is placed. I'm not aware we support ubuntu as host system.

fromanirh:
the blocking thing here is having e2e test coverage. I'm working to have them.

I'd love to be able to depend on crio alone, but I don't think is the best approach. I stated my reservations, and they all stands to date. It's all about 1. guarantees (or lack thereof) and 2. declared state (vs assumptions, albeit likely. But "likely" is "best effort"). 

But I'm willing to split the unit addition in its own PR. I think we will quickly agree about having the `irqbalance` change + e2e tests (bar discussions over them). The commits are already arranged to trivially postpone the unit change.

fromanirh:
cri-o improvements: https://github.com/cri-o/cri-o/pull/6105

fromanirh:
unit moved in #413

fromanirh:
```
WARN[2022-08-03T12:31:25Z] Failed to get build src.                      error=Get "[https://172.30.0.1:443/apis/build.openshift.io/v1/namespaces/ci-op-215vp9th/builds/src](https://172.30.0.1/apis/build.openshift.io/v1/namespaces/ci-op-215vp9th/builds/src)": http2: client connection lost
WARN[2022-08-03T12:32:14Z] Failed to get build src.                      error=the server is currently unable to handle the request (get builds.build.openshift.io src)
WARN[2022-08-03T12:33:30Z] Failed to get build src.                      error=the server was unable to return a response in the time allotted, but may still be processing the request (get builds.build.openshift.io src)
WARN[2022-08-03T12:34:10Z] Failed to get build src.                      error=Get "[https://172.30.0.1:443/apis/build.openshift.io/v1/namespaces/ci-op-215vp9th/builds/src](https://172.30.0.1/apis/build.openshift.io/v1/namespaces/ci-op-215vp9th/builds/src)": dial tcp 172.30.0.1:443: i/o timeout
WARN[2022-08-03T12:34:15Z] Could not start a watch on our test namespace.  error=Get "[https://172.30.0.1:443/api/v1/namespaces?fieldSelector=metadata.name%3Dci-op-215vp9th&watch=true](https://172.30.0.1/api/v1/namespaces?fieldSelector=metadata.name%3Dci-op-215vp9th&watch=true)": dial tcp 172.30.0.1:443: connect: connection refused - error from a previous attempt: dial tcp 172.30.0.1:443: i/o timeout
WARN[2022-08-03T12:34:15Z] Failed to get build src.                      error=Get "[https://172.30.0.1:443/apis/build.openshift.io/v1/namespaces/ci-op-215vp9th/builds/src](https://172.30.0.1/apis/build.openshift.io/v1/namespaces/ci-op-215vp9th/builds/src)": context canceled
```
infra issue :\

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -0,0 +1,315 @@
...
+			Expect(err).ToNot(HaveOccurred(), "failed to get Online CPUs list")
+
+			// expect no irqbalance run in the system already, AKA start from pristine conditions.
+			// This is not an hard requirement, just the easier state to manage and check
+			Expect(smpAffinitySet.Equals(onlineCPUsSet)).To(BeTrue(), "found default_smp_affinity %v, expected %v - IRQBalance already run?", smpAffinitySet, onlineCPUsSet)
+
+			origBannedCPUsFile := "/etc/sysconfig/orig_irq_banned_cpus"
+			By(fmt.Sprintf("Checking content of %q on node %q", origBannedCPUsFile, node.Name))
+			expectFileEmpty(node, origBannedCPUsFile)
+		})
```
bartwensley, line 171: The orig_irq_banned_cpus is private crio data and shouldn't be referenced from an NTO test case. This kind of testing should be done from the crio repo.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -0,0 +1,315 @@
...
+			_, err = nodes.ExecCommandOnNode([]string{"echo", "IRQBALANCE_BANNED_CPUS=2,3", ">", "/rootfs/" + irqBalanceConfFile}, node)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Preparing fake data for the irqbalance cpu ban list file")
+			origBannedCPUsFile := "/etc/sysconfig/orig_irq_banned_cpus"
+			restoreBanned := makeBackupForFile(node, origBannedCPUsFile)
+			defer restoreBanned()
+
+			_, err = nodes.ExecCommandOnNode([]string{"/usr/bin/truncate", "-s", "0", "/rootfs/" + origBannedCPUsFile}, node)
+			Expect(err).ToNot(HaveOccurred())
```
bartwensley, line 209: This should be tested without touching private crio files. Why not follow the same procedure used in the "Verify irqbalance configuration handling" and create a pod that updates the banned cpus and then restart crio? Or add this kind of testcase to crio itself since this is really testing crio behaviour.

bartwensley (changes requested):

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
fromanirh: Because we don't have any control over the content of the `orig_irq_banned_cpus`. So the closest we can do is to create a controlled environment.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
fromanirh: In the context of this fix, we need to have some component resetting the irqbalance ban mask to zero. This is what this change is about and this is what functests want to cover. We don't have APIs or any way to check this besides checking the private data of crio, until the crio improvements land.

fromanirh:
in the last upload, I split the controversial tests in their own commit. If the outcome of the review is that we prefer to have less test coverage and to blindly trust the content of the `orig_irq_banned_cpus`, we will be able to trivially remove the last commit and move forward.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
bartwensley: I guess we can agree to disagree - I'm not an owner so you don't need my approval.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
jmencak: > I'm not an owner so you don't need my approval.

This can easily change.  `OWNERS` file here is only really used for the bot to pick reviewers.  I believe anyone in the OpenShift org can LGTM.

fromanirh:
need to fix the (legit) test failure

fromanirh:
> I'm marking this as approved because I don't see any other way to remove the "request changes" flag and I don't want to hold up this PR. I still don't agree that the systemd unit is required and would prefer to just see the irqbalance plugin disabled - especially if this is going to be backported to earlier releases. But I do not want to hold this up since I seem to have the minority opinion here.

Thanks. But I want to point out that so far it's only the two of us which we happen to disagree: no other maintainer/reviewer commented on this specific matter, so rather than being the minority opinion, both opinions have equal weight at this point.

In turn, I don't think that it's wise to go without the systemd unit (or any other way to declare and possibly enforce the intended value). I explained on record my point and the rationale, perhaps even too verbosely.
If however, the consensus is that we don't want this extra and we want to go on with only the profile fix, so be it.

The unit was split out for this reasons, and the e2e tests verifying that the implicit intent is honored are put in their own commit to make this as easy as possible.
Once CI passes I'm going to actually move the extra e2e tests in the same PR adding the unit for this very reason.

fromanirh:
> > I'm marking this as approved because I don't see any other way to remove the "request changes" flag and I don't want to hold up this PR. I still don't agree that the systemd unit is required and would prefer to just see the irqbalance plugin disabled - especially if this is going to be backported to earlier releases. But I do not want to hold this up since I seem to have the minority opinion here.
> 
> Thanks. But I want to point out that so far it's only the two of us which we happen to disagree: no other maintainer/reviewer commented on this specific matter, so rather than being the minority opinion, both opinions have equal weight at this point.
> 
> In turn, I don't think that it's wise to go without the systemd unit (or any other way to declare and possibly enforce the intended value). I explained on record my point and the rationale, perhaps even too verbosely. If however, the consensus is that we don't want this extra and we want to go on with only the profile fix, so be it.
> 
> The unit was split out for this reasons, and the e2e tests verifying that the implicit intent is honored are put in their own commit to make this as easy as possible. Once CI passes I'm going to actually move the extra e2e tests in the same PR adding the unit for this very reason.

done. Content moved on https://github.com/openshift/cluster-node-tuning-operator/pull/413

bartwensley (approved):
Thanks Francesco - looks good.

fromanirh:
/hold
to let other reviewers chime in.

fromanirh:
e2e tests are unstable: it seems the new test pollutes the cluster state. Still chasing a proper reproducer to narrow down the root cause. PR is reviewable and almost final (only fixes to the e2e test are pending)

fromanirh:
after quite a lot of investigation, it seems the e2e test is exposing a real bug, likely in the crio code, which
triggers at least on CI:
```
I0811 14:09:23.293538   42525 request.go:601] Waited for 1.049685011s due to client-side throttling, not priority and fairness, request: GET:$MISSING
Running Suite: Performance Addon Operator e2e tests
===================================================
Random Seed: 1660219760
Will run 1 of 55 specs

SSSSS
------------------------------
[performance] Checking IRQBalance settings Verify irqbalance configuration handling 
  Should not overwrite the banned CPU set on tuned restart
  /home/fromani/go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/irqbalance.go:70
Aug 11 14:09:26.021: [INFO]: MCP "worker" is targeting 3 node(s)
Aug 11 14:09:26.590: [INFO]: MCP "worker" is targeting 3 node(s)
Aug 11 14:09:26.725: [INFO]: node "ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw" found among candidates, picking
STEP: verifying worker node "ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw"
Aug 11 14:09:26.862: [INFO]: found mcd machine-config-daemon-k7g2s for node ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw
Aug 11 14:09:27.849: [INFO]: IRQ Default affinity on "ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw" when test begins: {f}
Aug 11 14:09:27.986: [INFO]: found mcd machine-config-daemon-k7g2s for node ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw
Aug 11 14:09:28.962: [WARNING]: cannot find the CPU ban list in the configuration (
# irqbalance is a daemon process that distributes interrupts across
# CPUS on SMP systems. The default is to rebalance once every 10
# seconds. This is the environment file that is specified to systemd via the
# EnvironmentFile key in the service unit file (or via whatever method the init
# system you're using has.
#
# ONESHOT=yes
# after starting, wait for a minute, then look at the interrupt
# load and balance it once; after balancing exit and do not change
# it again.
#IRQBALANCE_ONESHOT=

#
# IRQBALANCE_BANNED_CPUS
# 64 bit bitmask which allows you to indicate which cpu's should
# be skipped when reblancing irqs. Cpu numbers which have their
# corresponding bits set to one in this mask will not have any
# irq's assigned to them on rebalance
#
#IRQBALANCE_BANNED_CPUS=

#
# IRQBALANCE_ARGS
# append any args here to the irqbalance daemon as documented in the man page
#
#IRQBALANCE_ARGS=

)

Aug 11 14:09:28.962: [INFO]: banned CPUs on "ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw" when test begins: {}
Aug 11 14:09:29.098: [INFO]: found mcd machine-config-daemon-k7g2s for node ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw
Aug 11 14:09:30.215: [INFO]: found mcd machine-config-daemon-k7g2s for node ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw
Aug 11 14:09:31.250: [INFO]: using testpod:
{"metadata":{"generateName":"test-","namespace":"performance-addon-operators-testing","creationTimestamp":null,"labels":{"test":""},"annotations":{"cpu-quota.crio.io":"disable","irq-load-balancing.crio.io":"disable"}},"spec":{"containers":[{"name":"test","image":"quay.io/openshift-kni/cnf-tests:4.9","command":["sleep","10h"],"resources":{"limits":{"cpu":"2","memory":"256Mi"}}}],"nodeName":"ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw","runtimeClassName":"performance-perfprof"},"status":{}}
Aug 11 14:09:48.812: [WARNING]: ->  AddedInterface Add eth0 [10.131.0.21/23] from ovn-kubernetes
Aug 11 14:09:48.812: [WARNING]: ->  Pulling Pulling image "quay.io/openshift-kni/cnf-tests:4.9"
Aug 11 14:09:48.812: [WARNING]: ->  Pulled Successfully pulled image "quay.io/openshift-kni/cnf-tests:4.9" in 13.818036237s
Aug 11 14:09:48.812: [WARNING]: ->  Created Created container test
Aug 11 14:09:48.812: [WARNING]: ->  Started Started container test
Aug 11 14:09:48.948: [INFO]: found mcd machine-config-daemon-k7g2s for node ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw
Aug 11 14:09:50.048: [WARNING]: cannot find the CPU ban list in the configuration (
# irqbalance is a daemon process that distributes interrupts across
# CPUS on SMP systems. The default is to rebalance once every 10
# seconds. This is the environment file that is specified to systemd via the
# EnvironmentFile key in the service unit file (or via whatever method the init
# system you're using has.)

Aug 11 14:09:50.048: [INFO]: banned CPUs on "ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw" just before the tuned restart: {}
Aug 11 14:09:50.185: [INFO]: found mcd machine-config-daemon-k7g2s for node ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw
Aug 11 14:09:51.215: [INFO]: IRQ Default affinity on "ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw" when test ends: {5}
Aug 11 14:09:51.215: [INFO]: deleting pod "test-x6td8"
Aug 11 14:10:22.894: [INFO]: found mcd machine-config-daemon-k7g2s for node ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw
Aug 11 14:10:24.004: [INFO]: banned CPUs setting: "IRQBALANCE_BANNED_CPUS=\"00000000,000000f0\""
Aug 11 14:10:24.005: [INFO]: banned CPUs: "00000000,000000f0"
Aug 11 14:10:24.005: [INFO]: banned CPUs on "ci-ln-2h3mjpb-72292-2wnt7-worker-c-zfmvw" when test ends: {4-7}
```

I'll investigate the crio code to narrow down a repro (and file a crio bug etc. etc.)

fromanirh:
proposed crio fix: https://github.com/cri-o/cri-o/pull/6146

since this change needs to be backportable, I'll work on a workaround on the e2e tests.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -0,0 +1,247 @@
...
+		if !strings.HasPrefix(line, "IRQBALANCE_BANNED_CPUS") {
+			continue
+		}
+		return line
+	}
+	return ""
+}
+
+func pickNodeIdx(nodes []corev1.Node) int {
+	name, ok := os.LookupEnv("E2E_PAO_TARGET_NODE")
```
yanirq, line 228: is E2E_PAO_TARGET_NODE a new Env var ? or are we using it in cnf-tests ? (documented?)

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		if !strings.HasPrefix(line, "IRQBALANCE_BANNED_CPUS") {
+			continue
+		}
+		return line
+	}
+	return ""
+}
+
+func pickNodeIdx(nodes []corev1.Node) int {
```
yanirq, line 227: consider moving to utils , not mandatory.

yanirq:
Added some small nit questions but given the discussion above and the crio bug discovered this LTGM.
Feel free to remove hold if it seems fit.
/lgtm

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
fromanirh: is a new env var. What's the best place to document the env vars?

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
fromanirh: Intentionally avoided because https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common
and I don't have enough material yet to start a new package

MarSik (approved):

fromanirh:
/hold cancel
agreed offline to move forward

PR #400, opened 2022-07-22
WIP: tuned: disable irqbalance, handle ban cpus earlier - using crio

fromanirh:
counterpart of https://github.com/openshift/cluster-node-tuning-operator/pull/396 leveraging existing crio functionality

The tuned irqbalance plugin clears the irqbalance banned CPUs list
when tuned starts. The list is then managed dynamically by the runtime
handlers.

On node restart, the tuned pod can be started AFTER the workload pods
(kubelet nor kubernetes offers ordering guarantees when recovering the
node state); clearing the banned CPUs list while pods are running and
compromising the IRQ isolation guarantees. Same holds true if the
NTO pod restarts for whatever reason.

The fix is twofold:
1. disable the irqbalance plugin, so tuned doesn't clear the irqbalance
   banned list anymore.
2. explicitly and predictably set the banned CPU list early in the boot
   leveraging crio functionality.

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -38,8 +38,10 @@ transparent_hugepages=never
 
 {{if not .GloballyDisableIrqLoadBalancing}}
 [irqbalance]
-#> Override the value set by cpu-partitioning with an empty one
-banned_cpus=""
+# Disable the plugin entirely, which was enabled by the parent profile `cpu-partitioning`.
+# It can be racy if TuneD restarts for whatever reason.
+#> cpu-partitioning
+enabled=false
 {{end}}
 
 [scheduler]
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -1,0 +1,139 @@
+package __performance
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/mcps"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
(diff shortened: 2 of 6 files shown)


fromanirh:
WIP because we need to crosscheck if the existing crio functionality actually covers the requirement as it seems

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -38,8 +38,8 @@ transparent_hugepages=never
 
 {{if not .GloballyDisableIrqLoadBalancing}}
 [irqbalance]
-#> Override the value set by cpu-partitioning with an empty one
-banned_cpus=""
+#> Disable the plugin entirely. It can be racy if TuneD restarts for whatever reason. We depend now on a systemd unit.
```
yanirq, line 41: We can add a comment for future devs about the fact that this plugin is enabled in the parent profile (cpu-partitioning) thus needing to disable it in the child

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -0,0 +1,137 @@
...
+			postCreateBannedCPUs, err := nodes.BannedCPUs(node)
+			Expect(err).ToNot(HaveOccurred(), "failed to extract the banned CPUs from node %q", node.Name)
+			Expect(postCreateBannedCPUs.IsEmpty()).To(BeFalse(), "banned CPUs %v not empty on node %q", postCreateBannedCPUs, node.Name)
+
+			By(fmt.Sprintf("getting a TuneD Pod running on node %s", node.Name))
+			pod, err := util.GetTunedForNode(cs, &node)
+			Expect(err).NotTo(HaveOccurred())
+
+			By(fmt.Sprintf("causing a restart of the tuned pod (deleting the pod) on %s", node.Name))
+			_, _, err = util.ExecAndLogCommand("oc", "delete", "pod", "-n", pod.Namespace, pod.Name)
```
jmencak, line 119: So, these days the `oc/kubectl delete` commands block until a resource is deleted unless one says `--wait=false`.  However, this still makes me a bit uneasy because it didn't use to be that way in the past, in which case we'd have a race here.  Do we want/need to say `--wait=true` to make it more obvious/explicit there is no race here?

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -0,0 +1,137 @@
...
+			Expect(err).ToNot(HaveOccurred(), "failed to extract the banned CPUs from node %q", node.Name)
+			Expect(bannedCPUs.IsEmpty()).To(BeTrue(), "banned CPUs %v not empty on node %q", bannedCPUs, node.Name)
+
+			smpAffinitySet, err := nodes.GetDefaultSmpAffinitySet(&node)
+			Expect(err).ToNot(HaveOccurred(), "failed to get default smp affinity")
+
+			onlineCPUsSet, err := nodes.GetOnlineCPUsSet(&node)
+			Expect(err).ToNot(HaveOccurred(), "failed to get Online CPUs list")
+
+			Expect(smpAffinitySet.Equals(onlineCPUsSet)).To(BeTrue(), "found default_smp_affinity %v, expected %v", smpAffinitySet, onlineCPUsSet)
```
jmencak, line 89: I'm not sure I understand the test here.  Perhaps a comment would be helpful?

assets/performanceprofile/tuned/openshift-node-performance
fromanirh: Sure, I will update both PRs.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
fromanirh: sure thing (will update both PRs)

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
fromanirh: right, added (and clarified a bit the error message)

bartwensley:
Should probably remove (or reword) this part of the commit message:
> 2. explicitly and predictably set the banned CPU list early in the boot
> leveraging crio functionality.

Since that functionality is already there in cri-o, it isn't really part of the fix. Maybe indicate that this is existing functionality?

PR #403, opened 2022-07-26
Add function to node utils to fetch Core Siblings

mrniranjan:
GetCoreSiblings function fetches core siblings per
numa Node

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/nodes/nodes.go
+++ b/test/e2e/performanceprofile/functests/utils/nodes/nodes.go
@@ -46,8 +46,9 @@ type NumaNodes struct {
 
 // NodeCPU Structure
 type NodeCPU struct {
-	CPU  string `json:"cpu"`
 	Node string `json:"node"`
+	Core string `json:"core"`
+	CPU  string `json:"cpu"`
 }
 
 // GetByRole returns all nodes with the specified role
@@ -275,7 +276,7 @@ func GetSMTLevel(cpuID int, node *corev1.Node) int {
 
 // GetNumaNodes returns the number of numa nodes and the associated cpus as list on the node
 func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
-	lscpuCmd := []string{"lscpu", "-e=cpu,node", "-J"}
+	lscpuCmd := []string{"lscpu", "-e=node,core,cpu", "-J"}
 	cmdout, err := ExecCommandOnNode(lscpuCmd, node)
 	var numaNode, cpu int
 	if err != nil {
@@ -297,6 +298,29 @@ func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
 		numaCpus[numaNode] = append(numaCpus[numaNode], cpu)
 	}
 	return numaCpus, err
+}
+
+// GetCoreSiblings returns the siblings of core per numa node
+func GetCoreSiblings(node *corev1.Node) (map[int]map[int][]int, error) {
+	lscpuCmd := []string{"lscpu", "-e=node,core,cpu", "-J"}
+	out, err := ExecCommandOnNode(lscpuCmd, node)
+	var result NumaNodes
+	var numaNode, core, cpu int
+	coreSiblings := make(map[int]map[int][]int)
+	err = json.Unmarshal([]byte(out), &result)
+	if err != nil {
+		return nil, err
+	}
+	for _, value := range result.Cpus {
+		if numaNode, err = strconv.Atoi(value.Node); err != nil {
+			break
+		}
+		if coreSiblings[numaNode] == nil {
+			coreSiblings[numaNode] = make(map[int][]int)
+		}
+		coreSiblings[numaNode][core] = append(coreSiblings[numaNode][core], cpu)
+	}
+	return coreSiblings, err
 }
 
 //TunedForNode find tuned pod for appropriate node


test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -46,8 +46,9 @@ type NumaNodes struct {
 
 // NodeCPU Structure
 type NodeCPU struct {
-	CPU  string `json:"cpu"`
 	Node string `json:"node"`
+	Core string `json:"core"`
+	CPU  string `json:"cpu"`
```
fromanirh, line 51: for the future: can these be `int`s?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -275,7 +276,7 @@ func GetSMTLevel(cpuID int, node *corev1.Node) int {
 
 // GetNumaNodes returns the number of numa nodes and the associated cpus as list on the node
 func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
-	lscpuCmd := []string{"lscpu", "-e=cpu,node", "-J"}
+	lscpuCmd := []string{"lscpu", "-e=node,core,cpu", "-J"}
```
fromanirh, line 279: we don't seem to need this new data in this function, so maybe omit this change?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -299,6 +300,37 @@ func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
...
+		}
+		if cpu, err = strconv.Atoi(value.CPU); err != nil {
+			break
+		}
+		if coreSibilings[numaNode] == nil {
+			coreSibilings[numaNode] = make(map[int][]int)
+			coreSibilings[numaNode][core] = append(coreSibilings[numaNode][core], cpu)
+		} else {
+			coreSibilings[numaNode][core] = append(coreSibilings[numaNode][core], cpu)
+		}
```
fromanirh, line 329: maybe:
```
if coreSibilings[numaNode] == nil {
    coreSibilings[numaNode] = make(map[int][]int)
}
coreSibilings[numaNode][core] = append(coreSibilings[numaNode][core], cpu)
```

ffromani (commented):
looks good, I'll have another check later

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
marioferh, line 279: not needed.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -299,6 +300,37 @@ func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
...
```
marioferh, line 329: `		if coreSibilings[numaNode] == nil {
			coreSibilings[numaNode] = make(map[int][]int)
		}
		coreSibilings[numaNode][core] = append(coreSibilings[numaNode][core], cpu)`

marioferh (commented):
its possible to add some test?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -299,6 +300,37 @@ func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
 	return numaCpus, err
 }
 
+// GetCoreSiblings returns the siblings of core per numa node
+func GetCoreSiblings(node *corev1.Node) (map[int]map[int][]int, error) {
+	lscpuCmd := []string{"lscpu", "-e=node,core,cpu", "-J"}
+	out, err := ExecCommandOnNode(lscpuCmd, node)
+	var result NumaNodes
+	var numaNode, core, cpu int
+	coreSibilings := make(map[int]map[int][]int)
```
marioferh, line 309: typo: siblings

marioferh, line 309: maybe simpler something that this:
`// GetCoreSiblings returns the siblings of core per numa node
func GetCoreSiblings(node *corev1.Node) ([][]int, error) {
	lscpuCmd := []string{"lscpu", "-e=node,core,cpu", "-J"}
	out, err := ExecCommandOnNode(lscpuCmd, node)
	if err != nil {
		return nil, err
	}
	var result NumaNodes
	coreSiblings := make([][]int, len(result.Cpus)/2)
	err = json.Unmarshal([]byte(out), &result)
	for _, value := range result.Cpus {
		var core, cpu int
		if core, err = strconv.Atoi(value.Core); err != nil {
			break
		}
		if cpu, err = strconv.Atoi(value.CPU); err != nil {
			break
		}
		if core != cpu {
			coreSiblings = append(coreSiblings, []int{core, cpu})
		}
	}
	return coreSiblings, err
}`

not tested, but the idea its not use that complicated map

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: We need it because, we want lscpu to give us the extended output in the order asked and i  use this order using the -J(json) and then dump(unmarshal) it in the structures fields.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: It can't be int's because the lscpu -e=node,core,cpu -J returns json output in strings 

example:

`
{
   "cpus": [
      {"node": "0", "core": "0", "cpu": "0"},
      {"node": "1", "core": "1", "cpu": "1"},
      {"node": "0", "core": "2", "cpu": "2"},
      {"node": "1", "core": "3", "cpu": "3"},
      {"node": "0", "core": "4", "cpu": "4"},
      {"node": "1", "core": "5", "cpu": "5"},
      {"node": "0", "core": "6", "cpu": "6"},
      {"node": "1", "core": "7", "cpu": "7"},
      {"node": "0", "core": "8", "cpu": "8"},

`

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: we do need i think because we need per numa node per core siblings which lscpu -e=node,core,cpu . In the GetNumaNodes we don't use the core fields but since GetnNumaNodes and GetCoreSiblings share the same Structure we are using the same command.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: The full output of lscp -e=node,core,cpu can be found here: https://paste.centos.org/view/83ec74d8

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: Yeah the map is complicated because we need Per numa node, per core siblings . 
So we get output like this:

```
map[0:map[0:[0 40] 10:[10 50] 12:[12 52] 14:[14 54] 16:[16 56]....
map[1:[1 41] 11:[11 51] 13:[13 53] 15:[15 55] 17:[17 57] 19:[19 59] ...
```

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: Updated in the latest commit

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: Fixed in the latest commit

marioferh: Ok, I don't like so much the complexity but if it has to be like that...

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
marioferh: understood.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -299,6 +300,37 @@ func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
 	return numaCpus, err
 }
 
+// GetCoreSiblings returns the siblings of core per numa node
+func GetCoreSiblings(node *corev1.Node) (map[int]map[int][]int, error) {
+	lscpuCmd := []string{"lscpu", "-e=node,core,cpu", "-J"}
+	out, err := ExecCommandOnNode(lscpuCmd, node)
+	var result NumaNodes
+	var numaNode, core, cpu int
+	coreSiblings := make(map[int]map[int][]int)
```
marioferh, line 309: Still thinking that we can simplify the map at least with types:
`type siblings []int
type cpuSiblings map[int]siblings
type nodeCpuSiblings map[int]cpuSiblings
`
`coreSiblings := make(nodeCpuSiblings)`
something like that

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -299,6 +300,37 @@ func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
...
+	err = json.Unmarshal([]byte(out), &result)
+	if err != nil {
+		return nil, err
+	}
+	for _, value := range result.Cpus {
+		if numaNode, err = strconv.Atoi(value.Node); err != nil {
+			break
+		}
+		if core, err = strconv.Atoi(value.Core); err != nil {
```
marioferh, line 318: this is redundant right? 
I cannot think in an scenario where returns an error, only if cpu is offline numanode would be -

mrniranjan: I have addressed this in the latest patch

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -299,6 +300,37 @@ func GetNumaNodes(node *corev1.Node) (map[int][]int, error) {
...
+			break
+		}
+		if cpu, err = strconv.Atoi(value.CPU); err != nil {
+			break
+		}
+		if coreSiblings[numaNode] == nil {
```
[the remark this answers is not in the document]
mrniranjan: Addressed this in the latest patch

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: I tried this but this looks like creating map in one steps we are doing it in multiple steps. Also one problem i hit using the above approach is, if i do like above, i don't know how to return coreSiblings because coresiblings is of type nodeCpuSiblings which is local to that function and is not a predeclared type like maps

marioferh (commented):
/lgtm

PR #408, opened 2022-07-31
Bug 2102450: set rcutree.kthread_prio to ksoftirqd prio

vitus133:
When a performance profile is applied, the priority of the
ksoftirqd and rcuc realtime attributes are set to to FIFO:11.
The rcutree.kthread_prio must be set accordingly to avoid
RCU stall situations with rt kernel

Signed-off-by: Vitaly Grinberg <vgrinber@redhat.com>

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -138,7 +138,7 @@ cmdline_isolation=+isolcpus=managed_irq,${isolated_cores}
 {{end}}
 
 {{if .RealTime}}
-cmdline_realtime=+nohz_full=${isolated_cores} tsc=nowatchdog nosoftlockup nmi_watchdog=0 mce=off skew_tick=1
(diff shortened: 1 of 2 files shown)


yanirq:
I think we are hitting a test error in the operator lane due to https://github.com/redhat-performance/tuned/pull/423 
we should align the test ([reboots][machine_config_labels] Node Tuning Operator machine config labels) and re-asses the values we set for the performance profile.
(cc @MarSik @jmencak )

jmencak:
> I think we are hitting a test error in the operator lane due to [redhat-performance/tuned#423](https://github.com/redhat-performance/tuned/pull/423) we should align the test ([reboots][machine_config_labels] Node Tuning Operator machine config labels) and re-asses the values we set for the performance profile. (cc @MarSik @jmencak )

Thank you Yanir.  Fix in: https://github.com/openshift/cluster-node-tuning-operator/pull/410

PR #409, opened 2022-08-01
Fix CPU's overlapping error output

marioferh:
## Description:

Output when forbidden CPU's overlapping is not formatted and it shows the object value

## Type of change

- [x] Change error output

## Testing steps
Launch PerformanceProfile as this one:
spec:
  cpu:
    isolated: "0-2"
    reserved: "0-2"

apply -f *.yaml
```
The PerformanceProfile "performance" is invalid: spec.cpu: Invalid value: v2.CPU{Reserved:(*v2.CPUSet)(0xc001017d30), Isolated:(*
v2.CPUSet)(0xc001017d20), BalanceIsolated:(*bool)(nil), Offlined:(*v2.CPUSet)(nil)}: **reserved and isolated cpus overlap**: [0 1 2] 
```

Also the code shows error as invalid, and it should be forbidden.

Fixed output:

```
The PerformanceProfile "performance" is invalid: spec.cpu: Forbidden: reserved and isolated cpus overlap: [0 1 2]
```

Also fixed for "reserved and offlined" and "isolated and offlined".

### DIFF
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
@@ -147,16 +147,16 @@ func (r *PerformanceProfile) validateCPUs() field.ErrorList {
 				}
 
 				if overlap := components.Intersect(cpuLists.GetIsolated(), cpuLists.GetReserved()); len(overlap) != 0 {
-					allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu"), r.Spec.CPU, fmt.Sprintf("reserved and isolated cpus overlap: %v", overlap)))
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.cpu"), fmt.Sprintf("reserved and isolated cpus overlap: %v", overlap)))
 				}
 			}
 
 			if r.Spec.CPU.Offlined != nil {
 				if overlap := components.Intersect(cpuLists.GetReserved(), cpuLists.GetOfflined()); len(overlap) != 0 {
-					allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu"), r.Spec.CPU, fmt.Sprintf("reserved and offlined cpus overlap: %v", overlap)))
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.cpu"), fmt.Sprintf("reserved and offlined cpus overlap: %v", overlap)))
 				}
 				if overlap := components.Intersect(cpuLists.GetIsolated(), cpuLists.GetOfflined()); len(overlap) != 0 {
-					allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu"), r.Spec.CPU, fmt.Sprintf("isolated and offlined cpus overlap: %v", overlap)))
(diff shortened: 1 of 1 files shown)


shajmakh:
@marioferh thanks for the fix! can you add in the PR description the CPU spec that was configured in the performance profile?
something like:
cpu:
    isolated:  ..
    reserved: ..
    offlined: ..

marioferh:
> @marioferh thanks for the fix! can you add in the PR description the CPU spec that was configured in the performance profile? something like: cpu: isolated: .. reserved: .. offlined: ..

done

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -147,16 +147,16 @@ func (r *PerformanceProfile) validateCPUs() field.ErrorList {
 				}
 
 				if overlap := components.Intersect(cpuLists.GetIsolated(), cpuLists.GetReserved()); len(overlap) != 0 {
-					allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu"), r.Spec.CPU, fmt.Sprintf("reserved and isolated cpus overlap: %v", overlap)))
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.cpu"), fmt.Sprintf("reserved and isolated cpus overlap: %v", overlap)))
```
fromanirh, line 150: what if we call the `String()` methods explicitely? (adding them if needed, or string helpers)
```
allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.cpu"), r.Spec.CPU.String(), fmt.Sprintf("reserved and isolated cpus overlap: %v", overlap)))
```

marioferh: You are talking about cpuset of kubelet. This is spec.cpu of API and has no String method. In fact `type CPUSet string`

It could be valid like this:
`allErrs = append(allErrs, field.Invalid(field.NewPath("spec.cpu"), string(*r.Spec.CPU.Reserved), fmt.Sprintf("reserved and isolated cpus overlap: %v", overlap)))`

But in any case, invalid would be the intersection, not the Reserved cpu, for that I've changed it to Forbidden.

fromanirh: I'm also wondering if the field should actually be forbidden (why?) rather than invalid. No strong feelings so far.

fromanirh: I don't think it's _necessary_. I do however think that keeping the context here will help and improve the UX, locally for this issue and in general.
Finding a good string representation for the CPU object can be nontrivial, so we can do in a later PR.

marioferh: Then if you agree we can move forward with this one and open other one to format the output of all invalid errors of this API
Also another thing because I've changed to forbidden its you don't have to add the arg "r.Spec.CPU"

fromanirh: sounds good.

PR #410, opened 2022-08-02
e2e: check for tsc=reliable instead of tsc=nowatchdog

jmencak:
Recent TuneD realtime profile is using `tsc=reliable`
instead of `tsc=nowatchdog`.  Adjust
`[reboots][machine_config_labels]` to reflect this.  See:
https://github.com/redhat-performance/tuned/pull/423

Signed-off-by: Jiri Mencak <jmencak@users.noreply.github.com>


yanirq:
do we still to maintain and align the assets ?  e.g. https://github.com/openshift/cluster-node-tuning-operator/blob/2b8674b972a169e737d3bb10e61f294d7933ae74/assets/tuned/daemon/profiles/realtime/tuned.conf#L55

PR #415, opened 2022-08-03
Implement new workloadHint perPodPowerManagement

marioferh:
## Description

Add new workloadHint PerPodPowerManagement

## Motivation and context
PerPodPowerManagement workloadHint allows power management to be configured on a per pod (v.s. per node) basis. Using this workloadHint will configure the node to run with c-states and p-states enabled, instead of being disabled (which is the default).

WorkloadHints HighPowerConsumption and PerPodPowerManagement cannot be enabled at the same time.

## Type of change

- [X] Add PerPodPowerManagement API type
- [X] Add logic
- [X] Modify tuned template to include new WorkloadHint
- [X] Add basic unit test and 
- [X] Add basic functional test

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -19,6 +19,10 @@ isolated_cores={{.IsolatedCpus}}
 
 not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
 
+{{if .PerPodPowerManagement}}
+[cpu]
+enabled=false
+{{else}}
 [cpu]
 #> latency-performance
 #> (override)
@@ -26,6 +30,7 @@ force_latency=cstate.id:1|3
 governor=performance
 energy_perf_bias=performance
 min_perf_pct=100
+{{end}}
 
 {{if .RealTimeHint}}
 [service]
@@ -144,7 +149,7 @@ cmdline_realtime=+nohz_full=${isolated_cores} tsc=nowatchdog nosoftlockup nmi_wa
 {{end}}
 
 {{if .HighPowerConsumption}}
-cmdline_power_performance=+processor.max_cstate=1 intel_idle.max_cstate=0 intel_pstate=disable
+cmdline_power_performance=+processor.max_cstate=1 intel_idle.max_cstate=0
 {{end}}
 
 {{if and .HighPowerConsumption .RealTimeHint}}
@@ -154,3 +159,11 @@ cmdline_idle_poll=+idle=poll
 cmdline_hugepages=+{{if .DefaultHugepagesSize}} default_hugepagesz={{.DefaultHugepagesSize}} {{end}} {{if .Hugepages}} {{.Hugepages}} {{end}}
 
 cmdline_additionalArg=+{{if .AdditionalArgs}} {{.AdditionalArgs}} {{end}}
+
+{{if .PerPodPowerManagement}}
+cmdline_pstate=+intel_pstate=passive
+{{end}}
+
+{{if .HighPowerConsumption}}
+cmdline_pstate=+intel_pstate=disable
+{{end}}
--- a/docs/performanceprofile/performance_profile.md
+++ b/docs/performanceprofile/performance_profile.md
@@ -180,5 +180,6 @@ WorkloadHints defines the set of upper level flags for different type of workloa
 | ----- | ----------- | ------ | -------- |
 | highPowerConsumption | HighPowerConsumption defines if the node should be configured in high power consumption mode. The flag will affect the power consumption but will improve the CPUs latency. | *bool | false |
 | realTime | RealTime defines if the node should be configured for the real time workload. | *bool | false |
+| perPodPowerManagement | PerPodPowerManagement defines if the node should be configured in per pod power management. PerPodPowerManagement and HighPowerConsumption hints can not be enabled together. | *bool | false |
 
 [Back to TOC](#table-of-contents)
--- a/manifests/20-performance-profile.crd.yaml
+++ b/manifests/20-performance-profile.crd.yaml
@@ -161,6 +161,9 @@ spec:
                     highPowerConsumption:
                       description: HighPowerConsumption defines if the node should be configured in high power consumption mode. The flag will affect the power consumption but will improve the CPUs latency.
                       type: boolean
+                    perPodPowerManagement:
+                      description: PerPodPowerManagement defines if the node should be configured in per pod power management. PerPodPowerManagement and HighPowerConsumption hints can not be enabled together.
+                      type: boolean
                     realTime:
                       description: RealTime defines if the node should be configured for the real time workload.
                       type: boolean
@@ -465,6 +468,9 @@ spec:
                   properties:
                     highPowerConsumption:
                       description: HighPowerConsumption defines if the node should be configured in high power consumption mode. The flag will affect the power consumption but will improve the CPUs latency.
+                      type: boolean
+                    perPodPowerManagement:
+                      description: PerPodPowerManagement defines if the node should be configured in per pod power management. PerPodPowerManagement and HighPowerConsumption hints can not be enabled together.
                       type: boolean
                     realTime:
                       description: RealTime defines if the node should be configured for the real time workload.
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
@@ -323,5 +323,11 @@ func (r *PerformanceProfile) validateWorkloadHints() field.ErrorList {
 		}
 	}
 
+	if r.Spec.WorkloadHints.HighPowerConsumption != nil && *r.Spec.WorkloadHints.HighPowerConsumption {
+		if r.Spec.WorkloadHints.PerPodPowerManagement != nil && *r.Spec.WorkloadHints.PerPodPowerManagement {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec.workloadHints.HighPowerConsumption"), r.Spec.WorkloadHints.HighPowerConsumption, "Invalid WorkloadHints configuration: HighPowerConsumption and PerPodPowerManagement can not be both enabled"))
+		}
+	}
+
 	return allErrs
 }
--- a/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
@@ -31,6 +31,7 @@ const (
 	nfConntrackHashsize                     = "nf_conntrack_hashsize=131072"
 	templateRealTimeHint                    = "RealTimeHint"
 	templateHighPowerConsumption            = "HighPowerConsumption"
+	templatePerPodPowerManagement           = "PerPodPowerManagement"
 )
 
 func new(name string, profiles []tunedv1.TunedProfile, recommends []tunedv1.TunedRecommend) *tunedv1.Tuned {
@@ -168,12 +169,17 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 	}
 
 	if profile.Spec.WorkloadHints != nil {
-		if profile.Spec.WorkloadHints.RealTime != nil && *profile.Spec.WorkloadHints.RealTime {
+		if IsRealTimeHintEnabled(profile) {
 			templateArgs[templateRealTimeHint] = "true"
 		}
 
-		if profile.Spec.WorkloadHints.HighPowerConsumption != nil && *profile.Spec.WorkloadHints.HighPowerConsumption {
+		if IsHighPowerConsumptionHintEnabled(profile) && IsPerPodPowerManagementEnabled(profile) {
+			err := fmt.Errorf("Invalid WorkloadHints configuration: HighPowerConsumption is %t and PerPodPowerManagement is %t", *profile.Spec.WorkloadHints.HighPowerConsumption, *profile.Spec.WorkloadHints.PerPodPowerManagement)
+			return nil, err
+		} else if IsHighPowerConsumptionHintEnabled(profile) {
 			templateArgs[templateHighPowerConsumption] = "true"
+		} else if IsPerPodPowerManagementEnabled(profile) {
+			templateArgs[templatePerPodPowerManagement] = "true"
 		}
 	}
 
@@ -216,4 +222,16 @@ func getProfileData(tunedTemplate string, data interface{}) (string, error) {
 
 func IsIRQBalancingGloballyDisabled(profile *performancev2.PerformanceProfile) bool {
 	return profile.Spec.GloballyDisableIrqLoadBalancing != nil && *profile.Spec.GloballyDisableIrqLoadBalancing
+}
+
+func IsRealTimeHintEnabled(profile *performancev2.PerformanceProfile) bool {
+	return profile.Spec.WorkloadHints.RealTime != nil && *profile.Spec.WorkloadHints.RealTime
+}
+
+func IsHighPowerConsumptionHintEnabled(profile *performancev2.PerformanceProfile) bool {
+	return profile.Spec.WorkloadHints.HighPowerConsumption != nil && *profile.Spec.WorkloadHints.HighPowerConsumption
+}
+
+func IsPerPodPowerManagementEnabled(profile *performancev2.PerformanceProfile) bool {
+	return profile.Spec.WorkloadHints.PerPodPowerManagement != nil && *profile.Spec.WorkloadHints.PerPodPowerManagement
 }
--- a/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
@@ -23,16 +23,18 @@ const expectedMatchSelector = `
 `
 
 var (
-	cmdlineCPUsPartitioning       = "+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} systemd.cpu_affinity=${not_isolated_cores_expanded} intel_iommu=on iommu=pt"
-	cmdlineWithStaticIsolation    = "+isolcpus=domain,managed_irq,${isolated_cores}"
-	cmdlineWithoutStaticIsolation = "+isolcpus=managed_irq,${isolated_cores}"
-	cmdlineRealtime               = "+nohz_full=${isolated_cores} tsc=nowatchdog nosoftlockup nmi_watchdog=0 mce=off skew_tick=1 rcutree.kthread_prio=11"
-	cmdlineHighPowerConsumption   = "+processor.max_cstate=1 intel_idle.max_cstate=0 intel_pstate=disable"
-	cmdlineIdlePoll               = "+idle=poll"
-	cmdlineHugepages              = "+ default_hugepagesz=1G   hugepagesz=1G hugepages=4"
-	cmdlineAdditionalArgs         = "+ audit=0 processor.max_cstate=1 idle=poll intel_idle.max_cstate=0"
-	cmdlineDummy2MHugePages       = "+ default_hugepagesz=1G   hugepagesz=1G hugepages=4 hugepagesz=2M hugepages=0"
-	cmdlineMultipleHugePages      = "+ default_hugepagesz=1G   hugepagesz=1G hugepages=4 hugepagesz=2M hugepages=128"
+	cmdlineCPUsPartitioning           = "+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} systemd.cpu_affinity=${not_isolated_cores_expanded} intel_iommu=on iommu=pt"
+	cmdlineWithStaticIsolation        = "+isolcpus=domain,managed_irq,${isolated_cores}"
+	cmdlineWithoutStaticIsolation     = "+isolcpus=managed_irq,${isolated_cores}"
+	cmdlineRealtime                   = "+nohz_full=${isolated_cores} tsc=nowatchdog nosoftlockup nmi_watchdog=0 mce=off skew_tick=1 rcutree.kthread_prio=11"
+	cmdlineHighPowerConsumption       = "+processor.max_cstate=1 intel_idle.max_cstate=0"
+	cmdlineIdlePoll                   = "+idle=poll"
+	cmdlineHugepages                  = "+ default_hugepagesz=1G   hugepagesz=1G hugepages=4"
+	cmdlineAdditionalArgs             = "+ audit=0 processor.max_cstate=1 idle=poll intel_idle.max_cstate=0"
+	cmdlineDummy2MHugePages           = "+ default_hugepagesz=1G   hugepagesz=1G hugepages=4 hugepagesz=2M hugepages=0"
+	cmdlineMultipleHugePages          = "+ default_hugepagesz=1G   hugepagesz=1G hugepages=4 hugepagesz=2M hugepages=128"
+	cmdlinePerPodPowerManagementHint  = "+intel_pstate=passive"
+	cmdlineHighPowerConsumptionPstate = "+intel_pstate=disable"
 )
 
 var _ = Describe("Tuned", func() {
@@ -160,24 +162,12 @@ var _ = Describe("Tuned", func() {
 					bootLoader, err := tunedData.GetSection("bootloader")
 					Expect(err).ToNot(HaveOccurred())
 					Expect(bootLoader.Key("cmdline_power_performance").String()).To(Equal(cmdlineHighPowerConsumption))
+					Expect(bootLoader.Key("cmdline_pstate").String()).To(Equal(cmdlineHighPowerConsumptionPstate))
 				})
 			})
 
 			When("realtime workload enabled", func() {
 				It("should not contain idle=poll cmdline", func() {
-					profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
-						RealTime:             pointer.BoolPtr(false),
-						HighPowerConsumption: pointer.BoolPtr(false),
-					}
-					tunedData := getTunedStructuredData(profile)
-					bootLoader, err := tunedData.GetSection("bootloader")
-					Expect(err).ToNot(HaveOccurred())
-					Expect(bootLoader.Key("cmdline_idle_poll").String()).ToNot(Equal(cmdlineIdlePoll))
-				})
-			})
-
-			When("realtime workload disabled", func() {
-				It("should contain idle=poll cmdline", func() {
 					profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
 						HighPowerConsumption: pointer.BoolPtr(true),
 						RealTime:             pointer.BoolPtr(true),
@@ -187,6 +177,58 @@ var _ = Describe("Tuned", func() {
 					Expect(err).ToNot(HaveOccurred())
 					Expect(bootLoader.Key("cmdline_idle_poll").String()).To(Equal(cmdlineIdlePoll))
 				})
+			})
+
+			When("realtime workload disabled", func() {
+				It("should contain idle=poll cmdline", func() {
+					profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
+						HighPowerConsumption: pointer.BoolPtr(true),
+						RealTime:             pointer.BoolPtr(false),
+					}
+					tunedData := getTunedStructuredData(profile)
+					bootLoader, err := tunedData.GetSection("bootloader")
+					Expect(err).ToNot(HaveOccurred())
+					Expect(bootLoader.Key("cmdline_idle_poll").String()).ToNot(Equal(cmdlineIdlePoll))
+				})
+			})
+
+			When("perPodPowerManagement Hint to true", func() {
+				It("should fail as PerPodPowerManagement and HighPowerConsumption can not be set to true", func() {
+					profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
+						HighPowerConsumption:  pointer.BoolPtr(true),
+						PerPodPowerManagement: pointer.BoolPtr(true),
+					}
+					_, err := NewNodePerformance(profile)
+					Expect(err.Error()).To(ContainSubstring("Invalid WorkloadHints configuration: HighPowerConsumption is true and PerPodPowerManagement is true"))
+				})
+			})
+		})
+
+		When("perPodPowerManagement Hint is false", func() {
+			It("should not contain perPodPowerManagement related parameters", func() {
+				profile.Spec.WorkloadHints.PerPodPowerManagement = pointer.BoolPtr(false)
+				tunedData := getTunedStructuredData(profile)
+				cpuSection, err := tunedData.GetSection("cpu")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(cpuSection.Key("enabled").String()).ToNot(Equal("false"))
+				bootLoaderSection, err := tunedData.GetSection("bootloader")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(bootLoaderSection.Key("cmdline_pstate").String()).ToNot(Equal(cmdlinePerPodPowerManagementHint))
+				Expect(bootLoaderSection.Key("cmdline_pstate").String()).ToNot(Equal(cmdlineHighPowerConsumptionPstate))
+			})
+		})
+
+		When("perPodPowerManagement Hint to true", func() {
+			It("should contain perPodPowerManagement related parameters", func() {
+				profile.Spec.WorkloadHints.PerPodPowerManagement = pointer.BoolPtr(true)
+				tunedData := getTunedStructuredData(profile)
+				cpuSection, err := tunedData.GetSection("cpu")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(cpuSection.Key("enabled").String()).To(Equal("false"))
+				bootLoader, err := tunedData.GetSection("bootloader")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(bootLoader.Key("cmdline_pstate").String()).To(Equal(cmdlinePerPodPowerManagementHint))
+				Expect(bootLoader.Key("cmdline_pstate").String()).ToNot(Equal(cmdlineHighPowerConsumptionPstate))
 			})
 		})
 
--- a/pkg/performanceprofile/utils/testing/testing.go
+++ b/pkg/performanceprofile/utils/testing/testing.go
@@ -81,8 +81,9 @@ func NewPerformanceProfile(name string) *performancev2.PerformanceProfile {
 				"nodekey": "nodeValue",
 			},
 			WorkloadHints: &performancev2.WorkloadHints{
-				HighPowerConsumption: pointer.BoolPtr(false),
-				RealTime:             pointer.BoolPtr(true),
+				HighPowerConsumption:  pointer.BoolPtr(false),
+				RealTime:              pointer.BoolPtr(true),
+				PerPodPowerManagement: pointer.BoolPtr(false),
 			},
 			AdditionalKernelArgs: additionalKernelArgs,
 		},
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -14,6 +14,7 @@ import (
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"gopkg.in/ini.v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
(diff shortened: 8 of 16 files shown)


test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -632,11 +632,12 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			})
 		})
 
-		When("realtime and high power consumption enabled", func() {
-			It("[test_id:50993][crit:high][vendor:cnf-qe@redhat.com][level:acceptance]should update kernel arguments and tuned accordingly", func() {
+		When("realtime and high power consumption enabled and perPodPowerManagent disabled", func() {
+			It("should update kernel arguments and tuned accordingly", func() {
```
mrniranjan, line 636: Request to retain the Test ID. These test id 's are required to match with test results with our Test case management system

pkg/performanceprofile/utils/testing/testing.go
```diff
@@ -19,9 +19,10 @@ const (
 	// ReservedCPUs defines the reserved CPU set used for tests
 	ReservedCPUs = performancev2.CPUSet("0-3")
 	// OfflinedCPUs defines the Offline CPU set used for tests
-	OfflinedCPUs     = performancev2.CPUSet("6-7") // SingleNUMAPolicy defines the topologyManager policy used for tests
-	SingleNUMAPolicy = "single-numa-node"
+	OfflinedCPUs = performancev2.CPUSet("6-7")
```
mrniranjan, line 22: Though these values that we add to ReservedCPUs , OfflinedCPUs, IsolatedCPUs are fine, But when running these tests in Baremetal, we can't use these values and we will have to fill them with what Baremetal systems have.  Because if i run them as is, tuned will start complaining if we don't account for the rest of the cpus

marioferh: Yes sorry, it was for local testing purposes, fixed.

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -142,13 +147,21 @@ cmdline_realtime=+nohz_full=${isolated_cores} tsc=nowatchdog nosoftlockup nmi_wa
...
-cmdline_power_performance=+processor.max_cstate=1 intel_idle.max_cstate=0 intel_pstate=disable
+cmdline_power_performance=+processor.max_cstate=1 intel_idle.max_cstate=0 
 {{end}}
 
 {{if and .HighPowerConsumption .RealTime}}
 cmdline_idle_poll=+idle=poll
 {{end}}
 
+{{if .PerPodPowerManagement}}
+cmdline_pstate=+intel_pstate=passive
```
MarSik, line 158: Do we want to set this explicitly or just rely on default? Or is some other part of the tuned profile inheriting something we must override? Please add a comment explaining the rationale here.

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -142,13 +147,21 @@ cmdline_realtime=+nohz_full=${isolated_cores} tsc=nowatchdog nosoftlockup nmi_wa
...
+{{end}}
+
+{{if and (not .PerPodPowerManagement) .HighPowerConsumption}}
```
MarSik, line 161: The template should contain as little logic as possible. I would personally make sure high power consumption and perpodpowermanagement are exclusive in the first place and move the logic to the operator.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -112,11 +114,50 @@ var _ = Describe("Tuned", func() {
...
 				By("Populating high power consumption cmdline")
 				Expect(cmdlineHighPowerConsumption.MatchString(manifest)).To(BeTrue())
 				By("Populating idle poll cmdline")
 				Expect(cmdlineHighPowerConsumptionWithRealtime.MatchString(manifest)).To(BeTrue())
+				By("Populating pstate cmdline")
+				Expect(cmdlineHighPowerConsumptionWithoutPerPodPowerManagement.MatchString(manifest)).To(BeTrue())
+			})
+		})
+
+		When("additional args", func() {
```
MarSik, line 128: The context here is a bit vague, what are you testing here?

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -112,11 +114,50 @@ var _ = Describe("Tuned", func() {
...
+				profile.Spec.AdditionalKernelArgs = nil
+				manifest := getTunedManifest(profile)
+				Expect(manifest).ToNot(ContainSubstring("audit=0"))
+				Expect(manifest).ToNot(ContainSubstring("processor.max_cstate=1"))
+				Expect(manifest).ToNot(ContainSubstring("idle=poll"))
+				Expect(manifest).ToNot(ContainSubstring("intel_idle.max_cstate=0"))
+			})
+		})
+
+		When("perPodPowerManagement Hint to false", func() {
```
MarSik, line 147:
```suggestion
		When("perPodPowerManagement Hint is false", func() {
```

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -112,11 +114,50 @@ var _ = Describe("Tuned", func() {
...
+			It("should not contain perPodPowerManagement related parameters", func() {
+				manifest := getTunedManifest(profile)
+				Expect(manifest).ToNot(ContainSubstring("enabled=false"))
+				Expect(cmdlinePerPodPowerManagementHint.MatchString(manifest)).ToNot(BeTrue())
+			})
+		})
+
+		When("perPodPowerManagement Hint to true", func() {
```
MarSik, line 155:
```suggestion
		When("perPodPowerManagement Hint is true", func() {
```

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -112,11 +114,50 @@ var _ = Describe("Tuned", func() {
...
+			It("should contain perPodPowerManagement related parameters", func() {
+				profile.Spec.WorkloadHints.PerPodPowerManagement = pointer.BoolPtr(true)
+				manifest := getTunedManifest(profile)
+				Expect(manifest).To(ContainSubstring("enabled=false"))
```
MarSik, line 159: This is too generic I think, we can use enabled/disabled in other parts of the profile too

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -112,11 +114,50 @@ var _ = Describe("Tuned", func() {
...
+				Expect(manifest).ToNot(ContainSubstring("processor.max_cstate=1"))
+				Expect(manifest).ToNot(ContainSubstring("idle=poll"))
+				Expect(manifest).ToNot(ContainSubstring("intel_idle.max_cstate=0"))
+			})
+		})
+
+		When("perPodPowerManagement Hint to false", func() {
+			It("should not contain perPodPowerManagement related parameters", func() {
+				manifest := getTunedManifest(profile)
+				Expect(manifest).ToNot(ContainSubstring("enabled=false"))
```
MarSik, line 150: Too generic, we can use enabled/disabled in other parts of the profile

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -669,6 +670,48 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+				By("Applying changes in performance profile and waiting until mcp will start updating")
+				mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+
+				By("Waiting when mcp finishes updates")
+				mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+
+				By("Verifying node kernel arguments")
+				cmdline, err := nodes.ExecCommandOnMachineConfigDaemon(&workerRTNodes[0], []string{"cat", "/proc/cmdline"})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(cmdline).To(ContainSubstring("intel_pstate=passive"))
```
MarSik, line 700: As above, I am not sure about this part.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: Right, @bartwensley point that. I am just figuring out how to check in the right way.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -669,6 +670,48 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+
+				By("Verifying tuned profile")
+				key := types.NamespacedName{
+					Name:      components.GetComponentName(profile.Name, components.ProfileNamePerformance),
+					Namespace: components.NamespaceNodeTuningOperator,
+				}
+				tuned := &tunedv1.Tuned{}
+				err = testclient.Client.Get(context.TODO(), key, tuned)
+				Expect(err).ToNot(HaveOccurred(), "cannot find the Cluster Node Tuning Operator object")
+				Expect(*tuned.Spec.Profile[0].Data).To(ContainSubstring("enabled=false"))
```
MarSik, line 711: enabled= is too generic, we should check if C states are enabled or that the profile does not use min_perf_pct for example.

assets/performanceprofile/tuned/openshift-node-performance
bartwensley: I would prefer this is explicit. I believe the cpu-partitioning profile is setting this to intel_pstate=disable so we need to override it.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
bartwensley: I think here we need to check that the [cpu] plugin is disabled.

assets/performanceprofile/tuned/openshift-node-performance
marioferh: Right, there is and if else for that:

`{{if .PerPodPowerManagement}}
{{end}}

{{if and (not .PerPodPowerManagement) .HighPowerConsumption}}
{{end}}`

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: I'm testing when it has aadditional args field, it can be confused...

marioferh: This not correspond to this PR, better do it in a new one:
https://github.com/openshift/cluster-node-tuning-operator/pull/420

assets/performanceprofile/tuned/openshift-node-performance
marioferh: Well, there a a bunch of if in the template, it could be a good idea do a separate PR to move al the logic to the tuned.go. Agree?

marioferh: Move it to tuned, I think it is less clear but...

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: done

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: done

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: Add a regex, not very happy with it but I don't see other option

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: same

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
marioferh: same

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
marioferh: same as above, do it with a regex

test/e2e/performanceprofile/testdata/render-expected-output/manual_tuned.yaml
```diff
@@ -54,7 +54,7 @@ spec:
       tuned.non_isolcpus=${not_isolated_cpumask} systemd.cpu_affinity=${not_isolated_cores_expanded}
       intel_iommu=on iommu=pt\n\n\ncmdline_isolation=+isolcpus=managed_irq,${isolated_cores}\n\n\n\ncmdline_realtime=+nohz_full=${isolated_cores}
       tsc=nowatchdog nosoftlockup nmi_watchdog=0 mce=off skew_tick=1 rcutree.kthread_prio=11\n\n\n\n\n\n\ncmdline_hugepages=+
-      default_hugepagesz=1G   hugepagesz=2M hugepages=128 \n\ncmdline_additionalArg=+\n"
+      default_hugepagesz=1G   hugepagesz=2M hugepages=128 \n\ncmdline_additionalArg=+\n\n\n\n\n"
```
bartwensley, line 57: ???

marioferh: When tuned template [assets/performanceprofile/tuned/openshift-node-performance](https://github.com/openshift/cluster-node-tuning-operator/pull/415/files/0c12451b26164b9edec463eac428b64494b0edd5#diff-e0d5dac27460da1c7759fdee98e915a6c1a4eed9a8969202562add4c6d41b292) the output string to check the test in tuned_test.go. If no \n are added, test fails.
I am thinking in a better way to test the changes in the template creating a struct parsing the yaml but it's not easy because the way template is parsed.

marioferh: I have this commit to improve tuned test https://github.com/marioferh/cluster-node-tuning-operator/commit/6eacd3c30f24323eda9f8e419fb5b57c05556f56

But the test above is about render command that the output is a string with the whole tuned kubernetes object in string format, I don't know even if this patch is ugly, if it worth the time to parse the whole string and check every field

bartwensley: OK - but prior to your changes, the yaml file was being rendered without the extra newlines - right? So doesn't that mean there is something wrong with this PR in that it is now introducing a bunch of extra newlines?

marioferh: The issue is due because these lines:
`
{{if .HighPowerConsumption}}
{{end}}

{{if .PerPodPowerManagement}}
{{end}}
`

Before both are = false, the output of the test is: 

`                                          "data": strings.Join({                                                                                                                                                             
                                                ... // 4119 identical bytes                                                                                                                                                
                                                " rcutree.kthread_prio=11\n\n\n\n\n\n\ncmdline_hugepages=+ default_hugep",                                                                                                 
                                                "agesz=1G   hugepagesz=2M hugepages=128 \n\ncmdline_additionalArg=+",                                                                                                      
        -                                       "\n\n\n\n",                                                                                                                                                                
                                                "\n",                                                                                                                                                                      
                                        }, ""),                                                                                                                                                                            
                                        "name": string("openshift-node-performance-manual"),      `

I don't see another solution. 
And I've said before, it makes no sense for mi parser a output string of binary to an struct just for this test.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: Fixed here: https://github.com/openshift/cluster-node-tuning-operator/pull/426

assets/performanceprofile/tuned/openshift-node-performance
marioferh: Move again from tuned to here. There is a parsers for the template that fails:
`configparser.DuplicateOptionError: While reading from '/etc/tuned/openshift-node-performance-manual/tuned.conf' [line 162]: option 'cmdline_pstate' in section 'bootloader' already exists
 `

marioferh (commented):
/hold
depends on https://github.com/openshift/cluster-node-tuning-operator/pull/426/files
https://github.com/openshift/cluster-node-tuning-operator/pull/426/files

manifests/20-performance-profile.crd.yaml
```diff
@@ -161,6 +161,9 @@ spec:
                     highPowerConsumption:
                       description: HighPowerConsumption defines if the node should be configured in high power consumption mode. The flag will affect the power consumption but will improve the CPUs latency.
                       type: boolean
+                    perPodPowerManagement:
+                      description: perPodPowerManagement TODO
```
bartwensley, line 165: Since this will hopefully be merged soon, you should probably fill in all the missing "TODO" text (there are several in the file).

marioferh: Sure

marioferh: Done

assets/performanceprofile/tuned/openshift-node-performance
marioferh: Sorry I missunestood the issue. Now is fixing, the logic has been moved to tuned.go and both hints can not set to enabled at the same time.

marioferh: Now:
```
{{if .PerPodPowerManagement}}
marioferh marked this conversation as resolved.
{{end}}

{{if .HighPowerConsumption}}
{{end}}
```
I do not mark as resolve to let you see the changes.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -167,14 +168,17 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 		}
 	}
 
-	if profile.Spec.WorkloadHints != nil {
-		if profile.Spec.WorkloadHints.RealTime != nil && *profile.Spec.WorkloadHints.RealTime {
-			templateArgs[templateRealTimeHint] = "true"
-		}
+	if IsRealTimeHintEnabled(profile) {
```
fromanirh, line 171: minor, nonblocking: in general, refactoring changes should be in their own commit, and logic changes should be in their own separate commit.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -217,3 +221,15 @@ func getProfileData(tunedTemplate string, data interface{}) (string, error) {
 func IsIRQBalancingGloballyDisabled(profile *performancev2.PerformanceProfile) bool {
 	return profile.Spec.GloballyDisableIrqLoadBalancing != nil && *profile.Spec.GloballyDisableIrqLoadBalancing
 }
+
+func IsRealTimeHintEnabled(profile *performancev2.PerformanceProfile) bool {
+	return profile.Spec.WorkloadHints.RealTime != nil && *profile.Spec.WorkloadHints.RealTime
```
fromanirh, line 226: can `profile.Spec.WorkloadHints` be nil? we should probably add some unit tests, mostly to document the behavior

fromanirh: additionally, the way we are handling the hints conflates "unspecified" with "false". We treat them the same. Is it what we want to do? possibly yes, let's just explicitely agree about it.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -166,28 +168,67 @@ var _ = Describe("Tuned", func() {
...
 			})
+
+			When("perPodPowerManagement Hint to true", func() {
+				It("should fail as PerPodPowerManagement and HighPowerConsumption can not be set to true", func() {
+					profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
+						HighPowerConsumption:  pointer.BoolPtr(true),
+						PerPodPowerManagement: pointer.BoolPtr(true),
+					}
+					_, err := NewNodePerformance(profile)
+					Expect(err).To(HaveOccurred())
```
fromanirh, line 201: do we want to test if the error we get actually means the two hints are conflicting? otherwise we can get unrelated errors and misinterpret it. Can be done later if requires too deep changes.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -166,28 +168,67 @@ var _ = Describe("Tuned", func() {
...
+			It("should contain perPodPowerManagement related parameters", func() {
+				profile.Spec.WorkloadHints.PerPodPowerManagement = pointer.BoolPtr(true)
+				tunedData := getTunedStructuredData(profile)
+				cpuSection, err := tunedData.GetSection("cpu")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(cpuSection.Key("enabled").String()).To(Equal("false"))
+				bootLoader, err := tunedData.GetSection("bootloader")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(bootLoader.Key("cmdline_pstate").String()).To(Equal(cmdlinePerPodPowerManagementHint))
+				Expect(bootLoader.Key("cmdline_pstate").String()).ToNot(Equal(cmdlineHighPowerConsumptionPstate))
```
fromanirh, line 230: when `cmdline_pstate` should be equal to `cmdlineHighPowerConsumptionPstate`?

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -669,6 +671,50 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+				By("Patching the performance profile with workload hints")
+				workloadHints, err := json.Marshal(profile.Spec.WorkloadHints)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(testclient.Client.Patch(context.TODO(), profile,
+					client.RawPatch(
+						types.JSONPatchType,
+						[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec/workloadHints", "value": %s }]`, workloadHints)),
+					),
+				)).ToNot(HaveOccurred())
```
fromanirh, line 690: let's make sure to revert this patch if the test fails early - and yes, other tests may have the same issue, so I'd like to introduce this test already fixed and then fix the others/file a issue to fix the others separately

docs/performanceprofile/performance_profile.md
```diff
@@ -180,5 +180,6 @@ WorkloadHints defines the set of upper level flags for different type of workloa
 | ----- | ----------- | ------ | -------- |
 | highPowerConsumption | HighPowerConsumption defines if the node should be configured in high power consumption mode. The flag will affect the power consumption but will improve the CPUs latency. | *bool | false |
 | realTime | RealTime defines if the node should be configured for the real time workload. | *bool | false |
+| perPodPowerManagement | perPodPowerManagement defines if the node should be configured in per pod power management. | *bool | false |
```
bartwensley, line 183: Should this mention that perPodPowerManagement is not compatible with highPowerConsumption?

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -669,6 +671,50 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+
+				By("Verifying tuned profile")
+				key := types.NamespacedName{
+					Name:      components.GetComponentName(profile.Name, components.ProfileNamePerformance),
+					Namespace: components.NamespaceNodeTuningOperator,
+				}
+				tuned := &tunedv1.Tuned{}
+				err = testclient.Client.Get(context.TODO(), key, tuned)
+				Expect(err).ToNot(HaveOccurred(), "cannot find the Cluster Node Tuning Operator object")
+				cpuEnabledFalse := regexp.MustCompile(`\s*force_latency=cstate.id:1|3\ngovernor=performance\nenergy_perf_bias=performance\nmin_perf_pct=100\\n\\nenabled=false\s*`)
```
bartwensley, line 713: It would be better to look for "[cpu]" followed by anything other than "[" followed by "enabled=false".

marioferh: done, changed as in tuned_test.go

docs/performanceprofile/performance_profile.md
marioferh: I'm not sure if we should include this in the API or in the docs or release notes @MarSik

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: Added in previous test when only highPowerConsumption is to true:

`			When("default realtime workload settings", func() {
				FIt("should contain high power consumption related parameters", func() {
					profile.Spec.WorkloadHints = &performancev2.WorkloadHints{HighPowerConsumption: pointer.BoolPtr(true)}
					tunedData := getTunedStructuredData(profile)
					bootLoader, err := tunedData.GetSection("bootloader")
					Expect(bootLoader.Key("cmdline_power_performance").String()).To(Equal(cmdlineHighPowerConsumption))
					Expect(bootLoader.Key("cmdline_pstate").String()).To(Equal(cmdlineHighPowerConsumptionPstate))
				})
			})`

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: good catch, thx. Done

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
marioferh: #_1 done
#_2 yes, default realTime Hint cannot be unspecified because default is = true, and the other ones unspecified is = false

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
marioferh: Yes sure, many changes before it, if necessary I can do it in a separate commit or PR but I did like that because is clearer even if mix differentes goals in the same commit.

fromanirh: improving the current state (like extracting these helpers) and ensuring this change keeps the behavior consistent should be done separately from logic changes, because it makes review and git bisection easier. Anyway: not a big deal for this case, surely not worth a split *now*.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
fromanirh: If we do that in AfterEach, that's enough indeed.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
marioferh: thanks

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -26,7 +26,9 @@ force_latency=cstate.id:1|3
 governor=performance
 energy_perf_bias=performance
 min_perf_pct=100
-
+{{if .PerPodPowerManagement}}
+enabled=false
```
bartwensley, line 30: nit: Don't need to change this now, but the resulting profile would look better if the "enabled=false" was the only line in the [cpu] section. Have you confirmed that tuned handles this properly when several configuration options are provided for a plugin, followed by "enabled=false"?

marioferh: `Have you confirmed that tuned handles this properly when several configuration options are provided for a plugin, followed by "enabled=false"?`

Yes, the plugin is disabled.

`nit: Don't need to change this now, but the resulting profile would look better if the "enabled=false" was the only line in the [cpu] section. `

Right, if test fail I can update it, if passed I can do a refactor in the future

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -669,6 +672,55 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+				tuned := &tunedv1.Tuned{}
+				err = testclient.Client.Get(context.TODO(), key, tuned)
+				Expect(err).ToNot(HaveOccurred(), "cannot find the Cluster Node Tuning Operator object")
+				tunedData := getTunedStructuredData(profile)
+				cpuSection, err := tunedData.GetSection("cpu")
+				Expect(err).ToNot(HaveOccurred())
+				Expect((cpuSection.Key("force_latency").String())).To(Equal("cstate.id:1|3"))
+				Expect((cpuSection.Key("governor").String())).To(Equal("performance"))
+				Expect((cpuSection.Key("energy_perf_bias").String())).To(Equal("performance"))
+				Expect((cpuSection.Key("min_perf_pct").String())).To(Equal("100"))
```
bartwensley, line 720: This should be checking that "enabled=false". The other options don't matter and shouldn't be checked here.

marioferh: oh, then it was wrong and working :S
fixed

assets/performanceprofile/tuned/openshift-node-performance
marioferh: done

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -669,6 +672,51 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+					Name:      components.GetComponentName(profile.Name, components.ProfileNamePerformance),
+					Namespace: components.NamespaceNodeTuningOperator,
+				}
+				tuned := &tunedv1.Tuned{}
+				err = testclient.Client.Get(context.TODO(), key, tuned)
+				Expect(err).ToNot(HaveOccurred(), "cannot find the Cluster Node Tuning Operator object")
+				tunedData := getTunedStructuredData(profile)
+				cpuSection, err := tunedData.GetSection("cpu")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(cpuSection.Key("enabled").String()).ToNot(Equal("false"))
```
bartwensley, line 716: Shouldn't this be "To(Equal("false"))" instead of "ToNot"?

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
fromanirh: So how do we handle `profile.Spec.WorkloadHints` possibly being nil?

fromanirh (commented):
I still have some questions about the implementation. The overall design LGTM, and I'll defer the choice of values in the profile proper.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
marioferh: I didn't realize and I've removed it with the refactor but I have added id again:
[
`if profile.Spec.WorkloadHints != nil {`](https://github.com/openshift/cluster-node-tuning-operator/blob/59698befa86989be3295f825c2faa3e8c5b5d4bb/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go#L171)

fromanirh: thanks, this is good enough indeed.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -160,34 +162,74 @@ var _ = Describe("Tuned", func() {
...
-						RealTime:             pointer.BoolPtr(false),
-						HighPowerConsumption: pointer.BoolPtr(false),
+						HighPowerConsumption: pointer.BoolPtr(true),
+						RealTime:             pointer.BoolPtr(true),
 					}
 					tunedData := getTunedStructuredData(profile)
 					bootLoader, err := tunedData.GetSection("bootloader")
 					Expect(err).ToNot(HaveOccurred())
-					Expect(bootLoader.Key("cmdline_idle_poll").String()).ToNot(Equal(cmdlineIdlePoll))
+					Expect(bootLoader.Key("cmdline_idle_poll").String()).To(Equal(cmdlineIdlePoll))
```
fromanirh, line 178: do we want to keep the old negative test? we're changing it to positive

fromanirh (commented):
/lgtm
/hold

fromanirh:
/approve
OK for the general direction, and we need this change

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
marioferh: The test was wrong.

https://github.com/openshift/cluster-node-tuning-operator/blob/59698befa86989be3295f825c2faa3e8c5b5d4bb/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go#L169

and the negative one:

https://github.com/openshift/cluster-node-tuning-operator/blob/59698befa86989be3295f825c2faa3e8c5b5d4bb/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go#L182

fromanirh: This is in general good information to convey in the commit message: why the old test was wrong, and how we are fixing it.

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -154,3 +159,11 @@ cmdline_idle_poll=+idle=poll
 cmdline_hugepages=+{{if .DefaultHugepagesSize}} default_hugepagesz={{.DefaultHugepagesSize}} {{end}} {{if .Hugepages}} {{.Hugepages}} {{end}}
 
 cmdline_additionalArg=+{{if .AdditionalArgs}} {{.AdditionalArgs}} {{end}}
+
+{{if .PerPodPowerManagement}}
```
yanirq, line 163: I am not sure how will this be translated to tuned final object in case both `PerPodPowerManagement` and `HighPowerConsumption` are set to true. 
Is that a valid scenario ?  and if so, how will the final `cmdline_pstate` line will look like in the generated tuned ?

marioferh: It is not possible to set both to true:

https://github.com/openshift/cluster-node-tuning-operator/blob/59698befa86989be3295f825c2faa3e8c5b5d4bb/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go#L176

Where I should document that?

docs/performanceprofile/performance_profile.md
yanirq: I can see we have a test that states they cannot be added together. maybe worth adding a note here

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -160,34 +162,74 @@ var _ = Describe("Tuned", func() {
...
 			})
+
+			When("perPodPowerManagement Hint to true", func() {
+				It("should fail as PerPodPowerManagement and HighPowerConsumption can not be set to true", func() {
+					profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
+						HighPowerConsumption:  pointer.BoolPtr(true),
+						PerPodPowerManagement: pointer.BoolPtr(true),
+					}
+					_, err := NewNodePerformance(profile)
+					Expect(err.Error()).To(ContainSubstring("Invalid WorkloadHints configuration: HighPowerConsumption is true and PerPodPowerManagement is true"))
```
yanirq, line 202: Do we have a validation hook that checks this behavior ? where do we prevent the user from setting both hints as true ?  should probably go under https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/apis/performanceprofile/v2/performanceprofile_validation.go

assets/performanceprofile/tuned/openshift-node-performance
yanirq: I think we should prevent that at profile creation time (aka validation hook)

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -323,5 +324,11 @@ func (r *PerformanceProfile) validateWorkloadHints() field.ErrorList {
 		}
 	}
 
+	if r.Spec.WorkloadHints.HighPowerConsumption != nil && *r.Spec.WorkloadHints.HighPowerConsumption {
```
yanirq, line 327: don'y we need to check if `r.Spec.WorkloadHints != nil`  first to avoid an exception ?

yanirq: I would rap the whole section if so

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -323,5 +324,11 @@ func (r *PerformanceProfile) validateWorkloadHints() field.ErrorList {
...
+		if r.Spec.WorkloadHints.PerPodPowerManagement != nil && *r.Spec.WorkloadHints.PerPodPowerManagement {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec.workloadHints.HighPowerConsumption"), r.Spec.WorkloadHints.HighPowerConsumption, "Invalid WorkloadHints configuration: HighPowerConsumption to enabled and PerPodPowerManagement enabled"))
+		}
```
yanirq, line 330: suggestion for the error message:  HighPowerConsumption and PerPodPowerManagement can not be both enabled

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
marioferh: I don't think so, it is done like this in other part of the code

marioferh: I mean there is a check before:
`if r.Spec.WorkloadHints == nil {
		return allErrs
	}`

fromanirh:
/lgtm
since last lgtm it seems only rebases
let's invest a bit more in the commit message and commit splitting next time maybe

PR #417, opened 2022-08-03
Bug 2108222 - Add offlined cpus to v1 api

marioferh:
## Description

Field spec.cpu.offlined must be also in v1 API as v1 and v2 API must be compatible

## Type of change

- [X] Add offlined cpu to v1 API
- [X] Generate new API and docs
- [] Add test

## Testing steps


MarSik (approved):
/lgtm

PR #419, opened 2022-08-04
Bug 1986681: Added verification on spec.cpu.reserved is required

shaior:
Automates https://bugzilla.redhat.com/show_bug.cgi?id=1986681


fromanirh:
/ok-to-test
don't we have coverage in the existing suite for this flow already?

shaior:
This bug was already implemented, closing it for now.

PR #421, opened 2022-08-04
Bug 2114779: openshift-tuned: remember recommended profile

jmencak:
Initially, `openshift-tuned` was aimed to be as staleless as possible.
For the state information it needed, commands such as `tuned-adm recommend`
were used.  Over time, we realized we need to keep state one
way or another in `openshift-tuned`.  Recent rhbz#2114779 shows, that the
output from `tuned-adm recommend` can sometimes be misleading.  We did
not yet find a root cause for this.  Therefore, we remember the
recommended profile for TuneD, rather than querying for it via
`tuned-adm recommend` command.  This should have a nice side effect of
using less CPU as we do not need to run an external command.

Other fixes/changes:
  - Fix the initial timeout for tunedTicker.

Resolves rhbz#2114779

### DIFF
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -138,6 +138,8 @@ type Controller struct {
 		stderr string
 		// stopping is true while the controller tries to stop the TuneD daemon.
 		stopping bool
+		// the TuneD profile we wish to be applied.
+		recommendedProfile string
 	}
 
 	tunedCmd     *exec.Cmd       // external command (tuned) being prepared or run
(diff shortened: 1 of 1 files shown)


jmencak:
Issue also discussed here: https://coreos.slack.com/archives/CNNFPNXBR/p1659447890826089

jmencak:
Sharing the log where the original issue is directly visible:
```
2022-07-24T02:36:35.348826897Z I0724 02:36:35.348677    2504 controller.go:511] written "/etc/tuned/recommend.d/50-openshift.conf" to set TuneD profile openshift-node
2022-07-24T02:36:35.675707772Z I0724 02:36:35.675511    2504 controller.go:816] active and recommended profile (virtual-guest) match; profile change will not trigger profile reload
```
i.e. `TuneD` did not see the recommended/requested profile.

jmencak:
/hold
needs more testing

jmencak:
After 14h of testing and cca 300k writes to `/etc/tuned/recommend.d/50-openshift.conf`, running `tuned-adm recommend` immediately after and comparing the values, I was not able to reproduce the issue across 3 hosts.  I.e. cca 1M tests.  This was without `fsync()`.  Dropping the added `fsync()` from this PR.  I wonder if the issue was some underlying HW issue we have no control over.  Anyway, keeping the value in memory vs. reading again from the container's filesystem should be more robust.

/hold cancel
QE for awareness:
/cc @liqcui

PR #422, opened 2022-08-05
Bug 2115802: Get latest profile before updating and use nodes.GetKubeletConfig 

mrniranjan:
- Get latest profile before modifying the profile .
- use nodes.GetKubeletConfig to get all the kubeletconfig values

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -420,34 +420,75 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 		})
 	})
 
-	// TODO: we have a dependency between tests(that in general bad practice, but saves us some tests run time),
-	// once we will want to run tests in the random order or without failFast we will need to refactor tests
 	Context("Updating of nodeSelector parameter and node labels", func() {
 		var mcp *machineconfigv1.MachineConfigPool
 		var newCnfNode *corev1.Node
 
 		newRole := "worker-test"
 		newLabel := fmt.Sprintf("%s/%s", testutils.LabelRole, newRole)
+		var labelsDeletion = false
 		newNodeSelector := map[string]string{newLabel: ""}
 
-		testutils.BeforeAll(func() {
+		//fetch the latest profile
+		profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+		Expect(err).ToNot(HaveOccurred())
+		var oldMcpSelector, oldNodeSelector map[string]string
+		var removeLabels func(map[string]string, *corev1.Node) error
+
+		//fetch existing MCP Selector if exists in profile
+		BeforeEach(func() {
+			//fetch existing MCP Selector if exists in profile
+			if profile.Spec.MachineConfigPoolSelector != nil {
+				oldMcpSelector = profile.Spec.DeepCopy().MachineConfigPoolSelector
+			}
+
+			//fetch existing Node Selector
+			if profile.Spec.NodeSelector != nil {
+				oldNodeSelector = profile.Spec.DeepCopy().NodeSelector
+			}
+			removeLabels = func(nodeSelector map[string]string, targetNode *corev1.Node) error {
+				for l := range nodeSelector {
+					delete(targetNode.Labels, l)
+				}
+				label, err := json.Marshal(targetNode.Labels)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(testclient.Client.Patch(context.TODO(), targetNode,
+					client.RawPatch(
+						types.JSONPatchType,
+						[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/metadata/labels", "value": %s }]`, label)),
+					),
+				)).ToNot(HaveOccurred())
+				mcps.WaitForCondition(testutils.RoleWorker, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+
+				By("Waiting when MCP Worker complete updates and verifying that node reverted back configuration")
+				mcps.WaitForCondition(testutils.RoleWorker, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+				return nil
+			}
+
 			nonPerformancesWorkers, err := nodes.GetNonPerformancesWorkers(profile.Spec.NodeSelector)
 			Expect(err).ToNot(HaveOccurred())
 			if len(nonPerformancesWorkers) != 0 {
 				newCnfNode = &nonPerformancesWorkers[0]
 			}
-		})
-
-		JustBeforeEach(func() {
 			if newCnfNode == nil {
 				Skip("Skipping the test - cluster does not have another available worker node ")
 			}
-		})
 
-		It("[test_id:28440]Verifies that nodeSelector can be updated in performance profile", func() {
 			nodeLabel = newNodeSelector
 			newCnfNode.Labels[newLabel] = ""
+
 			Expect(testclient.Client.Update(context.TODO(), newCnfNode)).ToNot(HaveOccurred())
+
+			//Remove the MCP Selector if exists
+			if profile.Spec.MachineConfigPoolSelector != nil {
+				By("Removing Machine Config Selector")
+				Expect(testclient.Client.Patch(context.TODO(), profile,
+					client.RawPatch(
+						types.JSONPatchType,
+						[]byte(fmt.Sprintf(`[{"op": "remove", "path": "/spec/%s"}]`, "machineConfigPoolSelector")),
+					),
+				)).ToNot(HaveOccurred())
+			}
 
 			By("Creating new MachineConfigPool")
 			mcp = mcps.New(newRole, newNodeSelector)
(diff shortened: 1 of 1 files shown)


MarSik (approved):
/lgtm

ffromani (commented):
/lgtm

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -439,6 +439,9 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			if newCnfNode == nil {
 				Skip("Skipping the test - cluster does not have another available worker node ")
 			}
+			if profile.Spec.MachineConfigPoolSelector != nil {
```
MarSik, line 442: This is fine from code perspective, but are we sure someone will invoke this test without MCPselector set too?

mrniranjan: yes atleast in our CI we can invoke the test without mcp selector

mrniranjan: Also the default profile that gets generated by config.go doesn't have mcp selector. So the default profile that gets created when no performance profile exists will have this tests run.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -536,6 +553,20 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				),
 			)).ToNot(HaveOccurred())
 
+			if mcpSelectorRemoved {
```
fromanirh, line 556: but what guarantees that this will run **after** test_id:28440 , so the revert will actually happen? ginkgo by defaults run specs in random order: https://onsi.github.io/ginkgo/#spec-randomization

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -456,21 +461,32 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
 			mcps.WaitForCondition(newRole, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
 
 			By("Waiting when MCP finishes updates and verifying new node has updated configuration")
 			mcps.WaitForCondition(newRole, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
 
-			kblcfg, err := nodes.ExecCommandOnNode(chkKubeletConfig, newCnfNode)
-			Expect(err).ToNot(HaveOccurred(), "failed to execute %s", chkKubeletConfig)
-			Expect(kblcfg).To(ContainSubstring("topologyManagerPolicy"))
+			kubeletConfig, err := nodes.GetKubeletConfig(newCnfNode)
+			Expect(kubeletConfig.TopologyManagerPolicy).ToNot(BeEmpty())
```
fromanirh, line 489: this seems like a good change and we can merge it separately anytime. But it will require the BZ dance so likely not worth the extra hassle

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
mrniranjan: @fromanirh i can try moving this code as AfterEach Section ?

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -420,35 +420,64 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
 
-	// TODO: we have a dependency between tests(that in general bad practice, but saves us some tests run time),
-	// once we will want to run tests in the random order or without failFast we will need to refactor tests
 	Context("Updating of nodeSelector parameter and node labels", func() {
 		var mcp *machineconfigv1.MachineConfigPool
 		var newCnfNode *corev1.Node
 
 		newRole := "worker-test"
 		newLabel := fmt.Sprintf("%s/%s", testutils.LabelRole, newRole)
+		var mcpSelectorRemoved, labelsDeletion = false, false
```
fromanirh, line 429: these values will be initialized once, we need to make sure they are initialized to value we expect before _each_ run

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -420,35 +420,64 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
 		newNodeSelector := map[string]string{newLabel: ""}
 
-		testutils.BeforeAll(func() {
+		existingMcpSelector := map[string]string{
+			fmt.Sprintf("%s/%s", "machineconfiguration.openshift.io", "role"): testutils.RoleWorkerCNF,
+		}
+		var removeLabels = func(nodeSelector map[string]string, targetNode *corev1.Node) error {
```
fromanirh, line 435: this works, but at glance it seems this code can be a regular function? IOW, why do we need to define this inline here?

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -420,35 +420,64 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
 		var mcp *machineconfigv1.MachineConfigPool
 		var newCnfNode *corev1.Node
 
 		newRole := "worker-test"
 		newLabel := fmt.Sprintf("%s/%s", testutils.LabelRole, newRole)
+		var mcpSelectorRemoved, labelsDeletion = false, false
 		newNodeSelector := map[string]string{newLabel: ""}
 
-		testutils.BeforeAll(func() {
+		existingMcpSelector := map[string]string{
```
fromanirh, line 432: wouldn't it be better/safer to fetch the actual mcp selector - whatever it is and store it here?
is this possible?

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -561,7 +581,27 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 
 			// revert node label to have the expected value
 			nodeLabel = testutils.NodeSelectorLabels
+
+			if mcpSelectorRemoved {
+				By("Restoring Machine config selector")
+				profile.Spec.MachineConfigPoolSelector = existingMcpSelector
+				spec, err := json.Marshal(profile.Spec)
+				Expect(err).ToNot(HaveOccurred())
```
fromanirh, line 589: this works, but I think we should prepare the revert _before_ to commit the change. This way we maximize the steps we prepared, and we minimize the possible failures in the restore path.
IOW:
1. fetch the old mcp selector
2. JSON marshal it - prepate the data for the restoration
3. only now, only if *both* the above succeeded, change the cluster MCP selector.

This way in the restore path the only thing that can fail is the patch operation, but we can't avoid, it's the minimum we need. Applies also to the label handling.

ffromani (commented):
looks reasonnable, possible improvements inside

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
mrniranjan: I have updated the code as per the above requirements.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
mrniranjan: Because this code is very specific to this test case .Also having this function inline within this context, its easy to understand the code easily. This helps in having 2 test cases independent.

mrniranjan: Also this code is very specific to the test case of checking the configuration after the node labels are reverted.  so the steps are :
1. Remove the node labels 
2. check the configuration test 
3. Revert the node labels 
Since step1 is mostly a setup part , we want to put in Before Each. But at the same time we want this to be executed only if this test case is executed. So we do not want this code to be executed if we are not calling this test case. 
Also step3 we have put in AfterEach. and we want this to be executed only when Step1 happens. Hence we have the variable "labelsDeletion" which acts as a flag. 
Due to the above reasons , i believe the below code suits the above requirements.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -420,35 +420,70 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
-		testutils.BeforeAll(func() {
+		//fetch the latest profile
+		profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+		Expect(err).ToNot(HaveOccurred())
+		oldMcpSelector := make(map[string]string)
+
+		//fetch existing MCP Selector if exists in profile
+		if profile.Spec.MachineConfigPoolSelector != nil {
+			oldMcpSelector = profile.Spec.DeepCopy().MachineConfigPoolSelector
+		}
```
fromanirh, line 440: we should have only variable declarations in the `Context` block. Trivial initializations (using constants, for examples) are OK, but more complex code, like getting the profile, should be moved in `BeforeEach` - or everywhere else suitable.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
fromanirh: not a fan, but let's keep this direction for now.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -420,35 +420,70 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+		var removeLabels = func(nodeSelector map[string]string, targetNode *corev1.Node) error {
+			for l := range nodeSelector {
+				delete(targetNode.Labels, l)
+			}
+			label, err := json.Marshal(targetNode.Labels)
+			if err != nil {
+				return err
```
fromanirh, line 447: mixing `return err` and `Expect*` is a smell. Let's pick either and be consistent within the single function

fromanirh: additionally, this is fragile. A bit better:
1. clone the labels of the `targetNode`
2. remove the labels as needed
3. marshal the new labels
4. if and only if we succeeded this far, update the `targetNode` object (or possibly a `DeepCopy()` of it) and do the Patch operation

ffromani (commented):
thanks for the updates. Almost there.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
mrniranjan: @fromanirh I have updated the as per above requirements.

ffromani (commented):
/lgtm
/approve
let's postpone further improvements

mrniranjan:
@fromanirh Updating code a minor fix with regard to variable declarations.

fromanirh:
should the lane fail again, I'll review the failure(s) and check if a override makes sense.

fromanirh:
doublecheck: https://github.com/openshift/cluster-node-tuning-operator/pull/498

fromanirh:
/override ci/prow/e2e-aws-ovn
the lane is failing for 1. apparently unrelated reasons and 2. in a very similar way wrt https://github.com/openshift/cluster-node-tuning-operator/pull/498 - which can't break anything and surely not OVN.

Considering how long this PR was open and how much effort was poured into it, I'm overriding the lane.
In case this PR actually breaks OVN (!!!!!) I will be accountable for a fix/revert.

fromanirh:
it's unfortunate the upgrade lane is red so much - failures really seem unrelated. There's not enough supporting evidence to override it yet. Worth noting this PR only changes **test** code, so it's very unlikely this PR is the culprit. I'll keep monitoring the lane.

PR #429, opened 2022-08-09
e2e: performance: configure isolated cpus in the profile

shajmakh:
Some tests in the suites require setting the number of CPUs to be used by the latency pod. If the available cpus is not sufficient, the tests will be skipped.
The goal of the suite is to verify that the main tests of the latency tools are executed as expected.
So to overcome this, update the isolated CPUs for this suite to CPU set that guarantees to execute the tests properly, and restore the initial set when finished.


shajmakh:
missing the part of skipping the suite, likely missed when splitting the commits 
marking as WIP

Tal-or:
LGTM

shajmakh:
/lgtm
trying to see who has permissions.

yanirq:
please squash the commits into one.

PR #431, opened 2022-08-10
WIP: feat: added support for generating workload partitioning files

eggfoobar:
Signed-off-by: ehila <ehila@redhat.com>


eggfoobar:
/hold
Waiting for pr openshift/api#1284

eggfoobar:
/assign @fromanirh 
/cc @haircommander @rphillips

Hey guys, would you be able to give a review on this PR in its current state while we wait for the API addition.

vendor/github.com/openshift/api/config/v1/0000_10_config-operator_01_infrastructure.crd.yaml
```diff
@@ -220,6 +220,13 @@ spec:
                     - HighlyAvailable
                     - SingleReplica
                     - External
+                cpuPartitioning:
+                  description: cpuPartitioning expresses that CPU partitioning is a currently enabled feature in the cluster.
+                  type: string
```
haircommander, line 225: should there be pattern recognition to filter out bogus options at API validation step? though, the pattern may be too copmlicated because there are a variable number of `,` and ranges allowed..

haircommander:
one question, otherwise LGTM

vendor/github.com/openshift/api/config/v1/0000_10_config-operator_01_infrastructure.crd.yaml
eggfoobar: That's a good point, I think David might have had the same input in the API PR for this, is this what you had in mind https://github.com/openshift/api/pull/1284#discussion_r971268312

haircommander: not quite, I was talking about validating that the cpuset is in a format we epxect like `0,2-4,7`  but come to think of it, the validation would not be trivial, so I am fine without.

manifests/20-performance-profile.crd.yaml
```diff
@@ -382,6 +382,15 @@ spec:
                     reserved:
                       description: Reserved defines a set of CPUs that will not be used for any container workloads initiated by kubelet.
                       type: string
+                    workloads:
```
MarSik, line 385: Why is this in the cpu section? All the other fields here accept cpu ids, not annotation names. I think this will confuse users.

pkg/apis/performanceprofile/v2/performanceprofile_types.go
```diff
@@ -110,8 +114,17 @@ type CPU struct {
 	// Offline defines a set of CPUs that will be unused and set offline
 	// +optional
 	Offlined *CPUSet `json:"offlined,omitempty"`
+	// Define workload partitioning configurations.
+	// The specified workloads annotation will be pinned to the defined "reserved" CPUs
+	// +listType=atomic
+	// +optional
+	Workloads []PinnedWorkloads `json:"workloads,omitempty"`
```
MarSik, line 121: Same as above. I do not think this belongs to the cpu section.

manifests/20-performance-profile.crd.yaml
```diff
@@ -382,6 +382,15 @@ spec:
...
+                      description: Define workload partitioning configurations. The specified workloads annotation will be pinned to the defined "reserved" CPUs
+                      type: array
+                      items:
+                        description: WorkloadAnnotation defines which workload annotation to pin
```
MarSik, line 389: Should you maybe document that only one value is supported atm?

MarSik (commented):
This is a huge change without a finished design. And I stumbled on it by accident. I understand the rationale, however we need to talk about some of the flows and design decisions much more. Especially since this is irreversible, will only work on certain clusters, changes stable API (v2), ... the user story here needs to be really really good before the code can be merged.

MarSik:
I commented more in the enhancement

manifests/20-performance-profile.crd.yaml
eggfoobar: Resolving: Changed approach no longer includes PA API Change

manifests/20-performance-profile.crd.yaml
eggfoobar: Resolving: Changed approach no longer includes PA API Change

manifests/40-rbac.yaml
```diff
@@ -66,7 +66,7 @@ rules:
   verbs: ["create","get","list","watch"]
 # Needed by every CVO-managed operator.
```
jmencak, line 67: This comment is no longer true based on the "infrastructures/status" addition, can we adjust it?

jmencak (commented):
I only looked at the integration with NTO (not at the PerformanceAddon controller pieces) and it looks good to me.  Only a minor nit.  We should get some feedback from the PerformanceAddon controller side.

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -258,6 +265,64 @@ func removePerformanceOLMOperator(cfg *rest.Config) error {
 	return nil
 }
 
+// During upgrade from 4.12 -> 4.13, this logic will update the authoritative flag in the new
+// Infrastructures.Status.CPUPartitioning to migrate Single Node clusters to the new method.
+// TODO: Revisit after 4.13 to remove logic when no longer needed.
+func migratePinnedSingleNodeInfraStatus(cfg *rest.Config, scheme *apiruntime.Scheme) error {
```
haircommander, line 271: should this PR be targeted at release-4.13 instead, as 4.13 and master have already branched? the master one could not have these pieces, and then the 4.13 one could, ensuring a correct transition.

assets/performanceprofile/configs/99-workload-partitioning.conf
```diff
@@ -0,0 +1,6 @@
+{{- if . }}
+[crio.runtime.workloads.{{ .Workload }}]
+activation_annotation = "target.workload.openshift.io/{{.Workload}}"
+annotation_prefix = "resources.workload.openshift.io"
+resources = { "cpushares" = 0, "cpuset" = "{{ .ReservedCpus}}" }
```
haircommander, line 5: The not adding of cpushares is not self documenting here. Not sure the best place to say it, but adding a blurb somewhere like
```
cpushares must be set to zero here. Not setting it causes the TOML to not be parsed by CRI-O. Further, it should not be configurable through the API, as the Kubelet will inject the correct cpu share annotations according to the pod spec, rom carried patches in https://github.com/openshift/kubernetes/pull/706
```
or something like that. Can you think of a good place to put such a note @eggfoobar

haircommander: I mention both of those because I forgot both of them and had to refresh my memory :)

manifests/40-rbac.yaml
eggfoobar: That's a good point, Since I think we want to revisit this permission, I moved it to it's own stanza and added a TODO for it.

cmd/cluster-node-tuning-operator/main.go
eggfoobar: Oh boy, I'm not sure actually @jmencak do you have any thoughts on this? I was under the impression the branching was happening next week Friday?

assets/performanceprofile/configs/99-workload-partitioning.conf
eggfoobar: Great idea, I added comments to the method that generates this files, as well as a golang template comment to help out any future developers that look at this file.

cmd/cluster-node-tuning-operator/main.go
jmencak: We already have release-4.13 and release-4.14 branches, **but**, they are still [frozen for merging](https://github.com/openshift/cluster-node-tuning-operator/issues/97) (i.e. read-only for us) and are [being fast-forwarded to from the master branch every hour](https://docs.ci.openshift.org/docs/architecture/branching/).  FWIU, what will happen on 3rd March unless the date is pushed out is that:

- we stop fast-forwarding master to release-4.13 branch ("feature-complete" == "feature-freeze" && "code-freeze" in 4.13, i.e. 0 days between ff & cf)
- release-4.14 branch opens for development (master will defacto become 4.14 -- it will continue being fast-forwarded to 4.14)

eggfoobar: That's great, we should be good then, resolving this.

haircommander:
LGTM

from CRI-O side

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -162,7 +190,35 @@ func render(inputDir, outputDir string) error {
...
 				return err
 			}
+			klog.Info(fileName)
+		}
+	}
+
+	return nil
+}
+
+func genCPUPartitioningManifests(pinnedMode *apicfgv1.CPUPartitioningMode, outputDir string, mcpNames ...string) error {
```
MarSik, line 200: This generates Workload partitioning manifests, right? I suggest a change of name here and maybe a short overall comment explaining the function and the use case.

assets/performanceprofile/configs/99-workload-partitioning.conf
```diff
@@ -0,0 +1,12 @@
+{{- /*
```
MarSik, line 1: I think it is time we add the intended destination for these configs if we can. This is a crio.conf.d file, right?

assets/performanceprofile/configs/openshift-workload-pinning
```diff
@@ -0,0 +1,7 @@
+{{- if . }}
```
MarSik, line 1: And this is kubernetes config file?

pkg/performanceprofile/cmd/render/render.go
MarSik: The name should probably make the configured component and generated object obvious. Is this method used for both crio and kubelet (253 and 260 in getIgnitionConfig?

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -437,3 +464,92 @@ func renderCrioConfigSnippet(profile *performancev2.PerformanceProfile, src stri
...
+
+	pinningConfigData := &bytes.Buffer{}
+	if err := profileTemplate.Execute(pinningConfigData, vars); err != nil {
+		return nil, err
+	}
+
+	return pinningConfigData.Bytes(), nil
+}
+
+func BootstrapWorkloadMC(role string, pinningMode *apiconfigv1.CPUPartitioningMode) (*machineconfigv1.MachineConfig, error) {
```
MarSik, line 504: Same here, why WorkloadMC? Should that be WorkloadPartitioningMC?

MarSik:
I am a bit confused by the  use of CPUPartitioning vs. WorkloadPartitioning. I know it comes from the API, but it mixes terms together.

Apart from that the overal logic looks sane enough, but the commit messages need cleaning up and I would prefer at least a higher level summary of the enhancement in this PR or even the commits.

assets/performanceprofile/configs/openshift-workload-pinning
eggfoobar: Right, this is the file we use as a flag to singal to kubelet

assets/performanceprofile/configs/99-workload-partitioning.conf
eggfoobar: Yup, this is a crio file, @MarSik do you mean that we should have this file in `assets/performanceprofile/configs/crio/` in the code, or are you suggesting we just add to the comment that we should mention this file will be created in `crio.conf.d`?

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Updated with comments, this method is the bootstrap method during render so it's the one that generates the default state, the regular `GetNewComponents` takes on the responsibility of generating the configs derived from normal operations when customers supply a PerformanceProfile.

assets/performanceprofile/configs/99-workload-partitioning.conf
eggfoobar: Added it as a comment to the file, just incase.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
eggfoobar: Good point, went and clarified it.

assets/performanceprofile/configs/99-workload-partitioning.conf
MarSik: Just a comment is enough for me.

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -162,7 +190,40 @@ func render(inputDir, outputDir string) error {
...
+	}
+
+	return nil
+}
+
+// genBootstrapWorkloadPinningManifests is used to generate the appropriate bootstrap workload pinning manifests
+// based on the cluster CPU Partitioning Mode and the given MachineConfigPool names.
+// The created manifests are the default state configs for the node which make no assumption for which CPU's are used for pinning.
+// This is done to ensure that when we generate the specific WorkloadPinning driven by the PerformanceProfile, we'll always have this
+// default state as a fall back if the user was to delete the PerformanceProfile by accident.
```
MarSik, line 204: I am not sure this will prevent accidents, because all objects generated here get ownershipReference that points to the PerfProfile. So deletion of the profile removes the MC as well (have you tested this scenario?).

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -437,3 +464,92 @@ func renderCrioConfigSnippet(profile *performancev2.PerformanceProfile, src stri
...
+// The template files listed below
+// [./assets/performanceprofile/configs/99-workload-pinning.conf] - CRI-O Config
+// [./assets/performanceprofile/configs/openshift-workload-pinning] - Kubelet Config
+//
+// Note:
+// The CRI-O Config for `cpushares` must be set to zero.
+// Not setting it causes the TOML to not be parsed by CRI-O which is why the template value is set to zero.
+// Further, it should not be configurable through the API, as the Kubelet will inject the correct cpu share annotations according to the pod spec.
+// Carried patches in https://github.com/openshift/kubernetes/pull/706
+func renderCPUPinningConfig(cpuv2 *performancev2.CPU, src string) ([]byte, error) {
```
MarSik, line 479: Just a suggestion to separate this from the regular cpu pinning terminology to not confuse readers.

```suggestion
func renderManagementCPUPinningConfig(cpuv2 *performancev2.CPU, src string) ([]byte, error) {
```

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Right, in this situation, these are the preliminary configs and do not belong to a PerformanceProfile, they're there to tell kubelet to correctly identify itself for CPUPinning

When the user creates a PerformanceProfile, that will generate an owned config that is tied to that performance profile which overrides the bootstrap configs. So when they delete that PerformanceProfile, the bootstrap config doesn't get touched and we revert back to the preliminary configs.

I have tested this behavior and it was consistent in reverting back the configs. What we might want to do in the future is decide if we want to introduce a static resource management component to the operator to enforce the preliminary configs more rigidly.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
eggfoobar: Sure thing, updated!

pkg/performanceprofile/cmd/render/render.go
eggfoobar: @MarSik Just to clarify, since these files are bootstrap files with the intention of being user defined later, we give them a lower priority in the MachineConfig file hierarchy. So for a CPU Pinned cluster we always generate these regardless if the user supplied a PerformanceProfile during bootstrapping or not, since the PerformanceProfile will generate it's own config that will override the bootstrap ones.

MarSik: Oh, so this MC does not have the ownership references. I think that is an important piece of information to be mentioned in the code. This file stays in the cluster forever, right?

eggfoobar: Right, exactly. I'll add that information I explained here to the comment above.

MarSik (approved):
/lgtm

I do not see anything significant to improve now, lets merge and iterate from here.

PR #432, opened 2022-08-16
convert core and cpu strings to integer

mrniranjan:
This is minor fix to the PR#403 where
only node was converted to string, Cpu and core
information from lscpu also needs to be
converted to integer.


marioferh (commented):
/lgtm
it was like that at the very begging of previous pr right?

mrniranjan:
> /lgtm it was like that at the very begging of previous pr right?

yes, while addressing reviews, I mistakenly removed this condition. which was required

PR #437, opened 2022-08-19
e2e: perfprof: crio mask fix workaround part 2

fromanirh:
In commit c5cf0bd33c5afae7695fd86d08c62aef269105d8 we added
a workaround for the incorrect crio mask padding (see:
https://github.com/cri-o/cri-o/issues/6145 )
But the fix implemented here is partial. Address the gap.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #436, opened 2022-08-19
podsecurity: disable OCP label sync

fromanirh:
Per last OCP recommendation (internal doc) is not sufficient to
declare the pod security labels, we also need to disable the
label sync, which we do in this PR.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #438, opened 2022-08-19
e2e: perfprof: unbreak the e2e-gcp PAO lane

fromanirh:
this PR combine both #436 and #437. We need _both_ to unbreak the lane, but either alone is not sufficient, and they are unrelated to each other


jmencak (commented):
Thank you Francesco,
/lgtm

MarSik (approved):

PR #445, opened 2022-08-24
Bug 2109965: low-latency-hook script optimization

Tal-or:
Measurements show that the script spends most of its time in entering/leaving the containers` network namespace.

This PR contains a couple of modifications that are trying to minimize entering the network namespace as much as possible.

A detailed explanation about each modification is listed in the commits themselves.

### DIFF
--- a/assets/performanceprofile/scripts/low-latency-hooks.sh
+++ b/assets/performanceprofile/scripts/low-latency-hooks.sh
@@ -2,6 +2,7 @@
 
 JQ="/usr/bin/jq"
 IP="/usr/sbin/ip"
+WORKDIR=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
 
 mask="${1}"
 [ -n "${mask}" ] || { logger "${0}: The rps-mask parameter is missing" ; exit 0; }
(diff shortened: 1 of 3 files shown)


assets/performanceprofile/scripts/low-latency-hooks.sh
```diff
@@ -1,5 +1,22 @@
 #!/usr/bin/env bash
 
+function rps_mask_set() {
+    local pid=$1
+    local mask=$2
+    local rps_files
+
+    rps_files=$(find /proc/"${pid}"/root/sys/devices/virtual -type f -name rps_cpus)
```
Tal-or, line 8: My main concern here is that this path is not modified for other containers (once the RPS mask of the network ns already set for other container in which share the same ns) and still represents the old rps mask.

So this function will always return `false` and the hook will run for each container in the pod and not only once as intended.

Tal-or:
/hold
WIP

assets/performanceprofile/scripts/low-latency-hooks.sh
```diff
@@ -14,7 +14,9 @@ ns=$(${IP} netns identify "${pid}" 2>&1)
 [[ $? -eq 0 && -n "${ns}" ]] || { logger "${0} Failed to identify the namespace: ${ns}"; exit 0; }
 
 # Updates the container veth RPS mask on the node
-netns_link_indexes=$(${IP} netns exec "${ns}" ${IP} -j link | ${JQ} ".[] | select(.link_index != null) | .link_index")
+jlink=$(${IP} netns exec "${ns}" ${IP} -j link)
```
pabeni, line 39: this could be:

jlink=$(${IP} -n "${ns}" -j link)

which should be faster

assets/performanceprofile/scripts/set-ns-rps-mask.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if ! [[ -w /sys ]]; then
+  if ! res=$(mount -o remount,rw /sys 2>&1); then
+    logger "${0}: Failed to remount /sys as rw: ${res}"; exit 0;
+  fi
+  REMOUNT="true"
+fi
+
+# /sys/class/net can't be used recursively to find the rps_cpus file, use /sys/devices/virtual instead
+if ! res=$(find /sys/devices/virtual -type f -name rps_cpus -exec sh -c "echo ${MASK} > {}" \; 2>&1) || [[ -n $res ]]; then
```
pabeni, line 14: for dev in /sys/devices/virtual/net/*; do
          echo ${mask} > ${dev}/queues/rx-0/rps_cpus || logger "${0}: Failed to apply the RPS mask on ${dev}:"
done

assets/performanceprofile/scripts/set-ns-rps-mask.sh
```diff
@@ -0,0 +1,23 @@
...
+  logger "${0}: Failed to apply the RPS mask: ${res}"
+fi
+
+if [[ ${REMOUNT} == "true" ]]; then
```
pabeni, line 18: I think this part is not needed, the running process is not sharing the mount namespace with the pod, and the pod's view of the /sys FS should still be RO. The running process is not going to touch /sys after this point, so no need to remount it

Tal-or: Keeping it as writable is ok?

Tal-or:
The PR looks good, tested manually and I haven't seen any regression.
@MarSik I think we're good to go here

Tal-or:
Closing. We came with a complete solution here:
https://github.com/openshift/cluster-node-tuning-operator/pull/479

PR #453, opened 2022-08-31
[WIP] PAO Skip rt parameters if rt kernel is not present

marioferh:
Applied RealTime parameters when rt kernel is disabled show different errors. 

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -29,6 +29,7 @@ min_perf_pct=100
 
 {{if .RealTime}}
 [service]
+#uname_regex=rt
 service.stalld=start,enable
 {{end}}
 
@@ -59,6 +60,7 @@ default_irq_smp_affinity = ignore
 {{end}}
 
 [sysctl]
+#uname_regex=rt
 {{if .RealTime}}
 #> cpu-partitioning #realtime
 kernel.hung_task_timeout_secs=600


assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -29,6 +29,7 @@ min_perf_pct=100
 
 {{if .RealTime}}
 [service]
+uname_regex=rt
```
MarSik, line 32: stalld is not RT kernel specific, it can be useful for any system with FIFO workloads

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -59,6 +60,7 @@ default_irq_smp_affinity = ignore
 {{end}}
 
 [sysctl]
+uname_regex=rt
```
MarSik, line 63: You would have to split the section here, not all the options are RT kernel specific. Moreover, this will not work, because we inherit from other profiles that have the same sysctls. This needs to be fixed on tuned level.

MarSik (commented):
/hold

I do not think this is a proper fix or at least not in the proper place.

assets/performanceprofile/tuned/openshift-node-performance
marioferh: I am getting an error when applied stalld enabled in a vm without rt kernel.

PR #460, opened 2022-09-07
Remove RPS configuration wihen realtime workload hint is explicitly set to false

yanirq:
When workload realtime hint is explicitly set to false there is no reason to set RPS realtime related configuration. In that case, the entire RPS configuration will not be set including systemd RPS device services and low latency crio hook.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -135,6 +135,7 @@ func GetMachineConfigName(profile *performancev2.PerformanceProfile) string {
 }
 
 func getIgnitionConfig(profile *performancev2.PerformanceProfile) (*igntypes.Config, error) {
+	var scripts []string
 	ignitionConfig := &igntypes.Config{
 		Ignition: igntypes.Ignition{
 			Version: defaultIgnitionVersion,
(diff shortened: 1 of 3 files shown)


test/e2e/performanceprofile/testdata/render-expected-output/manual_machineconfig.yaml
```diff
@@ -24,121 +24,121 @@ spec:
...
+          [Install]
+          WantedBy=multi-user.target
+        enabled: true
+        name: clear-irqbalance-banned-cpus.service
   extensions: null
   fips: false
   kernelArguments: null
   kernelType: realtime
-  osImageURL: ""
+  osImageURL: ""
```
marioferh, line 144: newline

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -145,8 +146,14 @@ func getIgnitionConfig(profile *performancev2.PerformanceProfile) (*igntypes.Con
 	}
 
 	// add script files under the node /usr/local/bin directory
+	if profile.Spec.WorkloadHints != nil && profile.Spec.WorkloadHints.RealTime != nil && !*profile.Spec.WorkloadHints.RealTime {
```
marioferh, line 149: Maybe is more readable compare with *profile.Spec.WorkloadHints.RealTime also because realTime Hint is default to true

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -164,28 +171,48 @@ func getIgnitionConfig(profile *performancev2.PerformanceProfile) (*igntypes.Con
...
-	var rpsRulesContent []byte
-	if profileutil.IsRpsEnabled(profile) {
-		rpsRulesContent, err = assets.Configs.ReadFile(filepath.Join("configs", udevPhysicalRpsRules))
-	} else {
-		rpsRulesContent, err = assets.Configs.ReadFile(filepath.Join("configs", udevRpsRules))
-	}
-	if err != nil {
-		return nil, err
+	// do not add RPS handling when realtime is explicitly disabled by workload hint
+	if profile.Spec.WorkloadHints == nil || profile.Spec.WorkloadHints.RealTime == nil || *profile.Spec.WorkloadHints.RealTime {
```
marioferh, line 175: Maybe do same if comparison as others

marioferh (commented):
/lgtm
Minor comments
There is a typo in the title: "wihen"

test/e2e/performanceprofile/testdata/render-expected-output/manual_machineconfig.yaml
yanirq: this passes the no cluster tests which generates and compare this file

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
yanirq: we first check if its nil then check the value

PR #461, opened 2022-09-08
Fix RPS e2e test condition

yanirq:
RPS e2e test never actually ran since the enter condition was wrong


mrniranjan:
Looks good to me

mrniranjan:
@yanirq We addressed this condition in another patch https://github.com/openshift/cluster-node-tuning-operator/pull/483

yanirq:
close in favor of https://github.com/openshift/cluster-node-tuning-operator/pull/483

PR #464, opened 2022-09-12
Add verification on kernel param rcutree.kthread

shaior:
Automates https://bugzilla.redhat.com/show_bug.cgi?id=2114681

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -1041,6 +1041,14 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			})
 		})
 	})
+
+	It("[test_id:54083] Should have kernel param rcutree.kthread", func() {
+		for _, node := range workerRTNodes {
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -1041,6 +1041,27 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 			})
 		})
 	})
+
+	It("[test_id:54083] Should have kernel param rcutree.kthread", func() {
```
marioferh, line 1045: I is feasible do it in the same way that the other ones?

`		It("[test_id:31198] Should set CPU affinity kernel argument", func() {
			for _, node := range workerRTNodes {
				cmdline, err := nodes.ExecCommandOnMachineConfigDaemon(&node, []string{"cat", "/proc/cmdline"})
				// since systemd.cpu_affinity is calculated on node level using tuned we can check only the key in this context.
				Expect(string(cmdline)).To(ContainSubstring("systemd.cpu_affinity="))
			}
		})`

shaior: Thanks for the comment, fixed.

marioferh (commented):
/lgtm

PR #468, opened 2022-09-13
workloadHints not included in validateFields()

marioferh:
workloadHints are not included in validateFields()
Thus, are not applying correctly. 
Also validateFields is not being testing.

- [X]  Add validateWorkloadHints to function.
- [X]  Add tests.

Related: https://github.com/openshift/cluster-node-tuning-operator/issues/339

### DIFF
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
@@ -380,6 +380,61 @@ var _ = Describe("PerformanceProfile", func() {
 			})
 		})
 	})
+
+	Describe("validation of validateFields function", func() {
+		It("should check all fields", func() {
+			// config all specs to rise an error in every func inside validateFields()
+			reservedCPUs := CPUSet("")
+			isolatedCPUs := CPUSet("0-6")
+			offlinedCPUs := CPUSet("7")
+			profile.Spec.CPU.Reserved = &reservedCPUs
+			profile.Spec.CPU.Isolated = &isolatedCPUs
+			profile.Spec.CPU.Offlined = &offlinedCPUs
+
+			profile.Spec.MachineConfigLabel["foo"] = "bar"
+
+			incorrectDefaultSize := HugePageSize("!#@")
+			profile.Spec.HugePages.DefaultHugePagesSize = &incorrectDefaultSize
+
+			profile.Spec.WorkloadHints = &WorkloadHints{
+				RealTime: pointer.BoolPtr(false),
+			}
+			profile.Spec.RealTimeKernel = &RealTimeKernel{
+				Enabled: pointer.BoolPtr(true),
+			}
+
+			invalidVendor := "123"
+			invalidDevice := "0x12345"
+			profile.Spec.Net.Devices[0].InterfaceName = pointer.StringPtr("")
+			profile.Spec.Net.Devices[0].VendorID = pointer.StringPtr(invalidVendor)
+			profile.Spec.Net.Devices[0].DeviceID = pointer.StringPtr(invalidDevice)
+
+			errors := profile.validateFields()
+
+			type void struct{}
+			var member void
+			errorMsgs := make(map[string]void)
+
+			errorMsgs["reserved CPUs can not be empty"] = member
+			errorMsgs["you should provide only 1 MachineConfigLabel"] = member
+			errorMsgs[`hugepages default size should be equal to "1G" or "2M"`] = member
+			errorMsgs["device name cannot be empty"] = member
+			errorMsgs[fmt.Sprintf("device vendor ID %s has an invalid format. Vendor ID should be represented as 0x<4 hexadecimal digits> (16 bit representation)", invalidVendor)] = member
+			errorMsgs[fmt.Sprintf("device model ID %s has an invalid format. Model ID should be represented as 0x<4 hexadecimal digits> (16 bit representation)", invalidDevice)] = member
+			errorMsgs["realtime kernel is enabled, but realtime workload hint is explicitly disable"] = member
+
+			for _, err := range errors {
+				_, exists := errorMsgs[err.Detail]
+				if exists {
+					delete(errorMsgs, err.Detail)
+				}
+			}
+
+			for errorMsg, ok := range errorMsgs {
+				Expect(ok).To(BeTrue(), "missing expected error: %q", errorMsg)
+			}
+		})
+	})
 })
 
 func setValidNodeSelector(profile *PerformanceProfile) {
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
@@ -108,6 +108,7 @@ func (r *PerformanceProfile) validateFields() field.ErrorList {
 	allErrs = append(allErrs, r.validateHugePages()...)
 	allErrs = append(allErrs, r.validateNUMA()...)
 	allErrs = append(allErrs, r.validateNet()...)
+	allErrs = append(allErrs, r.validateWorkloadHints()...)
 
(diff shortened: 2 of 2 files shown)


marioferh (commented):
/hold
Needed a test for validate function

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -369,6 +369,61 @@ var _ = Describe("PerformanceProfile", func() {
...
+			//check validateNet errors
+			Expect(errors[3].Error()).To(ContainSubstring(fmt.Sprintf("device name cannot be empty")))
+			Expect(errors[4].Error()).To(ContainSubstring(fmt.Sprintf("device vendor ID %s has an invalid format. Vendor ID should be represented as 0x<4 hexadecimal digits> (16 bit representation)", invalidVendor)))
+			Expect(errors[5].Error()).To(ContainSubstring(fmt.Sprintf("device model ID %s has an invalid format. Model ID should be represented as 0x<4 hexadecimal digits> (16 bit representation)", invalidDevice)))
+
+			// check validateWorkloadHints erros
+			Expect(errors).NotTo(BeEmpty())
+			Expect(errors[6].Error()).To(ContainSubstring("realtime kernel is enabled, but realtime workload hint is explicitly disable"))
+
+			Expect(len(errors)).To(Equal(7))
```
fromanirh, line 424: this is fragile, and so is expecting the error to happen on a specific order, unless the number of errors and their ordering is part of the API contract (and it should not, because it leaks unnecessary implementation details).
It is better to just look for the errors we expect in the error list, regardless of the order and of the total count.
Arguably, also checking for the error string is fragile, but this is yet another story.

fromanirh (commented):
/approve
good enough for now, let's iterate later

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
marioferh: You are right, I'll add an issue to track it.

MarSik: Mario, you are adding validation of workload hints, but the test here is actually not checking that part.

MarSik: (You have the other PR where some combination of workload hints is prohibited)

MarSik: Ok, NVM I totally missed the line 422.

marioferh: :D

fromanirh:
do we have the option to remove the dependency on the ordering of errors now?

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -369,6 +369,62 @@ var _ = Describe("PerformanceProfile", func() {
...
+
+			invalidVendor := "123"
+			invalidDevice := "0x12345"
+			profile.Spec.Net.Devices[0].InterfaceName = pointer.StringPtr("")
+			profile.Spec.Net.Devices[0].VendorID = pointer.StringPtr(invalidVendor)
+			profile.Spec.Net.Devices[0].DeviceID = pointer.StringPtr(invalidDevice)
+
+			errors := profile.validateFields()
+
+			errorMsgs := map[string]bool{}
```
fromanirh, line 403: this works. The golang idiom to implement a set is to use `struct{}` as element because it enables better compiler optimizations, possibly avoiding to use any space for the value element.
Alternatively, use `https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets`

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -369,6 +369,62 @@ var _ = Describe("PerformanceProfile", func() {
...
+			}
+
+			testIsCorrect := true
+			for _, element := range errorMsgs {
+				if !element {
+					testIsCorrect = false
+				}
+			}
+
+			Expect(testIsCorrect).To(BeTrue())
```
fromanirh, line 425: let's have an annotated assetion to make it obvious which error is missing. We can do in many ways. One of them is
```
for errorMsg, ok := range errorMsgs {
    Expect(ok).To(BeTrue(), "missing expected error: %q", errorMsg)
}
```

fromanirh (commented):
thanks, good improvement. Let's improve the assertion and we're ready.

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
marioferh: Done with set

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -380,6 +380,61 @@ var _ = Describe("PerformanceProfile", func() {
...
+
+			errorMsgs["reserved CPUs can not be empty"] = member
+			errorMsgs["you should provide only 1 MachineConfigLabel"] = member
+			errorMsgs[`hugepages default size should be equal to "1G" or "2M"`] = member
+			errorMsgs["device name cannot be empty"] = member
+			errorMsgs[fmt.Sprintf("device vendor ID %s has an invalid format. Vendor ID should be represented as 0x<4 hexadecimal digits> (16 bit representation)", invalidVendor)] = member
+			errorMsgs[fmt.Sprintf("device model ID %s has an invalid format. Model ID should be represented as 0x<4 hexadecimal digits> (16 bit representation)", invalidDevice)] = member
+			errorMsgs["realtime kernel is enabled, but realtime workload hint is explicitly disable"] = member
+
+			for _, err := range errors {
```
marioferh, line 426: maybe it is not the best way to do it.
Also we are not checking possible duplicates...

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -380,6 +380,61 @@ var _ = Describe("PerformanceProfile", func() {
...
+				_, exists := errorMsgs[err.Detail]
+				if exists {
+					delete(errorMsgs, err.Detail)
+				}
+			}
+
+			for errorMsg, ok := range errorMsgs {
```
marioferh, line 433: here it only shows firs error, we can accumulate the errors but I think does not worth.

fromanirh: showing all of them would be nicer, but this is good enough.

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
fromanirh: yep, but to really improve we need to treat errors as values, which is indeed a bigger change. LGTM for now.

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -380,6 +380,61 @@ var _ = Describe("PerformanceProfile", func() {
...
+			invalidDevice := "0x12345"
+			profile.Spec.Net.Devices[0].InterfaceName = pointer.StringPtr("")
+			profile.Spec.Net.Devices[0].VendorID = pointer.StringPtr(invalidVendor)
+			profile.Spec.Net.Devices[0].DeviceID = pointer.StringPtr(invalidDevice)
+
+			errors := profile.validateFields()
+
+			type void struct{}
+			var member void
+			errorMsgs := make(map[string]void)
```
fromanirh, line 416: a bit clunky, but good enough for now.

fromanirh (commented):
/approve
/lgtm

PR #471, opened 2022-09-15
e2e: Fix metrics_cert_rotation.go

jmencak:
Fix `metrics_cert_rotation.go` on systems where the host OS libraries do not match container libraries.  An alternative is to rely on container-shipped `openssl`, but we probably do not wish to add this dependency to the container image.

Other fixes:
  - insufficient number of arguments in `test/e2e/util/util.go`

/cc @dagrayvid


jmencak:
https://github.com/openshift/release/pull/32317 would have caught one of the issues this change is fixing now.

PR #474, opened 2022-09-21
Clean output of kernel args if no hugepages or additional args are present

marioferh:
If hugepages or additional arguments cmdline is no present in the perfromanceProfile, the ouptut of /proc/cmdline is dirty with "+" as:

Before fix:
[...]
rcutree.kthread_prio= 11 + + intel_pstate=passive

After fix:
[...]
rcutree.kthread_prio= 11 intel_pstate=passive


fromanirh (commented):
/lgtm

PR #475, opened 2022-09-23
Set default RealTime hint to true and add to conversion

marioferh:
Default value of RealTime hit is enabled.

- [X] RealTimeHint its not set to correct value by default
- [X] Add changes in logic
- [X] Add test to check the behavior

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
@@ -168,19 +168,17 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 		}
 	}
 
-	if profile.Spec.WorkloadHints != nil {
-		if IsRealTimeHintEnabled(profile) {
-			templateArgs[templateRealTimeHint] = "true"
-		}
+	if IsRealTimeHintEnabled(profile) {
+		templateArgs[templateRealTimeHint] = "true"
+	}
 
-		if IsHighPowerConsumptionHintEnabled(profile) && IsPerPodPowerManagementEnabled(profile) {
-			err := fmt.Errorf("Invalid WorkloadHints configuration: HighPowerConsumption is %t and PerPodPowerManagement is %t", *profile.Spec.WorkloadHints.HighPowerConsumption, *profile.Spec.WorkloadHints.PerPodPowerManagement)
-			return nil, err
-		} else if IsHighPowerConsumptionHintEnabled(profile) {
-			templateArgs[templateHighPowerConsumption] = "true"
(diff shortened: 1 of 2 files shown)


pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -168,9 +168,11 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 		}
 	}
 
+	templateArgs[templateRealTimeHint] = "true"
```
marioferh, line 171: should I add a comment about the reason to set it to true or it's enough with the api documentation?

fromanirh: it's sufficient to explain the rationale in the commit message.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -224,8 +226,8 @@ func IsIRQBalancingGloballyDisabled(profile *performancev2.PerformanceProfile) b
 	return profile.Spec.GloballyDisableIrqLoadBalancing != nil && *profile.Spec.GloballyDisableIrqLoadBalancing
 }
 
-func IsRealTimeHintEnabled(profile *performancev2.PerformanceProfile) bool {
-	return profile.Spec.WorkloadHints.RealTime != nil && *profile.Spec.WorkloadHints.RealTime
+func IsRealTimeHintDisabled(profile *performancev2.PerformanceProfile) bool {
+	return profile.Spec.WorkloadHints.RealTime != nil && !*profile.Spec.WorkloadHints.RealTime
```
MarSik, line 230: Why not like this?

func IsRealTimeHintEnabled
    return profile.Spec.WorkloadHints == nil || profile.Spec.WorkloadHints.RealTime == nil || *profile.Spec.WorkloadHints.RealTime

templateArgs[templateRealTimeHint] = IsRealtimeHintEnabled()

MarSik: I think it would be more readable than set default / delete when disabled

marioferh: The issue es I check before if workloadhints == nil in 173:
`if profile.Spec.WorkloadHints != nil {`
Then if is nil the func isRealTimeHintEnabled is not reached

MarSik: I know, but it does not have to stay this way. My version goes out of the "global" if and makes the logic explicit and readable.

You read the code more often than write it, so optimize for readability :) At least that is what I would do.

MarSik (approved):
/lgtm

PR #476, opened 2022-09-28
Adjust nto_custom_profiles:count metric to count by _id

dagrayvid:
This is needed in HyperShift management clusters when HyperShift is configured to have hosted control plane metrics scraped by the platform monitoring prometheus. In this case, metrics from all NTOs in all hosted control plane namespaces will be summed into one count. We want one count per cluster. This has no impact on standalone OCP.


jmencak (approved):
Untested, but makes sense to me.  Thank you, David.
/lgtm

PR #478, opened 2022-10-05
Remove omitempty on reserved CPU API v1

marioferh:
Remove omitempty on reserved CPU also on API v1

Related: https://github.com/openshift-kni/performance-addon-operators/pull/792


yanirq:
I am not sure this is a change we need unless I am missing something.
V2 is the version to use and V1 is kept for backward compatibility reasons.
There are also conversion hooks from v1 to v2 that might need to be changed accordingly if this change is in.

Is there a specific reason for this update on V1 ?

PR #479, opened 2022-10-06
OCPBUGSM-47141: set RPS for veth only at host level 

Tal-or:
RPS handling on the pod container level using crio-hooks causes long delay times when running the low latency script to set the RPS mask.

For RAN low latency solution it's sufficient to set  RPS only at the host level and avoid setting it inside the container while utilizing RSS behavior.

In the past, a low latency hook was added with additional RPS settings on virtual devices since there was an issue where the start and shutdown of a big amount of pods will initiate the creation of a systemd service that should update the new interfaces rps_cpus mask and can create an additional CPU load under the cluster (https://github.com/openshift-kni/performance-addon-operators/pull/659)
This is not the case anymore thus we may revert the aforementioned hook.

Co-authored-by: Yanir Quinn <yquinn@redhat.com>


assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -33,4 +33,4 @@ function find_dev_dir {
 [ -d "${dev_dir}" ] || { sleep 5; find_dev_dir; }  # search failed, wait a little and try again
 [ -d "${dev_dir}" ] || { echo "${dev_dir}" directory not found >&2 ; exit 0; } # the interface disappeared, not an error
 
-find "${dev_dir}"/queues -type f -name rps_cpus -exec sh -c "echo ${mask} | cat > {}" \;
\ No newline at end of file
+echo "${mask}" > "${dev_dir}"/queues/rx-0/rps_cpus
```
fromanirh, line 36: why only the queue number 0?

browsell:
LGTM

assets/performanceprofile/scripts/set-rps-mask.sh
fromanirh: actually answered in the commit message :)

fromanirh:
the change per se makes sense, but this is a delicate area. It would be nice to know: 1. how this was tested, especially to avoid regressions and 2. do we need or want a override mechanism to restore this functionalty.
IOW, what's our contingency plan should we find a regression later? issue a hotfix with a partial revert? What our options are here?

MarSik:
I think the commit message should better explain not what we do to lower the performance hit of the hook, but also why it is a correct change. Explanation about why RPS needs to be set for veth 2 in  Pod - veth1 <-> veth2 - Host scenario would be enough. (eg. what cpus pay for the processing, which side RPS is consulted, which side cares about latency etc..)

PR #480, opened 2022-10-07
OCPBUGS-2167: Workload hints backwards compatibility intel_pstate=disable

marioferh:
## Description

Maintain backwards compatibility after workloadHints change https://github.com/openshift/cluster-node-tuning-operator/pull/339

## Motivation and Context

The workloads hints feature introduced in 4.11 does not preserve the default tuning settings from 4.10: intel_pstate=disabled.

So if the customer wants to preserve the behavior they have in 4.10, they have to update their performance profile and incur another reboot.

The correct behavior the default setting should be preserved.

## Type of change

Add intel_pstate=disable to default configuration with no hints

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -165,8 +165,6 @@ cmdline_additionalArg=+{{.AdditionalArgs}}
 
 {{if .PerPodPowerManagement}}
 cmdline_pstate=+intel_pstate=passive
-{{end}}
-
-{{if .HighPowerConsumption}}
+{{else if .RealTimeHint}}
 cmdline_pstate=+intel_pstate=disable
 {{end}}
--- a/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
@@ -204,9 +204,10 @@ var _ = Describe("Tuned", func() {
 			})
 		})
 
-		When("perPodPowerManagement Hint is false", func() {
+		When("perPodPowerManagement Hint is false realTime Hint false", func() {
 			It("should not contain perPodPowerManagement related parameters", func() {
 				profile.Spec.WorkloadHints.PerPodPowerManagement = pointer.BoolPtr(false)
+				profile.Spec.WorkloadHints.RealTime = pointer.BoolPtr(false)
 				tunedData := getTunedStructuredData(profile)
 				cpuSection, err := tunedData.GetSection("cpu")
 				Expect(err).ToNot(HaveOccurred())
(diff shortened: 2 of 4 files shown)


pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -214,7 +214,7 @@ var _ = Describe("Tuned", func() {
 				bootLoaderSection, err := tunedData.GetSection("bootloader")
 				Expect(err).ToNot(HaveOccurred())
 				Expect(bootLoaderSection.Key("cmdline_pstate").String()).ToNot(Equal(cmdlinePerPodPowerManagementHint))
-				Expect(bootLoaderSection.Key("cmdline_pstate").String()).ToNot(Equal(cmdlineHighPowerConsumptionPstate))
+				Expect(bootLoaderSection.Key("cmdline_pstate").String()).To(Equal(cmdlineHighPowerConsumptionPstate))
```
marioferh, line 217: these tests have change, maybe it worth to add some more

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -661,7 +661,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 					"kernel.sched_rt_runtime_us":    "950000",
 					"vm.stat_interval":              "10",
 				}
```
MarSik, line 663: Just checking: is this an exhaustive list? Because intel_pstate=disable should be present for high power consumption case. As should idle=poll iirc.

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -165,8 +165,8 @@ cmdline_additionalArg=+{{.AdditionalArgs}}
 
 {{if .PerPodPowerManagement}}
 cmdline_pstate=+intel_pstate=passive
-{{end}}
-
-{{if .HighPowerConsumption}}
+{{else}}
+{{if .RealTimeHint}}
```
MarSik, line 169: I am not entirely happy with how this nested condition looks like, I will not block on it, but I would prefer something more readable if possible.

marioferh: done, I didn't realize that it was else if in go templates

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
bartwensley: How did this testcase pass before you made this change? Prior to your change, it seems like "intel_pstate=disable" would not have been present in this case, but it looks like this code is checking for it to be there? Or am I misunderstanding how this testcase works?

marioferh: It was working because before this change template was:
`{{if .HighPowerConsumption}}
{{end}}`

Now is:
`{{if .PerPodPowerManagement}}
{{else if .RealTimeHint}}
{{end}}
`

bartwensley: Right - thanks Mario.

MarSik: Oh, right. Then I just wonder.. do we even allow high power consumption = true with realtime = false? This test might be code correct, but semantically broken if we do not.

marioferh: Well, it is allowed to do that, maybe show a warning?

fromanirh (commented):
/approve
the context is very very terse, so I need to defer the LGTM to someone which has more information available.

fromanirh:
/lgtm
/hold

fromanirh:
thanks for the extra context. Next time please update the commit message (thus reupload the change) *and* the PR description, not just the latter.
This is because git history should be self sufficient; if we need to go back on github to get the context, well, the git history is not self sufficient :)

MarSik:
Yeah, this should be in the commit message too.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
MarSik: Nope, fine for now, we can rethink that separately.

fromanirh:
/lgtm
thanks for updating the commit message

PR #481, opened 2022-10-07
HyperShift: avoid unnecessary mcfg creation

dagrayvid:
This PR has two changes to be more careful about when MachineConfigs are created by NTO when running in HyperShift:
* In HyperShift never create MachineConfigs for default profile. Previously this would only happen if an admin user on the hosted cluster manually edits the profile to add status.bootcmdline, but with this PR it should never happen.
* In HyperShift never create or update MachineConfigs when Profile.generation is greater than generation when last updated by the operator + 1.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -652,6 +652,7 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 			profileMf.Spec.Config.Debug = operand.Debug
 			profileMf.Spec.Config.TuneDConfig = operand.TuneDConfig
 			profileMf.Status.Conditions = tunedpkg.InitializeStatusConditions()
+			c.pc.state.profileLastUpdatedGeneration[profileMf.Name] = 1
 			_, err = c.clients.Tuned.TunedV1().Profiles(ntoconfig.WatchNamespace()).Create(context.TODO(), profileMf, metav1.CreateOptions{})
 			if err != nil {
 				return fmt.Errorf("failed to create Profile %s: %v", profileMf.Name, err)
@@ -676,19 +677,27 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 		return fmt.Errorf("failed to get ProviderName: %v", err)
 	}
 
+	// profileLastUpdatedGeneration is set for each Profile when it is created or updated.
+	// In the case where it is not already set due to an operator restart, set it and
+	// clear profile.Status.Bootcmdline instead of syncing MachineConfigs.
+	_, lastUpdateGenSet := c.pc.state.profileLastUpdatedGeneration[profile.Name]
+
 	if ntoconfig.InHyperShift() {
 		// nodePoolName is the name of the NodePool which the Node corresponding to this Profile
 		// is a part of. If nodePoolName is the empty string, it either means that Node label
 		// based matching was used, or we don't know the NodePool, so we should not sync the
 		// MachineConfigs.
-		if nodePoolName != "" {
-			// In HyperShift
-			if profile.Status.TunedProfile == tunedProfileName && profileApplied(profile) {
-				// Synchronize MachineConfig only once the (calculated) TuneD profile 'tunedProfileName'
-				// has been successfully applied.
-				err := c.syncMachineConfigHyperShift(nodePoolName, profile)
-				if err != nil {
-					return fmt.Errorf("failed to update Profile %s: %v", profile.Name, err)
+		if nodePoolName != "" && lastUpdateGenSet {
+			if profile.Generation > c.pc.state.profileLastUpdatedGeneration[profile.Name]+1 {
+				klog.Infof("refusing to sync MachineConfig because Profile %s is more than 1 generation ahead of last update by operator", profile.Name)
+			} else {
+				if profile.Status.TunedProfile == tunedProfileName && profileApplied(profile) {
+					// Synchronize MachineConfig only once the (calculated) TuneD profile 'tunedProfileName'
+					// has been successfully applied.
+					err := c.syncMachineConfigHyperShift(nodePoolName, profile)
+					if err != nil {
+						return fmt.Errorf("failed to update Profile %s: %v", profile.Name, err)
+					}
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/operator/profilecalculator.go
```diff
@@ -254,9 +258,12 @@ func (pc *ProfileCalculator) calculateProfileHyperShift(nodeName string) (string
...
 
 		// If recommend.Match is empty, NodePool based matching is assumed
-		// or this is the default profile
 		if recommend.Match == nil {
-			klog.V(3).Infof("calculateProfileHyperShift: NodePool based matching or default profile used for node: %s, tunedProfileName:  %s, nodePoolName: %s", nodeName, *recommend.Profile, nodePoolName)
+			if *recommend.Profile == defaultProfile {
+				// Don't set nodepool for default profile, no MachineConfigs should be generated.
+				return *recommend.Profile, "", recommend.Operand, nil
+			}
+			klog.V(3).Infof("calculateProfileHyperShift: NodePool based matching used for node: %s, tunedProfileName:  %s, nodePoolName: %s", nodeName, *recommend.Profile, nodePoolName)
```
jmencak, line 266: Nit: `s/tunedProfileName:  %s/tunedProfileName: %s/`  # Two spaces vs 1

jmencak (commented):
Thank you David, the changes look good to me and I also briefly tested them.  Apart from the nit, you may want to initialize `pc.state.profileLastUpdatedGeneration` in `syncProfile()` based on the real `profile.Generation` .  The reason for this is the valid updates would not be missed in the corner case when the operator restarts for some reason in the middle of updates, but it would mean a little more complexity and slightly higher chance for the "confused" admin to get his change through.  The code change may not be worth it, but at least something to be aware of.

dagrayvid:
> Apart from the nit, you may want to initialize `pc.state.profileLastUpdatedGeneration` in `syncProfile()` based on the real `profile.Generation`

This is a good point. I think in the case where NTO restarts `pc.state.profileLastUpdatedGeneration` won't be initialized at all. We should initialize it if it's not already set, in syncProfile() before calling `syncMachineConfigHyperShift()`.
```
if _, ok := c.pc.state.profileLastUpdatedGeneration[profile.Name]; !ok {
    c.pc.state.profileLastUpdatedGeneration[profile.Name] = profile.Generation + 1
}
```

In my view, this doesn't really increase the "attack surface" as you would need access to the management cluster in order to make the operator restart.

jmencak:
> In my view, this doesn't really increase the "attack surface" as you would need access to the management cluster in order to make the operator restart.

The operator can restart on its own for various reasons, but I agree there are more advantages of initialization than disadvantages.

pkg/operator/controller.go
```diff
@@ -676,6 +677,11 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 		return fmt.Errorf("failed to get ProviderName: %v", err)
 	}
 
+	// Won't be set in case of operator restart
```
jmencak, line 680: Can we improve the terse comment a bit?  I know it is obvious what won't be set, but perhaps we want to be more explicit?

pkg/operator/controller.go
```diff
@@ -676,6 +677,11 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
...
+	if _, ok := c.pc.state.profileLastUpdatedGeneration[profile.Name]; !ok {
+		c.pc.state.profileLastUpdatedGeneration[profile.Name] = profile.Generation + 1
```
jmencak, line 682: I didn't test, but are you sure we want to initialize to `profile.Generation + 1`, not just `profile.Generation` ?

jmencak (commented):
Thank you for the update, David!  I believe it needs a little bit of extra work.

pkg/operator/controller.go
dagrayvid: I think you're right, good catch. Still testing this...

dagrayvid:
> you may want to initialize pc.state.profileLastUpdatedGeneration in syncProfile() based on the real profile.Generation
@jmencak, I added an additional commit to expand on this. 

Without this commit, if the operator restarts the profileLastUpdatedGeneration would be initialized in syncProfile() to the current generation which means that in the absence of any other updates to the Profile, the next change to the profile.status.bootcmdline would be accepted by the operator as valid, even if it came from a confused admin editing the Profile directly.

With this commit, if c.pc.state.profileLastUpdatedGeneration[profile.Name] is not set, the operator will clear the status.bootcmdline and update the Profile, triggering the operand to resync / set the bootcmdline to the correct value, which will then trigger the operator to reconcile it one extra time. The tradeoff is slightly more complexity, but I think it's probably worth it. 

Let me know what you think about this, and if you think it's a good change I will squash the commits.

pkg/operator/controller.go
```diff
@@ -676,19 +677,28 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
...
-				// Synchronize MachineConfig only once the (calculated) TuneD profile 'tunedProfileName'
-				// has been successfully applied.
-				err := c.syncMachineConfigHyperShift(nodePoolName, profile)
-				if err != nil {
-					return fmt.Errorf("failed to update Profile %s: %v", profile.Name, err)
+		if nodePoolName != "" && lastUpdateGenSet {
+			if profile.Generation > c.pc.state.profileLastUpdatedGeneration[profile.Name]+1 {
+				klog.Infof("refusing to sync MachineConfig because Profile %s is more than 1 generation ahead of last update by operator", profile.Name)
+			} else {
+				// In HyperShift
```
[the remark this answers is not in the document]
dagrayvid: Thanks, I removed this comment. I think it was there from when the conditions were reversed, but it's pretty redundant to say 
```
if ntoconfig.InHyperShift() {
    // In Hypershift
    ...
```

pkg/operator/controller.go
```diff
@@ -721,8 +735,12 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 	profile.Spec.Config.TuneDConfig = operand.TuneDConfig
 	profile.Spec.Config.ProviderName = providerName
 	profile.Status.Conditions = tunedpkg.InitializeStatusConditions()
+	if !lastUpdateGenSet {
+		profile.Status.Bootcmdline = ""
```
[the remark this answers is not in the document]
dagrayvid: You're right, calling Update() is enough to trigger the operand to sync, we don't need to clear the status.

dagrayvid:
@jmencak I think this PR is ready to go, feel free to leave a lgtm if you agree, and I will unhold when I'm done testing.

pkg/operator/controller.go
```diff
@@ -708,10 +717,14 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 		}
 	}
 
+	// Only update Profile if the spec needs to be changed, or !lastUpdateGenSet,
+	// which means we need to clear the profile.Status.Bootcmdline and wait for
```
jmencak, line 721: Can we fix this comment since we are no longer clearing `profile.Status.Bootcmdline`?  After that, I believe this should be ready to go.

dagrayvid: Sorry about that, done now.

jmencak:
Thank you for the changes, David!
/lgtm

PR #482, opened 2022-10-12
udev-rule: rationale for applying the rule on virtual interfaces

Tal-or:
…interfaces

We should apply the rule on both virtual and physical interfaces of the host, because the RPS mask that would be consulted, is the one on the RX side of the veth in the host. Consider the following diagram:
Pod A <veth1 - veth2> host <veth3 - veth4> Pod B
veth2's RPS affinity is the one determining the CPUs that are handling the packet processing when sending data from Pod A to pod B.

Additional common scenarios:
1. Pod A = sender, host = receiver The RPS affinity of the host side should be consulted (because it’s the receiver) and it should be the RPS of the physical NIC.

2. Pod A = receiver, host = sender In case of no RPS mask on the receiver side, the sender needs to pay the price and do all the processing on its cores.


assets/performanceprofile/configs/99-netdev-rps.rules
```diff
@@ -1 +1,12 @@
+# We should apply the rule on both virtual and physical interfaces of the host, because the RPS mask that would be consulted,
```
MarSik, line 1: Why physical devices? The traffic for physical devices is handled via RSS and real interrupts and we already control the affinity for those. RPS is only needed for host side veth devices to fix the pod to host traffic.

assets/performanceprofile/configs/99-netdev-rps.rules
```diff
@@ -1 +1,12 @@
...
+# Consider the following diagram:
+# Pod A <veth1 - veth2> host <veth3 - veth4> Pod B
+#  veth2's RPS affinity is the one determining the CPUs that are handling the packet processing when sending data from Pod A to pod B.
+# Additional common scenarios:
+# 1. Pod A = sender, host = receiver
+#  The RPS affinity of the host side should be consulted (because it’s the receiver) and it should be the RPS of the physical NIC.
+#
+# 2. Pod A = receiver, host = sender
+#  In case of no RPS mask on the receiver side, the sender needs to pay the price and do all the processing on its cores.
 SUBSYSTEM=="net", ACTION=="add", ENV{ID_BUS}!="pci", TAG+="systemd", ENV{SYSTEMD_WANTS}="update-rps@%k.service"
```
MarSik, line 12: Btw, notice here - we specifically remove the physical devices from the rule.

Tal-or: The `!="pci"` part?

MarSik: Yes

MarSik (approved):
/lgtm

Tal-or:
@MarSik We need to override the upgrade lane again and have your final approval

MarSik (approved):
/lgtm

MarSik:
/override ci/prow/e2e-upgrade
adding exception because trivial documentation PR and the test failures are not related at all.

PR #483, opened 2022-10-13
OCPBUGS-3324: RPS mask test fixes

mrniranjan:
Fixed the if condition which returned if the spec.CPU.Reserved was not nil Added Test case ids

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -16,7 +16,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	nodev1 "k8s.io/api/node/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
@@ -291,22 +290,31 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 	})
 
 	Context("RPS configuration", func() {
-		It("Should have the correct RPS configuration", func() {
-			if profile.Spec.CPU == nil || profile.Spec.CPU.Reserved != nil {
-				return
+		It("[test_id:55012] Should have the correct RPS configuration", func() {
+			if profile.Spec.CPU == nil || profile.Spec.CPU.Reserved == nil {
+				Skip("Test Skipped due nil Reserved cpus")
 			}
 			if profile.Spec.WorkloadHints != nil && profile.Spec.WorkloadHints.RealTime != nil &&
 				!*profile.Spec.WorkloadHints.RealTime && !profileutil.IsRpsEnabled(profile) {
-				return
+				Skip("realTime Workload Hints is not enabled")
 			}
 
 			expectedRPSCPUs, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
 			Expect(err).ToNot(HaveOccurred())
 			for _, node := range workerRTNodes {
 				// Verify the systemd RPS service uses the correct RPS mask
-				cmd := []string{"sed", "-n", "s/^ExecStart=.*echo \\([A-Fa-f0-9]*\\) .*/\\1/p", "/rootfs/etc/systemd/system/update-rps@.service"}
-				serviceRPSCPUs, err := nodes.ExecCommandOnNode(cmd, &node)
+				var maskContent string
+				cmd := []string{"cat", "/rootfs/etc/systemd/system/update-rps@.service"}
+				unitFileContents, err := nodes.ExecCommandOnNode(cmd, &node)
 				Expect(err).ToNot(HaveOccurred())
+				for _, line := range strings.Split(unitFileContents, "\n") {
+					if strings.Contains(line, "ExecStart=/usr/local/bin/set-rps-mask.sh") {
+						maskContent = line
+					}
+				}
+				rpsMaskContent := strings.TrimSuffix(maskContent, "\r")
+				Expect(len(strings.Split(rpsMaskContent, " "))).To(Equal(3), "systemd unit file rps mask value is empty")
+				serviceRPSCPUs := strings.Split(rpsMaskContent, " ")[2]
 
 				rpsCPUs, err := components.CPUMaskToCPUSet(serviceRPSCPUs)
 				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -291,8 +291,9 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 	})
 
 	Context("RPS configuration", func() {
-		It("Should have the correct RPS configuration", func() {
-			if profile.Spec.CPU == nil || profile.Spec.CPU.Reserved != nil {
+		It("[test_id:54190] Should have the correct RPS configuration", func() {
+			if profile.Spec.CPU == nil || profile.Spec.CPU.Reserved == nil {
+				testlog.Warning("Test Skipped due nil Reserved cpus")
```
fromanirh, line 296: since we want to fix it for good, let's just use a `ginkgo.Skip` here instead of reinventing it partially.

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -291,8 +291,9 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
 	})
 
 	Context("RPS configuration", func() {
-		It("Should have the correct RPS configuration", func() {
-			if profile.Spec.CPU == nil || profile.Spec.CPU.Reserved != nil {
+		It("[test_id:54190] Should have the correct RPS configuration", func() {
+			if profile.Spec.CPU == nil || profile.Spec.CPU.Reserved == nil {
+				Skip("Test Skipped due nil Reserved cpus")
```
fromanirh, line 296: the `return` is now redundant (code will never be reached) but we can clean up later.
We now require both fields to be != nil.
Question: is it OK to run the test if CPU != nil && reserved == nil ? If I'm not mistaken, the old code allowed that (intentionally or by side effect?)

ffromani (commented):
we're almost there, just a question inside

test/e2e/performanceprofile/functests/1_performance/performance.go
mrniranjan: I had the same query, Actually i think the if condition is not required, because we can't have profile without reserved cpus in the first place.

fromanirh: fair enough

fromanirh:
failures seem real.

yanirq:
check the comparison content in https://github.com/openshift/cluster-node-tuning-operator/blob/925e77766265a12609d708a63f9addc8de163dfb/test/e2e/performanceprofile/functests/1_performance/performance.go#L338 

`rpsCPUs` is a CPUSet  vs.  `expectedRPSCPUs` which is a parsed CPUSet: https://github.com/openshift/cluster-node-tuning-operator/blob/925e77766265a12609d708a63f9addc8de163dfb/test/e2e/performanceprofile/functests/1_performance/performance.go#L304

We might get different formats for string comparison here, worth checking in other places as well since this may be an overlooked issue from the past.

mrniranjan:
> check the comparison content in
> 
> https://github.com/openshift/cluster-node-tuning-operator/blob/925e77766265a12609d708a63f9addc8de163dfb/test/e2e/performanceprofile/functests/1_performance/performance.go#L338
> 
> `rpsCPUs` is a CPUSet vs. `expectedRPSCPUs` which is a parsed CPUSet:
> 
> https://github.com/openshift/cluster-node-tuning-operator/blob/925e77766265a12609d708a63f9addc8de163dfb/test/e2e/performanceprofile/functests/1_performance/performance.go#L304
> 
> We might get different formats for string comparison here, worth checking in other places as well since this may be an overlooked issue from the past.

Will look in to it, But the other thing. is the output of the below command is `0` or ` 1` insted of bitmask `0000,00000000,00000001`:
`find /rootfs/sys/devices/virtual -type f -name rps_cpus  -exec  cat {};`

yanirq:
> > check the comparison content in
> > https://github.com/openshift/cluster-node-tuning-operator/blob/925e77766265a12609d708a63f9addc8de163dfb/test/e2e/performanceprofile/functests/1_performance/performance.go#L338
> > 
> > `rpsCPUs` is a CPUSet vs. `expectedRPSCPUs` which is a parsed CPUSet:
> > https://github.com/openshift/cluster-node-tuning-operator/blob/925e77766265a12609d708a63f9addc8de163dfb/test/e2e/performanceprofile/functests/1_performance/performance.go#L304
> > 
> > We might get different formats for string comparison here, worth checking in other places as well since this may be an overlooked issue from the past.
> 
> @yanirq Will look in to it, But the other thing. is the output of the below command is `0` or ` 1` insted of bitmask `0000,00000000,00000001`: `find /rootfs/sys/devices/virtual -type f -name rps_cpus -exec cat {};`

The default value of the rps_cpus file is zero. 
The rps_cpus files use comma-delimited CPU bitmaps. Therefore, to allow a CPU to handle interrupts for the receive queue on an interface, set the value of their positions in the bitmap to 1. For example, to handle interrupts with CPUs 0, 1, 2, and 3, set the value of rps_cpus to 00001111 (1+2+4+8), or f (the hexadecimal value for 15).

mrniranjan:
> 
> The default value of the rps_cpus file is zero. The rps_cpus files use comma-delimited CPU bitmaps. Therefore, to allow a CPU to handle interrupts for the receive queue on an interface, set the value of their positions in the bitmap to 1. For example, to handle interrupts with CPUs 0, 1, 2, and 3, set the value of rps_cpus to 00001111 (1+2+4+8), or f (the hexadecimal value for 15).

Which means after the profile is applied rps mask was not applied to the pods interface. because its not reading bitmaps

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -291,22 +290,31 @@ var _ = Describe("[rfe_id:27368][performance]", func() {
...
+				cmd := []string{"cat", "/rootfs/etc/systemd/system/update-rps@.service"}
+				unitFileContents, err := nodes.ExecCommandOnNode(cmd, &node)
 				Expect(err).ToNot(HaveOccurred())
+				for _, line := range strings.Split(unitFileContents, "\n") {
+					if strings.Contains(line, "ExecStart=/usr/local/bin/set-rps-mask.sh") {
+						maskContent = line
+					}
+				}
+				rpsMaskContent := strings.TrimSuffix(maskContent, "\r")
+				Expect(len(strings.Split(rpsMaskContent, " "))).To(Equal(3), "systemd unit file doesn't have proper rpsmask")
```
yanirq, line 316: @mrniranjan what is the expected mask content here ?  will its length always be the constant 3 ?

mrniranjan: @yanirq Yes, Example ["ExecStart=/usr/local/bin/set-rps-mask.sh", "%i", "0000001f"]

yanirq: maybe change the error message to :
systemd unit file rps mask value is empty

mrniranjan: ack

PR #484, opened 2022-10-13
WIP: Hypershift: Performance Profile Controller Hypershift adaptation

jlojosnegros:
Performance Profile Controller need to be adapted to properly work on Hypershift deployments.

This PR will handle changes described in openshift/enhancements#1244 at `Design Details -> Performance Profile Controller`

Depends on : openshift/hypershift#1782 - changes in Hypershift operator to handle `PerformanceProfile` objects


jlojosnegros:
rebase

jlojosnegros:
/hold

Still not finished but It could use some review.

jlojosnegros:
rebase over master

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,650 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+//REVIEW - Most of these already are defined in 'pkg/operator/hypershift.go' maybe we should avoid duplication.
+//         What about creating a file with all these constants and labels and reference them both here and in the
+//         NTO hypershift code?
```
dagrayvid, line 55: Yes, let's make a pkg/hypershift/hypershift.go for these constants, as well as any functions that can be reused between the two controllers. We could make this a separate PR.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,650 @@
...
+	performanceProfileFromConfigMap, err := parsePerformanceProfileManifest([]byte(performanceProfileString), nodePoolName)
+	if err != nil {
+		klog.Warningf("failed to parseTunedManifests in ConfigMap %s: %v", instance.ObjectMeta.Name, err)
+		// Return and don't requeue
+		return reconcile.Result{}, nil
+	}
+
+	if err := updatePerformanceProfile(performanceProfileFromConfigMap, ctx, r.ManagementClient, instance, nodePoolName); err != nil {
+		klog.Errorf("failed to update performance profile from configMap %q components: %v", instance.Name, err)
+		//REVIEW - Should we record this events? and if so where? in the CM or the PP?
```
dagrayvid, line 187: For NTO we would just log such an error, I think.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,650 @@
...
+	componentSet, err := manifestset.GetNewComponents(performanceProfileFromConfigMap, nil)
+	if err != nil {
+		klog.Errorf("failed to deploy performance profile from configMap %q components: %v", instance.Name, err)
+		//REVIEW - Should we record this events? and if so where? in the CM or the PP?
+		return reconcile.Result{}, err
+	}
+
+	tunedEncoded, err := encodeTuned(componentSet.Tuned)
+	if err != nil {
+		klog.Warningf("failed to Encode Tuned in ConfigMap %s: %v", instance.ObjectMeta.Name, err)
```
dagrayvid, line 200:
```suggestion
		klog.Warningf("failed to encode Tuned in ConfigMap %s: %v", instance.ObjectMeta.Name, err)
```

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,650 @@
...
+	}
+}
+
+// parseManifests parses a YAML or JSON document that may contain one or more
+// kubernetes resources.
+func parsePerformanceProfileManifest(data []byte, nodePoolName string) (*performancev2.PerformanceProfile, error) {
+	scheme := runtime.NewScheme()
+	performancev2.AddToScheme(scheme)
+
+	//REVIEW - As serializer is used many times it could be worthy to create it just one time and keep it somewhere
```
dagrayvid, line 524: Maybe this could be moved to the shared hypershift pkg, and instantiated in init(). We would need to add multiple types to the scheme, but I think this is safe.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,650 @@
...
+
+	return nil
+}
+
+func hypershiftDeleteComponents(remoteClient client.Client, ctx context.Context, ppConfigMap *corev1.ConfigMap) error {
+	// ConfigMap is marked for deletion and waiting for finalizers to be empty
+	// so better to clean-up and delete the objects.
+
+	if ppConfigMap.Labels == nil {
+		return fmt.Errorf("ConfigMap %s has no Labels.Unable to finalize deletion procedure properly", ppConfigMap.Name)
```
dagrayvid, line 587:
```suggestion
		return fmt.Errorf("ConfigMap %s has no Labels. Unable to finalize deletion procedure properly", ppConfigMap.Name)
```

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,650 @@
...
+// parseNamespacedName expects a string with the format "namespace/name"
+// and returns the name only.
+// If given a string in the format "name" returns "name".
+func parseNamespacedName(namespacedName string) string {
+	parts := strings.SplitN(namespacedName, "/", 2)
+	if len(parts) > 1 {
+		return parts[1]
+	}
+	return parts[0]
+}
```
dagrayvid, line 650: Could be moved to a shared hypershift pkg

dagrayvid (commented):
Thanks @jlojosnegros. Just an initial review so far. I intend to also try running the code and will take a closer look.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: I have created a new issue #529 to track this.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: Handle this as part of #529

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: I have this questions because we have an EventRecorder as part of the PerformanceProfileReconciler in SNO that we use to record some of this failures ... and I do not know if this is applicable here. 

Anyway it could be added in a later PR if we think is needed.

WDYT? @yanirq, @fromanirh @Tal-or

fromanirh: having events in the management cluster, if feasible, is nice. Let's try to be consistent with what classic PAO/PPCtrl is doing.

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -57,14 +58,20 @@ const finalizer = "foreground-deletion"
 // PerformanceProfileReconciler reconciles a PerformanceProfile object
 type PerformanceProfileReconciler struct {
 	client.Client
-	Scheme   *runtime.Scheme
-	Recorder record.EventRecorder
+	// Should only be used when on Hypershift deployment. In StandAlone OCP will be nil
```
yanirq, line 61: is the comment for ManagementClient only ?  if so then worth explicitly mentioning it

jlojosnegros:
rebase

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -151,30 +152,46 @@ func operatorRun() {
...
 	metrics.RegisterVersion(version.Version)
 
-	if !config.InHyperShift() {
-		if err = (&paocontroller.PerformanceProfileReconciler{
-			Client:   mgr.GetClient(),
-			Scheme:   mgr.GetScheme(),
-			Recorder: mgr.GetEventRecorderFor("performance-profile-controller"),
-		}).SetupWithManager(mgr); err != nil {
-			klog.Exitf("unable to create PerformanceProfile controller: %v", err)
+	var mngClient client.Client
```
yanirq, line 155: I wonder if this should move to the var section at the start of the file

yanirq:
@jlojosnegros  please squash all commits into one

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: I have added the events in one of the last commits.

jlojosnegros:
`e2e-hypershift` seems to be failing due to timeouts that are hardly related with this PR
/test e2e-hypershift

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -17,20 +17,51 @@ limitations under the License.
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-const controllerGeneratedMachineConfig = "hypershift.openshift.io/performanceprofile-config"
+// REVIEW - Most of these already are defined in 'pkg/operator/hypershift.go' maybe we should avoid duplication.
+//
+//	What about creating a file with all these constants and labels and reference them both here and in the
+//	NTO hypershift code?
+//
+// REVIEW - Reorder constants to read them better.
```
ffromani, line 61: this is likely a good idea, let's do in a later PR as followup

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
 
 	return ctrl.NewControllerManagedBy(mgr).For(&corev1.ConfigMap{}, builder.WithPredicates(p)).Complete(r)
 }
+
+func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	if !ntconfig.InHyperShift() {
+		return reconcile.Result{}, fmt.Errorf("Using hypershift controller configuration while not in hypershift deployment")
+	}
```
ffromani: not blocking: this method is called only if the very same guard passed, do we need another check here?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+
+	instance := &corev1.ConfigMap{}
+
+	//REVIEW - Should be OperatorNamespace?
+	err := r.Get(ctx, req.NamespacedName, instance)
```
ffromani: please clarify

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			//FIXME - In hypershift there is no "owned" objects. So how do we delete created objects? or do we even have to?
+			//jlom - as we create a configmap for each of the elements created from a PerformanceProfile and we reference
+			//       these configmaps in NodePool so they could be "transfered" to the managed cluster, maybe "delete" them is
+			//       as simple as derreference them in NodePool and delete the ConfigMaps.
+			//NOTE - Adding a label with PerformanceProfile name on it to the configmaps could make this look up easier.
```
ffromani: still a fixme? or figured out the outcome? let's fix the comment in a followup

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+
+		// Return and don't requeue
+		return reconcile.Result{}, nil
+	}
+
+	cmNodePoolNamespacedName, ok := instance.Annotations[hypershiftNodePoolLabel]
+	if !ok {
+		klog.Warningf("failed to parse PerformanceProfileManifests in ConfigMap %s, no label %s", instance.ObjectMeta.Name, hypershiftNodePoolLabel)
+		// Return and don't requeue
+		return reconcile.Result{}, nil
```
ffromani: this looks serious enough to bubble up an error

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+	}
+	nodePoolName := parseNamespacedName(cmNodePoolNamespacedName)
+
+	performanceProfileFromConfigMap, err := parsePerformanceProfileManifest([]byte(performanceProfileString), nodePoolName)
+	if err != nil {
+		klog.Warningf("failed to parseTunedManifests in ConfigMap %s: %v", instance.ObjectMeta.Name, err)
+		// Return and don't requeue
+		return reconcile.Result{}, nil
```
ffromani: ditto

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+	ctrRuntime, err := r.getContainerRuntimeName(ctx, performanceProfileFromConfigMap)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("hypershift could not determine high-performance runtime class container-runtime for profile %q; %w", performanceProfileFromConfigMap.Name, err)
+	}
+	klog.Infof("hypershift using %q as high-performance runtime class container-runtime for profile %q", ctrRuntime, performanceProfileFromConfigMap.Name)
+
+	componentSet, err := manifestset.GetNewComponents(performanceProfileFromConfigMap, nil, &pinningMode, ctrRuntime)
+	if err != nil {
+		klog.Errorf("failed to deploy performance profile from configMap %q components: %v", instance.Name, err)
+		//REVIEW - Should we record this events? and if so where? in the CM or the PP?
```
ffromani: Let's use perfprofile.status. This looks material worthy to 1. bubble up as error 2. set as Degraded condition.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+		return reconcile.Result{}, err
+	}
+
+	//jlom - Now we have to create a ConfigMap for each of the elements in the componentSet and then handle them to
```
ffromani: no need to prefix clarifying comment with the author (git blame will tell us anyway if we're curious). If it's a development comment, though, let's remove it

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+	//       the different agents that would made them effective in the managed cluster.( where workers are)
+	tunedEncoded, err := encodeTuned(componentSet.Tuned)
+	if err != nil {
+		klog.Warningf("failed to Encode Tuned in ConfigMap %s: %v", instance.ObjectMeta.Name, err)
+		// Return and don't requeue
```
ffromani, line 243: again this should probably bubble up and be set as degraded condition

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+		return reconcile.Result{}, nil
+	}
+
+	tunedConfigMap := TunedConfigMap(instance, performanceProfileFromConfigMap.Name, cmNodePoolNamespacedName, string(tunedEncoded))
+
+	if err := createOrUpdateTunedConfigMap(tunedConfigMap, ctx, r.Client); err != nil {
+		klog.Error("failure on Tuned process: %v", err)
```
ffromani: ditto

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+		// Return and don't requeue
+		return reconcile.Result{}, nil
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func encodeTuned(tuned *tunedv1.Tuned) ([]byte, error) {
+	scheme := runtime.NewScheme()
+	tunedv1.AddToScheme(scheme)
```
ffromani: do we need to do this for every invocation?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+	)
+
+	buff := bytes.Buffer{}
+	err := yamlSerializer.Encode(obj, &buff)
+	return buff.Bytes(), err
+}
+
+func createOrUpdateTunedConfigMap(cm *corev1.ConfigMap, ctx context.Context, cli client.Client) error {
+	tunedConfigMapUpdateFunc := func(orig, dst *corev1.ConfigMap) error {
+		//REVIEW - Maybe here I should ensure the readed ConfifMap has the needed labels and annotations.
```
ffromani, line 346: seems like a good idea, yes. Can be done as followup though.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+		dst.Data[tunedConfigMapConfigKey] = orig.Data[tunedConfigMapConfigKey]
+		return nil
+	}
+	return createOrUpdateConfigMap(ctx, cli, cm, tunedConfigMapUpdateFunc)
+}
+
+func createOrUpdateConfigMap(ctx context.Context, cli client.Client, cm *corev1.ConfigMap, updateFunc func(origin, destination *corev1.ConfigMap) error) error {
```
ffromani, line 371: I think we have something very similar in controller-runtime, don't we?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+			tunedConfigMapConfigKey: tunedManifest,
+		},
+	}
+}
+
+// parseManifests parses a YAML or JSON document that may contain one or more
+// kubernetes resources.
+func parsePerformanceProfileManifest(data []byte, nodePoolName string) (*performancev2.PerformanceProfile, error) {
+	scheme := runtime.NewScheme()
+	performancev2.AddToScheme(scheme)
```
ffromani: do we need to do this for every invocation?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -71,3 +102,230 @@ func (r *PerformanceProfileReconciler) hypershiftSetupWithManager(mgr ctrl.Manag
...
+	// Propagate NodePool name from ConfigMap down to PerformanceProfile object
+	if performanceProfile.Labels == nil {
+		performanceProfile.Labels = make(map[string]string)
+	}
+	performanceProfile.Labels[hypershiftNodePoolNameLabel] = nodePoolName
+
+	return performanceProfile, nil
+}
+
+func hashStruct(o interface{}) string {
```
ffromani, line 616: but we use this to hash a string, don't we?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -189,6 +191,21 @@ func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context,
 		return reconcile.Result{}, nil
 	}
 
+	machineconfigEncoded, err := encodeMachineConfig(convertMachineConfig(componentSet.MachineConfig))
+	if err != nil {
+		klog.Warningf("failed to Encode MachineConfig in ConfigMap %s: %v", instance.ObjectMeta.Name, err)
```
ffromani: again seems worthy to bubble up the error and set degraded conditions

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -189,6 +191,21 @@ func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context,
...
+		// Return and don't requeue
+		return reconcile.Result{}, nil
+	}
+
+	machineconfigConfigMap := MachineConfigConfigMap(instance, performanceProfileFromConfigMap.Name, cmNodePoolNamespacedName, string(machineconfigEncoded))
+
+	if err := createOrUpdateMachineConfigConfigMap(machineconfigConfigMap, ctx, r.Client); err != nil {
```
ffromani: ditto

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -199,6 +216,13 @@ func encodeTuned(tuned *tunedv1.Tuned) ([]byte, error) {
 	return tunedEncoded, err
 }
 
+func encodeMachineConfig(machineConfig *mcfgv1.MachineConfig) ([]byte, error) {
+	scheme := runtime.NewScheme()
+	mcfgv1.AddToScheme(scheme)
```
ffromani: the scheme is not supposed to change during the lifetime of the server, so do we need to do this for every invocation?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -219,6 +243,15 @@ func createOrUpdateTunedConfigMap(cm *corev1.ConfigMap, ctx context.Context, cli
 	return createOrUpdateConfigMap(ctx, cli, cm, tunedConfigMapUpdateFunc)
 }
 
+func createOrUpdateMachineConfigConfigMap(cm *corev1.ConfigMap, ctx context.Context, cli client.Client) error {
+	machineconfigConfigMapUpdateFunc := func(orig, dst *corev1.ConfigMap) error {
+		//REVIEW - Maybe here I should ensure the readed ConfifMap has the needed labels and annotations.
```
ffromani, line 355: yes, seems like a good idea (also typo: configMap)

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -206,6 +206,21 @@ func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context,
 		return reconcile.Result{}, nil
 	}
 
+	kubeletconfigEncoded, err := encodeKubeletConfig(convertKubeletConfig(componentSet.KubeletConfig))
+	if err != nil {
+		klog.Warningf("failed to Encode KubeletConfig in ConfigMap %s: %v", instance.ObjectMeta.Name, err)
+		// Return and don't requeue
```
ffromani: usual comment about bubbling up and set degraded conditions (seems again a good idea to do so)

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -206,6 +206,21 @@ func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context,
...
+		return reconcile.Result{}, nil
+	}
+
+	kubeletconfigConfigMap := KubeletConfigConfigMap(instance, performanceProfileFromConfigMap.Name, cmNodePoolNamespacedName, string(kubeletconfigEncoded))
+
+	if err := createOrUpdateKubeletConfigConfigConfigMap(kubeletconfigConfigMap, ctx, r.Client); err != nil {
+		klog.Error("failure on KubeletConfig process: %w", err.Error())
+		// Return and don't requeue
+		return reconcile.Result{}, nil
```
ffromani: ditto

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -223,6 +238,13 @@ func encodeMachineConfig(machineConfig *mcfgv1.MachineConfig) ([]byte, error) {
 	return mcEncoded, err
 }
 
+func encodeKubeletConfig(kubeletConfig *mcfgv1.KubeletConfig) ([]byte, error) {
+	scheme := runtime.NewScheme()
+	mcfgv1.AddToScheme(scheme)
```
ffromani: and again about the scheme

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -252,6 +274,15 @@ func createOrUpdateMachineConfigConfigMap(cm *corev1.ConfigMap, ctx context.Cont
 	return createOrUpdateConfigMap(ctx, cli, cm, machineconfigConfigMapUpdateFunc)
 }
 
+func createOrUpdateKubeletConfigConfigConfigMap(cm *corev1.ConfigMap, ctx context.Context, cli client.Client) error {
+	kubeletconfigConfigMapUpdateFunc := func(orig, dst *corev1.ConfigMap) error {
+		//REVIEW - Maybe here I should ensure the readed ConfifMap has the needed labels and annotations.
```
ffromani, line 364: yes (same typo, copy/paste?)

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -355,6 +409,23 @@ func convertMachineConfig(origMC *mcov1.MachineConfig) *mcfgv1.MachineConfig {
...
 
+func convertKubeletConfig(origKC *mcov1.KubeletConfig) *mcfgv1.KubeletConfig {
+
+	return &mcfgv1.KubeletConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: mcfgv1.SchemeGroupVersion.String(),
+			Kind:       "KubeletConfig",
+		},
+		ObjectMeta: origKC.ObjectMeta,
+		//TODO - Is there any way to do this conversion in a type safe manner?
```
ffromani, line 530: which one specifically?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -155,6 +155,9 @@ func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context,
 		return reconcile.Result{}, nil
 	}
 
+	updatePerformanceProfileName(performanceProfileFromConfigMap, nodePoolName)
+	klog.Infof("[%s] PerformanceProfile name updated to %s)", instance.Name, performanceProfileFromConfigMap.Name)
```
ffromani, line 215: stray close parens?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -446,19 +449,16 @@ func parsePerformanceProfileManifest(data []byte, nodePoolName string) (*perform
...
 	// for example, if one ConfigMap is referenced in multiple NodePools
-	performanceProfile.SetName(performanceProfile.ObjectMeta.Name + "-" + hashStruct(nodePoolName))
-	klog.V(2).Infof("parsePerformanceProfileManifest: name: %s", performanceProfile.GetName())
-
-	// Propagate NodePool name from ConfigMap down to PerformanceProfile object
-	if performanceProfile.Labels == nil {
-		performanceProfile.Labels = make(map[string]string)
+	if !strings.HasSuffix(performanceProfile.ObjectMeta.Name, "-"+hashStruct(nodePoolName)) {
+		performanceProfile.SetName(performanceProfile.ObjectMeta.Name + "-" + hashStruct(nodePoolName))
+		klog.V(2).Infof("updatePerformanceProfile: name: %s", performanceProfile.GetName())
```
ffromani, line 569: this and the other log (line 159) seems redundant. Maybe consolidate in one? (and let's always log this rename)

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -139,6 +138,35 @@ func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context,
 		return reconcile.Result{}, err
 	}
 
+	if instance.DeletionTimestamp != nil {
+		// ConfigMap is marked for deletion and waiting for finalizers to be empty
+		// so better to clean-up and delete the objects.
+
+		//REVIEW - Is it ok to use the ctx Context with the remote client???
```
ffromani, line 158: why not? which issue do you see?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -159,8 +159,10 @@ func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context,
 		//REVIEW - Is it ok to use the ctx Context with the remote client???
 		if err := hypershiftDeleteComponents(r.Client, ctx, instance); err != nil {
 			klog.Errorf("[%s] Error: failed to delete components. err: %v", req.NamespacedName, err)
+			r.Recorder.Eventf(instance, corev1.EventTypeWarning, "Deletion failed", "[hypershift:%s] Failed to delete components: %v", req.NamespacedName, err)
```
ffromani, line 161: events are not bad, I'm in favor of them. But still setting conditions (especially for degraded) seems worthy

ffromani (commented):
codewise looks reasonnable, with quite a few places to improve, smoothen/streamline, but it's ok since this is the first PR implementing support for hypershift. We can iterate later.

What's surprising however is the lack of tests. Am I missing something huge? do we have already or plan to have e2e tests? We need to clarify the testing situation. Some unit tests would be welcome too.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: issue created: https://github.com/openshift/cluster-node-tuning-operator/issues/859

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: deleted

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: Namespace is correct.
Need to use ManagementClient instead.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: Figured out. 
OwnerRef is enough with all the created object but the RuntimeClass which needs to be explicitly deleted from hosted cluster (done in `hypershiftDeleteComponents`

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: yes it is used to hash a string.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: I am totally agree with that. 
The problem here is that PerformanceProfile is embedded inside a Configmap which is owned by the Hypershift's NodePoolController, so if I change anything of the configmap it will be reconciled again by the NodePoolController.

jlojosnegros:
@ffromani Thanks for your review.

Rebased over latest master and squashed into only one commit. 
Some comments addressed.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: issue opened.
comment deleted

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: error part done. 
for degraded conditions see https://github.com/openshift/cluster-node-tuning-operator/pull/484#discussion_r1409195931

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: error part done.
for degraded conditions see https://github.com/openshift/cluster-node-tuning-operator/pull/484#discussion_r1409195931

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: issue open.
comment deleted

pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go
```diff
@@ -78,7 +78,8 @@ func GetNewComponents(profile *performancev2.PerformanceProfile, opts *component
 	}
 
 	nodeConfig := node.NewNodeConfig(apiconfigv1.CgroupModeV1)
-	runtimeClass := runtimeclass.New(profile, machineconfig.HighPerformanceRuntime)
+	rtName := runtimeclass.BuildRuntimeClassName(profile.Name)
```
yanirq, line 81: is this a refactor or a needed change for the PR ?

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,594 @@
+/*
+
+
```
yanirq, line 3: Since this was cut from performanceprofile_controller.go a while ago  lets do a revisit/compare to see if we are fully aligned with the latest

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -72,7 +76,10 @@ func init() {
 
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(tunedv1.AddToScheme(scheme))
-	utilruntime.Must(mcov1.AddToScheme(scheme))
+	if !config.InHyperShift() {
+		klog.Infof("Running in NOT hypershift")
```
ffromani, line 80: nit: I'd avoid negatives phrased like this. What about `running on regular non-hypershift cluster`?

ffromani: In addition, what's the drawback of always register the mcov1 APIs?

pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go
ffromani: +1

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -19,15 +19,16 @@ package controller
...
 	"reflect"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+
 	apiconfigv1 "github.com/openshift/api/config/v1"
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
 	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
-	"github.com/openshift/cluster-node-tuning-operator/pkg/config"
+	ntconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
```
ffromani, line 31: nit: `ntoconfig`

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,594 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("performanceprofile_controller").
+		WatchesRawSource(source.Kind(managementCluster.GetCache(), &corev1.ConfigMap{}),
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(p)).Complete(r)
+}
+
+func (r *PerformanceProfileReconciler) hypershiftReconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	klog.Infof("[%s] *** Entering ReconcileLoop ***", req.NamespacedName)
+	defer klog.Infof("[%s] *** Exiting ReconcileLoop ***", req.NamespacedName)
```
ffromani, line 139: let's use `InfoS` if we can

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
```diff
@@ -0,0 +1,594 @@
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			klog.Infof("[%s] Error: Instance not found", req.NamespacedName)
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		klog.Infof("[%s] Error: Reading failure. err: %v", req.NamespacedName, err)
```
ffromani, line 153: this should probably be `ErrorS`

ffromani (commented):
before we dive too deep, we need to tackle the webhook thingy.
We should strive to offer the same experience on hypershift and on non-hypershift, if nothing else because doing this reduces the support cost.
I tend to believe a prerequisite to this PR should be to move away from webhooks in all the flows, moving the validation in the core reconcile loop and reporting using conditions.

pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go
jlojosnegros: In general I have tried to reuse as much code as possible so any change in the performance_profile controller's way of doing things will affect also the PP hypershift controller.
In this particular case [here](https://github.com/jlojosnegros/cluster-node-tuning-operator/blob/db3912baafe05eab080b65628c4d5dfc23e68958/pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go#L283) I need to get the rtclass name so I extracted that to a functions of its own for reuse.

pkg/performanceprofile/controller/performanceprofile_controller.go
jlojosnegros: thanks for the comment.
Done

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: thanks for the comment.
Done

go.mod
```diff
@@ -18,6 +19,7 @@ require (
 	github.com/openshift/build-machinery-go v0.0.0-20230306181456-d321ffa04533
 	github.com/openshift/client-go v0.0.0-20230807132528-be5346fb33cb
 	github.com/openshift/custom-resource-status v1.1.3-0.20220503160415-f2fdb4999d87
+	github.com/openshift/hypershift v0.1.9
```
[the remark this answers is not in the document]
jlojosnegros: Not sure about that. 
Right now (4.16/master) is the version used.

Maybe @csrwng or @jparrill could better answer this question.

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: Thanks for the comment.
Done

cmd/cluster-node-tuning-operator/main.go
jlojosnegros: In hypershift we are using also [this other mcAPI](https://github.com/jlojosnegros/cluster-node-tuning-operator/blob/98c7a75a15c31956544323702569b6aec329201e/pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go#L31) and I there is a conflict when we register them both

pkg/performanceprofile/controller/performanceprofile_controller_hypershift.go
jlojosnegros: In general I have tried to reuse as much code as possible so any change in the performance_profile controller's way of doing things will affect also the PP hypershift controller, but I will go through the last changes in PP to see if there is anything to add in hypershift.

jlojosnegros:
rebased over latest master (4.16)

go.mod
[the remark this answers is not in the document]
bryan-cox: I'm assuming this is based off our tags here - https://github.com/openshift/hypershift/tags?

We continually update those as we push new versions of the HyperShift Operator in ROSA. Those tags are not tied to an OCP release.

Tal-or:
/close 
superseded by #1032 and #1037

PR #486, opened 2022-10-14
e2e: perfprof: get cpus from node capacity

shajmakh:
Lately, the latency testing suite starting to fail and that was because the tests didn't reconfigure the performance profile to be compatible with the tests prerequisites, hence they were using the existing profile mainly missing the correct isolated CPUs value.

The reason for not changing the profile is that there were no nodes with sufficient CPUs on the compute nodes. This check was actually fetching the allocatable CPUs (total of isolated) and that is excluding the resreved CPUs, and this is expected because this is the configuration of the PP that was applied earlier on the cluster.

To fix that, fetch that CPUs value from node capacity instead in order to include all CPUs on the node.
Signed-off-by: shereenH <shajmakh@redhat.com>


MarSik (approved):
/lgtm

PR #487, opened 2022-10-17
OCPBUGS-764: [release-4.11] Add intel_pstate=disable to default configuration with no hints (#480)

marioferh:
Maintain backwards compatibility after workloadHints feature

The workloads hints feature introduced in 4.11 does not preserve the default tuning settings from 4.10: intel_pstate=disabled.

So if the customer wants to preserve the behavior they have in 4.10, they have to update their performance profile and incur another reboot.

The correct behavior the default setting should be preserved.

manual cherry-pick https://github.com/openshift/cluster-node-tuning-operator/pull/480:

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
@@ -26,12 +26,13 @@ var (
 	cmdlineWithStaticIsolation              = regexp.MustCompile(`\s*cmdline_isolation=\+\s*isolcpus=managed_irq,\${isolated_cores}\s*`)
 	cmdlineWithoutStaticIsolation           = regexp.MustCompile(`\s*cmdline_isolation=\+\s*isolcpus=domain,managed_irq,\${isolated_cores}\s*`)
 	cmdlineWithRealtime                     = regexp.MustCompile(`\s*cmdline_realtime=\+\s*nohz_full=\${isolated_cores}\s+tsc=nowatchdog\s+nosoftlockup\s+nmi_watchdog=0\s+mce=off\s+skew_tick=1\s*`)
-	cmdlineHighPowerConsumption             = regexp.MustCompile(`\s*cmdline_power_performance=\+\s*processor.max_cstate=1\s+intel_idle.max_cstate=0\s+intel_pstate=disable\s*`)
+	cmdlineHighPowerConsumption             = regexp.MustCompile(`\s*cmdline_power_performance=\+\s*processor.max_cstate=1\s+intel_idle.max_cstate=0\s*`)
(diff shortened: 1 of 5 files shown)


pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
```diff
@@ -69,8 +69,6 @@ var _ = Describe("Tuned", func() {
 			By("Populating hugepages cmdline")
 			Expect(cmdlineHugepages.MatchString(manifest)).To(BeTrue())
 			By("Populating empty additional kernel arguments cmdline")
-			Expect(manifest).To(ContainSubstring("cmdline_additionalArg="))
```
fromanirh, line 72: in the original PR we added a test (which is the right thing to do), while in this backport we are making an existing test weaker (?) and not adding new tests. Can we improve somehow?

marioferh: its not weaker, but the cherry pick takes also this change:
`{{if .AdditionalArgs}}
cmdline_additionalArg=+{{.AdditionalArgs}} 
{{end}}`
Then it's not necessary anymore to check if the is empty additionalArgs in cmdline. In any case is better delete this change and continue as before. But, if:
`{{if .RealTime}}
{{end}}`

is added, this check:
Expect(manifest).To(ContainSubstring("cmdline_additionalArg="))``

Fails, instead, if:
`intel_pstate=disable`

is added to the existing line:
`{{if .RealTime}}
cmdline_realtime=+nohz_full=${isolated_cores} tsc=nowatchdog nosoftlockup nmi_watchdog=0 mce=off skew_tick=1 rcutree.kthread_prio=11 intel_pstate=disable
{{end}}`

this fails:
`Expect(cmdlineHugepages.MatchString(manifest)).To(BeTrue())`

About the added test: 4.12 has perPodPowerManagement and doesn't have REGEX to check tests. If I add a test this wouldn't be a cherry-pick. It will be almost a new PR, because there are many changes until here...
I can do it but not sure if it is the best solution.

fromanirh:
/approve
we do want this backport

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
fromanirh: ok. Drifting too far from cherry-pick is indeed not good. If there's not a reasonable way to cherry-pick or backport tests, I guess we're good as we are.

marioferh: Thanks, I've tried to make the minimum possible changes but the test are different and changing things some of them get broke, also there is another hint in 4.12 that changes the cherry-pick.

MarSik (approved):
/lgtm

PR #444, opened 2022-08-24
irqbalance: add unit to clear the cpu ban list #413 

marioferh:
irqbalance: add unit to clear the cpu ban list

Add a oneshot unit to clear the CPU ban list once and explicitely per node reboot.
CRI-O has a facility to restore the irqbalance ban list, which only partially fit this scenario.

For this reason, to really ensure the desired behavior we tamper the CRI-O private irqbalance ban list, until CRI-O gains better
support for this flow.

Depends on https://github.com/openshift/cluster-node-tuning-operator/pull/396

Copy of https://github.com/openshift/cluster-node-tuning-operator/pull/413

PR #488, opened 2022-10-18
Fix banned cpu parsing

shajmakh:
The test goal is to check that the banned cpus list is overwritten after a CRI-O restart. After applying a fake data in the banned cpus list, the test restarts the CRI-O but expects to find a non-empty list. Fix the expectation to empty list.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -233,7 +233,7 @@ var _ = Describe("[performance] Checking IRQBalance settings", func() {
 					return false
 				}
 				return bannedCPUs.IsEmpty()
-			}).WithTimeout(5*time.Minute).WithPolling(10*time.Second).ShouldNot(BeTrue(), "banned CPUs %v not empty on node %q", bannedCPUs, targetNode.Name)
+			}).WithTimeout(5*time.Minute).WithPolling(10*time.Second).Should(BeTrue(), "banned CPUs %v not empty on node %q", bannedCPUs, targetNode.Name)
 		})
 	})
 })
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -63,6 +63,7 @@ var _ = Describe("[performance] Checking IRQBalance settings", func() {
 
 		nodeIdx := pickNodeIdx(workerRTNodes)
 		targetNode = &workerRTNodes[nodeIdx]
+		Expect(targetNode).ToNot(BeNil(), "missing target node")
 		By(fmt.Sprintf("verifying worker node %q", targetNode.Name))
 	})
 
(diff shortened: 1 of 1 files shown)


shajmakh:
@fromanirh can you take a look please? was this the intent of the test or did I misunderstand it?

fromanirh (commented):
great that we want to improve this test. Please add the following changes which I had lingering since forever:
```
diff --git a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
index 0be9c20c..e19edeed 100644
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -196,7 +196,6 @@ var _ = Describe("[performance] Checking IRQBalance settings", func() {
                })
 
                It("Should DO overwrite the banned CPU set on CRI-O restart", func() {
-
                        nodeIdx := pickNodeIdx(workerRTNodes)
                        node := &workerRTNodes[nodeIdx]
                        By(fmt.Sprintf("verifying worker node %q", node.Name))
@@ -244,7 +243,7 @@ var _ = Describe("[performance] Checking IRQBalance settings", func() {
                        Eventually(func() bool {
                                bannedCPUs, err = getIrqBalanceBannedCPUs(node)
                                if err != nil {
-                                       fmt.Fprintf(GinkgoWriter, "getting banned CPUS from %q: %v", node.Name, err)
+                                       testlog.Warningf("getting banned CPUS from %q: %v", node.Name, err)
                                        return false
                                }
                                return bannedCPUs.IsEmpty()
@@ -320,12 +319,12 @@ func makeBackupForFile(node *corev1.Node, path string) func() {
 
        out, err := nodes.ExecCommandOnNode([]string{"/usr/bin/cp", "-v", fullPath, savePath}, node)
        ExpectWithOffset(1, err).ToNot(HaveOccurred())
-       fmt.Fprintf(GinkgoWriter, "%s", out)
+       testlog.Infof("%s", out)
 
        return func() {
                out, err := nodes.ExecCommandOnNode([]string{"/usr/bin/mv", "-v", savePath, fullPath}, node)
-               fmt.Fprintf(GinkgoWriter, "%s", out)
+               testlog.Infof("%s", out)
        }
 }
 ```
(just in case: no need to credit them to me)

File: test/e2e/performanceprofile/functests/1_performance/irqbalance.go [hunk 1]
```diff
@@ -233,7 +233,7 @@ var _ = Describe("[performance] Checking IRQBalance settings", func() {
 					return false
 				}
 				return bannedCPUs.IsEmpty()
-			}).WithTimeout(5*time.Minute).WithPolling(10*time.Second).ShouldNot(BeTrue(), "banned CPUs %v not empty on node %q", bannedCPUs, targetNode.Name)
+			}).WithTimeout(5*time.Minute).WithPolling(10*time.Second).Should(BeTrue(), "banned CPUs %v not empty on node %q", bannedCPUs, targetNode.Name)
```
fromanirh, line 236: yes, the intent of the test should actually what you are fixing.
IOW the original boolean condition is inverted, apparently this was not caught during the review, and most worringly I'm not sure why this didn't fail constantly since then.

in #444 we merged code to make sure that `/etc/sysconfig/orig_irq_banned_cpus` always contains empty mask.
when crio starts with highperformace hooks enabled, it unconditionally restores the irqbalance config with the content of `/etc/sysconfig/orig_irq_banned_cpus`. In turn `/etc/sysconfig/orig_irq_banned_cpus` is created if missing the first time crio starts, with no control whatsoever on our side. This is why in #444 we effectively tamper it to make sure the content is predictable.

Since we are fixing this issue, let's also fix the error text, something like
```
"expected empty banned  CPUs on node %q, got %v instead". targetNode.Name, bannedCPUs,
```

fromanirh:
parallel investigation: https://github.com/openshift/cluster-node-tuning-operator/pull/493

fromanirh:
I'm afraid we need something more radical: https://github.com/openshift/cluster-node-tuning-operator/pull/494

shajmakh:
closing this PR for it's irrelevant following above comments

PR #490, opened 2022-10-18
Switch to OS-shipped stalld

jmencak:
OS-shipped `stalld` matured to version 1.15 in OCP 4.10 and fixes a number of issues, namely running the stalld process as `unconfined_service_t` type.


jmencak:
@liqcui, I've tested that `4.10.0-0.nightly-2022-11-01-031306` ships with `stalld-1.15-3.el8_4.x86_64` which uses `-t 20` and `initrc_t` SELinux type.  
Could you please provide `cherry-pick-approved` label?

PR #492, opened 2022-10-20
OCPBUGS-2437: Filter out copied csvs when removing pao CSV

yanirq:
OCPBUGS-2437: Clusters with large numbers of CSVs can cause crashloop,block upgrades.

This bug is caused since in upgrade procedure NTO lists all CSVs in all namespaces in order to remove PAO CSV.
Since PAO CSV is targeted to all namespaces and thus has a copy of its CSV in every namespace the list operation is expensive.

Using a filtered list to get the original non copied CSV should relieve the load.


fromanirh (commented):
/approve
assuming this flow is covered by some form of testing

fromanirh:
do we have some form of automated test covering this flow?

PR #494, opened 2022-10-21
e2e: perfprof: remove broken test

fromanirh:
We added a e2e test to ensure that the implicit and uncontrollable recover of the irqbalance settings performed by crio was doing what we expect (and actually assume) it would.

The test was becoming flakier and flakier over time ever since.

The test was well intentioned, but it tried to restart crio from within a container.
And this is a terrible idea in the first place, and a idea which could not possibly work in any sane environment in the second place.

A sane container environment just can't allow a container, no matter how privileged, to mess up with the runtime, for very obvious and very basic security reasons.

So the test, which still has its merits, need to be redesigned from scratch. This can take time, and most notably, would need a different approach. Possibly, the only sane way (or probably the most sane) is to request a node reboot, which is a very invasive and time consuming operation.

For all the reasons above, the only reasonnable way forward is to delete it until we have the better replacement.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -194,62 +194,6 @@ var _ = Describe("[performance] Checking IRQBalance settings", func() {
 			By(fmt.Sprintf("Checking content of %q on node %q", origBannedCPUsFile, node.Name))
 			expectFileEmpty(node, origBannedCPUsFile)
 		})
-
-		It("Should DO overwrite the banned CPU set on CRI-O restart", func() {
-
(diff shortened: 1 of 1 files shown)


fromanirh:
for historical record: leveraging `oc debug node/` can be a way out. Meaning, we can _possibly_ do what the test wants to do using a super privileged container, hence [the initial statement](https://github.com/openshift/cluster-node-tuning-operator/pull/494#issue-1418391884) can be challenged.
However, no easy or quick fix in sight, and even (or, especially) using a super privileged container will require proper design and scrutiny, so removing this test is the only viable short-term solution.

fromanirh:
branch 4.12 is still automatically synced

PR #495, opened 2022-10-23
OCPBUGSM-47141: Apply RPS mask for each RX queues

Tal-or:
On 4.12 a change was introduced wrt UDP GRO feature enabled by default.
This requires apply RPS mask to all queues (and not only rx-0).

We did some benchmark and find out that setting all the masks via a
simple \`for\` loop is faster than calling \`find\` and fork into a
subshell like it was implemented originally:
https://github.com/openshift/cluster-node-tuning-operator/blob/release-4.11/assets/performanceprofile/scripts/set-rps-mask.sh#L36


assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -33,4 +33,4 @@ function find_dev_dir {
 [ -d "${dev_dir}" ] || { sleep 5; find_dev_dir; }  # search failed, wait a little and try again
 [ -d "${dev_dir}" ] || { echo "${dev_dir}" directory not found >&2 ; exit 0; } # the interface disappeared, not an error
 
-echo "${mask}" > "${dev_dir}"/queues/rx-0/rps_cpus
+find "${dev_dir}"/queues -type f -name rps_cpus -exec sh -c "echo ${mask} | cat > {}" \;
```
fromanirh, line 36: I just wonder if a plain bash `for` loop would be better than calling `find + cat`. Did we try and benchmark the for loop approach?

Tal-or: There is no significant difference between using `for` and `find+cat`:
```
sh-4.4# time for i in /sys/class/net/*/queues/rx-*/rps_cpus; do sh -c echo test > /dev/null; done

real    0m0.292s
user    0m0.080s
sys     0m0.209s
sh-4.4# time find /sys/class/net/*/queues/ -type f -name rps_cpus -exec sh -c "echo test > /dev/null" \;

real    0m0.292s
user    0m0.104s
sys     0m0.177s
```

But!
There is a huge difference when we're not forking into a new shell process:
```
sh-4.4# time for i in /sys/class/net/*/queues/rx-*/rps_cpus; do echo test > /dev/null; done

real    0m0.004s
user    0m0.001s
sys     0m0.003s
```
Notice we remove the `sh -c`

fromanirh: yes, this is why I mentioned this approach. I assumed to do a for loop fully within spawning processes (so no `sh -c`). I should have been explicit since the beginning, sorry about that.

fromanirh: I believe the pure-shell for loop has its merit if we have reasonnable guarantees the glob expression will be valid for a long time. E.g. if the path is ALWAYS like
```bash
/sys/class/net/FOO/queues/rx-NUM/rps_cpus
/sys/class/net/BAR/queues/rx-NUM/rps_cpus
/sys/class/net/BAZ/queues/rx-NUM/rps_cpus
```

and like never ever nested devices:
```
/sys/class/net/FOO/BAR/queues/rx-NUM/rps_cpus
```

Tal-or: Yes. This is a stable API which I don't think would change

fromanirh: cool, then in a followup PR we can perhaps optimize further!

Tal-or: Why followup PR? I already apply the change

fromanirh: ah, great!

Tal-or:
/unhold
tested manually looks fine

assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -33,4 +33,6 @@ function find_dev_dir {
 [ -d "${dev_dir}" ] || { sleep 5; find_dev_dir; }  # search failed, wait a little and try again
 [ -d "${dev_dir}" ] || { echo "${dev_dir}" directory not found >&2 ; exit 0; } # the interface disappeared, not an error
 
-echo "${mask}" > "${dev_dir}"/queues/rx-0/rps_cpus
+for i in "${dev_dir}"/queues/rx-*/rps_cpus; do
```
fromanirh, line 36: `i` is not a great name here but considering how costly the testing is, I won't resubmit *just* for this. But let's keep in mind if you need to resubmit. Let's discuss the name, something like `rps_file` `mask_file` or the liking.

fromanirh: the only real concern I have is: this is bound to break if this path changes. How likely is this to change?

Tal-or: I don't know maybe @pabeni can tell us

MarSik: Well the alternative solution based on find works the same way, so I am not worried about this part.

MarSik: I was just wondering if the quotes should be extended to cover the whole path.

fromanirh: > Well the alternative solution based on find works the same way, so I am not worried about this part.

fair enough

Tal-or: > I was just wondering if the quotes should be extended to cover the whole path.

We had the same with find and it worked

pabeni: > the only real concern I have is: this is bound to break if this path changes. How likely is this to change?

do you mean the sysfs path/dir names? they are part of uAPI, are expected to be as stable as possible. Any change to that would be a serious kernel bug.

Tal-or: >do you mean the sysfs path/dir names? they are part of uAPI, are expected to be as stable as possible. Any change to that would be a serious kernel bug.

Thank you for your feedback, if that's the case I think we can move along with the current approach.

MarSik (approved):
/lgtm

PR #505, opened 2022-11-03
OCPBUGS-2209: Fix default hard eviction threshold when PCC is applied

marioferh:
After PCC is applied, hard evicition is different to the kubelet config default

Before: memory.available<100Mi nodefs.available<10% imagefs.available<15% nodefs.inodesFree<5% After PCC: memory.available<100Mi

After fix, hard eviction tresholds are default ones


test/e2e/performanceprofile/testdata/must-gather/must-gather.e2e/quay-io-openshift-kni-performance-addon-operator-must-gather-sha256-e0807f2ff1ff1754cdb7475e665c6c60fb1faa1f14eb0f2f7a791388ac8b33e8/cluster-scoped-resources/machineconfiguration.openshift.io/kubeletconfigs/performance-performance.yaml
```diff
@@ -132,7 +132,10 @@ spec:
       full-pcpus-only: "true"
     cpuManagerReconcilePeriod: 5s
     evictionHard:
+	    nodefs.available: 15%
```
fromanirh, line 135: weird spacing here, let's be consistent with the original indent

pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
```diff
@@ -26,14 +26,20 @@ const (
...
+	defaultKubeReservedMemory                    = "500Mi"
+	defaultSystemReservedMemory                  = "500Mi"
+	defaultHardEvictionThresholdMemory           = "100Mi"
+	defaultHardEvictionThresholdNodefs           = "10%"
+	defaultHardEvictionThresholdImagefs          = "15%"
+	defaultHardEvictionThresholdNodefsInodesFree = "5%"
+	evictionHardMemoryAvailable                  = "memory.available"
+	evictionHardNodefsAvaialble                  = "nodefs.available"
+	evictionHardImagefsAvailable                 = "imagefs.available"
+	evictionHardNodefsInodesFree                 = "nodefs.inodesFree"
```
fromanirh, line 42: I'm torn. On one hand, this is following the direction of the old code. On the other hand, we're duplicating k8s defaults, so this can get out of date, and most importantly, it can get out of date silently.
I think the minimum we should do is to add a test to ensure our defaults are still the same as k8s defaults.
The best option would be to read the defaults from the kubelet/k8s, but I'm not sure how practical it is.

MarSik: I hoped we would find out and document where the k8s defaults come from too. And yes, maintainability is the main concern here.

@marioferh Can you please add at least a comment close to the constants that will explain where the values came from? A k8s compliance test would be better and an automated way to read or update automatically the best. (Could go generate help here for example?)

test/e2e/performanceprofile/testdata/must-gather/must-gather.e2e/quay-io-openshift-kni-performance-addon-operator-must-gather-sha256-e0807f2ff1ff1754cdb7475e665c6c60fb1faa1f14eb0f2f7a791388ac8b33e8/cluster-scoped-resources/machineconfiguration.openshift.io/kubeletconfigs/performance-performance.yaml
marioferh: right, some error with IDE tabs

pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
marioferh: I have updated the vendor files to get defaults from kubelet files.

fromanirh (commented):
/approve
next time please amend the commits instead of adding a fixing commit at the end of the series

pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
```diff
@@ -26,20 +28,16 @@ const (
...
+	experimentalKubeletSnippetAnnotation = "kubeletconfig.experimental"
+	cpuManagerPolicyStatic               = "static"
+	cpuManagerPolicyOptionFullPCPUsOnly  = "full-pcpus-only"
+	memoryManagerPolicyStatic            = "Static"
+	defaultKubeReservedMemory            = "500Mi"
+	defaultSystemReservedMemory          = "500Mi"
+	evictionHardMemoryAvailable          = "memory.available"
+	evictionHardNodefsAvaialble          = "nodefs.available"
+	evictionHardImagefsAvailable         = "imagefs.available"
+	evictionHardNodefsInodesFree         = "nodefs.inodesFree"
```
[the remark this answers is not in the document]
marioferh: I didn't find it

pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
```diff
@@ -13,6 +13,8 @@ import (
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+
+	kubeletv1beta1 "k8s.io/kubernetes/pkg/kubelet/apis/config/v1beta1"
```
yanirq, line 17: is this the right version ? v1beta1?

fromanirh: note we also have `v1` nowadays https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/apis/config/v1

marioferh: there is a v1, but with no [defaults_test.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/v1beta1/defaults_test.go)

marioferh: Then the option is come back to the initial commit and hardcore the values.

yanirq: @fromanirh `v1` is not included under the 1.25 branch https://github.com/kubernetes/kubernetes/tree/release-1.25/pkg/kubelet/apis/config  so under kubelet context can we assume we need to use v1beta1 ?

fromanirh: the problem here is painting ourselves in a corner. This can be indeed be fine for 1.25, but what next?
- we know we will have to maintain this code
- we know we will have to move away from v1beta1 eventually
so this is an improvemente for the short term, but what about mid/long term?

fromanirh:
/approve cancel
concerns and open questions about the api version to use

pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go
marioferh: Well my opinion it would be do it as in the initial commit even it is not the best option. Use v1beta1 seems as we introduce two technical debs, the hardcored values an the dependency of v1beta1.

fromanirh: From my perspective replicating defaults is fine as long as we have a test which fails when our defaults drift away from kube's (because kube changes their defaults).

marioferh: Ok, I am updating with a test

marioferh: Add test: https://github.com/openshift/cluster-node-tuning-operator/blob/4a4875077e31778562e694916738c7a80c47a5af/pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig_test.go#L158

pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+	kubeletv1beta1 "k8s.io/kubernetes/pkg/kubelet/apis/config/v1beta1"
```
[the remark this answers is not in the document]
marioferh: If we want to add a test to check our hardcore values, I think so

fromanirh:
/approve
/lgtm
kube 1.26 is still using kubelet config api v1beta1, so it seems we can move forward and iterate later

PR #507, opened 2022-11-04
OCPBUGS-1334: Configure ktimers scheduler priority to match ksoftirqsd's

jlojosnegros:
New per-CPU kthread introduced to handle timer expiry named `ktimers/n` where `n` is the CPU id where it is executing.
Recommended priority is the same as ksoftirqsd.


MarSik:
Jose, would you mind preparing a PR for tuned itself as well please? We want to avoid PAO specific tuning code as much as possible.

PR #508, opened 2022-11-06
[release-4.11] [manual] OCPBUGS-3182: set RPS for veth on host level only

Tal-or:
RPS handling on pod container level using crio-hooks causes long delay times when running the low latency script to set the RPS mask (https://bugzilla.redhat.com/show_bug.cgi?id=2109965)

For RAN low latency solution it might be sufficient only to set the RPS on the host level and avoid setting it on the container level while utilizing RSS behavior.

In the past the low latency hook was added with RPS additional settings on virtual devices since there was an issue where the start and shutdown of big amount of pods will initiate the creation of the systemd service that should update the new interfaces rps_cpus mask and can create an additional CPU load under the cluster (https://github.com/openshift-kni/performance-addon-operators/pull/659)
This might not be the case any more thus we need to examine how the revert of the aforementioned PR will behave now.

Co-authored-by: Yanir Quinn <yquinn@redhat.com>


yanirq:
if there are relevant e2e-tests we can downport in the same PR it will be good for coverage.

Tal-or:
/hold cancel 
E2E tests were not merged to the master branch yet: https://github.com/openshift/cluster-node-tuning-operator/pull/501

PR #512, opened 2022-11-11
PPC: rt-kernel enabled, default power mode fails

marioferh:
PPCreator: If rt-kernel enabled, power consumption mode default cannot be selected
    
There is no such a constrain in the documentation


yanirq:
this tests needs to be aligned:
https://github.com/openshift/cluster-node-tuning-operator/blob/c11bb5c4db16beb5cd70293c53b4b915a670e869/test/e2e/performanceprofile/functests-performance-profile-creator/1_performance-profile_creator/ppc.go#L262

marioferh:
Sure, just waiting to get some feedback to understand if the msg is correct.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -506,8 +506,8 @@ func getProfileData(args ProfileCreatorArgs, cluster ClusterData) (*ProfileData,
 	case defaultLatency:
 		if profileData.rtKernel {
 			return nil, fmt.Errorf(
-				"please use one of %v power consumption modes together with the real-time kernel",
-				validPowerConsumptionModes[1:],
+				"%v power consumption mode is not available with real-time kernel, please use one of %v modes",
+				validPowerConsumptionModes[0], validPowerConsumptionModes[1:],
```
yanirq, line 510: if validPowerConsumptionModes[0] is always `defaultLatency`  maybe hardcode this value in the error message or use enums instead of integer indexes ?  (enums can be used for future flexibility)

yanirq:
if we are already fixing error messages, do you think this error should also be more accurate ? 
https://github.com/openshift/cluster-node-tuning-operator/blob/d033f0f11e2bfe7bca829ff01a56c61bbe1a7f14/cmd/performance-profile-creator/cmd/root.go#L520

fromanirh:
is the failure in the `no-cluster` lane real? at glance it seems so

fromanirh (commented):
/approve
/lgtm

PR #518, opened 2022-11-30
Fix two irqbalance tests - smp affinity vs online

MarSik:
On a system with offline cpus (like when nosmt is being used), the default_smp_affinity still shows all bits as one even though the respective cpus are offline.

[core@cnfdd7 ~]$ cat /proc/irq/default_smp_affinity ff,ffffffff,ffffffff,ffffffff
[core@cnfdd7 ~]$ cat /sys/devices/system/cpu/online 0-51

Our tests did not expect that.

The test is trying to check if all online cpus can handle interrupts before the tuning is applied and so this fix simply removes the offline cpus from the default affinity mask.


fromanirh (commented):
/approve
/lgtm
The new code is a stricter version of the previous, so this change seems good regardless of the specifics of how we discovered it.

fromanirh:
/lgtm
slow bot or picky bot?

PR #519, opened 2022-12-01
Add Containerfile for RHEL-9

jmencak:
This is a preliminary PR to add support for [RHEL-9](https://issues.redhat.com//browse/RHEL-9) image-based NTO.

Other changes:
  - switch from three to two-stage build for [RHEL-8](https://issues.redhat.com//browse/RHEL-8)

### DIFF
--- a/Dockerfile.rhel9
+++ b/Dockerfile.rhel9
@@ -1,9 +1,10 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.13 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.19-openshift-4.13 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-node-tuning-operator
 COPY . .
 RUN make build
 
-FROM registry.ci.openshift.org/ocp/4.13:base
+# Switch to registry.ci.openshift.org/ocp/4.13:base once 4.13:base is UBI9-based
+FROM registry.access.redhat.com/ubi9:latest
 COPY --from=builder /go/src/github.com/openshift/cluster-node-tuning-operator/_output/cluster-node-tuning-operator /usr/bin/
 COPY --from=builder /go/src/github.com/openshift/cluster-node-tuning-operator/_output/performance-profile-creator /usr/bin/
 COPY manifests/*.yaml manifests/image-references /manifests/


jmencak:
This is still WiP.  This PR will likely have consequences that will need to be fixed elsewhere as we're changing from 3 to 2 stage Containerfile.

jmencak:
Bump to golang 1.19 (as usual) causes "make verify" issues as usual.  Will fix in a separate PR.

jmencak:
Right, I actually cannot find any reason we should be worried about the switch from 3->2 stage Containerfile right now.  I found only https://github.com/openshift/cluster-node-tuning-operator/pull/191 where the bot missed updating the third Containerfile image URL.  Removing WiP.

Eventually, we'll need to adjust the upstream CI to use `Dockerfile.rhel9` and also make adjustments to the https://github.com/openshift/ocp-build-data repository unless this is done automatically for all operators.

/cc @dagrayvid

fromanirh:
Just asking in general because I happened to have a look. Any plan to rename Dockerfile* to Containerfile*? Is something we are expected or recommended to do?
obviously this is totally **NOT** blocking for this PR, asking in general.

jmencak:
> Any plan to rename Dockerfile* to Containerfile*? Is something we are expected or recommended to do?

I haven't seen a single core operator having done that yet.  It is a similar discussion as with master/main.  When starting a new project, I definitely use Containerfile* and main.

Now, I'm not against that change, but not sure about all the other consequences that will have.  Needs investigation and as everything, it is about priorities. :)  Right now though, I'm leaning towards _don't fix what ain't broken_.

fromanirh:
> > Any plan to rename Dockerfile* to Containerfile*? Is something we are expected or recommended to do?
> 
> I haven't seen a single core operator having done that yet. It is a similar discussion as with master/main. When starting a new project, I definitely use Containerfile* and main.
> 
> Now, I'm not against that change, but not sure about all the other consequences that will have. Needs investigation and as everything, it is about priorities. :) Right now though, I'm leaning towards _don't fix what ain't broken_.

Very fair. Thanks @jmencak for your insights.

jmencak:
@fromanirh If you agree, let's keep the current `Dockerfile.rhel8/9` and revisit this topic when we actually start using `Dockerfile.rhel9`.  Then we will know more and perhaps name it `Containerfile.rhel9` ?  Maybe alongside other renamings.

jmencak:
For future reference, changes to `openshift/ocp-build-data` were necessary after the 3->2 stage build after all.  Otherwise the Containerfile reconciliation was failing.  See: https://github.com/openshift/ocp-build-data/pull/2318

PR #522, opened 2022-12-02
CNF-6946: feat: expand support to render based on files in dir

eggfoobar:
This PR updates our render command to be more inline with a more flexible approach during the bootstrapping process.

Typically operators that need to generate files during bootstrapping might need some initial cluster state information in order to correctly render the resource files, some have static information that is retrieved and others more dynamic. In this approach we take the pattern used by MCO where the manifest directory is inspected and the relevant files are read in and parsed to render out the resulting configurations. ([see](https://github.com/openshift/machine-config-operator/blob/bff62e7ea81a3180860b28068fd33a7209475f14/pkg/controller/bootstrap/bootstrap.go#L52-L232))

This affords us some niceties like not being over prescriptive with file names for performance profile manifests that are supplied during bootstrap, as well as the ability to get the bootstrap cluster state and inform our rendering of the configuration files. 

Signed-off-by: ehila <ehila@redhat.com>

### DIFF
--- a/docs/performanceprofile/performance_controller.md
+++ b/docs/performanceprofile/performance_controller.md
@@ -62,7 +62,7 @@ The operator can render manifests for all the components it supposes to create, 
 You need to provide the following environment variables
 
 ```shell
-export PERFORMANCE_PROFILE_INPUT_FILES=<comma separated list of your Performance Profiles>
+export ASSET_INPUT_DIR=<input path containing performance profile manifests>
 export ASSET_OUTPUT_DIR=<output path for the rendered manifests>
 ```
 
@@ -75,7 +75,7 @@ _output/cluster-node-tuning-operator render
 Or provide the variables via command line arguments
 
 ```shell
-_output/cluster-node-tuning-operator render --performance-profile-input-files <path> --asset-output-dir<path>
+_output/cluster-node-tuning-operator render --asset-input-dir <path> --asset-output-dir <path>
 ```
 
 ## Troubleshooting
--- a/pkg/performanceprofile/cmd/render/cmd.go
+++ b/pkg/performanceprofile/cmd/render/cmd.go
@@ -1,0 +1,89 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package render
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"k8s.io/klog"
+)
+
+type renderOpts struct {
+	assetsInDir  string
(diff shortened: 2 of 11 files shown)


yanirq:
The render tests should be aligned to use the new render command  https://github.com/openshift/cluster-node-tuning-operator/blob/e43b1372c35826f65137978c51a0836994a9a2c9/test/e2e/performanceprofile/functests-render-command/1_render_command/render_test.go#L81

eggfoobar:
> The render tests should be aligned to use the new render command
> 
> https://github.com/openshift/cluster-node-tuning-operator/blob/e43b1372c35826f65137978c51a0836994a9a2c9/test/e2e/performanceprofile/functests-render-command/1_render_command/render_test.go#L81

Didn't see that there, updated, thanks! 

Also wanted to bring something to your attention, it seems we updated our runtime class struct to `node.k8s.io/v1` api, but we were still rendering the `node.k8s.io/v1beta1` in my testing that was causing errors when applying it during the bootstrapping process, so I updated it to `node.k8s.io/v1`. In current deployments it seems that gets updated to `node.k8s.io/v1` anyway, so this should be an okay change. What do you think?

https://github.com/openshift/cluster-node-tuning-operator/pull/522/files#diff-8fc0682a821429188c846138f0c000cf3ca68ac1807612f1b6306b33b251b26dR17

yanirq:
> > The render tests should be aligned to use the new render command
> > https://github.com/openshift/cluster-node-tuning-operator/blob/e43b1372c35826f65137978c51a0836994a9a2c9/test/e2e/performanceprofile/functests-render-command/1_render_command/render_test.go#L81
> 
> Didn't see that there, updated, thanks!
> 
> Also wanted to bring something to your attention, it seems we updated our runtime class struct to `node.k8s.io/v1` api, but we were still rendering the `node.k8s.io/v1beta1` in my testing that was causing errors when applying it during the bootstrapping process, so I updated it to `node.k8s.io/v1`. In current deployments it seems that gets updated to `node.k8s.io/v1` anyway, so this should be an okay change. What do you think?
> 
> https://github.com/openshift/cluster-node-tuning-operator/pull/522/files#diff-8fc0682a821429188c846138f0c000cf3ca68ac1807612f1b6306b33b251b26dR17

Actually we should not have any occurrences of [node.k8s.io/v1beta1](http://node.k8s.io/v1beta1) and this requires a re scan of our current code to track any additional rouge appearances of this deprecated version.
Thanks for bringing that up.

yanirq:
@eggfoobar  should we put this PR on hold until https://github.com/openshift/enhancements/pull/1213 gets a final approval ?

eggfoobar:
I think this should be fine on it's own, it's needed by the EP but it was already something that we wanted NTO to do during bootstrap. This code doesn't do anything with infrastructure or workload partitioning yet, I have that part of the #431 this is just refining our code to run during bootstrap. @MarSik What do you think?

MarSik:
> I think this should be fine on it's own, it's needed by the EP but it was already something that we wanted NTO to do during bootstrap. This code doesn't do anything with infrastructure or workload partitioning yet, I have that part of the #431 this is just refining our code to run during bootstrap. @MarSik What do you think?

I think merging this in isolation is fine as long as the commit message provides good enough rationale. Like pointing out to a common pattern everyone else uses. The directory in - directory out model was the indended one anyway, so I am ok with treating this as a standalone change not coupled to the enhancement.

eggfoobar:
Hey @yanirq, I went ahead and updated the commit messages and PR message to provide clear rational on this change.

yanirq:
> Hey @yanirq, I went ahead and updated the commit messages and PR message to provide clear rational on this change.

Thanks @eggfoobar.  Before we get to a final round of reviews and merging the PR it would be good to squash the commits first if possible.

eggfoobar:
Sure thing @yanirq, should be all set.

PR #526, opened 2022-12-09
OCPBUGS-5066: fix profile updates in latency tests

mrniranjan:
backup profile and use initial backup to revert
the profile after latency tests are run.

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
+++ b/test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
@@ -2,39 +2,37 @@ package profilesupdate
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 
+	. "github.com/onsi/gomega"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
 	profilecontroller "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
-	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
 	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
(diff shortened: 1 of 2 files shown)


test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
```diff
@@ -2,39 +2,37 @@ package profilesupdate
...
-func UpdateIsolatedReservedCpus(isolatedSet performancev2.CPUSet, reservedSet performancev2.CPUSet) error {
-	profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
-	if err != nil {
-		return fmt.Errorf("could not get the performance profile: %v", err)
-	}
-	updatedProfile := profile.DeepCopy()
-	updatedProfile.Spec.CPU = &performancev2.CPU{
+func UpdateIsolatedReservedCpus(profile *performancev2.PerformanceProfile, isolatedSet performancev2.CPUSet, reservedSet performancev2.CPUSet) error {
+
+	profile.Spec.CPU = &performancev2.CPU{
```
fromanirh, line 28: this seems a step back though. Why is trusting the given profile more reliable than getting a fresh copy?

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
```diff
@@ -56,7 +54,15 @@ func ApplyProfile(profile *performancev2.PerformanceProfile) error {
...
 	testlog.Info("Applying changes in performance profile and waiting until mcp will start updating")
-	profiles.UpdateWithRetry(profile)
+	spec, err := json.Marshal(profile.Spec)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(testclient.Client.Patch(context.TODO(), profile,
+		client.RawPatch(
+			types.JSONPatchType,
+			[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+		),
+	)).ToNot(HaveOccurred())
```
fromanirh, line 64: I'm not thrilled in adding expectations inside a utility method in this module. Actually I'm not thrilled about a patchwork approach, we should aim for some consistency.
But even before: why is Patch more reliable than Update?

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
mrniranjan: we are getting a fresh copy of the profile and passing that profile to this function.  We are doing this at:

https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/5_latency_testing/5_latency_testing_suite_test.go#L51

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
mrniranjan: I don't have answers now as to why patch works better than updateWithRetry except the issue is reproducible in this test case everytime. Also in the cases where we do use updateWithRetry , we have seen cases where tests do fail with profile updates fail .  I will investigate on this.  I can remove the exceptions and return err.

mrniranjan:
closing this PR as the issue is not reproducible

PR #528, opened 2022-12-13
performance-profile: render: make target for render sync

Tal-or:
For every change in PAO's assets, we should render the manifests and copy the data into `/test/e2e/performanceprofile/testdata/render-expected-output/`

This patch adds a make target to automate the process. For now, the make target itself should be called manually. If it proves itself useful, we can add it as a dependency of other targets.


fromanirh:
/approve
/lgtm
harmless, nice developer-quality-of-life enhancement.
Thanks!

PR #537, opened 2022-12-26
OCPBUGS-5529: Skip offline multiple ranges test if number of cores less than 20

mrniranjan:
Minor fix to GetCpuSiblings function to use core as key instead of cpu siblings

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -1018,7 +1018,6 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				}
 			}
 			isolatedCpus := strings.Join(isolated, ",")
-
 			// Create new performance with offlined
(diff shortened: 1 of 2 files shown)


test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -1093,9 +1091,11 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 					cores = append(cores, k)
 				}
 				sort.Ints(cores)
-				middleCoreIds := cores[len(cores)/2]
-				for i := middleCoreIds; i < middleCoreIds+10; i++ {
-					siblings := nodes.GetCpuSiblings(numaCoreSiblings, i)
+				if len(cores) < 20 {
```
yanirq, line 1094: s/atleast/at least  (missing white space)

yanirq:
Commit message only speakes about the minimal number of core.  
Worth adding also the info about the fix of the key under the same message.

PR #538, opened 2022-12-27
render: initialize klog flag

Tal-or:
klog flags are not initiated for the render command, so
passing klog arguments like `--v` would be ignored

In addition, added a few logs for debugging purposes.

### DIFF
--- a/hack/render-sync.sh
+++ b/hack/render-sync.sh
@@ -5,9 +5,8 @@ ARTIFACT_DIR=$(mktemp -d)
 
 cd "${WORKDIR}" || { echo "failed to change dir to ${WORKDIR}"; exit; }
 _output/cluster-node-tuning-operator render \
---performance-profile-input-files "${WORKDIR}"/test/e2e/performanceprofile/cluster-setup/manual-cluster/performance/performance_profile.yaml \
---asset-input-dir "${WORKDIR}"/build/assets \
+--asset-input-dir "${WORKDIR}"/test/e2e/performanceprofile/cluster-setup/manual-cluster/performance,"${WORKDIR}"/test/e2e/performanceprofile/cluster-setup/base/performance \
 --asset-output-dir "${ARTIFACT_DIR}"
 
-cp -r "${ARTIFACT_DIR}"/*  "${WORKDIR}"/test/e2e/performanceprofile/testdata/render-expected-output/
+cp "${ARTIFACT_DIR}"/*  "${WORKDIR}"/test/e2e/performanceprofile/testdata/render-expected-output
 rm -r "${ARTIFACT_DIR}"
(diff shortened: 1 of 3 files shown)


hack/render-sync.sh
```diff
@@ -5,9 +5,8 @@ ARTIFACT_DIR=$(mktemp -d)
...
 _output/cluster-node-tuning-operator render \
---performance-profile-input-files "${WORKDIR}"/test/e2e/performanceprofile/cluster-setup/manual-cluster/performance/performance_profile.yaml \
---asset-input-dir "${WORKDIR}"/build/assets \
+--asset-input-dir "${WORKDIR}"/test/e2e/performanceprofile/cluster-setup/manual-cluster/performance,"${WORKDIR}"/test/e2e/performanceprofile/cluster-setup/base/performance \
 --asset-output-dir "${ARTIFACT_DIR}"
 
-cp -r "${ARTIFACT_DIR}"/*  "${WORKDIR}"/test/e2e/performanceprofile/testdata/render-expected-output/
-rm -r "${ARTIFACT_DIR}"
+cp "${ARTIFACT_DIR}"/*  "${WORKDIR}"/test/e2e/performanceprofile/testdata/render-expected-output
+rm -rf "${ARTIFACT_DIR}"
```
yanirq, line 12: worth checking that the dir is populated with something right or at least have a default option. this is a bit risky

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -66,6 +66,7 @@ func render(inputDir, outputDir string) error {
 
 	// Read asset directory fileInfo
 	filePaths, err := listFiles(inputDir)
+	klog.V(4).Infof("listed files: %v", filePaths)
```
yanirq, line 69: why V(4) ?

Tal-or: this is for debugging purposes. I don't see the need for that to always be log

hack/render-sync.sh
Tal-or: I'll remove the `-f` but this directory is the artifact of  `mktemp` so I don't think there is something to worry about

pkg/performanceprofile/cmd/render/cmd.go
```diff
@@ -55,7 +56,7 @@ func NewRenderCommand() *cobra.Command {
 	}
 
 	renderOpts.AddFlags(cmd.Flags())
-
+	addKlogFlags(cmd)
```
fromanirh, line 59: (semi-)nit: klog flags, or in general flags pertaining to utility modules, are usually added first, then the application-specific flags.

fromanirh (commented):
/approve
/lgtm

pkg/performanceprofile/cmd/render/cmd.go
Tal-or: Done. Thanks

ffromani (commented):
/approve
/lgtm

ffromani:
/retest
infra issues (and known gcp-pao lane flakiness)

PR #540, opened 2023-01-05
Remove trailing space from test name

sshnaidm:
It confuses test parsers and just not nice.


yanirq:
/rename OCPBUGS-5771: Remove trailing space from test name

yanirq:
/retitle OCPBUGS-5771: Remove trailing space from test name

PR #541, opened 2023-01-06
OCPBUGS-4647: Add authentication to the /metrics endpoint

jmencak:
This change adds authentication to the /metrics endpoint by requiring client key and certificate.

Resolves: OCPBUGS-4647

### DIFF
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -4,26 +4,55 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"time"
 
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/cluster-node-tuning-operator/pkg/util"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"gopkg.in/fsnotify.v1"
-
-	"k8s.io/klog/v2"
 )
 
 const (
 	tlsSecretDir = "/etc/secrets"
 	tlsCert      = tlsSecretDir + "/tls.crt"
 	tlsKey       = tlsSecretDir + "/tls.key"
+
+	authCADir                = "/tmp/metrics-client-ca"
+	authCAFile               = authCADir + "/ca.crt"
+	AuthConfigMapNamespace   = "kube-system"
+	AuthConfigMapName        = "extension-apiserver-authentication"
+	AuthConfigMapClientCAKey = "client-ca-file"
 )
 
 type Server struct {
+}
+
+// DumpCA writes the root certificate bundle which is used to verify client certificates
+// on incoming requests to 'authCAFile' file.
+func DumpCA(ca string) error {
+	if err := util.Mkdir(authCADir); err != nil {
+		return fmt.Errorf("failed to create directory %q: %v", authCADir, err)
+	}
+
+	f, err := os.Create(authCAFile)
+	if err != nil {
+		return fmt.Errorf("failed to create file %q: %v", authCAFile, err)
+	}
+	defer f.Close()
+	if _, err = f.WriteString(ca); err != nil {
+		return fmt.Errorf("failed to write file %q: %v", authCAFile, err)
+	}
+
+	return nil
 }
 
 func buildServer(port int) *http.Server {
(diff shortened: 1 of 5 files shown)


simonpasquier (commented):
IIUC an update of the CA would require a restart of the operator. It would be better to watch for the configmap and dynamically reload the CA if needed?

jmencak:
> IIUC an update of the CA would require a restart of the operator. It would be better to watch for the configmap and dynamically reload the CA if needed?

Thank you for looking, @simonpasquier ! I see you point.  Ideally, I'd like to avoid adding an extra watch for the CA.  As far as I've seen, CA's lifetime is 10 years.  My assumption was that the server cert/key was issued by the same CA (update would be triggered then on server cert/key change), but that doesn't seem to be the case.

jmencak:
@dagrayvid / @simonpasquier , I've added metric's client auth CA reloading after it has been changed.  I believe it is ready for review.

dagrayvid (commented):
Thanks @jmencak. Aside from a couple of small suggestions on comments, this PR looks good to me. I put a hold on the PR to let you decide if you want more input from other reviewers before merging.

/lgtm

jmencak:
> Thanks @jmencak. Aside from a couple of small suggestions on comments, this PR looks good to me. I put a hold on the PR to let you decide if you want more input from other reviewers before merging.
> 
> /lgtm

Thank you @dagrayvid , accepted both suggestions and squashed the commits.  Hope this still looks good.
/hold cancel

dagrayvid:
/hold 
While I look into how to make this work in HyperShift

pkg/metrics/server.go
```diff
@@ -4,28 +4,57 @@ import (
...
-
-	"k8s.io/klog/v2"
 )
 
 const (
 	tlsSecretDir = "/etc/secrets"
 	tlsCert      = tlsSecretDir + "/tls.crt"
 	tlsKey       = tlsSecretDir + "/tls.key"
+
+	authCADir                = "/tmp/" + AuthConfigMapNamespace + "/" + AuthConfigMapName
```
dagrayvid, line 29:
```suggestion
	authCADir                = "/etc/metrics-client-ca"
```
In HyperShift, we will get the CA from a secret rather than from a file created by the NTO controller. Can we change this path to something more general?

jmencak:
/hold
I don't hink the operator will be able to create the directory in /etc as suggested by David by the latest comment.

jmencak:
> I don't hink the operator will be able to create the directory in /etc as suggested by David by the latest comment.

The alternative could be to make changes to the Containerfiles to have the `/var/lib/tuned` directory owned by uid `499` and let the operator create a directory there. The home directory contains some performance profile controller artifacts though:

```
sh-4.4$ pwd
/var/lib/tuned
sh-4.4$ ls
bin  performanceprofile  provider
sh-4.4$ find performanceprofile/
performanceprofile/
performanceprofile/assets.go
performanceprofile/configs
performanceprofile/configs/99-netdev-physical-rps.rules
performanceprofile/configs/99-netdev-rps.rules
performanceprofile/configs/99-runtimes.conf
performanceprofile/scripts
performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
performanceprofile/scripts/hugepages-allocation.sh
performanceprofile/scripts/set-cpus-offline.sh
performanceprofile/scripts/set-rps-mask.sh
performanceprofile/tuned
performanceprofile/tuned/openshift-node-performance
```

@yanirq :
1) are those file needed there
2) if they are needed, can we change the owner of the entire `/var/lib/tuned` to uid 499?

jmencak:
/hold cancel
Using "/tmp/metrics-client-ca".

pkg/metrics/server.go
```diff
@@ -4,28 +4,57 @@ import (
...
+	if err := util.Mkdir(authCADir); err != nil {
+		return fmt.Errorf("failed to create directory %q: %v", authCADir, err)
+	}
+
+	f, err := os.Create(authCAFile)
+	if err != nil {
+		return fmt.Errorf("failed to create file %q: %v", authCAFile, err)
+	}
+	defer f.Close()
+	if _, err = f.WriteString(ca); err != nil {
```
simonpasquier, line 51: Creating and writing to the file might generate several fsnotify events which could trigger several recreations of the server?

pkg/util/os.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import (
+	"os"
+)
+
+// Create a directory including its parents.
```
dagrayvid, line 7: I was confused for a moment since Mkdir is called both in DumpCA() and RunServer()

Might be worth mentioning the behaviour in the case where the directory already exists in the comment.
```suggestion
// Create a directory including its parents. Returns nil if directory already exists.
```

pkg/metrics/server.go
jmencak: I believe this is very unlikely.  I was able to write a 1M test file with a single WRITE event.  Also, inotify does only fire up after a write is completed.  What will happen though is two WRITE events when the `authCAFile` already exists.  The first one will create an empty file via `os.Create()` and the second WRITE will write the actual data.  However, we account for empty files in the code so no extra recreations of the serve will happen.

pkg/util/os.go
jmencak: It is "mkdir -p", so it does return 0/nil :) Will add the comment, thank you.

jmencak:
I believe it is time to
/override e2e-gcp-pao-updating-profile

PR #550, opened 2023-01-19
OCPBUGS-6057: Skip watch on CA ConfigMap in HyperShift

dagrayvid:
In HyperShift environments, don't watch the CA ConfigMap. 

In HyperShift the CA cert is in the hosted control plane namespace along with NTO, so we can mount it into the operator Pod.

/cc @jmencak

### DIFF
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -16,6 +16,7 @@ import (
 
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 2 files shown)


jmencak:
Thank you for the changes, David.  The changes make sense to me.  Please unhold when ready.
/lgtm
/hold
/retest

PR #556, opened 2023-01-29
OCPBUGS-4194: rps: reapply mask if not all queues are set

Tal-or:
In OCP >= 4.12 a new feature added named UDP GRO.
As part of this this feature implemention, virtual
interfaces get created with multiple queues.

From udev even perspective the flow look as follow:
  1. New veth gets created
  2. The kernel is adding all rx queues 0-N (N=<number of cpus>)
  3. The kernel is removing all rx queues 1-N
  4. The kernel is adding all rx queues 1-N again

A new veth creation event invokes the set-rps script,
but it might be running before all NIC's queues get created.
Hence the rps mask not apply to the uncreated queues, which
get created later on without the mask.

This change applies the mask again if not all queues were set.

This is an alternative approach to
https://github.com/openshift/cluster-node-tuning-operator/pull/527
which led to performance degradation on SNO cluster.


assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -33,6 +38,8 @@ function find_dev_dir {
 [ -d "${dev_dir}" ] || { sleep 5; find_dev_dir; }  # search failed, wait a little and try again
 [ -d "${dev_dir}" ] || { echo "${dev_dir}" directory not found >&2 ; exit 0; } # the interface disappeared, not an error
 
-for i in "${dev_dir}"/queues/rx-*/rps_cpus; do
-  echo "${mask}" > "${i}"
+set-rps-mask
+# check that all queues were setup correctly
+while [[ $(cat /sys/devices/virtual/net/*/queues/*/rps_cpus | sort -u | wc -l) != "1" ]]; do
```
MarSik, line 43: Isn't the main issue with this approach that there is a race condition here?

Tal-or: Theoretically yes, it's not perfect but it will decrease the problem rate significantly and make the CI much less flaky. The CI is falling most of the time because of the test that checks all queues are properly set. we can have this change til a proper solution will be found.

Tal-or:
We got proper fix on kernel side, but it still require some changes for interacting with the new API exposed by the kernel and deserves a separate PR. Closing.

PR #559, opened 2023-01-30
Configure OVS for dynamic cpu pinning

MarSik:
This is needed to avoid cpu starving the reserved cpus when a lot of kernel traffic is expected. OVS is by default under the system.slice that is statically restricted to the reserved cpus.

The separate slice will give OVS unrestricted access to all cpus and the code being introduced to OVN-K will link it to the cpu mask that is assigned to burstable pods by cpu manager.

The end result is dynamic scaling of resources available to OVS based on the amount of cpu pinned containers.

This implements the NTO part of https://github.com/openshift/enhancements/blob/master/enhancements/network/ovs-dynamic-cpu-pinning.md


MarSik:
This depends on https://github.com/openshift/ovn-kubernetes/pull/1697

assets/performanceprofile/scripts/cpuset-configure.sh
```diff
@@ -36,3 +39,25 @@ mkdir -p "$machine"
 
 # It's unlikely, but possible, that this cpuset already existed. Iterate just in case.
 for file in $(find "$machine" -name cpuset.cpus | sort -r); do echo "$reserved_set" > "$file"; done
+
+# OVS is running in its own slice that spans all cpus. The real affinity is managed by OVN-K ovnkube-node daemonset
+# Make sure this slice will not enable cpu balancing for other slice configured by this script.
+# This might seem counter-intuitive, but this will actually NOT disable cpu balancing for OVS itself.
+# - OVS has access to reserved cpus, but those have balancing enabled via the `system` cgroup created above
+# - OVS has access to isolated cpus that are currently not assigned to pinned pods. Those have balancing enabled by the
```
Tal-or, line 47: What about the case when workload partition enabled? 
Will it affine OVS to reserved?

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -318,9 +327,51 @@ func getIgnitionConfig(profile *performancev2.PerformanceProfile, pinningMode *a
...
+			return nil, err
+		}
+		addContent(ignitionConfig, content, ovsDynamicPinningTriggerHostFile, &ovsMode)
+	}
+
 	return ignitionConfig, nil
 }
 
+func MoveOvsIntoOwnSlice() (bool, string) {
+	// Make sure this does not interfere with SNO and workload partitioning
```
Tal-or, line 366: So restricting OVS to reserved cpus would be done at later stage by the OVN-K if it sees that WP enabled?

Tal-or:
Can we add e2e test for that? at least checking the files are in place and system configured correctly?

assets/performanceprofile/scripts/cpuset-configure.sh
```diff
@@ -36,3 +39,25 @@ mkdir -p "$machine"
...
+# Create the ovs.slice
+mkdir -p "$ovsslice"
+echo 0 > "$ovsslice"/cpuset.sched_load_balance
+cat "$root"/cpuset.cpus > "$ovsslice"/cpuset.cpus
+cat "$root"/cpuset.mems > "$ovsslice"/cpuset.mems
+
+# Move OVS over
+for process in $(cat "$ovsslice_systemd"/*/cgroup.procs | sort -r); do
+        echo $process > "$ovsslice"/cgroup.procs 2>&1 | grep -v "Invalid Argument" || true;
+done
```
ffromani, line 63: why do we need the sort and the grep?

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -318,9 +327,51 @@ func getIgnitionConfig(profile *performancev2.PerformanceProfile, pinningMode *a
...
+	// where OVS is intentionally still running in reserved only due to
+	// workload partitioning restricting the cpuset for OVN-K pods.
+	//
+	// This will be propagated by the ovkube-node cpu affinity logic
+	// and restrict OVS to reserved only.
+	return true, ovsSliceName
```
ffromani, line 372: why do we need a function to essentially return two constants?

ffromani (commented):
from my side, minor comments atm. How do we test this functionality, in general?

assets/performanceprofile/scripts/cpuset-configure.sh
MarSik: Yes, the ovnkube-node pod will only run on reserved and pass that affinity to OVS too.

MarSik: Once, at cluster startup time. Then it will never change.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
MarSik: It will not specifically care about WP, rather WP will make sure the affinity of the pod is right and the pod will just apply it to OVS too.

This will actually allow the accelerated startup to work properly, because the affinity will be wide during the boot sequence and then restricted once the accelerated startup is reverted.

assets/performanceprofile/scripts/cpuset-configure.sh
MarSik: The grep just filters out processes that might have existed meanwhile. The sort I honestly just copied from the existing logic Peter has developed for moving processes to the system slice.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
MarSik: We don't, but I wanted to have a clean code "hook point" for workload partitioning logic and the related comment. It is an artifact that I decided to keep for future development, because it allows nice grouping of the code for this feature.

MarSik:
@Tal-or The render test makes sure the files are present in the generated content. The e2e test is going to be more complex as we should test the behavior properly to avoid the same issues we are seeing with cpu balancing. And hoped we could develop it with Niranjan in a later PR.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
Tal-or: The ovn-k pods is passing the information about the allowed cpus to ovs?

MarSik: Yes. The ovnkube-node Pod is setting the OVS affinity.

assets/performanceprofile/scripts/cpuset-configure.sh
ffromani: ok

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
ffromani: having a "hook point" is no bother

ffromani (commented):
what ultimately bothers me is that this feature seems to be always enabled. Or am I missing something? If it is always enabled, we need to ensure the test coverage is sufficient to prevent (the most egregious) regressions. I'm not sure we have test coverage for that, do we?
I'd love to have an emergency disable switch of _any_ form - not necessarily surfacing at API layer (probably it shouldn't). How hard would be to add this global switch? Then if we're not sufficiently confident about the test coverage (and my understanding the e2e testing is struggling because other platform-level cgroups-related bugs) we can force the global switch to OFF, file a bug to enable it, make it depend on other platform bugs, and move forward and merge.

ffromani:
> what ultimately bothers me is that this feature seems to be always enabled. Or am I missing something? If it is always enabled, we need to ensure the test coverage is sufficient to prevent (the most egregious) regressions. I'm not sure we have test coverage for that, do we?

Thinking more about it, always enabled also means all the tests always run, and the lanes are green.
So, despite my fuzzy feeling about having the feature always enabled, it boils down to the test coverage. 
If the test coverage is enough to fully exercise the changes and make us reasonnably confident there are no regressions, we can move ahead. Is that the case? anything missing that would increase our confidence? Do we need more runs?

MarSik:
@ffromani 

> what ultimately bothers me is that this feature seems to be always enabled. Or am I missing something? 
>  I'd love to have an emergency disable switch of _any_ form

The switch is there, see this file:

https://github.com/openshift/cluster-node-tuning-operator/pull/559/files#diff-650f8a6609adebbf781f4fbb22f5113b523f58d356a8c94f224923764e57f94d

Deleting the file or at least the content (doable with a MC) stops the dynamic pinning at ovnkube-node level.

And as I said before, I wan't a functional e2e test and we have a test plan for it. But implementing it now would make the lanes RED because of https://issues.redhat.com/browse/OCPBUGS-15102. Making it conditional is possible, but since it will be quite a bit of work to prepare it I would prefer doing it in a separate PR.

MarSik:
Ok, I added a very rudimentary e2e test and lets see what happens.

MarSik:
@ffromani Basic e2e tests included now.

MarSik:
Just for the record:

This PR does not change the behavior or pinning of OVS in any way, unless ovnkube-node starts doing its work too.

OVS services were attached to the root slice in the past: that means all cpus, balancing enabled (4.12) or disabled (4.13). 
This PR moves OVS to its own slice, however it configures it in the same way - all cpus, cpu balancing disabled.

The only reason for doing this is to move OVS out of the scope of this https://github.com/openshift/cluster-node-tuning-operator/blob/master/assets/performanceprofile/scripts/cpuset-configure.sh#L28 loop that moves all system services to a new `system` cgroup and that limits the cpuset for those.

PR #561, opened 2023-02-02
Refactor IRQ load balancing enable/disable test

yanirq:
Moved enable/disable test under irqbalance file context for better grouping.

Change the use of nodes.BannedCPUs to getIrqBalanceBannedCPUs Inroder not to fail on empty values.
This will also spread the load between the e2e-pao and the e2e-pao-update lanes.


yanirq:
> could you please split the commit in
> 
>     1. trivial code movement without further changes
> 
>     2. changes?
> 
> 
> this will make the review and the backport/cherry-pick (if needed) way easier

I'm not a big fan of splitting commits unless we must have the split.
Also - this is currently a change that will be done in 4.13 and above (unless we split lanes in 4.12 too)
The changes are actually trivial as well since the only change is changing the use of getting banned cpus to a function that was already created for that in previous PR.

yanirq:
> > > could you please split the commit in
> > > ```
> > > 1. trivial code movement without further changes
> > > 
> > > 2. changes?
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > this will make the review and the backport/cherry-pick (if needed) way easier
> > 
> > 
> > I'm not a big fan of splitting commits unless we must have the split.
> 
> OK, but what's the value of keeping the (unrelated) changes squashed here?

Maybe the message is a bit over explaining and could be change. 
The major change here is to re  enable the skipped test + move the entire test to a different file (and lane) + change a use of one function to another existing function : BannedCPUs(node) -> getIrqBalanceBannedCPUs

PR #562, opened 2023-02-06
OCPBUGS-7066: Add test case to check single rx queue on veth interface

mrniranjan:
This  test checks on a specific veth interface
if a single rx-queue has rps mask equivalent of  reserved cpus mentioned in the profile

Include user-friendly comments in test cases that verify the partial application of RPS masks to all RX queues. This is a known Bug Refer: OCPBUGS-4194

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -290,15 +290,53 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 	})
 
 	Context("RPS configuration", func() {
-		It("[test_id:55012] Should have the correct RPS configuration", func() {
+		BeforeEach(func() {
 			if profile.Spec.CPU == nil || profile.Spec.CPU.Reserved == nil {
 				Skip("Test Skipped due nil Reserved cpus")
 			}
+		})
+		It("[test_id: 59572] Check RPS Mask is applied to atleast one single rx queue on all veth interface", func() {
 			if profile.Spec.WorkloadHints != nil && profile.Spec.WorkloadHints.RealTime != nil &&
 				!*profile.Spec.WorkloadHints.RealTime && !profileutil.IsRpsEnabled(profile) {
 				Skip("realTime Workload Hints is not enabled")
 			}
-
+			count := 0
+			expectedRPSCPUs, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
+			Expect(err).ToNot(HaveOccurred())
+			var vethInterfaces = []string{}
+			for _, node := range workerRTNodes {
+				allInterfaces, err := nodes.GetNodeInterfaces(node)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(allInterfaces).ToNot(BeNil())
+				// collect all vethinterfaces in a list
+				for _, iface := range allInterfaces {
+					if iface.Bridge == true && iface.Physical == false {
+						vethInterfaces = append(vethInterfaces, iface.Name)
+					}
+				}
+				//iterate over all the vethinterface and
+				//check if at least on single rx-queue has rps mask
+				for _, vethinterface := range vethInterfaces {
+					devicePath := fmt.Sprintf("%s/%s", "/rootfs/sys/devices/virtual/net", vethinterface)
+					getRPSMaskCmd := []string{"find", devicePath, "-type", "f", "-name", "rps_cpus", "-exec", "cat", "{}", ";"}
+					devsRPS, err := nodes.ExecCommandOnNode(getRPSMaskCmd, &node)
+					Expect(err).ToNot(HaveOccurred())
+					for _, devRPS := range strings.Split(devsRPS, "\n") {
+						rpsCPUs, err := components.CPUMaskToCPUSet(devRPS)
+						Expect(err).ToNot(HaveOccurred())
+						if rpsCPUs.String() == string(*profile.Spec.CPU.Reserved) {
+							count += 1
+						}
+					}
+					Expect(count > 0).To(BeTrue(), "Not a single receive queues have cpu mask %v", expectedRPSCPUs)
+				}
+			}
+		})
+		It("[test_id:55012] Should have the correct RPS configuration", func() {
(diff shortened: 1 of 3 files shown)


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -290,15 +290,47 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+			if profile.Spec.WorkloadHints != nil && profile.Spec.WorkloadHints.RealTime != nil &&
+				!*profile.Spec.WorkloadHints.RealTime && !profileutil.IsRpsEnabled(profile) {
+				Skip("realTime Workload Hints is not enabled")
+			}
+			expectedRPSCPUs, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
+			Expect(err).ToNot(HaveOccurred())
+			for _, node := range workerRTNodes {
+				//Get one veth interface
+				count := 0
+				cmd := []string{"/bin/sh", "-c", fmt.Sprintf("ip -c link show type veth | head -n 1")}
```
ffromani, line 308: this will get the first veth in the kernel order. So no guarantees about which veth will we work with. Is this good enough for this test? if so deserves a comment

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -290,15 +290,47 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+				vethInterfaceOutput, err := nodes.ExecCommandOnNode(cmd, &node)
+				Expect(err).ToNot(HaveOccurred())
+				vethParts := strings.Split(vethInterfaceOutput, " ")
+				output := strings.Split(vethParts[1], "@")[0]
+				// we need to ensure there are no escape characters
```
ffromani, line 313: from where the escape characters can come from?

ffromani (commented):
initial review

test/e2e/performanceprofile/functests/1_performance/performance.go
Tal-or: Maybe checking under `/sys/class/net/` is safer

test/e2e/performanceprofile/functests/1_performance/performance.go
Tal-or: the `ip`  command output

Tal-or (commented):
I think the original intent was to check the first queue for all the veth, isn't it?

test/e2e/performanceprofile/functests/1_performance/performance.go
mrniranjan: Some times i get the escape characters "\x1b[36m" in the ip command output

mrniranjan:
> I think the original intent was to check the first queue for all the veth, isn't it?

what if the rx-0 did not have the rps mask , I mean due to race conditions any rx queue can get missed right ?

test/e2e/performanceprofile/functests/1_performance/performance.go
mrniranjan: @Tal-or I have used /sys/clas/net in the latest commit. Please check

test/e2e/performanceprofile/functests/1_performance/performance.go
mrniranjan: @Tal-or @ffromani I have used the /sys/class/net instead of ip command. Please check the latest commit.

Tal-or:
> > I think the original intent was to check the first queue for all the veth, isn't it?
> 
> what if the rx-0 did not have the rps mask , I mean due to race conditions any rx queue can get missed right ?

The race-condition happened because the queues 1-N are removed and then get created again. But queue 0 is the default one who never gets deleted. We also saw it together - all queues in particular NIC weren't set with the correct mask but the first queue.

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -290,15 +290,54 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+			expectedRPSCPUs, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
+			Expect(err).ToNot(HaveOccurred())
+			var vethInterfaces = []string{}
+			for _, node := range workerRTNodes {
+				allInterfaces, err := nodes.GetNodeInterfaces(node)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(allInterfaces).ToNot(BeNil())
+				// collect all vethinterfaces in a list
+				for _, iface := range allInterfaces {
+					if iface.Bridge == true && iface.Physical == false {
```
Tal-or, line 313: ```iface.Bridge == true```? I believe you mean ```false```

mrniranjan: @Tal-or no it should be true as veth interfaces are part bridge ovs-system.

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -290,15 +290,54 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+						vethInterfaces = append(vethInterfaces, iface.Name)
+					}
+				}
+				//iterate over all the vethinterface and
+				//check if at least on single rx-queue has rps mask
+				for _, vethinterface := range vethInterfaces {
+					devicePath := fmt.Sprintf("%s/%s", "/rootfs/sys/devices/virtual/net", vethinterface)
+					getRPSMaskCmd := []string{"find", devicePath, "-type", "f", "-name", "rps_cpus", "-exec", "cat", "{}", ";"}
+					fmt.Println(getRPSMaskCmd)
```
Tal-or, line 322: Printing it every time for every device won't spam the log?
I would print the command only if the test failed, but this change can be done on later PR

mrniranjan: @Tal-or ah sorry, i was doing some debugging and added it, i can remove it , its a small fix, WDYT ?

yanirq:
/hold
lets fix the excessive print outs first

test/e2e/performanceprofile/functests/1_performance/performance.go
Tal-or: Thanks for the fix

PR #564, opened 2023-02-08
OCPBUGS-5529: Fix updating numa core siblings map in GetCpuSiblings function

mrniranjan:
move the delete operation outside of for loop.
Also minor fixes to offline tests to check if the worker node has 20 cores earlier in the test function

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -1442,7 +1442,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				// Select the last core id
 				higherCoreIds := cores[len(cores)-1]
 				// Get cpu siblings from the selected cores and delete the selected cores  from the map
-				cpusiblings := nodes.GetCpuSiblings(numaCoreSiblings, higherCoreIds)
+				cpusiblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, higherCoreIds)
 				offline = append(offline, cpusiblings...)
 			}
 			offlineCpus := strings.Join(offline, ",")
@@ -1450,7 +1450,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			for reservedCores := 0; reservedCores < 2; reservedCores++ {
 				// Get the cpu siblings from the selected core and delete the siblings
 				// from the map. Selected siblings of cores are saved in reservedCpus
-				cpusiblings := nodes.GetCpuSiblings(numaCoreSiblings, reservedCores)
+				cpusiblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, reservedCores)
 				reserved = append(reserved, cpusiblings...)
 			}
 			reservedCpus := strings.Join(reserved, ",")
@@ -1458,7 +1458,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			// numaCoreSiblings map is used in isolatedCpus
 			for key := range numaCoreSiblings {
 				for k := range numaCoreSiblings[key] {
-					cpusiblings := nodes.GetCpuSiblings(numaCoreSiblings, k)
+					cpusiblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, k)
 					isolated = append(isolated, cpusiblings...)
 				}
 			}
@@ -1512,14 +1512,14 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				}
 				sort.Ints(cores)
 				middleCoreIds := cores[len(cores)/2]
-				siblings := nodes.GetCpuSiblings(numaCoreSiblings, middleCoreIds)
+				siblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, middleCoreIds)
 				offline = append(offline, siblings...)
 			}
 			offlineCpus := strings.Join(offline, ",")
 			for reservedCores := 0; reservedCores < 2; reservedCores++ {
 				// Get the cpu siblings from the selected core and delete the siblings
 				// from the map. Selected siblings of cores are saved in reservedCpus
-				siblings := nodes.GetCpuSiblings(numaCoreSiblings, reservedCores)
+				siblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, reservedCores)
 				reserved = append(reserved, siblings...)
 			}
 			reservedCpus := strings.Join(reserved, ",")
@@ -1527,7 +1527,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			// numaCoreSiblings map is used in isolatedCpus
 			for key := range numaCoreSiblings {
 				for k := range numaCoreSiblings[key] {
-					siblings := nodes.GetCpuSiblings(numaCoreSiblings, k)
+					siblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, k)
 					isolated = append(isolated, siblings...)
 				}
 			}
@@ -1587,7 +1587,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 			for reservedCores := 0; reservedCores < 2; reservedCores++ {
 				// Get the cpu siblings from the selected core and delete the siblings
 				// from the map. Selected siblings of cores are saved in reservedCpus
-				siblings := nodes.GetCpuSiblings(numaCoreSiblings, reservedCores)
+				siblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, reservedCores)
 				reserved = append(reserved, siblings...)
 			}
 			reservedCpus := strings.Join(reserved, ",")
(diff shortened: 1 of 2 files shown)


test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -436,12 +436,14 @@ func GetByCpuCapacity(nodesList []corev1.Node, cpuQty int) []corev1.Node {
...
 func GetCpuSiblings(numaCoreSiblings map[int]map[int][]int, coreKey int) []string {
 	var cpuSiblings []string
+	var numanode int
 	for key := range numaCoreSiblings {
-		for _, c := range numaCoreSiblings[key][coreKey] {
-			cpuSiblings = append(cpuSiblings, strconv.Itoa(c))
-			delete(numaCoreSiblings[key], coreKey)
+		for _, sibling := range numaCoreSiblings[key][coreKey] {
+			cpuSiblings = append(cpuSiblings, strconv.Itoa(sibling))
+			numanode = key
```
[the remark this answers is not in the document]
MarSik: Oh, nevermind, I see what you did here.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -436,12 +436,14 @@ func GetByCpuCapacity(nodesList []corev1.Node, cpuQty int) []corev1.Node {
...
 		}
 	}
+	delete(numaCoreSiblings[numanode], coreKey)
```
MarSik, line 446: But shouldn't this be inside the numaCoreSibligs loop and use the corekey? Your code will not work if the siblings are not coming from the same NUMA node anyway.

MarSik: Why not simply use:

```
for key := range numaCoreSiblings {
    for _, c := range numaCoreSiblings[key][coreKey] {
        cpuSiblings = append(cpuSiblings, strconv.Itoa(c))
    }
    delete(numaCoreSiblings[key], coreKey)
}
```

mrniranjan: If we put the delete operation inside the outer for loop, the key will get modified because the for loop continues further . So we will have to add a break and end the loop , otherwise it will endup removing a some other core or fail

mrniranjan: I have updated the Patch with a if condition to check if the map has core key exists , if exist, we move to the internal for loop.

jlojosnegros: First I have my concerns about a "Get" function that modifies its input parameters... 

Regarding delete operation. If I understood correctly you are gonna end deleting all the CPUs of  a Core. All of them would usually belong to the same NUMA node so deleting them at the end should work ( I think) but it is confusing. 
Either you enforce the fact that all the cpus in a core must belong to one NumaNode or you create a list of numanodes in the for loops and then delete them afterwards... something like
```
func GetCpuSiblings(numaCoreSiblings map[int]map[int][]int, coreKey int) []string {
	var cpuSiblings []string
	var targetNumaNodes []int
	for numaNode := range numaCoreSiblings {
		for _, c := range numaCoreSiblings[numaNode][coreKey] {
			cpuSiblings = append(cpuSiblings, strconv.Itoa(c))
			targetNumaNodes = append(targetNumaNodes, numaNode)
		}
	}

	for numaNode := range targetNumaNodes {
		delete(numaCoreSiblings[numaNode], coreKey)
	}

	return cpuSiblings
}
```

or even better extract deletion to another function.

mrniranjan: @jlojosnegros  , let me explain the use case more elaborately :

 For example, i wanted to offline cpus from higher core ids 
 1. For the first time, i query the worker nodes and populate the map with numa node information which may bee like this:
 
`Map  =  map[0:map[0:[0 40] 2:[2 42] 4:[4 44] 6:[6 46] 8:[8 48] 10:[10 50] 12:[12 52] 14:[14 54] 16:[16 56] 18:[18 58] 20:[20 60] 22:[22 62] 24:[24 64] 26:[26 66] 28:[28 68] 30:[30 70] 32:[32 72] 34:[34 74] 36:[36 76] 38:[38 78]] 1:map[1:[1 41] 3:[3 43] 5:[5 45] 7:[7 47] 9:[9 49] 11:[11 51] 13:[13 53] 15:[15 55] 17:[17 57] 19:[19 59] 21:[21 61] 23:[23 63] 25:[25 65] 27:[27 67] 29:[29 69] 31:[31 71] 33:[33 73] 35:[35 75] 37:[37 77] 39:[39 79]]]`

Now i need pickup 38th core from node0 and 39th core from node 1 , this is stright forward, 
  a. loop to each numa node 
  b. sort all the cores from a numa node
  c. pick the highest core id  , gets its siblings, in a offlineCpuList 
  d. once fetched we remove that core from the Map 
  
Now i need cpusiblings from reserved cpus,  For that i am taking core 0, 1, 2 as my reserved cores: In this case core 0 , 2 fall on numa node 0 and core 1 on numa node 1 

Now in the GetCpuSiblings function, i passing the map and core id, (core ids are always unique). So i query each node, if core exist in that map, fetch the cpu siblings and remove them from the map .  

The reason i need to delete the coreid once fetched is because, when i  want cpus for isolated, I can use all the remaining core siblings, if i don't delete, i will get the cpu siblings which are already used in offline and reserved. 

The code  to me looks simple  , 
```
func GetCpuSiblings(numaCoreSiblings map[int]map[int][]int, coreKey int) []string {                                                                                                                           
         var cpuSiblings []string                                                                                                                                                                              
         // key here is the Numa Node                                                                                                                                                                          
         for key := range numaCoreSiblings {                                                                                                                                                                   
                 // Here we check in that specific numa node , the coreid exists                                                                                                                               
                 _, ok := numaCoreSiblings[key][coreKey]                                                                                                                                                       
                 // if exists                                                                                                                                                                                  
                 if ok {                                                                                                                                                                                       
                         for _, sibling := range numaCoreSiblings[key][coreKey] {                                                                                                                              
                                 cpuSiblings = append(cpuSiblings, strconv.Itoa(sibling))                                                                                                                      
                         }                                                                                                                                                                                     
                         // once the siblings are returned remove the core                                                                                                                                     
                         // from which the siblings are returned                                                                                                                                               
                         delete(numaCoreSiblings[key], coreKey)                                                                                                                                                
                 }                                                                                                                                                                                             
          }                                                                                                                                                                                                     
         return cpuSiblings                                                                                                                                                                                    
 }     
```

Now as you suggested its possible to remove the delete Operation to another function, in which case i have to duplicate some of the above like this:

```
func DeleteCpuSiblings(numaCoreSiblings map[int]map[int][]int, coreKey int) {                                                                                                                                 
         for key := range numaCoreSiblings {                                                                                                                                                                   
                _, ok := numaCoreSiblings[key][coreKey]                                                                                                                                                       
                 if ok {                                                                                                                                                                                       
                         delete(numaCoreSiblings[key], coreKey)                                                                                                                                                
                }                                                                                                                                                                                             
         }                                                                                                                                                                                                     
}   
```
I can do it this , let me know your views.

jlojosnegros: Ok ... I understand what you wanna do. 

My only point here, and maybe it is just a matter of my "taste" so to say, is that I would not expect a "get" function to modify its parameters. I know it is just a matter of naming ... so not a big deal if everybody else is ok with that.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -445,9 +445,16 @@ func GetByCpuCapacity(nodesList []corev1.Node, cpuQty int) []corev1.Node {
 // Also updates the map by deleting the cpu siblings returned
 func GetCpuSiblings(numaCoreSiblings map[int]map[int][]int, coreKey int) []string {
 	var cpuSiblings []string
+	// key here is Numa node id
 	for key := range numaCoreSiblings {
```
jlojosnegros, line 449: minor comment : as this is numa node id maybe using `numaKey` ?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
mrniranjan: @ffromani @jlojosnegros Thanks for inputs, i have renamed the function to GetAndRemoveCpuSiblingsFromMap ,

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -1442,23 +1442,23 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				// Select the last core id
 				higherCoreIds := cores[len(cores)-1]
 				// Get cpu siblings from the selected cores and delete the selected cores  from the map
-				cpusiblings := nodes.GetCpuSiblings(numaCoreSiblings, higherCoreIds)
+				cpusiblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, higherCoreIds)
```
ffromani, line 1445: next time: `Pop` may be a shorter sinonym of `GetAndRemove`. I won't resubmit for this.

ffromani (commented):
naming good enough for me.
@jlojosnegros for final LGTM

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
mrniranjan: Thanks and noted

ffromani:
failures in the `e2e-gcp-pao` are (very) unlikely to be caused or worsened by this PR and should be investigated in another PR.

PR #566, opened 2023-02-13
Report host-level sysctls in conflict with TuneD ones

jmencak:
Kernel parameters can be changed at runtime via various configuration files in `/run/sysctl.d/*.conf`, `/etc/sysctl.d/*.conf` and `/etc/sysctl.conf`

By default, NTO's node agents do not override these settings.  This matches the default TuneD behaviour `reapply_sysctl=1` in `tuned-main.conf`.

While NTO can be configured to set `reapply_sysctl=0` in `tuned-main.conf` to let TuneD manage all sysctls and override the host-level configuration files, many users do not know about this functionality.

This PR makes this potential issue more visible to users by reporting it issue via making the Profile Degraged with a status message set.

Also see:
* https://github.com/redhat-performance/tuned/pull/492
* https://github.com/redhat-performance/tuned/pull/501

### DIFF
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -41,6 +41,7 @@ const (
 	scApplied Bits = 1 << iota
 	scWarn
 	scError
+	scSysctlOverride
 	scTimeout
 	scUnknown
 )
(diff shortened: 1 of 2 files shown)


jmencak:
A note to reviewers/QE.  In order for this to work, `TuneD` which ships with at least https://github.com/redhat-performance/tuned/pull/492 is needed for this to work.

jmencak:
/cc @liqcui 
for pre-merge testing.

liqcui:
/lgtm
pre-merge testing passed, work as expected.

PR #576, opened 2023-03-01
kubectl explain to explicitly state workloadHint default values

andreaskaris:
kubectl explain to explicitly state worloadHint default values

Even though default values are set for the API, when kubebuilder builds
the CRDs it does not give end users a hint about the default values of
the fields. Improve the field comments to mention the default field
values. This way, admins who run oc explain can clearly see that the
realTime field has a different default value than the other 2 fields.

-----------------------

 Hacking instructions: Deploy custom tuning operator image on OpenShift

Add instructions to deploy the custom tuning operator image on
OpenShift.

-----------------------

With the change:
~~~
[akaris@linux perf-profile-creator]$ oc explain performanceprofile.spec.workloadHints 
KIND:     PerformanceProfile
VERSION:  performance.openshift.io/v2

RESOURCE: workloadHints <Object>

DESCRIPTION:
     WorkloadHints defines hints for different types of workloads. It will allow
     defining exact set of tuned and kernel arguments that should be applied on
     top of the node.

FIELDS:
   highPowerConsumption	<boolean>
     HighPowerConsumption defines if the node should be configured in high power
     consumption mode. The flag will affect the power consumption but will
     improve the CPUs latency. Defaults to false.

   perPodPowerManagement	<boolean>
     PerPodPowerManagement defines if the node should be configured in per pod
     power management. PerPodPowerManagement and HighPowerConsumption hints can
     not be enabled together. Defaults to false.

   realTime	<boolean>
     RealTime defines if the node should be configured for the real time
     workload. Defaults to true.
~~~

### DIFF
--- a/pkg/apis/performanceprofile/v2/performanceprofile_types.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_types.go
@@ -176,16 +176,16 @@ type RealTimeKernel struct {
 // WorkloadHints defines the set of upper level flags for different type of workloads.
 type WorkloadHints struct {
 	// HighPowerConsumption defines if the node should be configured in high power consumption mode.
-	// The flag will affect the power consumption but will improve the CPUs latency.
+	// The flag will affect the power consumption but will improve the CPUs latency. Defaults to false.
 	// +optional
 	HighPowerConsumption *bool `json:"highPowerConsumption,omitempty"`
-	// RealTime defines if the node should be configured for the real time workload.
+	// RealTime defines if the node should be configured for the real time workload. Defaults to true.
 	// +default=true
 	// +optional
 	RealTime *bool `json:"realTime,omitempty"`
 	// +optional
 	// PerPodPowerManagement defines if the node should be configured in per pod power management.
-	// PerPodPowerManagement and HighPowerConsumption hints can not be enabled together.
+	// PerPodPowerManagement and HighPowerConsumption hints can not be enabled together. Defaults to false.
 	PerPodPowerManagement *bool `json:"perPodPowerManagement,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


andreaskaris:
@MarSik

ffromani:
/lgtm

both worthy fixes, thanks. I'd have actually filed two PRs, but no big deal

MarSik (approved):
/lgtm

pkg/apis/performanceprofile/v2/performanceprofile_types.go
```diff
@@ -176,16 +176,16 @@ type RealTimeKernel struct {
...
 	HighPowerConsumption *bool `json:"highPowerConsumption,omitempty"`
-	// RealTime defines if the node should be configured for the real time workload.
+	// RealTime defines if the node should be configured for the real time workload. Defaults to true.
 	// +default=true
 	// +optional
 	RealTime *bool `json:"realTime,omitempty"`
 	// +optional
 	// PerPodPowerManagement defines if the node should be configured in per pod power management.
-	// PerPodPowerManagement and HighPowerConsumption hints can not be enabled together.
+	// PerPodPowerManagement and HighPowerConsumption hints can not be enabled together. Defaults to false.
```
ffromani, line 188: I'm not blocking for this, but I'm not completely sold about describing the default value in the CRD. On one hand, defaults are unlikely to change. On the other hand, I don't think this balongs into the CRD, which defines the purpose of the fields. Values should be ideally learned querying the objects (status) not the CRD.

andreaskaris: Yeah, I absolutely agree in most other circumstances. Unfortunately, the default values between the 3 fields are different with realTime being true, the other 2 being false. And the oc adm explain just doesn't show the default values to users ... and neither does our documentation which I think is largely derived from the oc explain output. I think this is one of the very, very rare cases where it makes sense in order to avoid user confusion, even though it is ugly

PR #577, opened 2023-03-01
Add a script to validate upstream TuneD

jmencak:
Adding `hack/test-upstream-tuned.sh` script to run NTO and PerformanceProfile controller tests on a cluster with a specific TuneD version from upstream.

Example invocations
```
ORG=quay-user hack/test-upstream-tuned.sh
ORG=quay-user TUNED_COMMIT=ce3d0926 hack/test-upstream-tuned.sh
```

This script assumes a deployed OpenShift cluster and an environment configured to write the locally built IMAGE to an image repository.

Also note `cvo_scale()` scales down CVO replica to 0 to prevent it from overriding the custom NTO image.  If you wish to upgrade your cluster, do not use this script.

### DIFF
--- a/hack/test-upstream-tuned.sh
+++ b/hack/test-upstream-tuned.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+set -e
+
+# Example invocation
+# ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


liqcui:
/lgtm
I have tested the script in my local environment, no issue
@jmencak Very useful script and make work easier

PR #578, opened 2023-03-02
Fix failing HyperShift presubmits

jmencak:
Recent PR431 broke HyperShift presubmits.  The code is not HyperShift aware.


jmencak:
Holding for @eggfoobar to unhold.

eggfoobar:
/unhold

We should be good with this change, the only concern I had was during the reconcile we check for `infra` resource, but it seems it will be there after HCCO comes up, if that's the case then the reconcile will eventually resolve if something were to trigger the PAO.

PR #583, opened 2023-03-15
tests: image: bump to the latest version

ffromani:
We should try to avoid floating tag (`latest`). Release-referring floating tags are the lesser evil, so it's OK(-ish).

Ideally, the tests should detect the cluster version and pick the tests version automatically by default.


ffromani:
@yanirq @mrniranjan looks like this tiny PR is still relevant

ffromani:
replaced by https://github.com/openshift/cluster-node-tuning-operator/pull/584

PR #585, opened 2023-03-15
OCPBUGS-10247: Update NTO-generated MC on MachineCount <= 1

jmencak:
PR558 added support for informing users about nodes with incompatible topologies within the same MCP.  However, a recovery from this state is not handled in a user-friendly way.  Currently, the entire MCP or NTO-generated MC needs to be deleted or manually adjusted to recover.

This PR fixes the situation so that when the MachineCount of the MCP which had nodes with conflicting topologies is equal or less than 1, updates to the NTO-generated MC are allowed even though the input configuration did not change and the new kernel parameters do not match the MC's kernel parameters.

Fixes OCPBUGS-10247

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -284,13 +284,8 @@ func (c *Controller) sync(key wqKey) error {
 		change, err := c.pc.nodeChangeHandler(key.name)
 		if err != nil {
 			if errors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


liqcui:
Pre-merge and review testing passed 
/lgtm

jmencak:
infra issues
/retest

PR #586, opened 2023-03-15
Make the enable-leader-election option work

jmencak:
The `--enable-leader-election=false` currently does not work.  Leader election is always on.  This change fixes the issue.


jmencak:
/cc @MarSik 
due to `git blame`

ffromani (commented):
/lgtm

ffromani:
do we need a 4.13 backport?

jmencak:
> do we need a 4.13 backport?

I believe it is not strictly needed.  I only ever use this functionality when developing.  Unless others care about this, of course.

PR #587, opened 2023-03-16
OCPBUGS-10217: Remove the preStop hook for openshift-tuned

jmencak:
The preStop hook for `openshift-tuned` and the socket interface to talk to it is not really needed.  The reason for having it in the first place was to ensure a rollback of `TuneD` settings.  However, realistically, the rollback of setting should never take more than a couple of seconds, often *much* less than that (typically ~0.1s).  This is well within the `terminationGracePeriodSeconds` of 30s.

This also fixes the issue of the TERM signal not reaching `openshift-tuned` during pod terminations by omitting the "exec" in the Pod startup script.  Now, we simply rely on the TERM signal to be sent to openshift-tuned, which catches the signal, handles `TuneD` termination (either gracefully or forcefully if it does not exit within 10s) and then `openshift-tuned` exits itself.

Also see OCPBUGS-10217

### DIFF
--- a/assets/bin/run
+++ b/assets/bin/run
@@ -1,6 +1,5 @@
 #!/bin/sh
 
-openshift_tuned_socket=/var/lib/tuned/openshift-tuned.sock
 export SYSTEMD_IGNORE_CHROOT=1
 
 start() {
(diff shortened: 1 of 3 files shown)


ffromani (commented):
/lgtm

codewise only

jmencak:
infra issues
/retest

PR #581, opened 2023-03-10
OCPBUGS-9959: check scheduler settings under /sys/kernel/debug/sched/ 

mrniranjan:
…sctl

On RHEL9.2 scheduler settings have been moved under  /sys/kernel/debug/sched/ instead of setting them using sysctl

PR #580, opened 2023-03-09
Update to the latest k8s and OpenShift deps

jmencak:
Update:
  - to k8s 1.26.2
  - to latest OpenShift dependencies
  - to latest MCO dependencies (unpinned)

Other changes:
  - `Makefile`:
      - reflect build-machinery-go changes
      - pin on controller-gen v0.6.0
  - `kubeletconfig_test.go` fix to reflect the latest k8s changes
  - `make update-manifests`
  - `go mod tidy && go mod vendor`

PR #588, opened 2023-03-20
OCPBUGS-10293: performance-profile: enable crun for high-performance runtime

Tal-or:
CRI-O can be configured to use crun as its low-level container runtime by
appliying ContainerRuntimeConfig CR.

Here, we are enabling high-performance runtime class to run with crun as
well based on the information from the ContainerRuntimeConfig CR.

- [x] make default runtime configurable for render 
- [x] e2e tests


pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -214,6 +274,44 @@ func (r *PerformanceProfileReconciler) getInfraPartitioningMode() (pinning apico
...
+func (r *PerformanceProfileReconciler) getContainerRuntimeName(profile *performancev2.PerformanceProfile) (mcov1.ContainerRuntimeDefaultRuntime, error) {
+	ctrcfgList := &mcov1.ContainerRuntimeConfigList{}
+	if err := r.List(context.Background(), ctrcfgList); err != nil {
+		if k8serros.IsNotFound(err) {
+			return mcov1.ContainerRuntimeDefaultRuntimeRunc, nil
+		}
+		return "", err
+	}
+
+	profileMCPSelector := labels.Set(profile.Spec.MachineConfigPoolSelector)
```
MarSik, line 286: I do not think this is enough. Our mcpSelector is an optional field and we actually have some logic to detect master /  worker pools too.

See https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go#L11 and the newer https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/performanceprofile/controller/performanceprofile_controller.go#L299

go.mod
```diff
@@ -6,119 +6,116 @@ require (
 	github.com/RHsyseng/operator-utils v0.0.0-20200213165520-1a022eb07a43
 	github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f
 	github.com/coreos/ignition v0.35.0
-	github.com/coreos/ignition/v2 v2.9.0
+	github.com/coreos/ignition/v2 v2.13.0
```
ffromani, line 9: why do we need this update? I'm asking because it's a bit unexpected to do it in the context of this PR. Perhaps we should have done it earlier?

manifests/40-rbac.yaml
```diff
@@ -75,7 +75,7 @@ rules:
   verbs: ["update"]
 # Needed by the core operator functionality.
 - apiGroups: ["machineconfiguration.openshift.io"]
-  resources: ["kubeletconfigs", "machineconfigs"]
+  resources: ["kubeletconfigs", "machineconfigs", "containerruntimeconfigs"]
```
ffromani, line 78: for the definitive fix, this needs to be on its own entry because we will totally need to list, get and watch `containerruntimeconfigs` objects, but we will not need to create, delete, update

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -201,6 +227,40 @@ func (r *PerformanceProfileReconciler) tunedProfileToPerformanceProfile(tunedPro
 	return requests
 }
 
+func (r *PerformanceProfileReconciler) ctrRuntimeConfToPerformanceProfile(ctrRuntimeConfObj client.Object) []reconcile.Request {
+	ctrcfg := &mcov1.ContainerRuntimeConfig{}
+
+	err := r.Get(context.TODO(), client.ObjectKeyFromObject(ctrRuntimeConfObj), ctrcfg)
```
ffromani, line 233: *NOT FOR THIS PR*: in the future we need to stop hardcoding `context.TODO()` and properly pass down the call chain in the reconcile loop the given context.

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -214,6 +274,44 @@ func (r *PerformanceProfileReconciler) getInfraPartitioningMode() (pinning apico
 	return infra.Status.CPUPartitioning, nil
 }
 
+func (r *PerformanceProfileReconciler) getContainerRuntimeName(profile *performancev2.PerformanceProfile) (mcov1.ContainerRuntimeDefaultRuntime, error) {
+	ctrcfgList := &mcov1.ContainerRuntimeConfigList{}
+	if err := r.List(context.Background(), ctrcfgList); err != nil {
```
ffromani, line 279: same, `context.Background()` is marginally better (if at all) than `context.TODO()` - not interruptable, and easy to miss when fixing all the contexts

pkg/performanceprofile/controller/performanceprofile_controller.go
ffromani: Yes, we need to review the matching logic and very likely to reuse it to make things consistent

ffromani (commented):
I agree with Martin than fixing/reusing the matching logic is the biggest pending issue

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: The `ContainerRuntimeCofing` object only has a `machineConfigPoolSelector`.
We can detect the MCP by the profile, and then look for labels in the MCP that are matching the `machineConfigPoolSelector`:

```
mcp, err := r.getMachineConfigPoolByProfile(profile)
...

//ctrcfg is mcov1.ContainerRuntimeConfig object
ctrcfgSelector, err := metav1.LabelSelectorAsSelector(ctrcfg.Spec.MachineConfigPoolSelector)
...

if ctrcfgSelector.Matches(labels.Set(mcp.Labels)) {
// we found the ContainerRuntimeConfig that is associate with the profile
}

Is it better now?

go.mod
Tal-or: Those dependencies updated by go mod automatically after I bumped
`github.com/openshift/machine-config-operator`
from: `v0.0.1-0.20220203091316-d3010b34d344`
to: `v0.0.1-0.20230317184452-40575b862f7b`

Tal-or: The bump itself is necessary because the `ContainerRuntimeConfig` object is relatively new under the  machine-config-operator api group

manifests/40-rbac.yaml
Tal-or: Done. Thanks!

go.mod
ffromani: Fair enough, we need the bump here so we're good.

ffromani:
Thanks for the updates. I'll review the matching logic again ASAP.

go.mod
Tal-or: i'll add the rational for the bump in the commit message

pkg/performanceprofile/controller/performanceprofile_controller.go
MarSik: Yeah, this looks better I think

Tal-or:
#580 should fix ci/prow/verify failure

Tal-or:
#581 should unblock `e2e-gcp-pao`

ffromani:
@Tal-or I added changes in the same vein as yours but now they are lost; please pull rebase next time you force push, we should sync better otherwise we actually waste CI cycles.

Tal-or:
> @Tal-or I added changes in the same vein as yours but now they are lost; please pull and rebase next time you force push, we should sync better otherwise we actually waste CI cycles.

Sorry about that 😔 I'm not going to touch this PR utill I comeback next Sunday (April 16th)

ffromani:
there's a bug in the e2e test. The test is not picking the correct MCP selector, so the ContainerRuntimeConfig object we create in the test is never applied and the controller never kicks in. I'll work on a fix.

ffromani:
/lgtm

/hold

ffromani:
/hold cancel

per offline conversation

PR #592, opened 2023-03-27
OCPBUGS-10635: workload-hints: disable stalld when rt disabled

Tal-or:
When RealTime workload hint configured as false, we should
explicitly disable stalld service

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -35,6 +35,9 @@ min_perf_pct=100
 {{if .RealTimeHint}}
 [service]
 service.stalld=start,enable
+{{else}}
+[service]
+service.stalld=stop,disable
 {{end}}
 
 [vm]
--- a/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go
@@ -119,8 +119,9 @@ var _ = Describe("Tuned", func() {
 			It("should not contain realtime related parameters", func() {
 				profile.Spec.WorkloadHints = &performancev2.WorkloadHints{RealTime: pointer.BoolPtr(false)}
 				tunedData := getTunedStructuredData(profile)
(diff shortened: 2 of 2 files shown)


ffromani:
/lgtm
it seems like a good idea in general

cc @MarSik for final approval

ffromani:
> /lgtm it seems like a good idea in general
> 
> cc @MarSik for final approval

...but did we test that the transitions enable->disable and disable->enable works as we expect?

Tal-or:
> ...but did we test that the transitions enable->disable and disable->enable works as we expect?
Yes manually. Plus the fact that https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go#L630 is passing

test/e2e/performanceprofile/testdata/render-expected-output/manual_machineconfig.yaml
```diff
@@ -6,10 +6,10 @@ metadata:
     machineconfiguration.openshift.io/role: worker-cnf
   name: 50-performance-manual
   ownerReferences:
-  - apiVersion: "performance.openshift.io/v2"
+  - apiVersion: performance.openshift.io/v2
     kind: PerformanceProfile
     name: manual
-    uid: ""
+    uid: 849928a9-63c5-4988-805a-1232f6067b89
```
MarSik, line 12: Will the UID be stable for CI? In other words, is the end result the same every run?

Tal-or: hmm, it  seems like the way render command works has changed and now it generates a uid as well. I dunno if that's intended but it should not be part of the manifests for the tests, because other wise every `render-sync` would generate a different test files.

Tal-or: https://github.com/openshift/cluster-node-tuning-operator/pull/594 issued a PR to fix that

Tal-or:
depends on #595

ffromani:
/hold cancel

my mistake, hold was meant for a different change.

PR #593, opened 2023-03-27
OCPBUGS-10931: add Liveness & Readiness probes

michael-topchiev:
This PR adds simple Liveness and Readiness probes so they can be used from any client application, like Hypershift.

### DIFF
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -32,6 +32,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
 	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/config"
@@ -128,6 +129,7 @@ func operatorRun() {
 		RetryPeriod:             &le.RetryPeriod.Duration,
 		RenewDeadline:           &le.RenewDeadline.Duration,
 		Namespace:               ntoNamespace,
+		HealthProbeBindAddress:  "0.0.0.0:6060",
 	})
 
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jmencak:
Thank you for the PR, @michael-topchiev .  So this basically allows *unauthenticated* probes to NTO from any pod in the cluster.  In the case of a CVE, this could potentially allow a DoS on the operator.  Do we absolutely really need this PR or can we work around this in some way?

jmencak:
Also, it is unclear to me how it actually helps address [OCPBUGS-10931](https://github.com/openshift/cluster-node-tuning-operator/pull/593#top) since this PR does not define [livenessProbe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) anywhere.  Could you please explain how does the kubelet knows it should restart the pod without `livenessProbe` defined?

michael-topchiev:
@jmencak yaml configuration for the probes will be defined by Hypershift in a separate PR once `cluster-node-tuning-operator` has the probes implementation. 

Wrt unauthenticated probes to NTO, I believe there are a few other OCP components that use very similar implementation over http or https bypassing authentication.

I think this probe is good to have, unless there is 100% certainty that NTO service never deadlocks or hangs on an outgoing network request. The requirement at IBM is to provide such probes to help prevent service HA problems.

jmencak:
Thank you for the extra information.

> Wrt unauthenticated probes to NTO, I believe there are a few other OCP components that use very similar implementation over http or https bypassing authentication.

Yes, I saw a couple of them, but I do not believe that is an argument for adding another attack vector also to NTO.

> I think this probe is good to have, unless there is 100% certainty that NTO service never deadlocks or hangs on an outgoing network request.

I've never seen the scenario you are talking about, but of course, you can hardly be sure of anything.  In the same way, we can never be sure the probe causes cascading failures by kubelet continuously restarting the operator.  While this is quite unlikely to happen with NTO, I've seen this many times with HAProxy/ingress in the past, when it was overloaded with incoming network requests.  Then the livenessProbe kicked in and made matters even worse.

> The requirement at IBM is to provide such probes to help prevent service HA problems.

I'd like to know more about this requirement before we approve this.  Could you perhaps share it over email/Slack?

jmencak:
How do you know that your code (in another thread than the probe runs in) actually really works (isn't deadlocked) even when your livenessProbe in another thread succeeds?  I general I have a problem with livenessProbes, because they are hardly ever are a true reflection of software health and they mostly just add additional problems.

jmencak:
> Wrt authentication, [this document](https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md) explicitly states `/healthz, /readyz, /livez should not require authorization.`

Absolutely.  No-one argues with that.

> I agree that this probes do little to actually reflect the NTO service health, it's just a minimal implementation similar to what a couple other CPO components have. If you guys have an idea of what would be the right logic, I'd happy to help with its implementation.

They reflect pretty much nothing, that why I'm wondering why we even need them.  This is my personal opinion.  We have more senior developers on NTO than me.  Please let's hear your thoughts.  I'm against adding more unneeded complexity.

michael-topchiev:
Dropping the PR because there is no hard evidence that the service really need the probes.

PR #594, opened 2023-03-28
render: remove uid from render-sync target

Tal-or:
The render command is generating manifests with uid now. we don't need it to be part of the data for our render-expected-output so let's remove it


ffromani (commented):
/lgtm

good enough for the short term, but we should undestand more about this and get to the root cause

yanirq:
@Tal-or  looking at the rendered changes I dont see the **uid** being removed,  only the **whitespaces** and **quotes** are being removed

Tal-or:
/retitle OCPBUGS-11852: render: remove uid from render-sync target

PR #595, opened 2023-03-28
OCPBUGS-10998: Fix updating nodeSelector test

yanirq:
The node selector tests are using a utility function that removes labels from nodes.
When the function tries to patch nodes with updated labels it will wait for the matching mcp to be updated. In case no update to node labels where done the update will time out.

The fix is to only wait for mcp updates if there was actually one.

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -378,8 +378,15 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				oldNodeSelector = profile.Spec.DeepCopy().NodeSelector
 			}
 			removeLabels = func(nodeSelector map[string]string, targetNode *corev1.Node) error {
+				patchNode := false
 				for l := range nodeSelector {
-					delete(targetNode.Labels, l)
+					if _, ok := targetNode.Labels[l]; ok {
+						patchNode = true
+						delete(targetNode.Labels, l)
+					}
+				}
+				if !patchNode {
+					return nil
 				}
 				label, err := json.Marshal(targetNode.Labels)
 				Expect(err).ToNot(HaveOccurred())


ffromani:
/approve
/lgtm

could be more idiomatic, but let's stop the bleeding first.
@mrniranjan feel free to remove the hold once you are happy

ffromani:
like:

```
removeLabels = func(nodeSelector map[string]string, targetNode *corev1.Node) error {
	patchNode := false
	for l := range nodeSelector {
		_, ok := targetNode.Labels[l]
		patchNode = ok
		delete(targetNode.Labels, l)
	}
	if !patchNode {
		return nil
	}
	label, err := json.Marshal(targetNode.Labels)
	Expect(testclient.Client.Patch(context.TODO(), targetNode,
		client.RawPatch(
			types.JSONPatchType,
			[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/metadata/labels", "value": %s }]`, label)),
		),
	)).ToNot(HaveOccurred())
	mcps.WaitForCondition(testutils.RoleWorker, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)

	By("Waiting when MCP Worker complete updates and verifying that node reverted back configuration")
	mcps.WaitForCondition(testutils.RoleWorker, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
	return nil
}
```

but again, let's stop the bleeding first.

ffromani:
/hold

my original suggestion introduced a bug: the bug is patchNode is reset, should be true if we delete at least a label
but if there are more labels in the map and we do like not-delete, delete, not-delete, patchNode is reset to false.
I fixed my suggestion, but too late.

yanirq:
> /hold
> 
> my original suggestion introduced a bug: the bug is patchNode is reset, should be true if we delete at least a label but if there are more labels in the map and we do like not-delete, delete, not-delete, patchNode is reset to false. I fixed my suggestion, but too late.

fixed. we can remove the hold
/hold cancel

PR #597, opened 2023-03-29
OCPBUGS-11125: Backup and revert profile when hugepages test completes

mrniranjan:
when testing profile parameters for hugepages count split between two numa nodes we don't take backup of existing profile.

In downstream we use a profile which takes in to account all the cpus and are updated accordingly in reserved and isolated. But this test overrides with only 4 cpus. but doesn't revert the profile once test completes causing the cluster to run on only 4 cpus causing tests to run very slow .

Thif patch takes the backup of the profile and reverts to the original profile.


ffromani:
/retest

failure in the images lane is infra issue

PR #601, opened 2023-03-30
OCPNODE-1539: perf profile: add script for preparing cgroups for CPU load balance disabling

haircommander:
on RHEL 9 and newer kernels, the sysfs previously relied on to disable CPU load balancing is no longer available. Instead, cpu load balancing must be configured through cgroups.

To configure a cpuset to have load balance disabled (on cgroup v1), a cpuset cgroup must have `cpuset.sched_load_balance` set to 0 (disable), and any cpuset that contains the same set as `cpuset.cpus` must also have `cpuset.sched_load_balance` set to disabled.

To do this, we must move systemd and all host processes to a separate cpuset off of the root cgroup, and manually configure it to use the reserved cpus set by set_cpuaffinity. This won't have any effect on the actual processes themselves, but will open the opportunity for the root cgroup to have load balance disabled.

Also, preconfigure the machine.slice (which will later be created by podman).

This PR adds a script to do these things and a systemd unit to enable it, which is created when there are reserved cpus in the performance profile.

### DIFF
--- a/assets/performanceprofile/scripts/cpuset-configure.sh
+++ b/assets/performanceprofile/scripts/cpuset-configure.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# cpuset-configure.sh configures three cpusets in preparation for allowing containers to have cpu load balancing disabled.
+# To configure a cpuset to have load balance disabled (on cgroup v1), a cpuset cgroup must have `cpuset.sched_load_balance`
+# set to 0 (disable), and any cpuset that contains the same set as `cpuset.cpus` must also have `cpuset.sched_load_balance` set to disabled.
+
+set -euo pipefail
+
+root=/sys/fs/cgroup/cpuset
+system="$root"/system
+machine="$root"/machine.slice
+
+# As such, the root cgroup needs to have cpuset.sched_load_balance=0. 
+echo 0 > "$root"/cpuset.sched_load_balance
+
+# However, this would present a problem for system daemons, which should have load balancing enabled.
+# As such, a second cpuset must be created, here dubbed `system`, which will take all system daemons.
+# Since systemd starts its children with the cpuset it is in, moving systemd will ensure all processes systemd begins will be in the correct cgroup.
+mkdir "$system"
+# cpuset.mems must be initialized or processes will fail to be moved into it.
+cat "$root/cpuset.mems" > "$system"/cpuset.mems
+# Retrieve the cpuset of systemd, and write it to cpuset.cpus of the system cgroup.
+reserved_set=$(taskset -cp  1  | awk 'NF{ print $NF }')
+echo "$reserved_set" > "$system"/cpuset.cpus
+
+# And move the system processes into it.
+# Note, some kernel threads will fail to be moved with "Invalid Argument". This should be ignored.
+for process in $(cat "$root"/cgroup.procs | sort -r); do
+	echo $process > "$system"/cgroup.procs 2>&1 | grep -v "Invalid Argument" || true;
+done
+
+# Finally, a the `machine.slice` cgroup must be preconfigured. Podman will create containers and move them into the `machine.slice`, but there's
+# no way to tell podman to update machine.slice to not have the full set of cpus. Instead of disabling load balancing in it, we can pre-create it.
+# with the reserved CPUs set ahead of time, so when isolated processes begin, the cgroup does not have an overlapping cpuset between machine.slice and isolated containers.
+mkdir "$machine" || true
+
+# It's unlikely, but possible, that this cpuset already existed. Iterate just in case.
+for file in $(find "$machine" -name cpuset.cpus | sort -r); do echo "$reserved_set" > "$file"; done
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -60,6 +60,7 @@ const (
 	setCPUsOffline            = "set-cpus-offline"
 	setRPSMask                = "set-rps-mask"
 	clearIRQBalanceBannedCPUs = "clear-irqbalance-banned-cpus"
+	cpusetConfigure           = "cpuset-configure"
 )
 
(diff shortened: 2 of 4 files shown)


browsell:
LGTM

yanirq:
@haircommander  can you please remove all 3 occurrences of `testutils.KnownIssueJira("OCPNODE-1538")` from the e2e tests so the relevant tests will run ? 

These are skips that were added to unblock CI until the issue is fixed (with this PR)

MarSik:
@yanirq This change is not enough to fix OCPNODE-1538, it must be accompanied by https://github.com/cri-o/cri-o/pull/6744 

So the removal of the KnownIssue marker will not make the test pass successfully.

yanirq:
> @yanirq This change is not enough to fix OCPNODE-1538, it must be accompanied by [cri-o/cri-o#6744](https://github.com/cri-o/cri-o/pull/6744)
> 
> So the removal of the KnownIssue marker will not make the test pass successfully.

@MarSik  but we do want to remove the skip and only if the tests pass successfully then merge this PR.
We can hold this PR as well until [cri-o/cri-o#6744](https://github.com/cri-o/cri-o/pull/6744) is resolved

Tal-or:
We need to add an e2e test for that. at least make sure the file is present and setup correctly.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -60,6 +60,7 @@ const (
 	setCPUsOffline            = "set-cpus-offline"
 	setRPSMask                = "set-rps-mask"
 	clearIRQBalanceBannedCPUs = "clear-irqbalance-banned-cpus"
+	cpusetConfigure           = "cpuset-configuration"
```
Tal-or, line 63: You should add the script itself to the system using [addContent](https://github.com/openshift/cluster-node-tuning-operator/blob/3dd2403b1617e612a09f9f3b9ff20a6ce0035b80/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go#L434)

Tal-or (commented):
I would recommend adding at least sanity e2e just to make sure the script is present and the systemd unit setup correctly

haircommander:
> @MarSik but we do want to remove the skip and only if the tests pass successfully then merge this PR.
We can hold this PR as well until https://github.com/cri-o/cri-o/pull/6744 is resolved 

agreed, we can use this as the final piece and merge kubelet, cri-o and runc pieces first

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
haircommander: I see I thought that was taken care of above but I was wrong, thanks!

assets/performanceprofile/scripts/cpuset-configure.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# cpuset-configure.sh configures three cpusets in preparation for allowing containers to have cpu load balancing disabled.
+# To configure a cpuset to have load balance disabled (on cgroup v1), a cpuset cgroup must have `cpuset.sched_load_balance`
+# set to 0 (disable), and any cpuset that contains the same set as `cpuset.cpus` must also have `cpuset.sched_load_balance` set to disabled.
+
+set -euo pipefail
+
+root=/sys/fs/cgroup/cpuset
+system="$root"/system
```
MarSik, line 10: I think the commonly used name is "system.slice".

assets/performanceprofile/scripts/cpuset-configure.sh
```diff
@@ -0,0 +1,38 @@
...
+reserved_set=$(taskset -cp  1  | awk 'NF{ print $NF }')
+echo "$reserved_set" > "$system"/cpuset.cpus
+
+# And move the system processes into it.
+# Note, some kernel threads will fail to be moved with "Invalid Argument". This should be ignored.
+for process in $(cat "$root"/cgroup.procs | sort -r); do
+	echo $process > "$system"/cgroup.procs 2>&1 | grep -v "Invalid Argument" || true;
+done
+
+# Finally, a the `machine.slice` cgroup must be preconfigured. Podman will create containers and move them into the `machine.slice`, but there's
```
MarSik, line 32: How is podman using the machine slice? Is that a workload partitioning specialty?

assets/performanceprofile/scripts/cpuset-configure.sh
haircommander: I specifically chose not to use `.slice` because slices are something unique to systemd and the name would imply it's being managed by systemd, which it's not

assets/performanceprofile/scripts/cpuset-configure.sh
haircommander: podman puts all of its systemd cgroup containers into the machine.slice by default. it was formally not at all configured for workload partitioning because the podman pieces run before guaranteed pods are in the picture. The slice remains on the node only as a remnant because nothing removes them

assets/performanceprofile/scripts/cpuset-configure.sh
haircommander: I specifically chose not to use `.slice` because slices are something unique to systemd and the name would imply it's being managed by systemd, which it's not

MarSik:
I am ok with the patch in general, but I think I would prefer if the slices were created via the systemd mechanism (a .slice file in /etc/systemd/system) so it is obvious where they come from though. If that is possible of course. system.slice might be a reserved name even though systemd is not creating it by default for the cpuset controller.

haircommander:
cri-o is out of date in 4.14 so the test won't pass yet

haircommander:
testing 4.13 on https://github.com/openshift/cluster-node-tuning-operator/pull/617 which has the required versions

ffromani (approved):
/approve
/lgtm

discussed extensively offline + followups already planned

yanirq:

test is reaching full timeout but non related to this PR

PR #567, opened 2023-02-14
OCPBUGS-7384: Remove optimization to allow full resync

jmencak:
There are edge cases (e.g. OCPBUGS-7384), where full resync seems to be necessary.  Remove an optimization to skip Tuned and Profile resync for NTO operands when the old and new Accessor are the same.  This will add on average 4% more CPU cycles, but should ensure more reliable operation.

PR #611, opened 2023-04-11
OCPBUGS-11150: Revert #567 and cleanup PPC-generated TuneD config

jmencak:
PerformanceProfile Controller (PPC) generates TuneD configuration out of template file `assets/performanceprofile/tuned/openshift-node-performance`.

This PR fixes several issues with this file:
  - Remove `TuneD [scheduler]` plug-in options that do not exist.
  - Add a note about sysctl existing only in non-rt kernels.

Also, as a temporary fix we revert the removal of optimizations done in #567 as a way of addressing OCPBUGS-7384.  However, we later found OCPBUGS-7384 was a user configuration issue so the change is not really needed at this point as we ran with it since OCP 4.5 with no issues.

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -59,10 +59,6 @@ group.rcuc=0:f:11:*:rcuc.*
 group.ktimers=0:f:11:*:ktimers.*
 sched_min_granularity_ns=10000000
 sched_migration_cost_ns=5000000
-numa_balancing=0
-{{if .RealTimeHint}}
-sched_rt_runtime_us=-1
-{{end}}
 {{if not .GloballyDisableIrqLoadBalancing}}
 default_irq_smp_affinity = ignore
 {{end}}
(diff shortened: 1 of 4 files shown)


jmencak:
As promised, Jose
/cc @jlojosnegros

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -59,10 +59,6 @@ group.rcuc=0:f:11:*:rcuc.*
 group.ktimers=0:f:11:*:ktimers.*
 sched_min_granularity_ns=10000000
 sched_migration_cost_ns=5000000
-numa_balancing=0
-{{if .RealTimeHint}}
-sched_rt_runtime_us=-1
```
ffromani, line 64: do we want to remove also this one?

jmencak: If you mean `sched_rt_runtime_us`, then yes.  There is no such `[scheduler]` directive.

ffromani: yes, this is what I meant. Looks like we need to do a better job verifying these directives when reviewing future PRs.

MarSik: I think the test needs to be updated to expect this modified content

jmencak: Thanks, adjusted based on a rendered `Tuned` CR.

jmencak:
I see, `performanceprofile/controller/performanceprofile/components/tuned/tuned_test.go` needs to be fixed at a minimum...

ffromani:
/lgtm

/hold 

MarSik (approved):
/lgtm

jmencak:
Thanks guys, let's see if automated cherry-pick to 4.13 succeeds.

/cherry-pick release-4.13

PR #615, opened 2023-04-13
Revert #567 and cleanup PPC-generated TuneD config

jmencak:
PerformanceProfile Controller (PPC) generates TuneD configuration out of template file `assets/performanceprofile/tuned/openshift-node-performance`.

This PR fixes several issues with this file:
  - Remove TuneD [scheduler] plug-in options that do not exist.
  - Add a note about sysctl existing only in non-rt kernels.

Also, as a temporary fix we revert the removal of optimizations done in #567 as a way of addressing OCPBUGS-7384.  However, we later found OCPBUGS-7384 was a user configuration issue so the change is not really needed at this point as we ran with it since OCP 4.5 with no issues.


jmencak:
/label backport-risk-assessed
@liqcui or @mrniranjan can we have the `cherry-pick-approved` label, please?

PR #617, opened 2023-04-14
[release-4.13] OCPNODE-1539: perf profile: add script for preparing cgroups for CPU load balance disabling

haircommander:
on RHEL 9 and newer kernels, the sysfs previously relied on to disable CPU load balancing is no longer available. Instead, cpu load balancing must be configured through cgroups.

To configure a cpuset to have load balance disabled (on cgroup v1), a cpuset cgroup must have `cpuset.sched_load_balance` set to 0 (disable), and any cpuset that contains the same set as `cpuset.cpus` must also have `cpuset.sched_load_balance` set to disabled.

To do this, we must move systemd and all host processes to a separate cpuset off of the root cgroup, and manually configure it to use the reserved cpus set by set_cpuaffinity. This won't have any effect on the actual processes themselves, but will open the opportunity for the root cgroup to have load balance disabled.

Also, preconfigure the machine.slice (which will later be created by podman).

This PR adds a script to do these things and a systemd unit to enable it, which is created when there are reserved cpus in the performance profile.

cherry-pick of https://github.com/openshift/cluster-node-tuning-operator/pull/601

### DIFF
--- a/assets/performanceprofile/scripts/cpuset-configure.sh
+++ b/assets/performanceprofile/scripts/cpuset-configure.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# cpuset-configure.sh configures three cpusets in preparation for allowing containers to have cpu load balancing disabled.
+# To configure a cpuset to have load balance disabled (on cgroup v1), a cpuset cgroup must have `cpuset.sched_load_balance`
+# set to 0 (disable), and any cpuset that contains the same set as `cpuset.cpus` must also have `cpuset.sched_load_balance` set to disabled.
+
+set -euo pipefail
+
+root=/sys/fs/cgroup/cpuset
+system="$root"/system
+machine="$root"/machine.slice
+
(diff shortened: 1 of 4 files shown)


yanirq:
/retitle [release-4.13] OCPNODE-1539: perf profile: add script for preparing cgroups for CPU load balance disabling

sdodson:
/hold
Just looking to see the 4.14 PR merge before this one does. Will clear this hold as soon as 4.14 merges.

sdodson:
Hmm, I thought the tests on master had passed but it looks like it failed again. I'm going to go ahead and let this go in if it passes CI on this branch. We can come back and figure out what's up with master branch ASAP and then determine if we need to backport anything to 4.13 too.

jmencak:
ci/prow/e2e-gcp-pao-updating-profile is broken, see:
https://redhat-internal.slack.com/archives/CQNBUEVM2/p1681878606439819

I'd recommend overriding 
ci/prow/e2e-gcp-pao-updating-profile 
to get this in.

PR #623, opened 2023-04-18
OCPBUGS-11914: Remove subPaths, they are broken

jmencak:
`subPath`s are broken in current versions of k8s and also suffer from other issues on RHOCP.  Do not use them.  See:
 - https://github.com/kubernetes/kubernetes/issues/89181
 - https://issues.redhat.com/browse/OCPBUGS-5255


jmencak:
/cc @dagrayvid 
We talked about this in the past.  I believe now is the time to remove them.

PR #626, opened 2023-04-19
OCPBUGS-11083: pao e2e: fix update test suit timeouts

yanirq:
This PR comes to fix :
- stalld flaky tests the occur due to race conditions
- afterAll function should be skipped if tests are skipped to save testing time and irrelevant errors.
- skip flaky update test due to OCPBUGS-12836

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/tuned/tuned.go
+++ b/test/e2e/performanceprofile/functests/utils/tuned/tuned.go
@@ -84,22 +84,22 @@ func GetPod(ctx context.Context, node *corev1.Node) (*corev1.Pod, error) {
 }
 
 func WaitForStalldTo(run bool, interval, timeout time.Duration, node *corev1.Node) error {
-	return wait.Poll(interval, timeout, func() (bool, error) {
-		cmd := []string{"/bin/bash", "-c", "pidof stalld || true"}
+	return wait.PollImmediate(interval, timeout, func() (bool, error) {
+		cmd := []string{"/bin/bash", "-c", "pidof stalld || echo \"stalld not running\""}
 		stalldPid, err := nodes.ExecCommandOnNode(cmd, node)
 		if err != nil {
-			return false, fmt.Errorf("failed to execute command %q on node: %q; %w", cmd, node.Name, err)
+			klog.Errorf("failed to execute command %q on node: %q; %v", cmd, node.Name, err)
+			return false, err
 		}
-
 		_, err = strconv.Atoi(stalldPid)
 		if !run { // we don't want stalld to run
 			if err == nil {
-				return false, fmt.Errorf("node=%q stalld_pid=%q stalld is running when it shouldn't", node.Name, stalldPid)
+				return false, nil
 			}
+			return true, nil
 		}
 		// we want stalld to run
 		if err != nil {
-			klog.Warningf("node=%q stalld_pid=%q is not a valid pid number: %v", node.Name, stalldPid, err)
 			return false, nil
 		}
 		return true, nil
@@ -107,15 +107,15 @@ func WaitForStalldTo(run bool, interval, timeout time.Duration, node *corev1.Nod
 }
 
 func CheckParameters(node *corev1.Node, sysctlMap map[string]string, kernelParameters []string, stalld, rtkernel bool) {
-	cmd := []string{"/bin/bash", "-c", "pidof stalld || true"}
+	cmd := []string{"/bin/bash", "-c", "pidof stalld || echo \"stalld not running\""}
 	By(fmt.Sprintf("Executing %q", cmd))
 	stalldPid, err := nodes.ExecCommandOnNode(cmd, node)
-	ExpectWithOffset(1, err).ToNot(HaveOccurred(), "failed to execute command %q on node: %q; %w", cmd, node.Name, err)
+	ExpectWithOffset(1, err).ToNot(HaveOccurred(), "failed to execute command %q on node: %q; %v", cmd, node.Name, err)
 
 	_, err = strconv.Atoi(stalldPid)
 	if stalld {
 		ExpectWithOffset(1, err).ToNot(HaveOccurred(),
-			"node=%q stalld_pid=%q is not a valid pid number: %w", node.Name, stalldPid, err)
+			"node=%q does not have a running stalld process", node.Name)
 	} else {
 		ExpectWithOffset(1, err).To(HaveOccurred(),
 			"node=%q stalld_pid=%q stalld is running when it shouldn't", node.Name, stalldPid)
@@ -131,11 +131,11 @@ func CheckParameters(node *corev1.Node, sysctlMap map[string]string, kernelParam
 		"cannot find the cluster Node Tuning Operator object "+key.String())
 
 	if stalld {
-		ExpectWithOffset(1, *tuned.Spec.Profile[0].Data).To(ContainSubstring("stalld"),
-			"node=%q cannot find substring stalld in tuned profile data", node.Name)
+		ExpectWithOffset(1, *tuned.Spec.Profile[0].Data).To(ContainSubstring("stalld=start,enable"),
+			"node=%q cannot find substring stalld=start,enable in tuned profile data", node.Name)
 	} else {
-		ExpectWithOffset(1, *tuned.Spec.Profile[0].Data).ToNot(ContainSubstring("stalld"),
-			"node=%q found substring stalld in tuned profile data when it shouldn't", node.Name)
+		ExpectWithOffset(1, *tuned.Spec.Profile[0].Data).To(ContainSubstring("stalld=stop,disable"),
+			"node=%q cannot find substring stalld=stop,disable in tuned profile data", node.Name)
 	}
 
 	for param, expected := range sysctlMap {
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -615,9 +615,9 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				By("Waiting for TuneD to start on nodes")
 				for i := 0; i < len(workerRTNodes); i++ {
 					node := &workerRTNodes[i]
+					wg.Add(1)
 					go func() {
 						defer GinkgoRecover()
-						wg.Add(1)
 						defer wg.Done()
 
 						pod, err := utilstuned.GetPod(context.TODO(), node)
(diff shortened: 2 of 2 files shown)


jlojosnegros (commented):
/lgtm

yanirq:
infra issues
/test e2e-gcp-pao-updating-profile

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
```diff
@@ -84,38 +84,34 @@ func GetPod(ctx context.Context, node *corev1.Node) (*corev1.Pod, error) {
...
 
 func WaitForStalldTo(run bool, interval, timeout time.Duration, node *corev1.Node) error {
-	return wait.Poll(interval, timeout, func() (bool, error) {
-		cmd := []string{"/bin/bash", "-c", "pidof stalld || true"}
-		stalldPid, err := nodes.ExecCommandOnNode(cmd, node)
-		if err != nil {
-			return false, fmt.Errorf("failed to execute command %q on node: %q; %w", cmd, node.Name, err)
-		}
+	return wait.PollImmediate(interval, timeout, func() (bool, error) {
+		cmd := []string{"/bin/bash", "-c", "pidof stalld"}
```
Tal-or, line 88: When stalld disabled, `pidof stalld` will return with an error, this is why I added the `|| true`.
Plus not checking the error returned from `ExecCommandOnNode` is also not so great.

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
```diff
@@ -84,38 +84,34 @@ func GetPod(ctx context.Context, node *corev1.Node) (*corev1.Pod, error) {
...
 		if !run { // we don't want stalld to run
 			if err == nil {
-				return false, fmt.Errorf("node=%q stalld_pid=%q stalld is running when it shouldn't", node.Name, stalldPid)
+				return false, nil
 			}
+			return true, nil
 		}
 		// we want stalld to run
 		if err != nil {
-			klog.Warningf("node=%q stalld_pid=%q is not a valid pid number: %v", node.Name, stalldPid, err)
```
Tal-or, line 102: Leave the warning it doesn't affect the test result and provide some good indication.

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
```diff
@@ -84,38 +84,34 @@ func GetPod(ctx context.Context, node *corev1.Node) (*corev1.Pod, error) {
...
-		}
+	return wait.PollImmediate(interval, timeout, func() (bool, error) {
+		cmd := []string{"/bin/bash", "-c", "pidof stalld"}
+		stalldPid, _ := nodes.ExecCommandOnNode(cmd, node)
 
-		_, err = strconv.Atoi(stalldPid)
+		_, err := strconv.Atoi(stalldPid)
 		if !run { // we don't want stalld to run
 			if err == nil {
-				return false, fmt.Errorf("node=%q stalld_pid=%q stalld is running when it shouldn't", node.Name, stalldPid)
```
Tal-or, line 97: Yes this was definitely a mistake since we need to wait some time for TuneD to catch up, but I would replace the error with a warning 
` klog.Warningf("node=%q stalld_pid=%q stalld is still running when it shouldn't", node.Name, stalldPid)`

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
```diff
@@ -84,38 +84,34 @@ func GetPod(ctx context.Context, node *corev1.Node) (*corev1.Pod, error) {
...
+				return false, nil
 			}
+			return true, nil
```
Tal-or, line 96: Here it means that `err != nil`. 
It can happens for various reasons if we're not checking that the output received from `ExecCommandOnNode` is OK

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
yanirq: there were some test cases where the the command itself with the condition actually failed. this is why we moved to a different kind of check (see the rest of the code)

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
yanirq: this will spam the log and the error message was moved to a higher level for better visibility.  Knowing the pid number itself does not provide too much helpful info in this particular case.

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
yanirq: I rather have this manifested as an actual error (see higher level)

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
yanirq: since we expect `ExecCommandOnNode` command to return error anyway (keep me honest here) then we will need to introduce some error handling here by error type, e.g : if error was for the execution itself or the error came out from not having stalld running

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
Tal-or: The problem is that we're not checking the error returned from `nodes.ExecCommandOnNode(cmd, node) ` hence we don't know whether the command execution was successful.

I don't mind changing the executed command  to something that doesn't contain pipes but we need to check the error anyway

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
Tal-or: Then you can returned it at the end of this function instead of write it down multiple times

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
Tal-or: > since we expect ExecCommandOnNode command to return error anyway 
No we don't that is my point exactly. if `ExecCommandOnNode` return with an error, you don't know whether it's because stalld is not running (which is sometimes a good thing, meaning the test passes)  or the execution itself failed. 
> need to introduce some error handling here by error type
We don't know what kind of errors might return so we perform such a check efficiently

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
Tal-or: It would at least give us some indication about where the test get stuck and how much time it takes.

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
yanirq: I would argue that that was the exact behavior even before this patch ,  we would still return true at the end of the function and in the last observation we have this was always the case since the syntax with || true was always failing. we can however get the error message from execute command on node and print it out as a warning to provide more debuging info.

yanirq:
/retest
infra issues

Tal-or (commented):
Nice catch!

yanirq:
test suit times out at the last test due to the long run time . 
Currently only the last test is not able to run on time ("ContainerRuntimeConfig") which on local tests always past.

Tal-or:
/lgtm
But we still need to figure out the timeout issue

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
```diff
@@ -84,38 +84,36 @@ func GetPod(ctx context.Context, node *corev1.Node) (*corev1.Pod, error) {
...
 	})
 }
 
 func CheckParameters(node *corev1.Node, sysctlMap map[string]string, kernelParameters []string, stalld, rtkernel bool) {
-	cmd := []string{"/bin/bash", "-c", "pidof stalld || true"}
+	cmd := []string{"/bin/bash", "-c", "pidof stalld"}
 	By(fmt.Sprintf("Executing %q", cmd))
-	stalldPid, err := nodes.ExecCommandOnNode(cmd, node)
-	ExpectWithOffset(1, err).ToNot(HaveOccurred(), "failed to execute command %q on node: %q; %w", cmd, node.Name, err)
+	stalldPid, _ := nodes.ExecCommandOnNode(cmd, node)
```
Tal-or, line 111: This is still an issue, we can't ignore the error here

Tal-or:
/hold
We still need to discuss about https://github.com/openshift/cluster-node-tuning-operator/pull/626#discussion_r1182131150

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
yanirq: ok, so I might bring back the pipe format that was non consistently failing  maybe due to the wg issue we had and test it

Tal-or: This is not an actual piping BTW, just an OR operator.
@jmencak Could you please elaborate when did you encounter issues with this `||` operator when it's part of a command executed by `node.ExecCommandOnNode`?

yanirq: sorry , I meant the OR operator

yanirq: Jiri have noticed that using || was flaky - depending also on stalld running or not.
but since I have discovered an issue with the parrallel runs of go procedure not properly blocked by the wg.wait (fixed in the last commit here) this might have been the real issue. 
I will test the previous use of || true first and will re-introduce it if successful.

yanirq: @Tal-or  ok , I have tested this and it does fail . node.ExecCommandOnNode does not play nicely with the OR condition

jmencak: > This is not an actual piping BTW, just an OR operator. @jmencak Could you please elaborate when did you encounter issues with this `||` operator when it's part of a command executed by `node.ExecCommandOnNode`?

I don't exactly remember, but if you test (as Yanir did), you should see it.  The test was completely stuck depending on whether `stalld` was running.

jmencak: I'd strongly suggest reviewing the e2e tests for this as they might be the reason for the tests sometimes taking too long/timing out.

Tal-or: What error do we get?
Then we should find another way catching error returning from `node.ExecCommandOnNode` 
Maybe we can postpone it to another PR

yanirq: The rest of the tests which are taking a significant chunck of time are tests pending on node reboots and waiting for mcp updates. we have a plan to reduce the wait time for mcps but this will be done in another PR after some desgin.
for the short term this PR should be merged + splitting the lanes further as suggested here : https://github.com/openshift/cluster-node-tuning-operator/pull/631

jmencak: > What error do we get?

In my case none.  It was just stuck/hanged forever until I pressed Ctrl+C.

yanirq: The error coming out from `node.ExecCommandOnNode(pidof stalld || true )` is: `timed out waiting for the condition`

Tal-or:
/lgtm
/hold cancel
Thank you @yanirq!

PR #641, opened 2023-04-29
OCPBUGS-10424: e2e: Fix RPS test for multi-worker cluster

zeeke:
Fix `[test_id: 59572] Check RPS Mask is applied to atleast one single rx queue on all veth interface` test bug that occurs when running against a cluster with multiple worker nodes.

@mrniranjan @jlojosnegros Please, take a look.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -303,8 +303,9 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 			count := 0
 			expectedRPSCPUs, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
 			Expect(err).ToNot(HaveOccurred())
-			var vethInterfaces = []string{}
+
 			for _, node := range workerRTNodes {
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -303,8 +303,9 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 			count := 0
```
ffromani, line 303:
```
Fix `[test_id: 59572] Check RPS Mask is applied to atleast
one single rx queue on all veth interface` test  bug
that occurs when running against a cluster with multiple
worker nodes.
```
I'd add:
```
the list of affected 'veth' interface is per-node, but the tests iterates over all the worker nodes and failed to reset the list when moving from a node to another, possibly checking stale interfaces.
```

zeeke: sounds good

mrniranjan:
looks good to me.

ffromani:
thanks @mrniranjan !

ffromani:

known failure, we have PRs tackling it already

PR #644, opened 2023-05-03
OCPBUGS-12978: use WatchNamespace() when deleting Profiles

dagrayvid:
In HyperShift, WatchNamespace() is the namespace in hosted cluster where Profile and Tuned objects exist, which is distinct from the OperatorNamespace() which is named based on the name of the HostedCluster. 

In standalone OCP, these will be the same namespace (`openshift-cluster-node-tuning-operator`) (by default, at least...)


jmencak:
Thank you for the fix, David.  Makes sense to me.
/lgtm

dagrayvid:
Infra issue.

/retest-required

PR #646, opened 2023-05-05
OCPBUGS-13148: Configure cpu balancing cpu sets for all clusters

MarSik:
The current code configures the cpu balancing management cpuset cgroup structure only when Workload partitioning is enabled.

That is wrong as even standard clusters use that feature.


browsell:
LGTM

ffromani (commented):
/lgtm

PR #650, opened 2023-05-10
OCPBUGS-4194: rps: use default rps mask kernel API

Tal-or:
In OCP >= 4.12 a new feature added named UDP GRO.
As part of this this feature implemention, virtual
interfaces get created with multiple queues.

A new veth creation event invokes the set-rps script,
but it might be running before all NIC's queues get created.
Hence the rps mask not apply to the uncreated queues, which
get created later on without the mask.

https://lore.kernel.org/netdev/cover.1676635317.git.pabeni@redhat.com/
exposes a new API to configure default rps mask for network devices.
By creating the veth with the desired RPS in as the default we're avoiding the 
problem described above.

We are configuring the rps mask directly from sysctl and not
from TuneD because at the boot process
sysctl spins up earlier than the veth devices.


assets/performanceprofile/configs/99-default-rps-mask.conf
```diff
@@ -1,5 +1,5 @@
-# We should apply the rule on the virtual interfaces of the host, because of the RPS mask that would be consulted,
-# is the one on the RX side of the veth in the host.
+# Apply the RPS mask on the virtual interfaces of the host by default, becasue
+# from the container perspective the RPS mask the would be consulted, is the one on the RX side of the veth in the host.
```
MarSik, line 2:
```suggestion
# from the container perspective the RPS mask that will be consulted, is the one on the RX side of the veth in the host.
```

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -109,6 +109,9 @@ vm.dirty_background_ratio=3
 #> latency-performance
 vm.swappiness=10
 
+#> rps configuration
```
MarSik, line 112: Please add a note that we also configure this via a sysctl.d file

Tal-or: Done. Thanks

PR #651, opened 2023-05-15
Expose TuneD socket to host.

jmencak:
This change exposes `TuneD` socket interface to the host OS.  This will, for example, allow to query and/or change a `TuneD` profile by using `TuneD` socket API.  For example, issuing the following (by the root user) from the host will return the active `TuneD` profile.

```
printf '{"jsonrpc": "2.0", "method": "active_profile", "id": 1}' | nc -U /run/tuned/tuned.sock
{"jsonrpc": "2.0", "id": 1, "result": "openshift-node"}
```

### DIFF
--- a/assets/bin/run
+++ b/assets/bin/run
@@ -2,6 +2,10 @@
 
 export SYSTEMD_IGNORE_CHROOT=1
 
+enable_tuned_unix_socket() {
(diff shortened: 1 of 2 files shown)


dagrayvid:
Thanks @jmencak, this change looks good to me now.

jmencak:
Thank you guys.  Can we get the LGTM label then?

PR #653, opened 2023-05-16
OCPBUGS-14137: e2e: perfprof: add SNO device recovery test

ffromani:
Add test to verify the device recovery on SNO nodes.
    
This test has unusual and potentially challenging requirements
- SRIOV stack fully configured
- SNO cluster
- needs to reboot the SNO cluster
    
For this reason, the test requires to be given explicit which node to use, and skips if a target node is not explicitely given using the `DEVICE_RECOVERY_TARGET_NODE` environment variable.
    
SRIOV setup is assumed, but the test makes no attempt to enforce, intentionally. The SRIOV device to consume can be overridden using the `DEVICE_RECOVERY_RESOURCE_NAME` environment variable, depending on the setup.    

The test needs to create test workload (deployment, pod) but more importantly needs to reboot the node. It tries hard to tolerate connection loss to the control plane (SNO rebooted -> control plane loss) but can't guarantee yet full reliability.

### DIFF
--- a/test/e2e/performanceprofile/functests/8_reboot/devices.go
+++ b/test/e2e/performanceprofile/functests/8_reboot/devices.go
@@ -1,0 +1,384 @@
+package __reboot_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	testnodes "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
+	testpods "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
+)
+
+// This test has a very high disruption potential: it wants to bypass the node cordoning et. al.
+// and kubernetes management and trigger a reboot straight from the node.
+// Additionally, it explicitly considers SNO, where the disruption potential is even higher.
+// Because of all the above, it must opted-in explicitly by supplying the `DEVICE_RECOVERY_TARGET_NODE`
+// variable pointing to a node in the cluster, SNO or worker node.
+
+const (
+	rebootNodeCommandMCD = "chroot /rootfs systemctl reboot"
+
+	// note NO DEFAULT for targetNode - intentionally
+	sriovDeviceResourceNameDefault = "openshift.io/dpdk_nic_1"
+	workloadContainerImageDefault  = "quay.io/fromani/devaccess:2023051001" // FIXME
+
+	targetNodeEnvVar              = "DEVICE_RECOVERY_TARGET_NODE"
+	sriovDeviceResourceNameEnvVar = "DEVICE_RECOVERY_RESOURCE_NAME"
+	workloadContainerImageEnvVar  = "DEVICE_RECOVERY_CONTAINER_IMAGE"
+
+	// this is the only timeout tunable because is a hard sleep. Every other long timeout is within a polling loop.
+	rebootCooldownEnvVar  = "REBOOT_COOLDOWN_TIME"
+	rebootCooldownDefault = 1 * time.Minute
+
+	// the more VFs, the more likely the bug to trigger
+	minimumAllocatableDevices = 32
+)
+
+var (
+	workloadLabels = map[string]string{
+		"app": "devaccess-test",
+	}
+)
+
+var _ = Describe("[disruptive][node][kubelet][devicemanager] Device management tests", func() {
+	defer GinkgoRecover()
+
+	var (
+		rebootCooldownTime = rebootCooldownDefault
+
+		targetNode              string
+		sriovDeviceResourceName = sriovDeviceResourceNameDefault
+		workloadContainerImage  = workloadContainerImageDefault
+	)
+
+	BeforeEach(func() {
+		targetNode = os.Getenv(targetNodeEnvVar)
+		if targetNode == "" {
+			Skip(fmt.Sprintf("Need an explicit target node name, got none (env var: %q)", targetNodeEnvVar))
+		}
+		testlog.Infof("target node name: %q", targetNode)
+
+		if val, ok := os.LookupEnv(sriovDeviceResourceNameEnvVar); ok {
+			sriovDeviceResourceName = val
+		}
+		testlog.Infof("(SRIOV) device resource name: %q", sriovDeviceResourceName)
+
+		if val, ok := os.LookupEnv(workloadContainerImageEnvVar); ok {
+			workloadContainerImage = val
+		}
+		testlog.Infof("workload container image: %q", workloadContainerImage)
+
+		var err error
+		if val, ok := os.LookupEnv(rebootCooldownEnvVar); ok {
+			rebootCooldownTime, err = time.ParseDuration(val)
+			Expect(err).ToNot(HaveOccurred(), "error parsing user-provided cooldown: %v", val)
+		}
+		testlog.Infof("reboot cooldown time: %v", rebootCooldownTime)
+
+		_, err = testnodes.GetByName(targetNode)
+		Expect(err).ToNot(HaveOccurred(), "error getting the target node %q", targetNode)
+	})
+
+	It("Verify that pods requesting devices are correctly recovered on node restart", func() {
+		namespace := testutils.NamespaceTesting
+
+		// refresh the targetNode object
+		node, err := testnodes.GetByName(targetNode)
+		Expect(err).ToNot(HaveOccurred(), "error getting the target node %q", targetNode)
+
+		// phase1: complete the node praparation: make sure we have enough devices. Short timeout, we should be idle now.
+		allocatableDevices := waitForNodeToReportResourcesOrFail("pre reboot", targetNode, sriovDeviceResourceName, 2*time.Minute, 2*time.Second)
+		Expect(allocatableDevices).To(BeNumerically(">=", minimumAllocatableDevices), "device %q has too low amount available - testing scenario unreliable", sriovDeviceResourceName, allocatableDevices)
+
+		// phase2: node is prepared, run the test workload and check it gets the device it expected
+		wlDp := makeWorkloadDeployment(namespace, workloadContainerImage, sriovDeviceResourceName, 1)
+		err = testclient.Client.Create(context.TODO(), wlDp)
+		Expect(err).ToNot(HaveOccurred(), "error creating workload deployment")
+
+		// short timeout: we are on idle cluster
+		wlDp = waitForDeploymentCompleteOrFail("pre reboot", *wlDp, 5*time.Minute)
+		testlog.Infof("deployment %s/%s running", wlDp.Namespace, wlDp.Name)
+
+		// phase3: the reboot, which trigger the very scenario we are after
+
+		// managed, clean restart (e.g. `reboot` command or `systemctl reboot`
+		// - details don't matter as long as this is a managed clean restart).
+		// Power loss scenarios, aka hard reboot, deferred to another test.
+		// intentionally ignoring error. We need to tolerate connection error or disconnect
+		// because the node is rebooting.
+		runCommandOnNodeThroughMCD(node, "reboot", rebootNodeCommandMCD)
+		// this is (likely) a SNO. We need to tolerate connection errors,
+		// because the apiserver is going down as well.
+		// we intentionally use a generous timeout.
+		// On Bare metal reboot can take a while.
+		waitForNodeReadyOrFail("post reboot", targetNode, 20*time.Minute, 3*time.Second)
+
+		// are we really sure? we can't predict if we will have state flapping,
+		// we can't predict if pods go back to containercreating and ideally we
+		// should have no flapping.
+		// Tracking all the state will make the test complex *and fragile*.
+		// The best we can do right now is to let the SNO cool down and check again.
+		testlog.Infof("post reboot: entering cooldown time: %v", rebootCooldownTime)
+		time.Sleep(rebootCooldownTime)
+		testlog.Infof("post reboot: finished cooldown time: %v", rebootCooldownTime)
+
+		// longer timeout. We expect things to be still on flux
+		waitForNodeToReportResourcesOrFail("post reboot", targetNode, sriovDeviceResourceName, 10*time.Minute, 2*time.Second)
+
+		// from now on we assume the node is somehow stable again and we stop swallowing the connection errors
+
+		// what happened to the previous workload? Note we need a long timeout, because we have no means to learn when the SNO is actually settled
+		wlDp = waitForDeploymentCompleteOrFail("post reboot", *wlDp, 30*time.Minute)
+		testlog.Infof("deployment %s/%s running", wlDp.Namespace, wlDp.Name)
+
+		// do we have any pods from the workload deployment on UnexpectedAdmissionError?
+		// we usually do, but note it's legit to not have them at all if we got lucky.
+		pods, err := listPodsByDeployment(testclient.Client, context.TODO(), *wlDp)
+		Expect(err).ToNot(HaveOccurred(), "error checking the pods belonging to deployment %s/%s", wlDp.Namespace, wlDp.Name)
+		admissionFailed := 0
+		for idx := range pods {
+			if isUnexpectedAdmissionError(pods[idx], sriovDeviceResourceName) {
+				admissionFailed++
+			}
+		}
+		testlog.Infof("workload deployment %s/%s got %d admission errors (created %d pods to go running)", wlDp.Namespace, wlDp.Name, admissionFailed, len(pods))
+
+		// phase4: sanity check that a new workload works as expected
+		wlPod := makeWorkloadPod(namespace, workloadContainerImage, sriovDeviceResourceName)
+
+		err = testclient.Client.Create(context.TODO(), wlPod)
+		Expect(err).ToNot(HaveOccurred(), "error creating workload pod post reboot")
+
+		// things should be settled now so we can use again a short timeout
+		testlog.Infof("post reboot: running a fresh pod %s/%s resource=%q", wlPod.Namespace, wlPod.Name, sriovDeviceResourceName)
+		updatedPod, err := testpods.WaitForPredicate(wlPod, 1*time.Minute, func(pod *corev1.Pod) (bool, error) {
+			return isPodReady(*pod), nil
+		})
+		Expect(err).ToNot(HaveOccurred(), "error checking the workload pod post reboot")
+		testlog.Infof("post reboot: newer workload pod %s/%s admitted: %s", updatedPod.Namespace, updatedPod.Name, extractPodStatus(updatedPod.Status))
+	})
+})
+
+func makeWorkloadPodSpec(imageName, resourceName string) corev1.PodSpec {
+	true_ := true
+	false_ := false
+	zero := int64(0)
+	rl := corev1.ResourceList{
+		corev1.ResourceCPU:                *resource.NewQuantity(2, resource.DecimalSI),
+		corev1.ResourceMemory:             resource.MustParse("512Mi"),
+		corev1.ResourceName(resourceName): *resource.NewQuantity(1, resource.DecimalSI),
+	}
+	return corev1.PodSpec{
+		RestartPolicy:                 corev1.RestartPolicyAlways,
+		TerminationGracePeriodSeconds: &zero,
+		Containers: []corev1.Container{{
+			Image:           imageName,
+			ImagePullPolicy: corev1.PullAlways,
+			Name:            "devaccess-container",
+			Resources: corev1.ResourceRequirements{
+				Limits:   rl,
+				Requests: rl,
+			},
+			SecurityContext: &corev1.SecurityContext{
+				Capabilities: &corev1.Capabilities{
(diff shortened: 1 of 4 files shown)


ffromani:
A clean run looks like this:
```
[disruptive][kubeelet][node] Device management tests Verify that pods requesting devices are correctly recovered on node restart
/home/fromani/go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/7_performance_kubelet_node/devices.go:96
  May 17 16:48:20.227: [INFO]: target node name: "sno.cnf.lab"
  May 17 16:48:20.227: [INFO]: (SRIOV) device resource name: "openshift.io/dpdk_nic_1"
  May 17 16:48:20.227: [INFO]: workload container image: "quay.io/fromani/devaccess:2023051001"
  May 17 16:48:20.227: [INFO]: reboot cooldown time: 1m0s
  May 17 16:48:20.503: [INFO]: pre reboot: wait for target node "sno.cnf.lab" to report resource "openshift.io/dpdk_nic_1"
  May 17 16:48:20.619: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=64
  May 17 16:48:20.619: [INFO]: pre reboot: reporting resources from node "sno.cnf.lab": openshift.io/dpdk_nic_1=64
  May 17 16:48:20.742: [INFO]: pre reboot: wait for deployment performance-addon-operators-testing/devaccess-workload-deployment to be ready
  May 17 16:48:20.838: [INFO]: deployment performance-addon-operators-testing/devaccess-workload-deployment replicas=1 updated=1 available=0
  May 17 16:48:25.936: [INFO]: deployment performance-addon-operators-testing/devaccess-workload-deployment replicas=1 updated=1 available=1
  May 17 16:48:25.936: [INFO]: pre reboot: reported deployment performance-addon-operators-testing/devaccess-workload-deployment ready
  May 17 16:48:25.936: [INFO]: deployment performance-addon-operators-testing/devaccess-workload-deployment running
  May 17 16:48:25.936: [INFO]: node "sno.cnf.lab": before reboot
  May 17 16:48:26.031: [INFO]: found mcd machine-config-daemon-t2tdm for node sno.cnf.lab
  May 17 16:48:42.345: [INFO]: node "sno.cnf.lab": output=[]
  May 17 16:48:42.345: [INFO]: node "sno.cnf.lab": after reboot
  May 17 16:48:42.345: [INFO]: post reboot: waiting for node "sno.cnf.lab": to be ready
  May 17 16:48:42.436: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:48:45.560: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:48:48.653: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:48:51.746: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:48:54.841: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:48:57.938: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:49:01.031: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:49:04.123: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:49:07.216: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:49:10.308: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:49:43.310: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:50:16.315: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:50:49.320: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:51:22.321: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:51:55.326: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:52:28.328: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:52:41.426: [INFO]: failed to get node "sno.cnf.lab": failed to get node for the node "sno.cnf.lab"
  May 17 16:52:44.797: [INFO]: node "sno.cnf.lab" ready=true
  May 17 16:52:44.797: [INFO]: post reboot: node "sno.cnf.lab": reported ready
  May 17 16:52:44.797: [INFO]: post reboot: entering cooldown time: 1m0s
  May 17 16:53:44.799: [INFO]: post reboot: finished cooldown time: 1m0s
  May 17 16:53:44.799: [INFO]: post reboot: wait for target node "sno.cnf.lab" to report resource "openshift.io/dpdk_nic_1"
  May 17 16:53:44.984: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:53:47.167: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:53:49.351: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:53:51.535: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:53:53.718: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:53:55.911: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:53:58.103: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:00.292: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:02.485: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:04.670: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:06.854: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:09.038: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:11.223: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:13.407: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:15.591: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:17.785: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:19.971: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:22.267: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:24.451: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:26.635: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:28.833: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:31.045: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:33.236: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:35.419: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:37.601: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:39.786: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:41.969: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:44.154: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:46.338: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:48.520: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:50.704: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:52.887: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:55.070: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:57.254: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:54:59.438: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:01.623: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:03.805: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:05.990: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:08.178: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:10.361: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:12.544: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:14.727: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:16.918: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=0
  May 17 16:55:19.102: [INFO]: node "sno.cnf.lab" resource=openshift.io/dpdk_nic_1 allocatable=64
  May 17 16:55:19.102: [INFO]: post reboot: reporting resources from node "sno.cnf.lab": openshift.io/dpdk_nic_1=64
  May 17 16:55:19.102: [INFO]: post reboot: wait for deployment performance-addon-operators-testing/devaccess-workload-deployment to be ready
  May 17 16:55:19.195: [INFO]: deployment performance-addon-operators-testing/devaccess-workload-deployment replicas=1 updated=1 available=0
  May 17 16:55:24.289: [INFO]: deployment performance-addon-operators-testing/devaccess-workload-deployment replicas=1 updated=1 available=0
  May 17 16:55:29.384: [INFO]: deployment performance-addon-operators-testing/devaccess-workload-deployment replicas=1 updated=1 available=1
  May 17 16:55:29.385: [INFO]: post reboot: reported deployment performance-addon-operators-testing/devaccess-workload-deployment ready
  May 17 16:55:29.385: [INFO]: deployment performance-addon-operators-testing/devaccess-workload-deployment running
  May 17 16:55:29.479: [INFO]: workload deployment performance-addon-operators-testing/devaccess-workload-deployment got 1 admission errors
  May 17 16:55:33.813: [INFO]: post reboot: newer workload pod / admitted: phase="" reason="" message=""
• [433.586 seconds]
```

test/e2e/performanceprofile/functests/8_reboot/devices.go
```diff
@@ -0,0 +1,388 @@
...
+
+		if val, ok := os.LookupEnv(workloadContainerImageEnvVar); ok {
+			workloadContainerImage = val
+		}
+		testlog.Infof("workload container image: %q", workloadContainerImage)
+
+		var err error
+		if val, ok := os.LookupEnv(rebootCooldownEnvVar); ok {
+			rebootCooldownTime, err = time.ParseDuration(val)
+			Expect(err).ToNot(HaveOccurred(), "error creating workload pod")
```
swatisehgal, line 88: Error message does not seem right here.

test/e2e/performanceprofile/functests/8_reboot/devices.go
```diff
@@ -0,0 +1,388 @@
...
+		waitForNodeReadyOrFail("post reboot", targetNode, 20*time.Minute, 3*time.Second)
+
+		// are we really sure? we can't predict if we will have state flapping,
+		// we can't predict if pods go back to containercreating and ideally we
+		// should have no flapping.
+		// Tracking all the state will make the test complex *and fragile*.
+		// The best we can do right now is to let the SNO cool down and check again.
+		testlog.Infof("post reboot: entering cooldown time: %v", rebootCooldownTime)
+		time.Sleep(rebootCooldownTime)
+		testlog.Infof("post reboot: finished cooldown time: %v", rebootCooldownTime)
```
swatisehgal, line 137: This is interesting, how did you come to the conclusion that a minute is sufficient for post reboot cooldown? Was it after some experimentation/testing with a real cluster?

test/e2e/performanceprofile/functests/8_reboot/devices.go
```diff
@@ -0,0 +1,388 @@
...
+			admissionFailed := 0
+			for idx := range pods {
+				if isUnexpectedAdmissionError(pods[idx], sriovDeviceResourceName) {
+					admissionFailed++
+				}
+			}
+			testlog.Infof("workload deployment %s/%s got %d admission errors", wlDp.Namespace, wlDp.Name, admissionFailed)
+		} else {
+			testlog.Infof("!!! workload deployment %s/%s got running at first try !!!", wlDp.Namespace, wlDp.Name)
+		}
```
swatisehgal, line 162: I don't feel too strongly but this seemed a bit complicated to me on the first go. Took me a minute to wrap my head around what's going on here. In general it is okay but perhaps it would be more straightforward to identify the pod name before node reboot happens and query for that specific pod via the API and crosscheck its status? 
Another reason for my suggestion is that we are creating the deployment with one replica so it makes sense to identify the pod that was created as part of the deployment pre-reboot and check for it later. If we had multiple replicas, what you have currently makes more sense.

swatisehgal (commented):
Thanks for your work on this! Really appreciate it. I tested this in my environment and it works like a charm :)

I left a few comments/suggestions but I think we are there already :)

test/e2e/performanceprofile/functests/8_reboot/devices.go
```diff
@@ -0,0 +1,388 @@
...
+	minimumAllocatableDevices = 32
+)
+
+var (
+	workloadLabels = map[string]string{
+		"app": "devaccess-test",
+	}
+)
+
+var _ = Describe("[disruptive][kubeelet][node] Device management tests", func() {
```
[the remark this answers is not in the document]
ffromani: sure, done. Intentionally used `lowercase` for consistency.

test/e2e/performanceprofile/functests/8_reboot/devices.go
ffromani: thanks, fixed

test/e2e/performanceprofile/functests/8_reboot/devices.go
ffromani: this is a tricky part. I did some experimentation and come up with "high enough" value. 
I'm trying to avoid the state flapping without adding fragile code, but I'm not so keen on sleeping.

test/e2e/performanceprofile/functests/8_reboot/devices.go
ffromani: I'm not keen of complex code (simplified a bit this area after your comment). I think it's important to capture this detail to implement a good testcase. The reason why we are doing this here post-reboot are:

1. we should check what happens to the deployment on the recovery path. Depending on unpredictable recovery ordering and cluster load, we may have few attempts from the deployment controller to create the single replica we need, with few failures for UnexpectedAdmissionError.
2. we should check the fate of all the attempts, trying to detect unexpected failures (e.g. we expect all the failures to be caused by the fix, not because other reasons)
3. on top of all the above, actually going running at the first try is legit!

For all these reasons I'm inclined to make the test check all the pods pertaining to the deployment post-reboot. I reckon this implementation of the concept may be suboptimal and I'm open to suggestions and improvements here.

swatisehgal: Fair enough, let's check for the whole deployment.

test/e2e/performanceprofile/functests/8_reboot/devices.go
swatisehgal: What I wasn't able to fully understand and am interested in learning here is that in spite of giving enough time (20 minutes) for the node to reboot why exactly did we need to wait another minute for the system to not be in flux? IOW, when you say "We expect things to be still on flux" what do we mean by that?

ffromani: The things I observed when SNO reboots is that  the state reported by apiserver flaps while the processes start up and expected communications resume. In practical terms, the apiserver SEEMS to answer queries with the latest consolidated status before reboot. For examples, pods are initially (= at the first query) reported Running, while it's not actually true, but the kubelet didn't update their status yet.

I must say I didn't dive too deep in here, but intuitively the theory has at least some merit.

This flapping may cause the test (or any other client) to misdetect things settled too early. The real pod status for example was observed going
Running -> Pending -> [a minute or so] -> Running again.

Tracking the state changes is hard and likely fragile, because on SNO reboot I didn't find a way to trust the apiserver until the cluster is actually settled, which is learned through a combination of factors (multiple parties reporting ready like node, all pods...).

Instead of chasing all these factors I added a sleep to let the cluster stabilize and bypass (some) flapping.
I think this can also be handled using a combination of `Eventually` and `Consistently`, but I didn't invest too much in this area yet.

swatisehgal: > The things I observed when SNO reboots is that the state reported by apiserver flaps while the processes start up and expected communications resume. In practical terms, the apiserver SEEMS to answer queries with the latest consolidated status before reboot. For examples, pods are initially (= at the first query) reported Running, while it's not actually true, but the kubelet didn't update their status yet.

I have noticed that as well and know exactly what you are talking about.
 
> This flapping may cause the test (or any other client) to misdetect things settled too early. The real pod status for example was observed going Running -> Pending -> [a minute or so] -> Running again.

What I noticed was that the recovered pod goes Running and then after the kubelet is back (in cases where device plugins are not up or haven't registered) it fails at  admission.

The new pod created is because of the deployment controller spinning up a new pod, goes into pending because the pod has to wait for the node allocatable to be updated and only when that happens the pod goes Running.

> Tracking the state changes is hard and likely fragile, because on SNO reboot I didn't find a way to trust the apiserver until the cluster is actually settled, which is learned through a combination of factors (multiple parties reporting ready like node, all pods...).
> 
> Instead of chasing all these factors I added a sleep to let the cluster stabilize and bypass (some) flapping. I think this can also be handled using a combination of `Eventually` and `Consistently`, but I didn't invest too much in this area yet.

I tend to be in the favor of using `Eventually` by waiting till the pod goes into admission error rather than putting an arbitrary sleep. I acknowledge that we could have cases where the pod correctly goes into running state when the SR-IOV devices are ready to be allocated. However, based on input from Sebastian and what I have noticed from my own testing it appears that if we have a large number of virtual function we can be fairly sure that it will take time for the config-daemon to bind them to the DPDK driver allowing us to be certain that the pod would fail at admission. Can we check for this in eventually as it is a more educated guess rather than an arbitrary sleep (which we cannot be sure about) ? 

Pretty much brainstorming here. Let me know what you think.

ffromani: > > This flapping may cause the test (or any other client) to misdetect things settled too early. The real pod status for example was observed going Running -> Pending -> [a minute or so] -> Running again.
> 
> What I noticed was that the recovered pod goes Running and then after the kubelet is back (in cases where device plugins are not up or haven't registered) it fails at admission.

This is exactly the problem: this pod is *first* wrongly reported Running, so `Eventually` won't help. We need `Eventually` + `Consistently`, both wrapped in a retry loop.

Pod goes running -> Eventually succeeds;
Pods STAYS running (is actually running for real) -> Consistently also succeeds

But if it's a flap, the Eventually succeeds and then the Consistently will fail (because the real state will put the pod back in Pending or even UnexpectedAdmissionError)
so we need to retry the pair a few times to really catch the correct sequence of events.

Doing like this looks more correct, but also looks more complex and fragile. Ignoring the state "for a while" seems a brutal yet surprisingly effective way to sidestep this issue.

> > Tracking the state changes is hard and likely fragile, because on SNO reboot I didn't find a way to trust the apiserver until the cluster is actually settled, which is learned through a combination of factors (multiple parties reporting ready like node, all pods...).
> > Instead of chasing all these factors I added a sleep to let the cluster stabilize and bypass (some) flapping. I think this can also be handled using a combination of `Eventually` and `Consistently`, but I didn't invest too much in this area yet.
> 
> I tend to be in the favor of using `Eventually` by waiting till the pod goes into admission error rather than putting an arbitrary sleep.

Note we do both: the sleep is "just" to skip the initial transient phase until components catch up.

>  I acknowledge that we could have cases where the pod correctly goes into running state when the SR-IOV devices are ready to be allocated. However, based on input from Sebastian and what I have noticed from my own testing it appears that if we have a large number of virtual function we can be fairly sure that it will take time for the config-daemon to bind them to the DPDK driver allowing us to be certain that the pod would fail at admission. Can we check for this in eventually as it is a more educated guess rather than an arbitrary sleep (which we cannot be sure about) ?

I agree we should improve here but I'm struggling to find a viable way forward :\

swatisehgal: Yep, this is a very tricky one indeed. Let's leave it as is for now and see if we can improve at a later stage.

swatisehgal (approved):
/lgtm

ffromani:
/cherry-pick release-4.13

let's see how it goes

PR #658, opened 2023-05-18
e2e: latency: Add new env var for latency tools path

jlojosnegros:
Latency test suite uses an executable to run latency tests with different parameters.

This file's complete filename is hardcoded in the suite.

Being able to configure the path would make building cnf-test latency test image easier and, as if the variable is not set or empty default value would be used, this change is backward compatible.

### DIFF
--- a/test/e2e/performanceprofile/functests/5_latency_testing/5_latency_testing_suite_test.go
+++ b/test/e2e/performanceprofile/functests/5_latency_testing/5_latency_testing_suite_test.go
@@ -31,9 +31,6 @@ import (
 	qe_reporters "kubevirt.io/qe-tools/pkg/ginkgo-reporters"
 )
 
-// TODO get commonly used variables from one shared file that defines constants
-const testExecutablePath = "../../../../../build/_output/bin/latency-e2e.test"
-
 var prePullNamespace = &corev1.Namespace{
 	ObjectMeta: metav1.ObjectMeta{
 		Name: "testing-prepull",
@@ -42,7 +39,7 @@ var prePullNamespace = &corev1.Namespace{
 var profile *performancev2.PerformanceProfile
 
 var _ = BeforeSuite(func() {
-	Expect(isTestExecutableFound()).To(BeTrue())
+	Expect(isTestExecutableFound(testutils.LatencyToolsPath)).To(BeTrue(), fmt.Sprintf("Unable to find latency tools at %s", testutils.LatencyToolsPath))
 	Expect(testclient.ClientsEnabled).To(BeTrue())
 
 	// update PP isolated CPUs. the new cpu set for isolated should have an even number of CPUs to avoid failing the pod on SMTAlignment error,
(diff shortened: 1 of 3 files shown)


test/e2e/performanceprofile/functests/5_latency_testing/5_latency_testing_suite_test.go
```diff
@@ -124,8 +121,9 @@ func createNamespace() error {
 	return err
 }
 
-func isTestExecutableFound() bool {
-	if _, err := os.Stat(testExecutablePath); os.IsNotExist(err) {
+func isTestExecutableFound(path string) bool {
+	testlog.Infof("[jlom] looking for tools at: %s", path)
```
ffromani, line 125: this can be a regular log, just drop the `jlom` tag

test/e2e/performanceprofile/functests/utils/consts.go
```diff
@@ -24,7 +26,16 @@ var NodesSelector string
 // ProfileNotFound is true when discovery mode is enabled and no valid profile was found
 var ProfileNotFound bool
 
+// Complete file name for the latency tools executable
+var LatencyToolsPath string
+
 func init() {
+
+	LatencyToolsPath = os.Getenv("LATENCY_TOOLS_PATH")
```
Tal-or, line 34: Usually https://pkg.go.dev/os#LookupEnv is better because it help differentiate between an empty value and unset value

jlojosnegros: Yeah. I used that but as the logic in case of variable being unset or being empty is the same, use default value, I changed it to `GetEnv`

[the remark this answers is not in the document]
shajmakh: sounds a good idea indeed. On the other hand, if the path did change, and test executable was compiled, it will be found only in the default path even though the env var LATENCY_TOOLS_PATH is set.  
A case that I can think of where this can be easily broken is running 5_latency_testing after specifying a new non-default path under LATENCY_TOOLS_PATH. this will result to skip the tests because the test executable was not found in the new path, but in fact in the default one because this https://github.com/openshift/cluster-node-tuning-operator/blob/master/hack/build-latency-test-bin.sh#L10 is kept as is. so test executable is generated but tests are skipped because they search for it in the wrong directory. Did you have a chance to test this?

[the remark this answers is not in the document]
jlojosnegros: Yes, I totally agree. 
If you run the test suite with a wrong environment it would fail, but I think that is applicable to any other env var.

Error logging can be improved so the problem, in case of happening, will be easier to debug.
I could export the `LATENCY_TOOLS_PATH` in `build-latency-test.sh` after the build but I am not sure if that would help at all.

One thing is certain ... if `LATENCY_TOOLS_PATH` is not set or empty everything should work as expected.

Tal-or: make sense

ffromani:
/retest-required

but let's deep dive if it fails again

PR #661, opened 2023-05-23
Latency

IlyaTyomkin:
oslat latency test fails when one of Env. variable LATENCY_TEST_DELAY was set with its value more than 120 sec. This PR automate and verify the state when on the test code the value of LATENCY_TEST_DELAY = 200 sec for all latency tests.It tests all latency tools with this new value.


IlyaTyomkin:
PR was closed due to the wrong rebase. The link to new one:
https://github.com/openshift/cluster-node-tuning-operator/pull/662

PR #665, opened 2023-05-29
Use RHEL9 as a base

jmencak:
Switch to RHEL9 as NTO base


jmencak:
Let's hear a better
/joke
than the Popsicle one.

/retest

PR #676, opened 2023-06-02
[release-4.11] OCPBUGS-14501: Fix updating numa core siblings map in GetCpuSiblings f…

mrniranjan:
…unction

Fix updating numa core siblings map in GetCpuSiblings function move the delete operation outside of for loop.

Also minor fixes to offline tests to check if the worker node has 20 cores earlier in the test function


MarSik (approved):
/lgtm
/approved
/backport-risk-assessed

PR #677, opened 2023-06-05
[test] [e2e] Check variables before reading them in the cleanup closures.

jlojosnegros:
When a failure occurs in a `BeforeEach`, `JustBeforeEach`, or `It` closure Ginkgo halts execution of the current spec and cleans up by invoking any registered `AfterEach` or `JustAfterEach` closures (and any registered `DeferCleanup` closures if applicable). 

We need to check variables before reading them in the cleanup closures.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -762,6 +762,7 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 			var globallyDisableIrqLoadBalancing bool
 
 			JustBeforeEach(func() {
+				//REVIEW - This is overwritting the "profile" var, is that ok?
 				key := types.NamespacedName{
 					Name:      profile.Name,
 					Namespace: profile.Namespace,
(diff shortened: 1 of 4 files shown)


ffromani (commented):
/lgtm

not addressing the root cause (profile being `nil` in the first place) but still worthy improvement

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -762,6 +762,7 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 			var globallyDisableIrqLoadBalancing bool
 
 			JustBeforeEach(func() {
+				//REVIEW - This is overwritting the "profile" var, is that ok?
```
jlojosnegros, line 765: This is overwriting `profile` variable which is defined in a higher context.
Not sure if that is done intentionally or it is just reusing a common name.

yanirq: IIRC  in 0_config we get either the initial profile created for all the tests (or get an existing one in the system)
This profile acts as the anchor for all tests so manipulations are done to it and in the end they get reverted.
This method is high maintenance as this PR demonstrates  and was embedded from the start.
The benefits might be always having the same data for most of the test that will not break them such as: isolated and reserved CPUs, hugepages, node selector , etc..

PR #680, opened 2023-06-08
Add golangci-lint

danielmellado:
This commit adds a new golangci-lint target to the makefile and sets up
a common config along the lines of other openshift operators.

- Address misspell violations
- Address staticheck violations

Signed-off-by: Daniel Mellado <dmellado@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ ASSETS=$(shell find assets -name \*.yaml)
 GO=GOOS=linux GO111MODULE=on GOFLAGS=-mod=vendor go
 GO_BUILD_RECIPE=$(GO) build -o $(OUT_DIR)/$(PACKAGE_BIN) -ldflags '-X $(PACKAGE)/version.Version=$(REV)' $(PACKAGE_MAIN)
 GOFMT_CHECK=$(shell find . -not \( \( -wholename './.*' -o -wholename '*/vendor/*' \) -prune \) -name '*.go' | sort -u | xargs gofmt -s -l)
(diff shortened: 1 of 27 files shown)


jlojosnegros:
/lgtm

cc @yanirq @MarSik

Makefile
```diff
@@ -122,6 +124,10 @@ else
 	@exit 1
 endif
 
+golangci-lint:
```
yanirq, line 127: I'm not sure we can directly use downloaded content here ,
@MarSik  WDYT?

jlojosnegros: we are doing something similar with ginkgo for other tests ...

jlojosnegros: @yanirq @MarSik 
kindly reminder

jmencak: Downloading is fine as long as the downloading does not happen in the (ART) environment production (nightly/release) images are built.

jmencak (approved):
@danielmellado Thank you for the PR, I like it.  This should make the code cleaner.  It would be great if we didn't unnecessarily download the linter binary every time, e.g. we probably want a new download rule with the binary as a target in the makefile.  If you could make this change, it would be great.  Adding
/hold
because of that, otherwise
/lgtm
/approve

ffromani:
I think to fix the failing lanes a rebase could be sufficient

jlojosnegros:
@danielmellado I think this can be closed as #793 has been already merged.
Thanks for your help :)

PR #682, opened 2023-06-13
OCPBUGS-14934: consistent use of ginkgo flags in Makefile

mrniranjan:
consistent use of ginkgo flags in Makefile

Also pass --junit-report=report.xml with other ginkgo options  instead of passing with -r

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -208,7 +208,7 @@ pao-functests-updating-profile: cluster-label-worker-cnf pao-functests-update-on
 pao-functests-update-only:
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/2_performance_update test/e2e/performanceprofile/functests/3_performance_status test/e2e/performanceprofile/functests/7_performance_kubelet_node test/e2e/performanceprofile/functests/8_reboot" -p "--v -r --fail-fast --flake-attempts=2 --timeout=5h --junit-report=report.xml" -m "Running Functional Tests"
+	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/2_performance_update test/e2e/performanceprofile/functests/3_performance_status test/e2e/performanceprofile/functests/7_performance_kubelet_node test/e2e/performanceprofile/functests/8_reboot" -p "-v -r --fail-fast --flake-attempts=2 --timeout=5h --junit-report=report.xml" -m "Running Functional Tests"
 
 .PHONY: pao-functests-performance-workloadhints
 pao-functests-performance-workloadhints: cluster-label-worker-cnf pao-functests-performance-workloadhints-only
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -208,7 +208,7 @@ pao-functests-updating-profile: cluster-label-worker-cnf pao-functests-update-on
 pao-functests-update-only:
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/2_performance_update test/e2e/performanceprofile/functests/3_performance_status test/e2e/performanceprofile/functests/7_performance_kubelet_node test/e2e/performanceprofile/functests/8_reboot" -p "-v -r -timeout=5h --fail-fast --flake-attempts=2" -m "Running Functional Tests" -r "--junit-report=report.xml"
+	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/2_performance_update test/e2e/performanceprofile/functests/3_performance_status test/e2e/performanceprofile/functests/7_performance_kubelet_node test/e2e/performanceprofile/functests/8_reboot" -p "--v -r -timeout=5h --fail-fast --flake-attempts=2 --junit-report=report.xml" -m "Running Functional Tests"
```
jmencak, line 211: Why the `s/-v/--v/` change?

jmencak (commented):
Can we have a description of what this PR does?

Makefile
mrniranjan: fixed in the latest commit

Makefile
```diff
@@ -217,13 +217,13 @@ pao-functests-performance-workloadhints: cluster-label-worker-cnf pao-functests-
 pao-functests-performance-workloadhints-only:
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/8_performance_workloadhints" -p "-v -r -timeout=5h --fail-fast --flake-attempts=2" -m "Running Functional WorkloadHints Tests" -r "--junit-report=report.xml"
+	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/8_performance_workloadhints" -p "--v -r --fail-fast --flake-attempts=2 -timeout=5h --junit-report=report.xml" -m "Running Functional WorkloadHints Tests"
```
ffromani: what is the semantics of "-r"?

ffromani:
/approve

I like the general direction

Makefile
```diff
@@ -208,7 +208,7 @@ pao-functests-updating-profile: cluster-label-worker-cnf pao-functests-update-on
 pao-functests-update-only:
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/2_performance_update test/e2e/performanceprofile/functests/3_performance_status test/e2e/performanceprofile/functests/7_performance_kubelet_node test/e2e/performanceprofile/functests/8_reboot" -p "--v -r -timeout=5h --fail-fast --flake-attempts=2 --junit-report=report.xml" -m "Running Functional Tests"
+	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/2_performance_update test/e2e/performanceprofile/functests/3_performance_status test/e2e/performanceprofile/functests/7_performance_kubelet_node test/e2e/performanceprofile/functests/8_reboot" -p "--v -r --fail-fast --flake-attempts=2 --timeout=5h --junit-report=report.xml" -m "Running Functional Tests"
```
jlojosnegros: Why the double `-` in timeout?
and the change in order?

Makefile
jlojosnegros: Report parameters where positioned after the required test suites in the original scripts. 
When I wrote them with the rest of parameters it did not work for me, so I added a different parameter in the script to be able to write them after the suites as in the original.

Makefile
mrniranjan: i am just going by ginkgo run --options, 

```
Debugging Tests
In addition to these flags, Ginkgo supports a few debugging environment
variables.  To change the parallel server protocol set GINKGO_PARALLEL_PROTOCOL
to HTTP.  To avoid pruning callstacks set GINKGO_PRUNE_STACK to FALSE.
  --dry-run
    If set, ginkgo will walk the test hierarchy without actually running
    anything.  Best paired with -v.
  --progress
    If set, ginkgo will emit progress information as each spec runs to the
    GinkgoWriter.
  --timeout [duration] (default: 1h)
    Test suite fails if it does not complete within the specified timeout.
  --output-interceptor-mode [dup, swap, or none]
    If set, ginkgo will use the specified output interception strategy when
    running in parallel.  Defaults to dup on unix and swap on windows.
  --until-it-fails
    If set, ginkgo will keep rerunning test suites until a failure occurs.
  --repeat [n] (default: 0 - i.e. no repetition, run only once)
    The number of times to re-run a test-suite.  Useful for debugging flaky
    tests.  If set to N the suite will be run N+1 times and will be required to
    pass each time.
```

PR #692, opened 2023-06-21
OCPBUGS-14946: Switch to rslave/HostToContainer volume mount propagation

jmencak:
Mount propagation of a volume is controlled by the `mountPropagation` field in `Container.volumeMounts`.  By default, `None/rprivate` mount propagation -- see `mount(8)`, is used when sharing volumes mounted by a container.  `None/rprivate` mount propagation seems to be causing issues in certain cases -- see [OCPBUGS-14946](https://issues.redhat.com/browse/OCPBUGS-14946)  Switch to `HostToContainer/rslave` volume mount propagation, which also aligns better with what we would like to see in the TuneD pods.

Resolves: OCPBUGS-14946

### DIFF
--- a/assets/tuned/manifests/ds-tuned.yaml
+++ b/assets/tuned/manifests/ds-tuned.yaml
@@ -37,27 +37,37 @@ spec:
         volumeMounts:
         - mountPath: /var/lib/tuned/profiles-data
           name: var-lib-tuned-profiles-data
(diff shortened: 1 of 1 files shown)


jmencak:
Only if you have a little time:
/cc @dagrayvid 
all tests passed and the issue seems to be fixed by this.

PR #693, opened 2023-06-22
e2e: memorymanager fix: check the hugepage size condition

mrniranjan:
Minor if condition fix, earlier it was checking the string "2mi", but the test cases are setting hugepages requried in field called hpgSize, which was not used . hence pod was not getting the right hugepages mounted.


ffromani (commented):
/lgtm
/approve

PR #697, opened 2023-06-27
e2e: reboot: add kubelet restart test

ffromani:
Add a test to check that a workload consuming devices is not disrupted by a kubelet restart.
The very same instance should be preserved across restarts.


ffromani:
/hold

need to run the test and finetune it

ffromani:
/cc @swatisehgal 

reviewers kindly be aware that we are intentionally duplicating the test added to the `e2e_node` u/s suite. One of the main reasons is this test can be run against any cluster (whose node OS uses systemd)

swatisehgal (approved):
/lgtm

Straightforward test, we are replicating pretty much most of the steps that we had for node reboot scenario except the use of a standalone pod which makes sense as we need to check for the same pod instance.

ffromani:
/hold

complete and ready for review now

swatisehgal (approved):
/lgtm

test/e2e/performanceprofile/functests/9_reboot/devices.go
```diff
@@ -171,6 +173,68 @@ var _ = Describe("[disruptive][node][kubelet][devicemanager] Device management t
...
+	It("Verify that pods requesting devices are not disrupted by a kubelet restart", func() {
+		namespace := testutils.NamespaceTesting
+
+		// refresh the targetNode object
+		node, err := testnodes.GetByName(targetNode)
+		Expect(err).ToNot(HaveOccurred(), "error getting the target node %q", targetNode)
+
+		// phase1: complete the node praparation: make sure we have enough devices. Short timeout, we should be idle now.
+		allocatableDevices := waitForNodeToReportResourcesOrFail("pre reboot", targetNode, sriovDeviceResourceName, 2*time.Minute, 2*time.Second)
+		Expect(allocatableDevices).To(BeNumerically(">=", minimumAllocatableDevices), "device %q has too low amount available - testing scenario unreliable", sriovDeviceResourceName, allocatableDevices)
```
shajmakh: is it recommended here too that we have this amount of minimal devices? if I get it right, we don't want care about that the device plugin takes time to get back up, I actually think it's easier (less exec time) to get enough with only the workload limits. Is my understanding correct?

ffromani: good point. We can relax/remove this constraint for this test.

test/e2e/performanceprofile/functests/9_reboot/devices.go
```diff
@@ -171,6 +173,68 @@ var _ = Describe("[disruptive][node][kubelet][devicemanager] Device management t
...
+		// short timeout: we are on idle cluster
+		updatedPod, err := testpods.WaitForPredicate(wlPod, 3*time.Minute, func(pod *corev1.Pod) (bool, error) {
+			return isPodReady(*pod), nil
+		})
+		Expect(err).ToNot(HaveOccurred(), "error waiting for the workload pod to be ready - pre restart")
+		podUID := updatedPod.UID // shortcut to the reference
+		testlog.Infof("pod %q %s/%s ready", podUID, updatedPod.Namespace, updatedPod.Name)
+
+		// phase3: the kubelet restart
+		runCommandOnNodeThroughMCD(node, "reboot", kubeletRestartCommandMCD)
```
shajmakh, line 202: nit: kubelet restart description or similar to clarify the performed action

PR #700, opened 2023-06-29
OCPVE-382: fix: add default state for crio config

eggfoobar:
The original implementation of workload partitioning tried to leverage default behavior for CRIO to allow full use of CPU Sets when no Performance Profile is supplied by the user while still being a CPU partitioned cluster. This works fine for CPU affinity however because we don't supply a config and allow the default behavior to kick in, CRIO does not alter the CPU share and gives all pods 2 CPU Share value. 

We need to supply a config for CRIO with an empty string for CPU Set to support both CPU share and CPU affinity behavior when NO performance profile is supplied, so that the `resource.requests` which get altered to CPU Share, are correctly being applied in a default state.

/assign @haircommander

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -546,24 +546,18 @@ func BootstrapWorkloadPinningMC(role string, pinningMode *apiconfigv1.CPUPartiti
 	}
 
 	mode := 420
-	/*
-		Contents for emptyKubeletConfig:
-		{
-			"management": {
-				"cpuset": ""
-			}
-		}
+	emptySet := performancev2.CPUSet("")
+	emptySetCPU := performancev2.CPU{Reserved: &emptySet}
 
-		Contents for emptyCrioConfig:
-		[crio.runtime.workloads.management]
-		activation_annotation = "target.workload.openshift.io/management"
-		annotation_prefix = "resources.workload.openshift.io"
-		resources = { "cpushares" = 0, "cpuset" = "" }
-	*/
-	emptyKubeletConfig := "ewogICJtYW5hZ2VtZW50IjogewogICAgImNwdXNldCI6ICIiCiAgfQp9Cg=="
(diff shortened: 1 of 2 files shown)


haircommander:
LGTM from CRI-O side

eggfoobar:
Validated behavior is correct on release build with this PR.

Steps:
- validated the new MC contains both config files for kubelet and crio
- spun up debug pod in node
- checked process with `taskset` to make sure CPU affinity behavior was not effected by change
- checked with `crictl inspect` to validate `cpu shares` and `cpu set` were correctly being applied during default behavior

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -540,7 +546,28 @@ func BootstrapWorkloadPinningMC(role string, pinningMode *apiconfigv1.CPUPartiti
...
+		}
+
+		Contents for emptyCrioConfig:
+		[crio.runtime.workloads.management]
+		activation_annotation = "target.workload.openshift.io/management"
+		annotation_prefix = "resources.workload.openshift.io"
+		resources = { "cpushares" = 0, "cpuset" = "" }
+	*/
+	emptyKubeletConfig := "ewogICJtYW5hZ2VtZW50IjogewogICAgImNwdXNldCI6ICIiCiAgfQp9Cg=="
+	emptyCrioConfig := "W2NyaW8ucnVudGltZS53b3JrbG9hZHMubWFuYWdlbWVudF0KYWN0aXZhdGlvbl9hbm5vdGF0aW9uID0gInRhcmdldC53b3JrbG9hZC5vcGVuc2hpZnQuaW8vbWFuYWdlbWVudCIKYW5ub3RhdGlvbl9wcmVmaXggPSAicmVzb3VyY2VzLndvcmtsb2FkLm9wZW5zaGlmdC5pbyIKcmVzb3VyY2VzID0geyAiY3B1c2hhcmVzIiA9IDAsICJjcHVzZXQiID0gIiIgfQo="
```
ffromani, line 564: I'm fine with the concept of this change but hardcoding the base64 encoded value seems unnecessarily obfuscated. I'd hardcoded the plaintext and base64 programmatically.
cc @MarSik @Tal-or

ffromani:
> Validated behavior is correct on release build with this PR.
> 
> Steps:
> 
>     * validated the new MC contains both config files for kubelet and crio
> 
>     * spun up debug pod in node
> 
>     * checked process with `taskset` to make sure CPU affinity behavior was not effected by change
> 
>     * checked with `crictl inspect` to validate `cpu shares` and `cpu set` were correctly being applied during default behavior

thanks for the validation, appreciated. Do we have (at least one) functests covering this flow? If not we should probably add them.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
eggfoobar: Sure thing, updated to use our current flow for empty states

eggfoobar:
I don't think there are func tests in this repo for empty state. I'll be adding CPU set and CPU share checks on origin to accompany this API feature going GA. If it's okay with you @ffromani I can add functests here in a different PR as part of that effort, what do you think?

Currently tracking that here https://issues.redhat.com/browse/OCPVE-318

MarSik:
Yeah, our test suite does not have a lane for checking the install bootstrapping of a WP SNO. I am ok with relying on the check of generated manifests.

/lgtm

PR #702, opened 2023-06-30
e2e: perf-prof: disable truncating gomega output

shajmakh:
When the tests fail, we want to have a clear full log of the output of the executed latency tool. The default format.MaxLength is 4000 but it isn't always sufficient to display the full string presentation especially when there is a failure.

Disable this limitation by setting the format.MaxLength to 0, to help with better troubleshooting of the failure.

For more info: https://onsi.github.io/gomega/#adjusting-output

### DIFF
--- a/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
+++ b/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
@@ -11,6 +11,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


shajmakh:
/hold 
for full testing

ffromani:
/approve

this is a necessary evil. The proper fix would be fix the output in all the relevant (and future) tests

ffromani:

this PR cannot cause this failure and is not related to hypershift in any way

shajmakh:
/hold cancel
Verified. The log is with a complete failure message now.

PR #703, opened 2023-06-30
e2e: RPS: fix expected cpu set

ffromani:
Because of a logic oversight, the expected cpu set was overwritten with the value expected for physical devices after the first loop iteration. This broke the test when running on clusters with more than a single worker node.


zeeke:
LGTM

Tal-or:
/lgtm
Good catch!

ffromani:

none of the failures are obviously related to this lane, which seems to be failing also with other (still unrelated) PRs

PR #710, opened 2023-07-05
OCPBUGS-15800:e2e: latency testing: increase the expected threshold and fix gomega truncating output

shajmakh:
lately some of the latency tests were failing and it turned out that the failure is due to that the maximum latency found was greater than anticipated. The tests in this suite do not care about the measured latency because we do not care to tune the systems in the first place. Its only goal is to run the tests of the latency tools with different values of environment variables and validate that the tool actually runs with these parameters.

Increase the maximum latency threshold so that the test doesn't fail in case of a high latency result.


shajmakh:
4.10 porting: https://github.com/openshift-kni/performance-addon-operators/pull/964

shajmakh:
/retitle OCPBUGS-15800:e2e: latency testing: increase the expected threshold and fix gomega truncating output

PR #714, opened 2023-07-07
 Build gather-sysinfo

MarSik:
The gather-sysinfo binary is used by must-gather to collect
structured hardware information data about an OCP node.

This data is then also used by the Performance Profile Creator
also present in this repository.

Must-gather also needs the lspci binary from the pciutils package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ clone-tuned:
 	  cd $(TUNED_DIR) && git checkout $(TUNED_COMMIT) && cd .. && \
 	  rm -rf $(TUNED_DIR)/.git)
 
-build: $(BINDATA) pkg/generated build-performance-profile-creator
+build: $(BINDATA) pkg/generated build-performance-profile-creator build-gather-sysinfo
 	$(GO_BUILD_RECIPE)
 	ln -sf $(PACKAGE_BIN) $(OUT_DIR)/openshift-tuned
 
@@ -100,7 +100,7 @@ test-e2e:
 	done
 
 .PHONY: test-e2e-local
-test-e2e-local: $(BINDATA) performance-profile-creator-tests
+test-e2e-local: $(BINDATA) performance-profile-creator-tests gather-sysinfo-tests
 	$(GO_BUILD_RECIPE)
 	for d in performanceprofile/functests-render-command/1_render_command; do \
 	  $(GO) test -v -timeout 40m ./test/e2e/$$d -ginkgo.v -ginkgo.no-color -ginkgo.fail-fast || exit; \
@@ -241,6 +241,19 @@ build-performance-profile-creator:
 performance-profile-creator-tests: build-performance-profile-creator
 	@echo "Running Performance Profile Creator Tests"
 	hack/run-test.sh -t "test/e2e/performanceprofile/functests-performance-profile-creator" -p "--v -r --fail-fast --flake-attempts=2" -m "Running Functional Tests" -r "--junit-report=/tmp/artifacts"
+
+# Gather sysinfo binary for use in must-gather
+.PHONY: build-gather-sysinfo
+build-gather-sysinfo:
+	@echo "Building gather-sysinfo"
+	LDFLAGS="-s -w -X ${PACKAGE}/cmd/gather-sysinfo/version.Version=${REV} "; \
+	$(GO) build -v $(LDFLAGS) -o $(OUT_DIR)/gather-sysinfo ./cmd/gather-sysinfo
+
+.PHONY: gather-sysinfo-tests
+gather-sysinfo-tests: build-gather-sysinfo
+	@echo "Running gather-sysinfo Tests"
+	$(GO) test -v ./cmd/gather-sysinfo
+
 
 .PHONY: render-sync
 render-sync: build
--- a/cmd/gather-sysinfo/gather-sysinfo.go
+++ b/cmd/gather-sysinfo/gather-sysinfo.go
@@ -1,0 +1,177 @@
+package main
+
+import (
+	"fmt"
+	"github.com/jaypipes/ghw/pkg/snapshot"
+	"github.com/openshift-kni/debug-tools/pkg/knit/cmd"
+	"github.com/openshift-kni/debug-tools/pkg/knit/cmd/k8s"
+	"github.com/openshift-kni/debug-tools/pkg/machineinformer"
+	"github.com/spf13/cobra"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+)
+
+const machineInfoFilePath string = "machineinfo.json"
+
+type snapshotOptions struct {
+	dumpList bool
+	output   string
+	rootDir  string
+}
+
+func main() {
+	root := cmd.NewRootCommand(newSnapshotCommand,
+		k8s.NewPodResourcesCommand,
+		k8s.NewPodInfoCommand,
+	)
+
+	if err := root.Execute(); err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
+
+func newSnapshotCommand(knitOpts *cmd.KnitOptions) *cobra.Command {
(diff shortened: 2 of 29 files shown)


cmd/gather-sysinfo/output
jlojosnegros, line 1: I think this file has been committed by mistake

cmd/gather-sysinfo/gather-sysinfo_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
```
ffromani, line 8: we should have moved away from ginkgo for these tests long ago

cmd/gather-sysinfo/testSnapshot.tgz
jlojosnegros, line 1: this file also

cmd/gather-sysinfo/gather-sysinfo_test.go
```diff
@@ -0,0 +1,93 @@
+package main
+
+import (
+	"io/ioutil"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
ffromani, line 7: ditto

ffromani (commented):
importing `gather-sysinfo` into the tree may and should be the chance for a good and long-needed cleanup, but I guess we don't have time/capacity for that, do we?

cmd/gather-sysinfo/testSnapshot.tgz
MarSik: Part of the tests afaik.

cmd/gather-sysinfo/output
MarSik: It is part of the test.

cmd/gather-sysinfo/gather-sysinfo_suite_test.go
MarSik: updated

cmd/gather-sysinfo/gather-sysinfo_test.go
MarSik: updated

cmd/gather-sysinfo/testSnapshot.tgz
MarSik: You were right, removed.

cmd/gather-sysinfo/output
MarSik: Removed.

go.mod
```diff
@@ -160,6 +162,7 @@ require (
 
 // Pinned to kubernetes-1.26.2
 replace (
+	github.com/google/cadvisor => github.com/google/cadvisor v0.43.0
```
ffromani, line 165: we previously consumed 0.46.0, I'd fix the code to use the newer version (should any change be needed) rather than jumping back

ffromani:
thanks for the updates! next time I recommend considering a verbatim import + fixing in the subsequent commits in order to preserve a good audit trail. In some cases this is unpractical but in this case it could have been.

go.mod
MarSik: I just did, this was the reason I had to fix debug tools first.

cmd/gather-sysinfo/gather-sysinfo.go
```diff
@@ -0,0 +1,183 @@
...
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/jaypipes/ghw/pkg/snapshot"
+	"github.com/openshift-kni/debug-tools/pkg/knit/cmd"
+	"github.com/spf13/cobra"
+)
```
ffromani, line 15: can we reorder imports?

cmd/gather-sysinfo/gather-sysinfo.go
```diff
@@ -0,0 +1,183 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return makeSnapshot(cmd, knitOpts, opts, args)
+		},
+		Args: cobra.NoArgs,
+	}
+	snap.Flags().StringVar(&opts.rootDir, "root", "", "pseudofs root - use this if running inside a container")
+	snap.Flags().StringVar(&opts.output, "output", "", "path to clone system information into")
+	snap.Flags().BoolVar(&opts.dumpList, "dump", false, "just dump the glob list of expected content and exit")
+	// use this to debug container behaviour
+	snap.Flags().IntVar(&opts.sleepTime, "sleep", 0, "amount of seconds to sleep once done, before exit")
```
ffromani, line 49: this raised concerns in review and we can probably replace anyway with something better. I'd remove it.

cmd/gather-sysinfo/gather-sysinfo_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"io/ioutil"
+	"k8s.io/utils/strings/slices"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/openshift-kni/debug-tools/pkg/knit/cmd"
+	"github.com/spf13/cobra"
```
ffromani, line 11: can we reorder imports?

cmd/gather-sysinfo/gather-sysinfo_test.go
```diff
@@ -0,0 +1,113 @@
...
+	err := makeSnapshot(cmd, knitOpts, opts, args)
+	if err == nil {
+		t.Errorf("Failure was expected when running without --output argument.")
+	}
+	t.Log(err)
+
+	opts.output = "testSnapshot.tgz"
+
+	// Delete old snapshot if it exists
+	_ = os.Remove(opts.output) // ignore error
```
ffromani, line 102: we leaked this in the past, should probably be moved into a `defer` block

MarSik: Isn't the leak intentional? To allow debugging of failed tests?

ffromani: right but if so (sloppy me) it should be controllable from outside (flags, env vars...)

Makefile
```diff
@@ -242,6 +242,19 @@ performance-profile-creator-tests: build-performance-profile-creator
 	@echo "Running Performance Profile Creator Tests"
 	hack/run-test.sh -t "test/e2e/performanceprofile/functests-performance-profile-creator" -p "--v -r --fail-fast --flake-attempts=2" -m "Running Functional Tests" -r "--junit-report=/tmp/artifacts"
 
+# Gather sysinfo binary for use in must-gather
+.PHONY: build-gather-sysinfo
+build-gather-sysinfo:
+	@echo "Building gather-sysinfo"
+	LDFLAGS="-s -w -X ${PACKAGE}/cmd/gather-sysinfo/version.Version=${REV} "; \
+	$(GO) build  -v $(LDFLAGS) -o $(OUT_DIR)/gather-sysinfo ./cmd/gather-sysinfo
```
jmencak, line 250: Nit: formatting double spaces.

Makefile
```diff
@@ -242,6 +242,19 @@ performance-profile-creator-tests: build-performance-profile-creator
...
+
+.PHONY: gather-sysinfo-tests
+gather-sysinfo-tests: build-gather-sysinfo
+	@echo "Running gather-sysinfo Tests"
+	$(GO) test  -v ./cmd/gather-sysinfo
```
jmencak, line 255: Nit: formatting double spaces.  I guess copied from `build-performance-profile-creator`

jmencak (commented):
The changes overall seem fine, but it is not clear to me how the collected information gets into must-gather.  The commit message only states:

    The gather-sysinfo binary is used by must-gather to collect
    structured hardware information data about an OCP node.

Can you expand on that a bit how must-gather gets that information?  Perhaps even in the commit message?  Saw some nits so I pointed them out.

Makefile
MarSik: Yeah, I assumed that was intentional :)

PR #718, opened 2023-07-14
change min_granularity_ns value to 3000000

mrniranjan:
With recent changes to kernel the value of
min_granularity_ns is changed from 10000000 to 3000000


mrniranjan:
closing this PR as again the kernel is showing the right value that we set through performance Profile.

PR #719, opened 2023-07-18
e2e: add expected max latancy to hwlatdetec test & rename constant

shajmakh:
The test started to fail frequently in CI. The reason for the failure is that there were measured latency samples that exceeds the maximal default threshold of hwlatdetec. Since we have enabled new CI machines, it seems that they are not well tuned hence causing in high latency.

To fix the test, just add a high threshold. Note that in these tests we do not care about latency measurments but about enabling the latency tools tests with differnt variables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,7 +225,7 @@ pao-functests-performance-workloadhints-only:
 pao-functests-latency-testing: dist-latency-tests
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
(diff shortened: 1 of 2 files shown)


shajmakh:
> Is the 20000 threshold high enough?
no, the test was getting latency result greater than 20000, so I set it to 900000

shajmakh:
failure in hypershift job, the changes are not related so:
/retest

MarSik (approved):

ffromani (commented):
/lgtm

PR #720, opened 2023-07-19
pao: status: do not take address of loop var

Tal-or:
Taking the address of the range loop variable is bad, because go reuses the loop variable address.

more on that here: https://www.evanjones.ca/go-gotcha-loop-variables.html


ffromani (commented):
/approve
/lgtm

good catch!

Tal-or:
/retest 
infra issue

yanirq:
nice,  lets cherry pick this (with a bug) if needed

PR #721, opened 2023-07-19
OCPBUGS-15044: e2e:irqbalance: wait for tuned profile to be ready

Tal-or:
We need to wait for tuned to apply the recent changes, before we check the irqbalance settings.

We poll the tuned profile and check the applied status. we failed the test if the status not consolidate to applied within the timeout value.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -18,6 +18,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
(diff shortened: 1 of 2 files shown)


jmencak:
Thank you for the PR, Talor.  Apart from the correct case ("Tuned Profile")  this looks good to me.  But the casing is a nit, so
/lgtm

Tal-or:
> Thank you for the PR, Talor. Apart from the correct case ("Tuned Profile") this looks good to me. But the casing is a nit, so /lgtm

Thanks @jmencak, what case? I'm not sure I understand

jmencak:
> > Thank you for the PR, Talor. Apart from the correct case ("Tuned Profile") this looks good to me. But the casing is a nit, so /lgtm
> 
> Thanks @jmencak, what case? I'm not sure I understand

CaSe SeNsItIvItY.

Tal-or:
> > > Thank you for the PR, Talor. Apart from the correct case ("Tuned Profile") this looks good to me. But the casing is a nit, so /lgtm
> > 
> > 
> > Thanks @jmencak, what case? I'm not sure I understand
> 
> CaSe SeNsItIvItY.

L97 and L103 in the error messages?

jmencak:
Thank you for the changes.
/lgtm

PR #724, opened 2023-07-21
Improve render error handling

jlojosnegros:
When looking for a MachineConfigPool match for each PerformanceProfile we can found no match at all.
We should check for this case before using the MCP.

This do not detect the root cause of the problem, an invalid PerformanceProfile. To tackle that problem we would need something more deep like a validator.


ffromani:
/approve
/lgtm

looks a nice improvement anyway to me

PR #727, opened 2023-07-24
OCPBUGS-15044:e2e:irqbalance: applied condition exists

Tal-or:
There was a subtle bug in the previous implementaion, where if the `tunedv1.TunedProfileApplied` does not exits in the slice we still return `true`

This behavior is logically wrong.
We should return `true` only if the condition
was found and it's condition status is positive.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -91,17 +91,21 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 				}
 
 				for _, node := range workerRTNodes {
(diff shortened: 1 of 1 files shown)


Tal-or:
/hold
Depends on: https://github.com/openshift/cluster-node-tuning-operator/pull/729

ffromani (commented):
another good catch!

/approve
/lgtm

Tal-or:
> another good catch!
> 
> /approve /lgtm

A self-catch doesn't count ;)

Tal-or:
/hold cancel
/test e2e-gcp-pao
 https://github.com/openshift/cluster-node-tuning-operator/pull/729 will be handled separately

PR #729, opened 2023-07-24
OCPBUGS-7980:e2e:ht-aware: exec on the correct worker node

Tal-or:
We  retrieve the worker node where the pod
located, before executing command against it.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -665,7 +665,7 @@ func checkPodHTSiblings(testpod *corev1.Pod) bool {
 		fmt.Sprintf("/bin/crictl inspect %s | /bin/jq -r '.info.runtimeSpec.linux.resources.cpu.cpus'", containerID),
 	}
 	output, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
-	Expect(err).ToNot(HaveOccurred(), "Unable to crictl inspect containerID %s", "containerID")
+	Expect(err).ToNot(HaveOccurred(), "Unable to crictl inspect containerID %q", containerID)
 
 	podcpus, err := cpuset.Parse(strings.Trim(fmt.Sprint(output), "\n"))
 	Expect(err).ToNot(
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -670,7 +670,7 @@ func checkPodHTSiblings(testpod *corev1.Pod) bool {
 	podcpus, err := cpuset.Parse(strings.Trim(fmt.Sprint(output), "\n"))
 	Expect(err).ToNot(
 		HaveOccurred(), "Unable to cpuset.Parse pod allocated cpu set from output %s", fmt.Sprint(output))
-	By(fmt.Sprintf("Test pod CPU list: %s", podcpus.String()))
+	testlog.Infof("Test pod CPU list: %s", podcpus.String())
 
 	// aggregate cpu sibling paris from the host based on the cpus allocated to the pod
 	By("Get host cpu siblings for pod cpuset")
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -718,6 +718,7 @@ func startHTtestPod(cpuCount int) *corev1.Pod {
 	testpod.Namespace = testutils.NamespaceTesting
 
 	By(fmt.Sprintf("Creating test pod with %d cpus", cpuCount))
+	testlog.Info(pods.DumpResourceRequirements(testpod))
 	err := testclient.Client.Create(context.TODO(), testpod)
 	Expect(err).ToNot(HaveOccurred())
 	testpod, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -664,12 +664,15 @@ func checkPodHTSiblings(testpod *corev1.Pod) bool {
 		"-c",
 		fmt.Sprintf("/bin/crictl inspect %s | /bin/jq -r '.info.runtimeSpec.linux.resources.cpu.cpus'", containerID),
 	}
-	output, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+	node, err := nodes.GetByName(testpod.Spec.NodeName)
+	Expect(err).ToNot(HaveOccurred(), "failed to get node %q", testpod.Spec.NodeName)
+	Expect(testpod.Spec.NodeName).ToNot(BeEmpty(), "testpod %s/%s still pending - no nodeName set", testpod.Namespace, testpod.Name)
+	output, err := nodes.ExecCommandOnNode(cmd, node)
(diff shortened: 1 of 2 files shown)


test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -665,7 +665,7 @@ func checkPodHTSiblings(testpod *corev1.Pod) bool {
 		fmt.Sprintf("/bin/crictl inspect %s | /bin/jq -r '.info.runtimeSpec.linux.resources.cpu.cpus'", containerID),
 	}
 	output, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
-	Expect(err).ToNot(HaveOccurred(), "Unable to crictl inspect containerID %s", "containerID")
+	Expect(err).ToNot(HaveOccurred(), "Unable to crictl inspect containerID %s", containerID)
```
ffromani, line 670: the change is correct. To make it even better use `%q` instead of `%s`

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -671,6 +671,7 @@ func checkPodHTSiblings(testpod *corev1.Pod) bool {
 	Expect(err).ToNot(
 		HaveOccurred(), "Unable to cpuset.Parse pod allocated cpu set from output %s", fmt.Sprint(output))
 	By(fmt.Sprintf("Test pod CPU list: %s", podcpus.String()))
+	testlog.Infof("Test pod CPU list: %s", podcpus.String())
```
ffromani, line 675: I agree. Do we want to add or replace? I'm slightly in favor of replacing. Your call.

ffromani: nevermind, you actually delete it in the next patch. Good!

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -664,13 +664,14 @@ func checkPodHTSiblings(testpod *corev1.Pod) bool {
 		"-c",
 		fmt.Sprintf("/bin/crictl inspect %s | /bin/jq -r '.info.runtimeSpec.linux.resources.cpu.cpus'", containerID),
 	}
-	output, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+	node, err := nodes.GetByName(testpod.Spec.NodeName)
```
ffromani, line 667: this is indeed more correct (and IMHO easier to follow too). I'm just wondering in which circumnstances the `workerRTNode` can be different from `testpod.Spec.NodeName`

ffromani (commented):
/approve
/lgtm
/hold

holding only to give you the chance to address the (minor, nonblocking) inline comments

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
Tal-or: We are not telling the `testpod` on which node we want it, so it can land on a node which is different than the `workerRTNode`.
But this PR actually exposed the real problem:
```
  [FAILED] failed to get node ""
  Unexpected error:
      <*errors.errorString | 0xc0008acb30>: {
          s: "failed to get node for the node \"\"",
      }
      failed to get node for the node ""
  occurred
  In [It] at: /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/cpu_management.go:668 @ 07/24/23 10:36:40.373```

seems like our pod is pending/failed to get scheduled for some reason because `testpod.Spec.NodeName` is empty

Comment (Tal-or) [reply test/e2e/performanceprofile/functests/1_performance/cpu_management.go:667]: My bad, we do set the nodeSelector so the original fix doesn't help.
So we know the pod is going to the correct node, but for some reason it failed to get admitted. i'll keep digging.

File: test/e2e/performanceprofile/functests/1_performance/cpu_management.go (hunk 1 above)
Comment (Tal-or) [reply test/e2e/performanceprofile/functests/1_performance/cpu_management.go]: Done, thanks

File: test/e2e/performanceprofile/functests/1_performance/cpu_management.go [hunk 4]
```diff
@@ -619,6 +619,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 
 				for _, cpuCount := range cpuCounts {
 					testpod = startHTtestPod(cpuCount)
+					key := client.ObjectKeyFromObject(testpod)
+					Expect(testclient.Client.Get(context.TODO(), key, testpod)).ToNot(HaveOccurred(), "failed to get pod: %q", key.String())
```
ffromani, line 623: This is not harmful but I wonder why is needed because `startHTtestPod` is supposed to return the updated pod, isn't it?

ffromani (commented):
a comment inside. What caused the admission failure?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
Tal-or: Here: https://github.com/openshift/cluster-node-tuning-operator/pull/729#discussion_r1272164393 we can see that although we were suppose to get the updated pod, we didn't (the `nodeName` was empty) The reason for that is still mistiry to me TBH

Tal-or:
/hold 
@ffromani https://github.com/openshift/cluster-node-tuning-operator/pull/729#discussion_r1273278298 in place. i'll have another cycle

Tal-or:
@ffromani I didn't manage to reproduce the flake in the last 4 runs. 
The change is good in general and did help to prevent some stale data https://github.com/openshift/cluster-node-tuning-operator/pull/729#discussion_r1273351746, but still doesn't explain the root cause.
I suggest to merge this one and track to see if the issue keep reproduce on the CI in the future. 
Meantime, we should not close the bug.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
ffromani: I find counter intuitive and unnecessary to had this code which should be redundant. Can be remove it? If this makes the failure resurface again we have narrowed down the search area

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
ffromani: no problem

ffromani (commented):
I agree with @Tal-or 's assessment, but I'd like to remove the allegedly redundant get before to merge this change. Other than this I'm fine with the proposed apprach

ffromani:
/approve
/lgtm

thanks for all your work on this fix

Tal-or:
depends on #744

ffromani:
the hypershift lane is failing on setup. This PR only changes test code so the hypershift failures are not really a blocker.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -719,8 +719,8 @@ func startHTtestPod(cpuCount int) *corev1.Pod {
 	By(fmt.Sprintf("Creating test pod with %d cpus", cpuCount))
 	err := testclient.Client.Create(context.TODO(), testpod)
 	Expect(err).ToNot(HaveOccurred())
-	testpod, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
 	testlog.Info(pods.DumpResourceRequirements(testpod))
+	testpod, err = pods.WaitForPhase(client.ObjectKeyFromObject(testpod), corev1.PodRunning, 10*time.Minute)
```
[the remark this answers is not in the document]
Tal-or: No reason anymore. let's keep the stricter

ffromani:
/retest

Error 504 is hardly caused by code changes

ffromani:
we had a clean run! but then it is retesting because another PR merged
image

PR #730, opened 2023-07-25
e2e:cpuloadbalance: deflake the test

Tal-or:
This PR contains bunch of improvement in order to deflake the test.
Most of the commits are cosmetics, but the verification of the sched
domains before the test begins commit.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -232,8 +232,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		var smtLevel int
 		var testpod *corev1.Pod
 
-		// getCPUswithLoadBalanceDisabled Return cpus which are not in any scheduling domain
-		getCPUswithLoadBalanceDisabled := func() (map[int]string, error) {
+		// getCPUsWithLoadBalanceDisabled Return cpus which are not in any scheduling domain
+		getCPUsWithLoadBalanceDisabled := func() (map[int]string, error) {
 			cmd := []string{"/bin/bash", "-c", "cat /proc/schedstat"}
 			schedstat, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
 			if err != nil {
@@ -269,7 +269,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			var err error
 			// It's possible that when this test runs the value of
 			// defaultCpuNotInSchedulingDomains is empty if no gu pods are running
-			defaultCpuNotInSchedulingDomains, err := getCPUswithLoadBalanceDisabled()
+			defaultCpuNotInSchedulingDomains, err := getCPUsWithLoadBalanceDisabled()
 			Expect(err).ToNot(HaveOccurred(), "Unable to fetch scheduling domains")
 			testlog.Infof("Default scheduling Domains are: %v", defaultCpuNotInSchedulingDomains)
 			annotations := map[string]string{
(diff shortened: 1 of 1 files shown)


Tal-or:
This PR suppose to deflake the test or make the failure consistent. 
either way, this is better than the flakiness we have right now.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -343,14 +350,14 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 					return true
 				} else {
 					for _, podcpu := range podCpus.ToSlice() {
-						for _, cpu := range cpusNotinSchedulingDomains {
-							if !strings.Contains(cpu, fmt.Sprint(podcpu)) {
```
Tal-or, line 347: This line checks if the `podcpu` is not substring of `cpu`.
podcpu is cpu id for example: `3` .
the `cpu` is single line output of the `/proc/schedstats` command, for example:
`[ cpu0 0 0 0 0 0 0 68186178574807 516377247436 331031573  cpu1 0 0 0 0 0 0 75970491002822 375072790117 330836684]`
The problem with this check is that if the string "3" not appears somewhere in this line, we return `true`, which is wrong, because although this line refer to `cpu0` and `cpu1` but the string `3` still appears.

Tal-or:
```
[performance]Hugepages [rfe_id:27354]Huge pages support for container workloads [It] [test_id:27477][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Huge pages support for container workloads
/go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/hugepages.go:123
  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc0004d3ca0>: {
          msg: "failed to run command [cat /rootfs/sys/fs/cgroup/hugetlb/hugetlb.1GB.usage_in_bytes]: output \"cat: /rootfs/sys/fs/cgroup/hugetlb/hugetlb.1GB.usage_in_bytes: No such file or directory\\r\\r\\n\"; error \"\"; command terminated with exit code 1",
          err: <exec.CodeExitError>{
              Err: <*errors.errorString | 0xc000342870>{
                  s: "command terminated with exit code 1",
              },
              Code: 1,
          },
      }
      failed to run command [cat /rootfs/sys/fs/cgroup/hugetlb/hugetlb.1GB.usage_in_bytes]: output "cat: /rootfs/sys/fs/cgroup/hugetlb/hugetlb.1GB.usage_in_bytes: No such file or directory\r\r\n"; error ""; command terminated with exit code 1
```
We need to check if this is another issue

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
ffromani: this deserves to be a comment in the code

ffromani (commented):
/lgtm

nice work!

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
Tal-or: I don't think the test is going to pass. 
I think this patch just float that the issue is real and consistent.

ffromani: Right, there's probably little point in documenting this even in the fixing patch.

ffromani:
please do _NOT_ rebase until both #729 and #750 merged

yanirq:
@Tal-or  we can rebase now

Tal-or:
Agreed with @mrniranjan to remove the check at the end of the test until we'll figure out why the kernel doesn't put the cpus back in the sched domain

Tal-or:
Agreed with @mrniranjan to remove the check at the end of the test until we'll figure out why the kernel doesn't put the cpus back in the sched domain

PR #731, opened 2023-07-26
OCPBUGS-14241: nto: pao avoid timeout when there are too many CSV

jlojosnegros:
Adding pagination to CSVs list operation while deleting PAO OLM operator elements.

From versions >= 4.11 PAO has been integrated into NTO
For that reason PAO OLM operator elements (ClusterServiceVersion, Subscriptions .. ) should be deleted

As part of this operation CSVs are listed. This operation can take too long in some clusters due to the number of elements to be listed, and so the operation will hit a timeout and the pod will crash. This could end on a `Crashbackloop`situation.

Adding pagination to the list operation should avoid this situation.

### DIFF
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -14,7 +14,6 @@ import (
 	paocontroller "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/machineconfig"
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
-	olmoperators "github.com/operator-framework/api/pkg/operators/install"
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -51,6 +50,8 @@ const (
 	webhookCertDir  = "/apiserver.local.config/certificates"
 	webhookCertName = "apiserver.crt"
 	webhookKeyName  = "apiserver.key"
+
+	performanceOperatorDeploymentName = "performance-operator"
 )
 
 var (
@@ -210,33 +211,22 @@ func removePerformanceOLMOperator(cfg *rest.Config) error {
 		return err
 	}
 
-	// Register OLM types to the client
-	olmoperators.Install(k8sclient.Scheme())
-
-	var performanceOperatorCSVs []olmv1alpha1.ClusterServiceVersion
-	csvs := &olmv1alpha1.ClusterServiceVersionList{}
 	csvSelector := labels.NewSelector()
 	req, err := labels.NewRequirement(olmv1alpha1.CopiedLabelKey, selection.DoesNotExist, []string{})
 	if err != nil {
 		return err
 	}
 	csvSelector.Add(*req)
-	if err := k8sclient.List(context.TODO(), csvs, client.MatchingLabelsSelector{Selector: csvSelector}); err != nil {
-		if !errors.IsNotFound(err) {
-			return err
-		}
+
+	//REVIEW - should this be an input parameter? or something configurable?
+	paginationLimit := uint64(1000)
+	options := []client.ListOption{
+		client.MatchingLabelsSelector{Selector: csvSelector},
 	}
-	for i := range csvs.Items {
-		csv := &csvs.Items[i]
-		deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
-		if deploymentSpecs != nil {
-			for _, deployment := range deploymentSpecs {
-				if deployment.Name == "performance-operator" {
-					performanceOperatorCSVs = append(performanceOperatorCSVs, *csv)
-					break
-				}
-			}
-		}
+
+	performanceOperatorCSVs, err := paocontroller.ListPerformanceOperatorCSVs(k8sclient, options, paginationLimit, performanceOperatorDeploymentName)
+	if err != nil {
+		return err
 	}
 
 	subscriptions := &olmv1alpha1.SubscriptionList{}
@@ -257,7 +247,7 @@ func removePerformanceOLMOperator(cfg *rest.Config) error {
 					}
 					subscriptionExists = false
 				}
-				klog.Infof("Removing performance-addon-operator related CSV %s/%s", csv.Name, csv.Namespace)
+				klog.Infof("Removing performance-addon-operator related CSV %s/%s", csv.Namespace, csv.Name)
 				if err := k8sclient.Delete(context.TODO(), &csv); err != nil {
 					return err
 				}
--- a/pkg/performanceprofile/controller/utils.go
+++ b/pkg/performanceprofile/controller/utils.go
@@ -1,0 +1,52 @@
+package controller
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	olmoperators "github.com/operator-framework/api/pkg/operators/install"
+	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+)
+
+func ListPerformanceOperatorCSVs(k8sclient client.Client, options []client.ListOption, paginationLimit uint64, performanceOperatorDeploymentName string) ([]olmv1alpha1.ClusterServiceVersion, error) {
+	// Register OLM types to the client
+	olmoperators.Install(k8sclient.Scheme())
+	var performanceOperatorCSVs []olmv1alpha1.ClusterServiceVersion
+
+	csvs := &olmv1alpha1.ClusterServiceVersionList{}
+
+	opts := []client.ListOption{}
+	copy(opts, options)
+	opts = append(options, client.Limit(paginationLimit))
+
+	for {
+		if err := k8sclient.List(context.TODO(), csvs, opts...); err != nil {
+			if !errors.IsNotFound(err) {
+				return performanceOperatorCSVs, err
+			}
+		}
+		continueToken := csvs.GetContinue()
+
+		for i := range csvs.Items {
+			csv := &csvs.Items[i]
+			deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+
+			for _, deployment := range deploymentSpecs {
+				if deployment.Name == performanceOperatorDeploymentName {
+					performanceOperatorCSVs = append(performanceOperatorCSVs, *csv)
+					break
+				}
+			}
+
+		}
+		if continueToken == "" {
+			// empty token means there is no more elements
+			break
+		}
+		// add new continuation token and keep looking
+		opts = append(opts, client.Continue(continueToken))
+	}
+	return performanceOperatorCSVs, nil
+}
--- a/test/e2e/basic/csvs.go
+++ b/test/e2e/basic/csvs.go
@@ -1,0 +1,190 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"os"
+	"strconv"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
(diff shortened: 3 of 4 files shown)


cmd/cluster-node-tuning-operator/main.go
```diff
@@ -219,48 +219,65 @@ func removePerformanceOLMOperator(cfg *rest.Config) error {
...
-				if deployment.Name == "performance-operator" {
-					performanceOperatorCSVs = append(performanceOperatorCSVs, *csv)
-					break
+	for {
+		if err := k8sclient.List(context.TODO(), csvs, opts...); err != nil {
+			if !errors.IsNotFound(err) {
+				return err
+			}
+		}
+		continueToken := csvs.GetContinue()
```
yanirq, line 235: can this panic / fail ?

jlojosnegros: mm I think it should not panic. 
error from `List` is checked just above and the marker to know if there is no more elements is an empty continuation token so ...

jlojosnegros:
> fix PR description: PAO is included in NTO since 4.11 (including)

thanks, done.

MarSik:
The logic looks sound to me. However this needs manual validation, because our CI will not test the proper scenario on the main branch.

/lgtm
/approved

jlojosnegros:
/hold 

lets try to add some testing.

yanirq:
@jlojosnegros  are you planning to add more tests to this PR ?

jlojosnegros:
> @jlojosnegros are you planning to add more tests to this PR ?

Yeah, I have added some e2e tests.
Some refactoring to make the code easier to test also

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -208,33 +207,22 @@ func removePerformanceOLMOperator(cfg *rest.Config) error {
...
 		return err
 	}
 	csvSelector.Add(*req)
-	if err := k8sclient.List(context.TODO(), csvs, client.MatchingLabelsSelector{Selector: csvSelector}); err != nil {
-		if !errors.IsNotFound(err) {
-			return err
-		}
+
+	//REVIEW - should this be an input parameter? or something configurable?
+	paginationLimit := uint64(1000)
```
ffromani, line 218: @jmencak any preference here?

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -247,15 +235,15 @@ func removePerformanceOLMOperator(cfg *rest.Config) error {
 		subscription := &subscriptions.Items[i]
 		subscriptionExists := true
 		for _, csv := range performanceOperatorCSVs {
-			if subscription.Namespace == csv.Namespace && subscription.Status.InstalledCSV == csv.Name {
+			if subscription.Namespace == csv.GetNamespace() && subscription.Status.InstalledCSV == csv.GetName() {
```
ffromani, line 238: moving to getters is a bit of sidegrade, but won't hurt

pkg/performanceprofile/controller/utils.go
```diff
@@ -0,0 +1,51 @@
+package controller
```
ffromani, line 1: can we avoid `utils.go` perhaps? (semi-related: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common)

pkg/performanceprofile/controller/utils.go
```diff
@@ -0,0 +1,51 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	olmoperators "github.com/operator-framework/api/pkg/operators/install"
+	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+)
+
+func ListPerformanceOperatorCSVs(k8sclient client.Client, options []client.ListOption, paginationLimit uint64) ([]olmv1alpha1.ClusterServiceVersion, error) {
+	// Register OLM types to the client
+	olmoperators.Install(k8sclient.Scheme())
```
ffromani, line 15: this should preferably be done once. Now, this function is still likely called once, but still.

pkg/performanceprofile/controller/utils.go
```diff
@@ -0,0 +1,51 @@
...
+
+	var performanceOperatorCSVs []olmv1alpha1.ClusterServiceVersion
+
+	csvs := &olmv1alpha1.ClusterServiceVersionList{}
+	initialOpts := append(options, client.Limit(paginationLimit))
```
ffromani, line 20: this _can_ still mutate the argument passed (https://go.dev/blog/slices-intro) likely NOT an issue considering how the function is implemented, but still.

pkg/performanceprofile/controller/utils.go
```diff
@@ -0,0 +1,51 @@
...
+			}
+		}
+		continueToken := csvs.GetContinue()
+
+		for i := range csvs.Items {
+			csv := &csvs.Items[i]
+			deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+
+			for _, deployment := range deploymentSpecs {
+				if deployment.Name == "performance-operator" {
```
ffromani, line 36: can we detect (maybe pass as argument) the expected name from a reliable source? IOW, is it safe to hardcode here?

pkg/performanceprofile/controller/utils.go
```diff
@@ -0,0 +1,51 @@
...
+				}
+			}
+
+		}
+		if continueToken == "" {
+			// empty token means there is no more elements
+			break
+		}
+		// add new continuation token and keep looking
+		opts = append(initialOpts, client.Continue(continueToken))
```
ffromani, line 48: append or replace? can we end up with more than 1 token in the options?

test/e2e/basic/csvs.go
```diff
@@ -0,0 +1,228 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/ginkgo/v2"
```
ffromani, line 9: duplicated import

test/e2e/basic/csvs.go
```diff
@@ -0,0 +1,228 @@
...
+				options := []client.ListOption{}
+				ret, err := paocontroller.ListPerformanceOperatorCSVs(cli, options, paginationLimit)
+				Expect(err).To(Succeed())
+				Expect(ret).To(HaveLen(0))
+
+			})
+		})
+
+		When("there are few CSVs in the cluster", func() {
+			const csvs_file = "../testing_manifests/csv-dummy.yaml"
```
ffromani, line 83: please use `camelCase`

test/e2e/basic/csvs.go
```diff
@@ -0,0 +1,228 @@
...
+			existingCSVs = len(performanceOperatorCSVs)
+
+			By(fmt.Sprintf("Delete all(%d) CSVs", existingCSVs))
+			for _, element := range performanceOperatorCSVs {
+				err = cli.Delete(context.TODO(), &element)
+				Expect(err).To(Succeed())
+			}
+
+			paginationLimit = uint64(1000)
+			By(fmt.Sprintf("Setting default paginationLimit to: %d", paginationLimit))
```
ffromani, line 57: deleting stuff makes me wary in general. Can we create the CSVs we need in out own testing namespace and leave the existing one alones?

test/e2e/basic/csvs.go
```diff
@@ -0,0 +1,228 @@
...
+				for _, key := range csvs_created {
+					obj := olmv1alpha1.ClusterServiceVersion{}
+					err = cli.Get(context.TODO(), key, &obj)
+					Expect(err).To(Succeed())
+
+					err = cli.Delete(context.TODO(), &obj)
+					Expect(err).To(Succeed())
+				}
+				csvs_created = nil
+			})
```
ffromani, line 128: if we use our own test-specific namespace, we can just delete the namespace and ensure there's no resource leak. There's anything test-specific which prevent us to use this approach?

test/e2e/basic/csvs.go
```diff
@@ -0,0 +1,228 @@
...
+
+			When("pagination value is set to ONE", func() {
+				var oldPaginationLimit uint64
+				BeforeEach(func() {
+					oldPaginationLimit = paginationLimit
+					paginationLimit = 1
+				})
```
ffromani, line 135: can we just shadow the pagination limit here in the innermost closure? would it be less readable? (I don't think so, but let's see)

test/e2e/basic/csvs.go
```diff
@@ -0,0 +1,228 @@
...
+				var oldPaginationLimit uint64
+				BeforeEach(func() {
+					oldPaginationLimit = paginationLimit
+					times := 100
+					paginationLimit = uint64(times * len(csvs_created))
+					By(fmt.Sprintf("Setting paginationLimit to (%d), that is %d times number of csvs created (%d)", paginationLimit, times, len(csvs_created)))
+				})
+				AfterEach(func() {
+					paginationLimit = oldPaginationLimit
+				})
```
ffromani, line 157: same comment about intentional shadowing

pkg/performanceprofile/controller/utils.go
jlojosnegros: I totally agree that fuzzy package names can lead to a jumble package with different functions but I didn't came with a better name for this and as `util` package is commonly used in our projects took that path.

I enclosed the functionality in a function to be able to test it easely ( not need to restart operator to execute the sut code )

pkg/performanceprofile/controller/utils.go
jlojosnegros: Totally agree ... 
My concern here is that this needs to be executed for the `List` operation to work, so either the function execute it or we assume it is a pre-requisite for the function to work. 

I do not know how to test if the scheme is properly configured so opted for this.

pkg/performanceprofile/controller/utils.go
jlojosnegros: yeah, sure we can ... 
it was hardcoded in the original code and I forgot to change it.

pkg/performanceprofile/controller/utils.go
jlojosnegros: AFAIU we can't because I am using `initialOpts` , which do not have the token, as "base" to append the token
A different thing is that maybe using replace could be better or at least more easy to read.

jlojosnegros: About the mutability scenario I was thinking on use `copy` to avoid it.
	opts := []client.ListOption{}
	copy(opts, options)
	opts = append(options, client.Limit(paginationLimit))
```

Also I think that even using just `opts` and appending new `Continue` tokens at the end will work, as the `List` method will use `ApplyOptions` to merge all the options into just one applying the last of them over the previous so the values of the latest elements would prevail. What I am not sure of is if we can rely on this or it is just an implementation detail

test/e2e/basic/csvs.go
jlojosnegros: done

test/e2e/basic/csvs.go
jlojosnegros: None. Thanks for the comment.
Done

test/e2e/basic/csvs.go
jlojosnegros: I find more readable the "backup&restore" operation cause I can see it is a deliberate operation while shadowing could be a typo error ... but in this case shadowing looks better.

test/e2e/basic/csvs.go
jlojosnegros: Thanks.

pkg/performanceprofile/controller/utils.go
jlojosnegros: Thanks for the info.

cmd/cluster-node-tuning-operator/main.go
jlojosnegros: Changed.

jlojosnegros:
@yanirq PTAL 
Thanks

yanirq:
/payload-job periodic-ci-openshift-release-master-ci-4.15-upgrade-from-stable-4.14-e2e-gcp-ovn-upgrade

jlojosnegros:
/hold cancel 
do not know why it still has the hold label

jlojosnegros:
code rebased over latest master.

jlojosnegros:
/lgtm

Label dropped due to rebase.

PR #733, opened 2023-07-26
NTO: force cgroupv1 with node config

haircommander:
NTO is not ready for cgroupv2, so we must add the object to prevent clusters with it from being on v2

### DIFF
--- a/test/e2e/performanceprofile/testdata/render-expected-output/manual_node.yaml
+++ b/test/e2e/performanceprofile/testdata/render-expected-output/manual_node.yaml
@@ -1,0 +1,13 @@
+apiVersion: config.openshift.io/v1
+kind: Node
+metadata:
+  creationTimestamp: null
+  name: cluster
+  ownerReferences:
+  - apiVersion: performance.openshift.io/v2
+    kind: PerformanceProfile
+    name: manual
+    uid: ""
+spec:
+  cgroupMode: v1
+status: {}
--- a/manifests/40-rbac.yaml
+++ b/manifests/40-rbac.yaml
@@ -97,6 +97,10 @@ rules:
 - apiGroups: ["operators.coreos.com"]
   resources: ["clusterserviceversions","operatorgroups","subscriptions"]
   verbs: ["get","delete","list","update","watch"] 
+# Needed to set the cgroup version, as NTO needs v1 for now
+- apiGroups: ["config.openshift.io"]
+  resources: ["nodes"]
+  verbs: ["create", "update"]
 ---
 
 # Bind the operator cluster role to its Service Account.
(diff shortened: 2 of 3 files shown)


yanirq:
please run `make render-sync` and update the test files under: https://github.com/openshift/cluster-node-tuning-operator/tree/master/test/e2e/performanceprofile/testdata/render-expected-output

haircommander:
failure seems unrelated
/retest

Tal-or:
Will the MachineConfig get generated automatically when we set this flag at the object?

harche:
> Will the MachineConfig get generated automatically when we set this flag at the object?

Yes, [syncNodeConfigHandler](https://github.com/openshift/machine-config-operator/blob/ddea245222e437e2259bb56f559a2b36fa7eaea0/pkg/controller/kubelet-config/kubelet_config_nodes.go#L63) in MCO should take care of that.

test/e2e/performanceprofile/testdata/render-expected-output/manual_node.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: config.openshift.io/v1
+kind: Node
```
MarSik, line 2: Kind: Node? Shouldn't that be NodeConfig?

MarSik: Btw, we have a makefile target to render and update the expected manifests (make render-sync iirc).

yanirq: render already mentioned here: https://github.com/openshift/cluster-node-tuning-operator/pull/733#issuecomment-1651969605

yanirq:
looks like latest failures are related and makes me wonder if the configuration for v1 as default actually gets applied:

`msg: "failed to run command [cat /rootfs/sys/fs/cgroup/cpuset/ovs.slice/cgroup.procs]: output \"cat: /rootfs/sys/fs/cgroup/cpuset/ovs.slice/cgroup.procs: No such file or directory\\r\\n\"; error \"\"; command terminated with exit code 1"`
`
msg: "failed to run command [cat /rootfs/sys/fs/cgroup/hugetlb/hugetlb.1GB.usage_in_bytes]: output \"cat: /rootfs/sys/fs/cgroup/hugetlb/hugetlb.1GB.usage_in_bytes: No such file or directory\\r\\r\\n\"; error \"\"; command terminated with exit code 1"`

haircommander:
it's likely we'll go for https://github.com/openshift/cluster-node-tuning-operator/pull/737 instead of this

PR #737, opened 2023-07-27
Update the `config.openshift.io/node` object's `cgroupMode` to "v1"

sairameshv:
cc: @harche @haircommander


haircommander:
```
 pkg/performanceprofile/controller/performanceprofile_controller.go:435:30: k8s.io/apimachinery/pkg/runtime/schema.GroupVersionKind struct literal uses unkeyed fields 
```
vet job failed

haircommander:
one potential limitation of this approach is that the masters will need to rollout a reboot as they are set to v1 as well, even if a perf profile is only set for workers. Given the node config is cluster wide, I don't think there's anything to do (there's no way to apply a node config to a mcp, right @harche @sairameshv )

sairameshv:
> one potential limitation of this approach is that the masters will need to rollout a reboot as they are set to v1 as well, even if a perf profile is only set for workers. Given the node config is cluster wide, I don't think there's anything to do (there's no way to apply a node config to a mcp, right @harche @sairameshv )

Yes, there would be a reboot observed on all the nodes where the Kernel configuration gets updated from "v2" to "v1"

haircommander:
as a note: this only applies to brand new 4.14 clusters that have perf profiles applied day 1, not 4.13 clusters that upgrade with one or 4.14 clusters that are installed with one. I think the number of clusters this applies to should not be very high, but FYI

haircommander:
/lgtm

pao test passed!

sairameshv:
/assign @dagrayvid 

Hello David, Could you PTAL at the changes and approve?

MarSik:
How disruptive will this be on the cluster? All nodes including masters will reboot once the first PerfProfile is applied?

MarSik:
@sairameshv But this PR is not updating the rendered manifests for the install time. So the user would have to do that manually.

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -422,6 +426,29 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	key := types.NamespacedName{
+		Name: nodeCfgName,
+	}
+	nodeCfg := &apiconfigv1.Node{}
+	nodeCfg.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "config.openshift.io",
+		Version: "v1",
+		Kind:    "Node",
+	})
+	err = r.Client.Get(context.Background(), key, nodeCfg)
```
MarSik, line 440: This is also executed during every reconciliation, right? Will the API team complain the number of requests from NTO got higher?

sairameshv:
> @sairameshv But this PR is not updating the rendered manifests for the install time. So the user would have to do that manually.

@MarSik , As I'm new to this code, Could you suggest any similar reference to render a manifest file for the install time?

pkg/performanceprofile/controller/performanceprofile_controller.go
sairameshv: @harche , @haircommander Do you have any thoughts on this?

MarSik:
> > @sairameshv But this PR is not updating the rendered manifests for the install time. So the user would have to do that manually.
> 
> @MarSik , As I'm new to this code, Could you suggest any similar reference to render a manifest file for the install time?

That is what the first PR did, all objects that NTO should create when PerfProfile is present are defined here:

https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go#L43

But it only supports creation, not updates.

harche:
> > > @sairameshv But this PR is not updating the rendered manifests for the install time. So the user would have to do that manually.
> > 
> > 
> > @MarSik , As I'm new to this code, Could you suggest any similar reference to render a manifest file for the install time?
> 
> That is what the first PR did, all objects that NTO should create when PerfProfile is present are defined here:
> 
> https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go#L43
> 
> But it only supports creation, not updates.

Interesting, but when we tested that [PR](https://github.com/openshift/cluster-node-tuning-operator/pull/733/files) we didn't observe the nodes config getting applied. Anything missing in that PR?

pkg/performanceprofile/controller/performanceprofile_controller.go
browsell: I think this needs to be addressed, unnecessary overhead. This really should only need to be done at startup

harche: We are looking for pointers on how to do that in NTO, https://github.com/openshift/cluster-node-tuning-operator/pull/737#issuecomment-1655797423

harche: Although this change will only get the nodes config object and only if it set to v2, it will try to set it to v1. This will not have any effect if the cluster is already in v1.

harche: The number of request to API (to fetch the nodes config object) would be proportional to the number of times performance profile is changed.

MarSik:
> Interesting, but when we tested that [PR](https://github.com/openshift/cluster-node-tuning-operator/pull/733/files) we didn't observe the nodes config getting applied. Anything missing in that PR?

Not sure, the rendered test object was kind: Node and I am not sure if you can have multiple NodeConfig objects on a cluster at the same time. So it might need to be a combination of the reconciliation update for day-2 and properly rendered object for install time.

harche:
> > Interesting, but when we tested that [PR](https://github.com/openshift/cluster-node-tuning-operator/pull/733/files) we didn't observe the nodes config getting applied. Anything missing in that PR?
> 
> Not sure, the rendered test object was kind: Node and I am not sure if you can have multiple NodeConfig objects on a cluster at the same time. So it might need to be a combination of the reconciliation update for day-2 and properly rendered object for install time.

Node object has a cluster wide scope but, it can be updated by anyone with right permissions.

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: We can watch this object using predicates and act accordingly in case it's set to v2
https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/performanceprofile/controller/performanceprofile_controller.go#L149 
BTW this code is only temporarily right?  until NTO catches up with cgroupsv2

yanirq: I don't think we need to add an additional watch here.  
What we see here are additional API calls (get/update) added into the reconcile loop (which has more triggers than only performance profile changes) 
Couldn't this be applied in the upper level ? 

for example, where we still have the upgrade code for removing PAO old operator:
https://github.com/openshift/cluster-node-tuning-operator/blob/96b067113eca80a24b649fe044ec0047b580eccb/cmd/cluster-node-tuning-operator/main.go#L138

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -422,6 +426,29 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	if err != nil {
+		klog.Errorf("failed to get config node object; name=%q err=%v", nodeCfg.GetName(), err)
+		nodeCfg.Name = nodeCfgName
+		nodeCfg.Spec.CgroupMode = apiconfigv1.CgroupModeV1
+		r.Client.Update(ctx, nodeCfg)
```
yanirq, line 445: if we failed getting the object, don't we want to use Client.Create here ?

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -53,7 +54,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-const finalizer = "foreground-deletion"
+const (
+	finalizer   = "foreground-deletion"
+	nodeCfgName = "cluster"
```
yanirq, line 59: will this always be the Node config name ?

pkg/performanceprofile/controller/performanceprofile_controller.go
sairameshv: Ideally yes. 
But in this case, the node config object is by default created, present in the cluster. Reference -> [here](https://github.com/openshift/cluster-config-operator/blob/master/empty-resources/0000_10_config-operator_02_node.cr.yaml) 
This error log is added just to be more cautious over any failure in getting the object.

pkg/performanceprofile/controller/performanceprofile_controller.go
sairameshv: Yes, the node config object named `cluster` is the only custom resource that gets created in the cluster.

pkg/performanceprofile/controller/performanceprofile_controller.go
yanirq: > This is also executed during every reconciliation, right? Will the API team complain the number of requests from NTO got higher?

@MarSik  Adding a Get operation in the reconcile loop is that expensive ?  assuming we cannot add this at startup and MUST have a performance profile present inorder to switch to v1

yanirq:
As it stands - the move from cgroups from v2 to v1 must happen ONLY if a performance profile is present in the cluster. 
Currently we do not have a more elegant solution than setting this under the reconcile loop with the additional Get operation cost in the reconcile loop (not a big toll for performance).

Unless there are not objections - we will approve this PR  which will unblock our CI.

MarSik:
/retitle OCPBUGS-16976: Update the config.openshift.io/node object's cgroupMode to "v1"

yanirq:
/retest
there are flakes that should be addressed with https://github.com/openshift/cluster-node-tuning-operator/pull/730

yanirq:
> Thanks for the review and approval @yanirq , @MarSik !! This code can be removed once cgroupsv2 is supported by the operator.

@sairameshv @haircommander  for future removal (and reference under this PR) -  will it be enough to remove the code here that changes the Node configuration CR to have cgroups v1 ? 

Meaning in the future, if a cluster will be upgraded to 4.15 after it was set to have cgroups v1 via cluster-node-tuning-operator in 4.14,  will openshift automatically be able to set it back to v2 ?  will we get this "for free" from the upgrade procedure ? (via MCO ?)

PR #738, opened 2023-07-28
Apply `cgroupsv1` configuration via `config.openshift.io`'s node object

sairameshv:
cc: @harche @haircommander 

Trying this out to avoid reboot, addition of code in the Reconcile function.

### DIFF
--- a/manifests/60-config-node.yaml
+++ b/manifests/60-config-node.yaml
@@ -1,0 +1,6 @@
+apiVersion: config.openshift.io/v1
+kind: Node
+metadata:
+  name: cluster
+spec:
+  cgroupMode: "v1"


MarSik:
Even if this gets applies it will negate the cgroup v2 default for common clusters. NTO is a core component and is always installed.

sairameshv:
> Even if this gets applies it will negate the cgroup v2 default for common clusters. NTO is a core component and is always installed.

Yeah, I launched(using cluster-bot) a cluster with this PR and verified the node config object and looks like the `cgroupMode` is observed to be `v2` and not `v1`

Hence, ruling out this option.
/close

PR #741, opened 2023-07-30
WIP: DNM: Set NTO cgroups mode to "v1" by default

yanirq:
Update the config.openshift.io/node object's cgroupMode to "v1" on startup


cmd/cluster-node-tuning-operator/main.go
```diff
@@ -134,6 +136,10 @@ func operatorRun() {
 		klog.Exit(err)
 	}
 
+	if err := setCgroupsV1(restConfig); err != nil {
```
MarSik, line 139: This will put all clusters into V1 mode, irrespective of a performance profile presence... The operator is always present.

yanirq: So this must be at the reconcile level

PR #696, opened 2023-06-26
e2e: remove image parameter from PAO must gather

SargunNarula:
The PAO must gather is being merged with the NTO and hence in the updated versions the image parameter is not required to be given with the oc adm command.

Adapt this change in tests and remove the image parameter from the different places.

PR #743, opened 2023-08-02
e2e: remove image parameter from must gather

SargunNarula:
The PAO must gather is being merged with the NTO and hence in the updated versions the image parameter is not required to be given with the oc adm command.


ffromani:
supercedes #696

PR #744, opened 2023-08-02
e2e:wait: return updated pod object explicitly

Tal-or:
In some of the tests we're implicitly relaying on the fact that the pod data returned from the `pods.WaitFor*` functions is the most updated one.

This is not the case in some of the functions on which we are storing the data on different pointer to pod object and never return it.

Instead of using different pod object, we should use the same one passed into the function.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -202,7 +202,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			err = testclient.Client.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
 
-			err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
+			testpod, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
 			logEventsForPod(testpod)
 			Expect(err).ToNot(HaveOccurred())
 
@@ -284,7 +284,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			err = testclient.Client.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
 
-			err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
+			testpod, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
 			logEventsForPod(testpod)
 			Expect(err).ToNot(HaveOccurred(), "failed to create guaranteed pod %v", testpod)
 
@@ -396,7 +396,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 
 			err = testclient.Client.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
-			err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
+			testpod, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
 			logEventsForPod(testpod)
 			Expect(err).ToNot(HaveOccurred())
 
@@ -451,7 +451,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			err := testclient.Client.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
 
-			err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
+			testpod, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
 			logEventsForPod(testpod)
 			Expect(err).ToNot(HaveOccurred())
 		})
@@ -542,7 +542,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			err := testclient.Client.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
 
-			currentPod, err := pods.WaitForPredicate(testpod, 10*time.Minute, func(pod *corev1.Pod) (bool, error) {
+			currentPod, err := pods.WaitForPredicate(client.ObjectKeyFromObject(testpod), 10*time.Minute, func(pod *corev1.Pod) (bool, error) {
 				if pod.Status.Phase != corev1.PodPending {
 					return true, nil
 				}
@@ -717,7 +717,7 @@ func startHTtestPod(cpuCount int) *corev1.Pod {
 	By(fmt.Sprintf("Creating test pod with %d cpus", cpuCount))
 	err := testclient.Client.Create(context.TODO(), testpod)
 	Expect(err).ToNot(HaveOccurred())
-	err = pods.WaitForCondition(testpod, corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
+	testpod, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
 	logEventsForPod(testpod)
 	Expect(err).ToNot(HaveOccurred(), "Start pod failed")
 	// Sanity check for QoS Class == Guaranteed
--- a/test/e2e/performanceprofile/functests/1_performance/hugepages.go
+++ b/test/e2e/performanceprofile/functests/1_performance/hugepages.go
@@ -14,6 +14,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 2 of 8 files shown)


Tal-or:
Thank you @ffromani for helping with this

ffromani (commented):
the intent of the change is fine.

I'm reluctant to merge as-is as I'm generally against mutating arguments if we can help it. I'd prefer to return the updated object, even if this change is more invasive, I think is also clearer in the long run and lead to better code.

Also note that by far the most likely reason we have a pointer to pod as first argument (which will indeed mean the function can mutate the argument) is that we were concerned about performance in argument passing.

IOW, we pass a pointer not because the function was meant to mutate (that was never really documented or decided) but as optimization.

Tal-or:
> the intent of the change is fine.
> 
> I'm reluctant to merge as-is as I'm generally against mutating arguments if we can help it. I'd prefer to return the updated object, even if this change is more invasive, I think is also clearer in the long run and lead to better code.
> 
> Also note that by far the most likely reason we have a pointer to pod as first argument (which will indeed mean the function can mutate the argument) is that we were concerned about performance in argument passing.
> 
> IOW, we pass a pointer not because the function was meant to mutate (that was never really documented or decided) but as optimization.

Yes I completely understand your intention.
I mutate the pointer itself because: 
1. I wanted to avoid changing the functions signature, because as you said it would required more invasive change (maybe in cnf-feature-deploy as well?)
2. mutating the pod pointer is common it's a controller-runtime-like approach.

Anyway I'm OK with what ever approach

ffromani:
>     1. I wanted to avoid changing the functions signature, because as you said it would required more invasive change (maybe in cnf-feature-deploy as well?)
> 
>     2. mutating the pod pointer is common and a controller-runtime-like approach.

You make good points. Even if I'm not a fan of this approach, I won't block it either. I've tagged the other reviewers, let's see if we have a preference collectively.

MarSik:
I wish we had an explicit way of marking the pointer as mutable (&mut Pod from Rust anyone? :)
I am OK with the approach, but I want to see that written in the doctext for the each of the functions.

Tal-or:
> I wish we had an explicit way of marking the pointer as mutable (&mut Pod from Rust anyone? :) I am OK with the approach, but I want to see that written in the doctext for the each of the functions.

`const pod *corev1.Pod` :)

yanirq:
> > ```
> > 1. I wanted to avoid changing the functions signature, because as you said it would required more invasive change (maybe in cnf-feature-deploy as well?)
> > 
> > 2. mutating the pod pointer is common and a controller-runtime-like approach.
> > ```
> 
> You make good points. Even if I'm not a fan of this approach, I won't block it either. I've tagged the other reviewers, let's see if we have a preference collectively.

I would not rule of the "invasive" approach yet , can we check to what extent is it invasive ? (maybe its not that large)

ffromani:
Rough guesstimation:
```
1016 11:29:20 fromani@musashi3 ~/go/src/github.com/openshift/cluster-node-tuning-operator $ git log -1
commit 0e3f9fca2f798654e3a0e8758241a299b27dae8d (HEAD -> master, upstream/release-4.15, upstream/release-4.14, upstream/master)
Author: SargunNarula <57625241+SargunNarula@users.noreply.github.com>
Date:   Wed Aug 2 18:14:03 2023 +0530

    e2e: remove image parameter from must gather (#743)
1017 11:29:24 fromani@musashi3 ~/go/src/github.com/openshift/cluster-node-tuning-operator $ git grep -w WaitForDeletion | grep -v vendor | grep pods | wc -l
7
1018 11:29:52 fromani@musashi3 ~/go/src/github.com/openshift/cluster-node-tuning-operator $ git grep -w WaitForCondition | grep -v vendor | grep pods | wc -l
15
1019 11:29:59 fromani@musashi3 ~/go/src/github.com/openshift/cluster-node-tuning-operator $ git grep -w WaitForPhase | grep -v vendor | grep pods | wc -l
3
```

swatisehgal (commented):
My thought on this is that for operations like the ones here (`WaitForDeletion`, `WaitForCondition` and `WaitForPhase`) it seems inappropriate to assume that the pod object returned is mutated (and is fresher than the object passed). Mutation should not be part of the equation here and the focus should be on **waiting** until the pod is deleted, meets a condition or attains a certain phase respectively. Ideally we should have had a function definition where we passed pod name instead of a pointer to pod object and that would have made it explicit that we don't expect pod object to be mutated.

If the caller relies on a fresh pod object, they should obtain it explicitly. The current changes seem to be further enforcing this (incorrect?) assumption rather than correcting or making it explicit. Now the question is are there many places where we have this assumption? IOW, what is the impact surface if we wanted to rectify this assumption of obtaining upto date pod objects when these functions are called? 

If the motivating factor here to ensure clarity and to facilitate the consumer of utils/pods package, how about we add a function to explicitly obtain an updated pod object like below and use that where we expect to have a fresh pod object:

```
func GetPod(podName, namespace string) (*corev1.Pod, error){
.....
}
```



I am curious about what exactly is the reason we want to have the latest pod object? Do we expect the pod object to change here? In theory I can see a few use cases (in place VPA allowing resources updates, updates to labels, taints, tolerations etc)  but I would like to understand it better in the context of the use cases we care about and our test suite.

ffromani:
> My thought on this is that for operations like the ones here (`WaitForDeletion`, `WaitForCondition` and `WaitForPhase`) it seems inappropriate to assume that the pod object returned is mutated (and is fresher than the object passed). Mutation should not be part of the equation here and the focus should be on **waiting** until the pod is deleted, meets a condition or attains a certain phase respectively. Ideally we should have had a function definition where we passed pod name instead of a pointer to pod object and that would have made it explicit that we don't expect pod object to be mutated.

I fully agree. It would be much better if the functions accept the pod identification data (= ObjectKey or namespace/name pair) rather than the full pod object.

> If the caller relies on a fresh pod object, they should obtain it explicitly. The current changes seem to be further enforcing this (incorrect?) assumption rather than correcting or making it explicit. Now the question is are there many places where we have this assumption? IOW, what is the impact surface if we wanted to rectify this assumption of obtaining upto date pod objects when these functions are called?

This would be indeed the most explicit approach.

> If the motivating factor here to ensure clarity and to facilitate the consumer of utils/pods package, how about we add a function to explicitly obtain an updated pod object like below and use that where we expect to have a fresh pod object:
> 
> ```
> func GetPod(podName, namespace string) (*corev1.Pod, error){
> .....
> }
> ```
> 
> I am curious about what exactly is the reason we want to have the latest pod object? Do we expect the pod object to change here? In theory I can see a few use cases (in place VPA allowing resources updates, updates to labels, taints, tolerations etc) but I would like to understand it better in the context of the use cases we care about and our test suite.

It's to save a Get which is not really costly, but redundant (yes, this is a form of premature optimization)

In a nutshell, some Wait* functions return a copy of the object after the waited condition is met, effectively saving the caller a Get. This is both an optimization and a implementation detail leaking (The Wait* functions are all essentially a Get+condition check wrapped in a poll loop with a timeout), but the saving grace is that it still kinda fits in the semantic, albeit not cleanest approach.

Tal-or:
Then it's settle we have a decision. please review that latest patch

Tal-or:
/test e2e-gcp-pao
New flake/failure, that we won't investigate as part of this PR
```
[rfe_id:27368][performance] Verify API Conversions  [JustBeforeEach] when the GloballyDisableIrqLoadBalancing field set to false should preserve the value during the v1 <-> v2 conversion
  [JustBeforeEach] /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:777
  [It] /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:810
  [FAILED] Failed to create profile
  Unexpected error:
      <*errors.StatusError | 0xc00037f720>: {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "Internal error occurred: failed calling webhook \"vwb.performance.openshift.io\": failed to call webhook: Post \"[https://performance-addon-operator-service.openshift-cluster-node-tuning-operator.svc:443/validate-performance-openshift-io-v2-performanceprofile?timeout=10s](https://performance-addon-operator-service.openshift-cluster-node-tuning-operator.svc/validate-performance-openshift-io-v2-performanceprofile?timeout=10s)\": no endpoints available for service \"performance-addon-operator-service\"",
              Reason: "InternalError",
              Details: {
                  Name: "",
                  Group: "",
                  Kind: "",
                  UID: "",
                  Causes: [
                      {
                          Type: "",
                          Message: "failed calling webhook \"vwb.performance.openshift.io\": failed to call webhook: Post \"https://performance-addon-operator-service.openshift-cluster-node-tuning-operator.svc:[443](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-node-tuning-operator/744/pull-ci-openshift-cluster-node-tuning-operator-master-e2e-gcp-pao/1688150438286725120#1:build-log.txt%3A443)/validate-performance-openshift-io-v2-performanceprofile?timeout=10s\": no endpoints available for service \"performance-addon-operator-service\"",
                          Field: "",
                      },
                  ],
                  RetryAfterSeconds: 0,
              },
              Code: 500,
          },
      }
      Internal error occurred: failed calling webhook "vwb.performance.openshift.io": failed to call webhook: Post "[https://performance-addon-operator-service.openshift-cluster-node-tuning-operator.svc:443/validate-performance-openshift-io-v2-performanceprofile?timeout=10s](https://performance-addon-operator-service.openshift-cluster-node-tuning-operator.svc/validate-performance-openshift-io-v2-performanceprofile?timeout=10s)": no endpoints available for service "performance-addon-operator-service"
  occurred
  In [JustBeforeEach] at: /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:793 @ 08/06/23 12:22:20.221
```

yanirq:
looks good to me . ill wait for others to give an ack before I add the approval

ffromani (approved):
/approve

thanks for the changes, IMO much better now

ffromani:
holding to let other reviewers chime in

PR #746, opened 2023-08-03
OCPBUGS-20368: E2E: Add tests for Dynamic ovs pinning

mrniranjan:


test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -95,3 +518,246 @@ func cpuSpecToString(cpus *performancev2.CPU) string {
...
+		if err != nil {
+			return nil, err
+		}
+		pid := strings.Fields(output)
+		pids = append(pids, pid...)
+	}
+	return pids, nil
+}
+
+/*// getCpuOfOvsServices returns cpus used by the ovs services ovs-vswitchd and ovs-dbserver
```
Tal-or, line 703: please remove the commented code

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -95,3 +518,246 @@ func cpuSpecToString(cpus *performancev2.CPU) string {
 	}
 	return sb.String()
 }
+
+func createMachineConfig(profile *performancev2.PerformanceProfile) (*machineconfigv1.MachineConfig, error) {
```
Tal-or, line 522: Since it's a tests code which used only once here, maybe it's simpler to read a complete MC manifest using `embed` and just apply it instead?

Tal-or (commented):
Initial review.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -1,48 +1,119 @@
...
+	activation_file         string = "/rootfs/var/lib/ovn-ic/etc/enable_dynamic_cpu_affinity"
+	performanceMCP          string
+	//go:embed scripts/*
+	Scripts embed.FS
+)
 
 var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
 	var onlineCPUSet cpuset.CPUSet
 
 	testutils.CustomBeforeAll(func() {
```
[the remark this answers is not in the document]
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -1,48 +1,119 @@
...
 		RunningOnSingleNode = isSNO
+
+		mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+		By("Waiting when mcp finishes updates")
+
+		mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+	})
+
+	AfterAll(func() {
+		By("Removing the crio fix workaround")
```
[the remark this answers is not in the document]
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -1,48 +1,119 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
-	"strings"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"embed"
```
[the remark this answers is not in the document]
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,6 +124,72 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				Expect(err).ToNot(HaveOccurred(), "Unable to fetch latest performance profile")
+				currentPolicy := profile.Spec.NUMA.TopologyPolicy
+				if *currentPolicy == "best-effort" {
+					policy = "restricted"
+				}
+				profile.Spec.NUMA = &performancev2.NUMA{
+					TopologyPolicy: &policy,
+				}
+				By("Updating the performance profile")
+				profiles.UpdateWithRetry(profile)
```
shajmakh, line 156: I see you revert the profile in aftereach, why not add a defer() here to revert it, it's only one test that modifies the profile. 
OTOH if you don't specifically need a PP update, then I believe system reboot on the node on which the pod is running can be more efficient here and cheaper.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,6 +124,72 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				By("Applying changes in performance profile and waiting until mcp will start updating")
+				mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+				By("Waiting for MCP being updated")
+				mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+				By("Checking Activation file")
+				cmd := []string{"ls", activation_file}
```
shajmakh, line 162: same here for linux commands

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +214,325 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, pid := range pidList {
+					cmd := []string{"/bin/bash", "-c", fmt.Sprintf("grep Cpus_allowed_list /proc/%s/status | awk '{print $2}'", pid)}
+					cpusOfovsServices, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+					Expect(cpusOfovsServices).ToNot(BeEmpty(), "unable to fetch cpus of ovs pid %s", pid)
+					testlog.Infof("cpus used by ovs services are: %s", cpusOfovsServices)
+					Expect(cpus == cpusOfovsServices).To(BeTrue(), "affinity of ovn kube node pods(%s) do not match with ovservices(%s)", cpus, cpusOfovsServices)
+					if err != nil {
+						testlog.Error(err.Error())
```
shajmakh, line 260: do we not fail here on purpose? (same for other similar places)

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +214,325 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				testpod, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod), corev1.PodConditionType(corev1.PodReady), corev1.ConditionTrue, 5*time.Minute)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(testpod.Status.QOSClass).To(Equal(corev1.PodQOSGuaranteed))
+				By("Getting the container cpuset.cpus cgroup")
+				containerID, err := pods.GetContainerIDByName(testpod, "test")
+				Expect(err).ToNot(HaveOccurred())
+
+				containerCgroup := ""
+				Eventually(func() string {
+					cmd := []string{"/bin/bash", "-c", fmt.Sprintf("find /rootfs/sys/fs/cgroup/cpuset/ -name *%s*", containerID)}
```
shajmakh, line 289: As 4.15 will be using cgroupv2, I believe we need to have some generic function to switch respectively according to what cgroup the system is configured to, yet I believe this is part of next release.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +214,325 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				testlog.Infof("Cpus used by ovn Containers are %s", cpus)
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, pid := range pidList {
+					cmd := []string{"/bin/bash", "-c", fmt.Sprintf("grep Cpus_allowed_list /proc/%s/status | awk '{print $2}'", pid)}
+					cpusOfovsServices, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+					Expect(cpusOfovsServices).ToNot(BeEmpty(), "unable to fetch cpus of ovs pid %s", pid)
+					testlog.Infof("cpus used by ovs services are: %s", cpusOfovsServices)
+					Expect(cpus == cpusOfovsServices).To(BeTrue(), "affinity of ovn kube node pods(%s) do not match with ovservices(%s)", cpus, cpusOfovsServices)
```
shajmakh, line 258: you can use To(Equal(..)) instead

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +214,325 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				ovnContainers, err := getOvnPodContainers(&ovnPod)
+				Expect(err).ToNot(HaveOccurred())
+				cpus := getCpusUsedByOvnContainer(ovnContainers[0])
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, pid := range pidList {
+					cmd := []string{"/bin/bash", "-c", fmt.Sprintf("grep Cpus_allowed_list /proc/%s/status | awk '{print $2}'", pid)}
+					cpusOfovServices, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+					Expect(cpus == cpusOfovServices).To(BeTrue(), "affinity of ovn kube node pods(%s) do not match with ovservices(%s)", cpus, cpusOfovServices)
```
shajmakh, line 311: same, you can use To(Equal(..)) instead

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +214,325 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				testlog.Infof("cpus used by ovn kube node pods %v", ovnContainerCpus)
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+
+				// Verify ovs-vswitchd and ovsdb-server process affinity is updated
+				for _, pid := range pidList {
+					cmd := []string{"/bin/bash", "-c", fmt.Sprintf("grep Cpus_allowed_list /proc/%s/status | awk '{print $2}'", pid)}
+					cpusOfovServices, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+					testlog.Infof("cpus used by ovs service %s is %s", pid, cpusOfovServices)
+					Expect(cpusOfovServices == ovnContainerCpus).To(BeTrue(), "affinity of ovn kube node pods(%s) do not match with ovservices(%s)", ovnContainerCpus, cpusOfovServices)
```
shajmakh, line 405: same, you can use To(Equal(..)) instead, and in other later places

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +214,325 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				}()
+
+				// Check the cpus of ovn kube node pods and ovs services
+				ovnContainerCpus, err := getOvnContainerCpus(workerRTNode)
+				Expect(err).ToNot(HaveOccurred(), "Unable to fetch cpus of the containers inside ovn kubenode pods")
+				testlog.Infof("cpus used by ovn kube node pods %v", ovnContainerCpus)
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+				for _, pid := range pidList {
+					cmd := []string{"/bin/bash", "-c", fmt.Sprintf("grep Cpus_allowed_list /proc/%s/status | awk '{print $2}'", pid)}
```
shajmakh, line 460: same here and in other places regarding depending on linux commands

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +214,325 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				// Note: this is used only to trigger reboot
+				policy := "best-effort"
+				// Need to make some changes to pp , causing system reboot
+				// and check if activation files is modified or deleted
+				profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+				Expect(err).ToNot(HaveOccurred(), "Unable to fetch latest performance profile")
+				currentPolicy := profile.Spec.NUMA.TopologyPolicy
+				if *currentPolicy == "best-effort" {
+					policy = "restricted"
+				}
```
shajmakh, line 480: same here regarding reboot, if the goal is to trigger a system reboot and not a PP specific update, then I'd do a system reboot on the node on which the pod runs on. This 1. won't need a revert 2. keep the tests running on consistent configuration (unless intended the opposite) 3. less dependency by directly rebooting the node instead of w/a.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +214,325 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+					cpusOfovServices, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+					testlog.Infof("cpus used by ovs services is %s", cpusOfovServices)
+					Expect(cpusOfovServices == ovnContainerCpus).To(BeTrue(), "affinity of ovn kube node pods(%s) do not match with ovservices(%s)", ovnContainerCpus, cpusOfovServices)
+					if err != nil {
+						testlog.Error(err.Error())
+					}
+				}
+			})
+			AfterEach(func() {
+				By("Reverting the Profile")
```
shajmakh, line 511: same regarding the revert vs reboot

shajmakh (commented):
Thanks for the PR! I have left few comments below for you. 
It would be good to have a short description on the PR, and squash commits into single units where it fits.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,6 +124,72 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
 		Expect(err).ToNot(HaveOccurred())
 	})
 
+	Describe("[rfe_id: 64006][Dynamic OVS Pinning]", Ordered, func() {
+		Context("[Performance Profile applied]", func() {
+			It("[test_id:64097] Activation file is created", func() {
+				cmd := []string{"ls", activation_file}
```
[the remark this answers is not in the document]
mrniranjan: This is being run on the worker-cnf node , we need to use linux tools.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: we can't run systemctl reboot from the pod, It doesn't allow 
For example:

```
[root@dell-r630-007 ~]# oc exec -it pods/machine-config-daemon-bvp6t -n openshift-machine-config-operator -- bash -c "systemctl reboot"
Defaulted container "machine-config-daemon" out of: machine-config-daemon, kube-rbac-proxy
Running in chroot, ignoring request: reboot
```
Also yes we are changing Profile to trigger reboot, but this is the most safest way to reboot as all the utility functions are well tested and used extensively.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: as mentioned above, we need to use the ls command to check the activation file exists on worker node .

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: please note these commands are required to run on worker node.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: As mentioned above using systemctl reboot or rebooting the node from a pod will cause  unspecified behavior.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
ffromani: re: reboot, you can do jumping through some hoops: https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/9_reboot/devices.go#L125

that said, can't tell if the reboot is the best approach here

mrniranjan: also executing commands through mcd pod is fine as long as you need some output, executing commands like "reboot", where you will instantly loose connection to the node , i find it not so appropriate. As i mentioned above, i can follow the same method if required.

shajmakh: I see your point, but there is a big difference in execution time with the PP update vs node reboot. when PP is updated it will trigger reboot on all (usually not less than 2) nodes, and later on the revert will do the same, so the time is x4 than rebooting only the node on which the pod is running. I do not expect the pod to be terminated on node reboot, otherwise it would likely be a bug

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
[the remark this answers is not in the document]
shajmakh: you're right, I missed that.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
shajmakh: okay, so I see this involves all nodes. following this, correct me if mistaken,  I think it is a good idea to cover that also when verifying the existence of the file not only on one node but on all relevant nodes, right?
ref: 
https://github.com/openshift/cluster-node-tuning-operator/pull/746/files/23f53670b348a210197ec8d933c7db6fa48934b5#diff-ce7d5cbe8bfc6cc4efbf81602d5f238b37bafcbb68bf2c1ad66f1e1b148a33c5R163

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: I think i have addressed this already on why doing system reboot is a bad idea.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: addressed above.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +171,326 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(ovsLoadBalancing).To(Equal("0"))
 		})
+		It("[test_id:64098] ovs-vswitchd and ovsdb-server are under ovs.slice cgroup", func() {
+			processCgroups := "cpuset:/ovs.slice"
```
MarSik, line 175: This is an implementation detail, we do not actually care about this as long as the cpus can escape the reserved cpuset.

So I suggest we update this to check the cpuset cpus of the processes span all available cpus. Also would it make sense to write this in a way that is compatible with the upcoming move to cgroups v2?

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +171,326 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+		Context("ovn-kubenode Pods affinity ", func() {
+			BeforeEach(func() {
+				initialProfile = profile.DeepCopy()
+			})
+			It("[test_id:64100] matches with ovs process affinity", func() {
+				ovnPod, err := getOvnPod(workerRTNode)
+				Expect(err).ToNot(HaveOccurred(), "Unable to get ovnPod")
+				ovnContainersids, err := getOvnPodContainers(&ovnPod)
+				Expect(err).ToNot(HaveOccurred())
+				// Generally there are many containers inside a kubenode pods
```
MarSik, line 203: I think we actually care about a specific container. Only one of them has the pinning logic inside. I just forgot which one it is.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
MarSik: +1

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +171,326 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+
+				containerCgroup := ""
+				Eventually(func() string {
+					cmd := []string{"/bin/bash", "-c", fmt.Sprintf("find /rootfs/sys/fs/cgroup/cpuset/ -name *%s*", containerID)}
+					containerCgroup, err = nodes.ExecCommandOnNode(cmd, workerRTNode)
+					Expect(err).ToNot(HaveOccurred())
+					return containerCgroup
+				}, (cluster.ComputeTestTimeout(30*time.Second, RunningOnSingleNode)), 5*time.Second).ShouldNot(BeEmpty())
+
+				By("Get cpus used by the pod")
```
MarSik, line 253: You can use a node agnostic method. Do the equivalent of "oc exec -it <pod> -- cat /sys/fs/cgroup/cpuset/cpuset.cpus" or "oc exec -it <pod> -- taskset -pc 1" and you get the cpus the container sees from inside.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +171,326 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				ovnPod, err := getOvnPod(workerRTNode)
+				Expect(err).ToNot(HaveOccurred(), "Unable to get ovnPod")
+				ovnContainers, err := getOvnPodContainers(&ovnPod)
+				Expect(err).ToNot(HaveOccurred())
+				cpus := getCpusUsedByOvnContainer(ovnContainers[0])
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, pid := range pidList {
+					cmd := []string{"/bin/bash", "-c", fmt.Sprintf("grep Cpus_allowed_list /proc/%s/status | awk '{print $2}'", pid)}
```
MarSik, line 267: Why not use taskset? Is that less reliable than grep, proc and awk?

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -95,3 +510,154 @@ func cpuSpecToString(cpus *performancev2.CPU) string {
...
+	return ovnKubeNodePodContainerids, err
+}
+
+// getCpusUsedByOvnContainer returns cpus used by the ovn kube node container
+func getCpusUsedByOvnContainer(ovnKubeNodePodCtnid string) string {
+	var cpus string
+	var err error
+	var containerCgroup = ""
+	Eventually(func() string {
+		cmd := []string{"/bin/bash", "-c", fmt.Sprintf("find /rootfs/sys/fs/cgroup/cpuset/ -name '*%s*'", ovnKubeNodePodCtnid)}
```
MarSik, line 571: Again, we will have to rewrite all this in the next release unless we make it more generic.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: @MarSik Addressed this in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed this in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed this in the latest commit

mrniranjan: @MarSik In this case  when i use the method of getting the cpus used by taskset,  i get only reserved cpus. Since the ovn kubenode pods contains a lot of containers, we don't have an api to execute commands on specific container in a pod.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+					mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+				}
+			})
+		})
+	})
+
+	Context("Verification of cgroup layout on the worker node", func() {
+		chkOvsCgrpProcs := func(node *corev1.Node) (string, error) {
+			testlog.Info("Verify cgroup.procs is not empty")
+			cmd := []string{"cat", "/rootfs/sys/fs/cgroup/cpuset/ovs.slice/cgroup.procs"}
```
MarSik, line 153: I am still not happy with this, we will have to update it once we switch to cgroups v2. (the cpuset part of the path will disappear).

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				ovnPod, err := getOvnPod(workerRTNode)
+				Expect(err).ToNot(HaveOccurred(), "Unable to get ovnPod")
+
+				ovnContainersids, err := getOvnPodContainers(&ovnPod)
+				Expect(err).ToNot(HaveOccurred())
+
+				// Generally there are many containers inside a kubenode pods
+				// we don't need to check cpus used by all the containers
+				// we take first container
+				cpus := getCpusUsedByOvnContainer(workerRTNode, ovnContainersids[0])
```
MarSik, line 231: There are more containers than just one in the OVN-K pod, are you sure the [0] will pick the right one?

MarSik: It probably works now, but there is a hidden assumption here that all are burstable.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+
+				cmd := []string{"taskset", "-pc", "1"}
+				testpodCpus, err := pods.ExecCommandOnPod(testclient.K8sClient, testpod, cmd)
+				testlog.Infof("%v pod is using %v cpus", testpod.Name, string(testpodCpus))
+
+				By("Get ovnpods running on the worker cnf node")
+				ovnPod, err := getOvnPod(workerRTNode)
+				Expect(err).ToNot(HaveOccurred(), "Unable to get ovnPod")
+
+				By("Get cpu used by ovn pod containers")
```
MarSik, line 276: There is a race condition here. CPU manager only runs the reconcile loop every 5 seconds or so.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				cpus := getCpusUsedByOvnContainer(workerRTNode, ovnContainers[0])
+
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
 
-			ovsCPUSet, err := cpuset.Parse(ovsCpuList)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(ovsCPUSet).To(Equal(onlineCPUSet))
+				for _, pid := range pidList {
```
MarSik, line 284: And another here, the ovnkube-node only checks its affinity once per a second (used to be 10 seconds).

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				testpod2, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod2), corev1.PodConditionType(corev1.PodReady), corev1.ConditionTrue, 5*time.Minute)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(testpod1.Status.QOSClass).To(Equal(corev1.PodQOSGuaranteed))
+
+				By("Getting the container cpuset.cpus cgroup")
+				testpod2Cpus, err := pods.ExecCommandOnPod(testclient.K8sClient, testpod2, tasksetcmd)
+				testlog.Infof("%v pod is using %v cpus", testpod2.Name, string(testpod2Cpus))
+
+				// Get cpus used by the ovnkubenode-pods containers
```
MarSik, line 349: The cpu manager race again.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+					Expect(ovnContainerCpus).To(Equal(maskSet.String()), "affinity of ovn kube node pods(%s) do not match with ovservices(%s)", ovnContainerCpus, maskSet.String())
+					if err != nil {
+						testlog.Error(err.Error())
+					}
+				}
+				// Delete testpod1
+				testlog.Infof("Deleting pod %v", testpod1.Name)
+				deleteTestPod(testpod1)
+
+				// Check the cpus of ovnkubenode pods
```
MarSik, line 374: Here it is complicated... we might hit couple of things:
- the cpu manager reconcile period race
- the kubernetes bug https://github.com/kubernetes/kubernetes/issues/107074

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+
+				defer func() {
+					// delete deployment
+					testlog.Infof("Deleting Deployment %v", dp.Name)
+					err := testclient.Client.Delete(context.TODO(), dp)
+					Expect(err).ToNot(HaveOccurred())
+				}()
+
+				// Check the cpus of ovn kube node pods and ovs services
+				ovnContainerCpus, err := getOvnContainerCpus(workerRTNode)
```
MarSik, line 421: Cpu manager race again.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				// Generally there are many containers inside a kubenode pods
+				// we don't need to check cpus used by all the containers
+				// we take first container
+				cpus := getCpusUsedByOvnContainer(workerRTNode, ovnContainersids[0])
+				testlog.Infof("Cpus used by ovn Containers are %s", cpus)
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, pid := range pidList {
+					cmd := []string{"taskset", "-pc", pid}
```
MarSik, line 237: This code is repeated multiple times in this file, could you maybe extract that to a function?

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+					},
+				}
+				testpod2.Spec.NodeSelector = map[string]string{testutils.LabelHostname: workerRTNode.Name}
+				err = testclient.Client.Create(context.TODO(), testpod2)
+				Expect(err).ToNot(HaveOccurred())
+				testpod2, err = pods.WaitForCondition(client.ObjectKeyFromObject(testpod2), corev1.PodConditionType(corev1.PodReady), corev1.ConditionTrue, 5*time.Minute)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(testpod1.Status.QOSClass).To(Equal(corev1.PodQOSGuaranteed))
+
+				By("Getting the container cpuset.cpus cgroup")
```
MarSik, line 345: This code for getting the Pod cpuset is repeated multiple times in the file, could you maybe extract it to a function?

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+					if err != nil {
+						testlog.Error(err.Error())
+					}
+				}
+
+				// Delete testpod2
+				deleteTestPod(testpod2)
+			})
+
+			It("[test_id:64103] ovs process affinity still excludes guaranteed pods after reboot", Label("t4"), func() {
```
MarSik, line 399: I wonder if this test is necessary. Basically we are making sure OVN-k starts and cpu manager obeys the commands even after reboot. As an e2e scenario it is probably fine, it is just expensive in terms of resources.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				}()
+
+				// Check the cpus of ovn kube node pods and ovs services
+				ovnContainerCpus, err := getOvnContainerCpus(workerRTNode)
+				Expect(err).ToNot(HaveOccurred(), "Unable to fetch cpus of the containers inside ovn kubenode pods")
+				testlog.Infof("cpus used by ovn kube node pods %v", ovnContainerCpus)
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+				for _, pid := range pidList {
+					cmd := []string{"taskset", "-pc", pid}
```
MarSik, line 427: And the ovnkube-node race again.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+
+				By("Updating the performance profile")
+				profiles.UpdateWithRetry(profile)
+				By("Applying changes in performance profile and waiting until mcp will start updating")
+				mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+				By("Waiting for MCP being updated")
+				mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+
+				// Check the cpus of ovn kube node pods and ovs services
+				ovnContainerCpus, err = getOvnContainerCpus(workerRTNode)
```
MarSik, line 463: This is racy on many levels:
- The reboot killed the guaranteed pods and the Deployment will take its time to restart them
- The cpu manager takes time to apply the cpu masks

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				Expect(err).ToNot(HaveOccurred(), "Unable to fetch cpus of the containers inside ovn kubenode pods")
+				testlog.Infof("cpus used by ovn kube node pods %v", ovnContainerCpus)
+				pidList, err = getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+
+				// Verify ovs-vswitchd and ovsdb-server process affinity is updated
+				for _, pid := range pidList {
+					cmd := []string{"taskset", "-pc", pid}
```
MarSik, line 471: Please extract this into a function...

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -1,48 +1,73 @@
...
 		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
 
+		performanceMCP, err = mcps.GetByProfile(profile)
+		Expect(err).ToNot(HaveOccurred())
+
+	})
+
+	BeforeEach(func() {
+		if discovery.Enabled() && testutils.ProfileNotFound {
```
Tal-or, line 66: This check can move to `BeforeAll` as well

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,6 +77,76 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				currentSpec, _ := json.Marshal(profile.Spec)
+				spec, _ := json.Marshal(initialProfile.Spec)
+				performanceMCP, err := mcps.GetByProfile(profile)
+				Expect(err).ToNot(HaveOccurred())
+				// revert only if the profile changes.
+				if !bytes.Equal(currentSpec, spec) {
+					Expect(testclient.Client.Patch(context.TODO(), profile,
+						client.RawPatch(
+							types.JSONPatchType,
+							[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
```
Tal-or, line 136: Since `spec` is nested, I think `%v` would display the information better

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -76,6 +171,326 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
 			Expect(ovsLoadBalancing).To(Equal("0"))
 		})
+		It("[test_id:64098] ovs-vswitchd and ovsdb-server are under ovs.slice cgroup", func() {
+			processCgroups := "cpuset:/ovs.slice"
+			pidList, err := getOVSServicesPid(workerRTNode)
+			Expect(err).ToNot(HaveOccurred())
+			for _, pid := range pidList {
+				cgroupPath := fmt.Sprintf("/proc/%s/cgroup", pid)
+				checkCgroupCmd := []string{"cat", cgroupPath}
+				cgroupInfo, err := nodes.ExecCommandOnNode(checkCgroupCmd, workerRTNode)
```
Tal-or: We are not checking the error here

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+	Context("Verification of cgroup layout on the worker node", func() {
+		chkOvsCgrpProcs := func(node *corev1.Node) (string, error) {
+			testlog.Info("Verify cgroup.procs is not empty")
+			cmd := []string{"cat", "/rootfs/sys/fs/cgroup/cpuset/ovs.slice/cgroup.procs"}
+			return nodes.ExecCommandOnNode(cmd, node)
+		}
+		chkOvsCgrpCpuset := func(node *corev1.Node) (string, error) {
 			cmd := []string{"cat", "/rootfs/sys/fs/cgroup/cpuset/ovs.slice/cpuset.cpus"}
-			ovsCpuList, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+			return nodes.ExecCommandOnNode(cmd, workerRTNode)
```
Tal-or, line 158: The `chkOvsCgrpCpuset` function's  parameter name is `node` but you are passing `workerRTNode` to the 
 `nodes.ExecCommandOnNode` function instead.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+		}
+
+		chkOvsCgroupLoadBalance := func(node *corev1.Node) (string, error) {
```
Tal-or, line 161: Same here, we should pass `node` to `nodes.ExecCommandOnNode`

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -1,48 +1,72 @@
...
-
-var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
-	var onlineCPUSet cpuset.CPUSet
-
-	testutils.CustomBeforeAll(func() {
-		isSNO, err := cluster.IsSingleNode()
-		Expect(err).ToNot(HaveOccurred())
-		RunningOnSingleNode = isSNO
-	})
+type checkCgroupFunction func(*corev1.Node) (string, error)
```
Tal-or, line 37: I couldn't find, where this function is being used?

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				}
+
+				if ovsProc {
+					result, err := chkOvsCgrpProcs(workerRTNode)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(result).To(Not(BeEmpty()))
+				}
+
+				if ovsCpus {
+					result, err := chkOvsCgrpCpuset(workerRTNode)
```
Tal-or, line 200: We are not checking the error here

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				testlog.Infof("Cpus used by ovn Containers are %s", cpus)
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, pid := range pidList {
+					cmd := []string{"taskset", "-pc", pid}
+					cpumask, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+					Expect(err).ToNot(HaveOccurred(), "failed to fetch cpus of %s", pid)
+					mask := strings.SplitAfter(cpumask, " ")
+					maskSet, err := cpuset.Parse(mask[len(mask)-1])
```
Tal-or, line 241: We are not checking the error here

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				smtLevel := nodes.GetSMTLevel(cpuID, workerRTNode)
+				if smtLevel < 2 {
+					Skip(fmt.Sprintf("designated worker node %q has SMT level %d - minimum required 2", workerRTNode.Name, smtLevel))
+				}
+				var testpod1, testpod2 *corev1.Pod
+				var err error
+
+				// Create testpod1
+				testpod1 = pods.GetTestPod()
+				testpod1.Namespace = testutils.NamespaceTesting
```
Tal-or, line 310: I see this pod spec mutation on 3 different places, maybe we can extract this code into separate function

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				pidList, err = getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+
+				// Verify ovs-vswitchd and ovsdb-server process affinity is updated
+				for _, pid := range pidList {
+					cmd := []string{"taskset", "-pc", pid}
+					cpumask, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+					Expect(err).ToNot(HaveOccurred(), "failed to fetch cpus of %s", pid)
+					mask := strings.SplitAfter(cpumask, " ")
+					maskSet, err := cpuset.Parse(mask[len(mask)-1])
```
Tal-or, line 387: `err` check is missing

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -52,30 +76,436 @@ var _ = Describe("[performance] Cgroups and affinity", Ordered, func() {
...
+				Expect(err).ToNot(HaveOccurred(), "Unable to fetch cpus of the containers inside ovn kubenode pods")
+				testlog.Infof("cpus used by ovn kube node pods %v", ovnContainerCpus)
+				pidList, err := getOVSServicesPid(workerRTNode)
+				Expect(err).ToNot(HaveOccurred())
+				for _, pid := range pidList {
+					cmd := []string{"taskset", "-pc", pid}
+					cpumask, err := nodes.ExecCommandOnNode(cmd, workerRTNode)
+					Expect(err).ToNot(HaveOccurred(), "failed to fetch cpus of %s", pid)
+					mask := strings.SplitAfter(cpumask, " ")
+					maskSet, err := cpuset.Parse(mask[len(mask)-1])
```
Tal-or, line 431: `err` check is missing

I see you're checking the error later in the flow, but this is done after you're already using `maskSet`

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -95,3 +525,147 @@ func cpuSpecToString(cpus *performancev2.CPU) string {
...
+
+// deleteTestPod removes guaranteed pod
+func deleteTestPod(testpod *corev1.Pod) {
+	// it possible that the pod already was deleted as part of the test, in this case we want to skip teardown
+	err := testclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(testpod), testpod)
+	if errors.IsNotFound(err) {
+		return
+	}
+
+	err = testclient.Client.Delete(context.TODO(), testpod)
```
Tal-or, line 537: Please pass the `context` from outside

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -95,3 +525,147 @@ func cpuSpecToString(cpus *performancev2.CPU) string {
...
+}
+
+// getOvnPod Get OVN Kubenode pods running on the worker cnf node
+func getOvnPod(workerNode *corev1.Node) (corev1.Pod, error) {
+	var ovnKubeNodePod corev1.Pod
+	ovnpods := &corev1.PodList{}
+	options := &client.ListOptions{
+		Namespace: "openshift-ovn-kubernetes",
+	}
+	err := testclient.Client.List(context.TODO(), ovnpods, options)
```
Tal-or, line 551: same, first parameter of the function should be `ctx context` and it should be passed to all fucntions which accepts `context`

test/e2e/performanceprofile/functests/utils/mcps/mcps.go
```diff
@@ -296,3 +299,29 @@ func Delete(name string) error {
...
+			},
+		},
+	}
+	return ctrcfg1
+}
+
+func GetValueFromCrioConfig(workercnfNode *corev1.Node, key, confPath string) ([]string, error) {
+	out, err := nodes.ExecCommandOnMachineConfigDaemon(workercnfNode, []string{"cat", filepath.Join("/rootfs", confPath)})
+	if err != nil {
+		msg := fmt.Sprintf("failed to fetch values from %s", confPath)
```
Tal-or, line 321: you can simplify:
`return fmt.Errorf("failed to fetch values from %s: %w", confPath, err)`

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: @MarSik We are picking the ovnkubenode-pod running on the worker-cnf node ,   there 6 containers in each ovnkubenode pod All the containers have the same the cpu affinity 

Example from worker-cnf node:

```
d9a47cd57d88b       254a65c3ca4ef4a15337e1ea6d9fcc009ae81ce11ad2385a7e57fcdcc00ab856   4 hours ago         Running             nbdb                                 11                  e9a21fe08292b       ovnkube-node-rx8ng
a02982b6a5c69       254a65c3ca4ef4a15337e1ea6d9fcc009ae81ce11ad2385a7e57fcdcc00ab856   4 hours ago         Running             northd                               11                  e9a21fe08292b       ovnkube-node-rx8ng
1f358afb3ca9a       39ba63a1ff406d50e8d3d0d67b4028d33be766e49d9fe0fcc77b5f3f2a24e51f   4 hours ago         Running             kube-rbac-proxy-ovn-metrics          11                  e9a21fe08292b       ovnkube-node-rx8ng
fe0c694d25dab       39ba63a1ff406d50e8d3d0d67b4028d33be766e49d9fe0fcc77b5f3f2a24e51f   4 hours ago         Running             kube-rbac-proxy-node                 11                  e9a21fe08292b       ovnkube-node-rx8ng
cac2002c520b2       254a65c3ca4ef4a15337e1ea6d9fcc009ae81ce11ad2385a7e57fcdcc00ab856   4 hours ago         Running             ovn-acl-logging                      11                  e9a21fe08292b       ovnkube-node-rx8ng
c9a851fc9db71       254a65c3ca4ef4a15337e1ea6d9fcc009ae81ce11ad2385a7e57fcdcc00ab856   4 hours ago         Running             ovn-controller                       11                  e9a21fe08292b       ovnkube-node-rx8ng
```

All the containers have same cpu affinity:

```[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]# cat cpuset.cpus
0-3,5,7-23
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]#  find /sys/fs/cgroup/cpuset -name crio-a02982b6a5c69*
/sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-a02982b6a5c69e3c67f763c8edb2afceede88cae47523853fedbed070b5884fc.scope
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]# cat /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-a02982b6a5c69e3c67f763c8edb2afceede88cae47523853fedbed070b5884fc.scope/cpuset.cpus
0-3,5,7-23
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]#  find /sys/fs/cgroup/cpuset -name crio-1f358afb3ca9a*
/sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-1f358afb3ca9a32726b03c101bbd6e1825feba8c7c2ccdde6a64a1aa7c1106d6.scope
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]# cat /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-1f358afb3ca9a32726b03c101bbd6e1825feba8c7c2ccdde6a64a1aa7c1106d6.scope/cpuset.cpus
0-3,5,7-23
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]#  find /sys/fs/cgroup/cpuset -name crio-fe0c694d25dab*
/sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-fe0c694d25dabe1eb9394a40242e1133673e5f3689f994cc1db464525a2047fb.scope
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]# cat /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-fe0c694d25dabe1eb9394a40242e1133673e5f3689f994cc1db464525a2047fb.scope/cpuset.cpus
0-3,5,7-23
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]#  find /sys/fs/cgroup/cpuset -name crio-cac2002c520b2*
/sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-cac2002c520b2c2d0c84c17b21a3cb4224ef024dc9f3c195b3e00813b1e8d94e.scope
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]# cat /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-cac2002c520b2c2d0c84c17b21a3cb4224ef024dc9f3c195b3e00813b1e8d94e.scope/cpuset.cpus
0-3,5,7-23
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]#  find /sys/fs/cgroup/cpuset -name crio-c9a851fc9db71*                                                                                                                          /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-c9a851fc9db71424838be9d5e8d6b9aa435bbb6db54e6bac1ea5733af038cfbb.scope
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]# cat /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-c9a851fc9db71424838be9d5e8d6b9aa435bbb6db54e6bac1ea5733af038cfbb.scope/cpuset.cpus
0-3,5,7-23
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]# find /sys/fs/cgroup/cpuset -name crio-d9a47cd57d88b*
/sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope
[root@ocp-worker-0 crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope]# cat /sys/fs/cgroup/cpuset/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod051b68ac_f4ab_4761_8979_5f6a9bf746e3.slice/crio-d9a47cd57d88b0114c63953bb95955db2c3797e059d71c0cb5aead177a332de9.scope/cpuset.cpus
0-3,5,7-23

```

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: @MarSik this is the flow of the test case 

1.  Create a gu testpod (no annotations are used)
2.  Get cpu used  this testpod (using taskset)
3.  Find the cgroup used by one of the containers of ovn-kubenode pod,
4. There is a  wait of 30seconds here 
5. check it's cpuset.cpus 
6. Verify the cpuset.cpus match with the ovs-vswitchd 

Can you let me know where i can hit the race condition

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
MarSik: Right now they do, but as I said, it is a hidden assumption that this will not change.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
MarSik: Where is the 30 second delay? The race is between the cpus being allocated to the gu pod in 1 and the cpu manager updating the cpu mask for ovnkube-node (should be about 5 seconds).

mrniranjan: @MarSik the 30 second delay is in the https://github.com/openshift/cluster-node-tuning-operator/pull/746/files#diff-ce7d5cbe8bfc6cc4efbf81602d5f238b37bafcbb68bf2c1ad66f1e1b148a33c5R580

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Okay do we have any other reliable method.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: fixed in the latest commit.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Yes it's expensive , that's why we run this test case only on systems with 2 numa nodes . So mostly this test case will be executed only in downstream CI.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Regarding race condition, i have replied above about the order of checking the ovs affinity.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -208,8 +193,7 @@ var s_ = Describe("[performance] Cgroups and affinity", Ordered, func() {
 					Expect(result).To(Equal("0"))
 				}
 			},
-			Entry("[test_id:68385] Verify ovs cgroup with crun ", true, true, true, true),
-			Entry("[test_id:64098] Verify ovs cgroup with runc", false, true, true, true),
+			Entry("[test_id:64098] Verify ovs cgroup with runc", true, true, true),
```
Tal-or, line 196: If we only have a single entry let's just run it as regular `It`

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -218,7 +202,7 @@ var s_ = Describe("[performance] Cgroups and affinity", Ordered, func() {
 			testutils.CustomBeforeAll(func() {
 				initialProfile = profile.DeepCopy()
 			})
-			It("[test_id:64100] matches with ovs process affinity", func() {
+			It("[test_id:64100] matches with ovs process affinity", Label("fun1"), func() {
```
Tal-or, line 205: `Label("fun1")` I would use more descriptive name

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
MarSik: Can we find the names of the containers and pick the right one?

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: I have addressed this by using wait condition to  check for deployment and verify all deployment pods are ready before we check any cpus.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed this in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: I am addressing this by checking first the deployment is created pods and the pods are in running state. before we do any checks further.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed this in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/utils/mcps/mcps.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Removed the label, it was used only for testing.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: @MarSik  i am not understanding the race condition here,  Because In this test case , we are not rebooting the host, we are just creating/deleting gu pods, 
The sequence of events are:
Before the test begins,  node already has performance profile
1. Create a gu pod testpod1
2. Get cpus used by testpod1 using taskset
3. Create a gu pod testpod2
4. Get cpus used by testpod2 using taskset
5. Get cpus used by ovs-vswitchd  (Here we do wait for 30 seconds before we fetch the cpus used by ovs-vswitchd )
6. Delete testpod1
7. Get cpus used by ovs-vswitchd 
8. We verify the cpuid's used by ovs-vswitchd is updated appropriately.   (Here we do wait for 30 seconds before we fetch the cpus used by ovs-vswitchd )

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: i am really not sure, how i should address without complicating the test code, for one i can't reproduce the race condition, so i don't even know how to address this.

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -1,49 +1,78 @@
...
-	BeforeEach(func() {
-		if discovery.Enabled() && testutils.ProfileNotFound {
-			Skip("Discovery mode enabled, performance profile not found")
-		}
+	var (
+		onlineCPUSet            cpuset.CPUSet
+		workerRTNode            *corev1.Node
+		workerRTNodes           []corev1.Node
+		profile, initialProfile *performancev2.PerformanceProfile
+		activation_file         string = "/rootfs/var/lib/ovn-ic/etc/enable_dynamic_cpu_affinity"
```
Tal-or, line 43: If this file is not changing it can be const

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
Tal-or: Why do we still check it in `BeforeEach`?

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
Tal-or: please address

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
Tal-or: you are creating new `context` here. you should pass `ctx` (the one on line 474)

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
Tal-or: Please change the `getOvnPod` to get `context` as it first argument and pass it to the `Client.List` call

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
mrniranjan: Addressed in the latest commit

test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go
```diff
@@ -1,49 +1,83 @@
...
 		workerRTNode = &workerRTNodes[0]
+
 		profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 		Expect(err).ToNot(HaveOccurred())
 
+		performanceMCP, err = mcps.GetByProfile(profile)
+		Expect(err).ToNot(HaveOccurred())
+
+		cgfs, err := nodes.GetCgroupFs(workerRTNode)
+		if cgfs == "tmpfs" {
```
MarSik, line 71: tmpfs? really?

ffromani: guess it comes from https://kubernetes.io/docs/concepts/architecture/cgroups/#check-cgroup-version

mrniranjan: @MarSik I have completely removed this check , the whole automation patch for now will be compatible with cgroupv1 only, I will do the cgroupv2 changes in separate PR as i need more time to test.

mrniranjan: @MarSik Request to review,  regarding cgroupv2 changes, i would like to take some time and test properly before sending any PR.

MarSik (approved):
I am pretty much ok with how it looks like now. I can still see possible improvements (cgroups v2, race prevention etc), however we need better test infra for some of that.

PR #753, opened 2023-08-09
e2e: irqbalance: improve test troubleshooting

ffromani:
Add traces and makes it easier to troubleshoot test flakes. These changes are not meant to fix the flakes yet - still investigating.


ffromani:

this change in test code can't break the hypershift lane this way

PR #754, opened 2023-08-09
OCPBUGS-17219: Render mode should not segfault w/ no matching MCP

MarSik:
There was a possibility of a segfault when the input directory for the render mode had some MCPs, but none of them matched the expected node selector.

### DIFF
--- a/pkg/performanceprofile/cmd/render/render.go
+++ b/pkg/performanceprofile/cmd/render/render.go
@@ -270,6 +270,10 @@ func selectMachineConfigPool(pools []*mcfgv1.MachineConfigPool, selectors map[st
 		}
 	}
 
+	if count == 0 {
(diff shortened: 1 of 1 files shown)


ffromani (commented):
/lgtm
/approve

MarSik:
I think we merged this prematurely, Jose actually fixed it in a better way here https://github.com/openshift/cluster-node-tuning-operator/pull/724

ffromani:
Jose's fix is likely better, but is this one harmful? or just redundant?

MarSik:
Well, it prevents the skipping logic in Jose's fix. Which might not be a bad thing, because Jose's patch prints out a warning and ignores the profile. My change fails the rendering. Not sure which way is better.

PR #756, opened 2023-08-12
OCPBUGS-16348: OSLAT latency spikes due to tsc karg setting

browsell:
Signed-off-by: Brent Rowsell <browsell@redhat.com>

A RHEL 9.x change added a timer to make sure TSC_adjust is always which introduces latency spikes. Update tsc karg to tsc=reliable. This aligns with the realtime tuned profile.


jmencak:
Based on the OCPBUG
/assign @MarSik

PR #758, opened 2023-08-14
OCPBUGS-14026: cgroup: Match the name of the cgroup to what is expected by kubelet

swatisehgal:
Ensure consistent naming of system cgroup in order to ensure that it matches cgroup name expected by Kubelet.


yanirq:
@swatisehgal  you might need to align the render command tests (see failing job for ci/prow/e2e-no-cluster)

yanirq:
/retitle OCPBUGS-14026: cgroup: Match the name of the cgroup to what is expected by kubelet

Tal-or:
Is it possible to add an e2e for verifying the fix?

swatisehgal:
/hold
The fix needs to be verified.

swatisehgal:
/unhold

As @browsell has verified the fix. We will defer adding the e2e test in a follow-up PR.

PR #691, opened 2023-06-20
OCPBUGS-4194:  Use appropriate device path  for rpsmask test

mrniranjan:
When using mco pod to execute find command to search for bridges we use /rootfs/sys/devices

PR #760, opened 2023-08-16
[release-4.13] [manual] OCPBUGS-17794: rps: use default rps mask kernel API

Tal-or:
Manual cherry pick of: https://github.com/openshift/cluster-node-tuning-operator/pull/650 
Plus the following fixes for the e2e test: https://github.com/openshift/cluster-node-tuning-operator/pull/691, https://github.com/openshift/cluster-node-tuning-operator/pull/703


assets/performanceprofile/configs/99-default-rps-mask.conf
```diff
@@ -1,5 +1,5 @@
-# We should apply the rule on the virtual interfaces of the host, because of the RPS mask that would be consulted,
-# is the one on the RX side of the veth in the host.
+# Apply the RPS mask on the virtual interfaces of the host by default, becasue
```
yanirq, line 1: @Tal-or  please note that there was a rename here, is that correct ?

Tal-or: yes why not?

PR #762, opened 2023-08-17
main: remove deprecated multi ns function

Tal-or:
The latest operator SDK version allow specifying
multiple namespaces directly as part of the cache options.

Functionality would remain the same.


jmencak:
Thank you for the PR.
/approve
/lgtm
/retest

PR #768, opened 2023-08-22
OCPBUGS-15044: e2e:irqloadbalance: wait for profile revert

Tal-or:
In the test we were reverting the profile but didn't wait for it to catch up and apply all the changes.

This led to pollution of the next tests execution which in turn leads to flaky tests.

In addition, we move the `defer` call into the highest point possible in the scope, so it won't get missed by the reader.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -138,6 +138,21 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 				return nil
 			}
 
+			defer func() { // return initial configuration
+				By("reverting the profile into its initial state")
+				spec, err := json.Marshal(initialProfile.Spec)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(testclient.Client.Patch(context.TODO(), profile,
+					client.RawPatch(
+						types.JSONPatchType,
+						[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+					),
+				)).ToNot(HaveOccurred())
+				// in the initial profile `GloballyDisableIrqLoadBalancing` value should be false,
+				// so we expect the banned cpu set to be empty
+				Eventually(verifyNodes).WithArguments(cpuset.New()).WithPolling(10 * time.Second).WithTimeout(1 * time.Minute).ShouldNot(HaveOccurred())
+			}()
+
 			var expectedBannedCPUs cpuset.CPUSet
 			if irqLoadBalancingDisabled {
 				expectedBannedCPUs = isolatedCPUSet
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -138,6 +138,21 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
...
+				spec, err := json.Marshal(initialProfile.Spec)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(testclient.Client.Patch(context.TODO(), profile,
+					client.RawPatch(
+						types.JSONPatchType,
+						[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
+					),
+				)).ToNot(HaveOccurred())
+				// in the initial profile `GloballyDisableIrqLoadBalancing` value should be false,
+				// so we expect the banned cpu set to be empty
```
ffromani, line 152: It is not needed, but still beneficial, to annotate the `Eventually` expectation to explicitly convey that if this fails, it's because the profile failed to revert

ffromani (approved):
/approve
thanks, this is a good improvement anyway so let's have it

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
Tal-or: Why do you think the wait is not necessary?

ffromani: The wait is necessary, I just think is better to be extra explicit about what the failure means.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -84,7 +84,13 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 
 			By(fmt.Sprintf("verifying the behavior with irqLoadBalancingDisabled=%v isolatedCPUSet=%v", irqLoadBalancingDisabled, isolatedCPUSet))
 
-			verifyNodes := func(expectedBannedCPUs cpuset.CPUSet) error {
+			verifyNodes := func(irqLoadBalancingDisabled bool) error {
```
ffromani, line 87: so it seems I broke it in e74e75ce075c81b538bbaef5d03eec39b640a420 ?

ffromani (commented):
/lgtm

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
Tal-or: No, because we already look at the external value even before https://github.com/openshift/cluster-node-tuning-operator/commit/e74e75ce075c81b538bbaef5d03eec39b640a420#diff-decd79950ca0682eefd520e1171ff0d6c0af15779a02cc59fc71c4dbe1012bfaL87

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
Tal-or: oh I see, I added the comment just to explain why we're passing `false` argument to the comment and not why we add `Eventually`

ffromani:
interesting, it seems the `e2e-no-cluster` lane is flaky? we will address in a separate PR

yanirq:
test change only - unblock CI

PR #774, opened 2023-08-24
[release-4.13] OCPBUGS-18063: cgroup: Match the name of the cgroup to what is expected by kubelet

swatisehgal:
Ensure consistent naming of system cgroup in order to ensure that it matches cgroup name expected by Kubelet.

- Manual cherry-pick of https://github.com/openshift/cluster-node-tuning-operator/pull/758.
- Note that the second commit is different from the corresponding commit in the original PR as the contents source for netdev physical rps rules had to be updated.


swatisehgal:
/assign @yanirq @MarSik 
for reviews

PR #775, opened 2023-08-29
OCPBUGS-17585: Tighten the rules for modifying Tuned Profiles

jmencak:
NTO operands allow updating Tuned Profiles.  This is intentional and by design
as some host information needs to be communicated back to the NTO operator,
but this also allows a successful local attacker potentially affect node-level
configuration of other cluster nodes.

This change addresses the situation in two ways.  First, scoped RBAC
permissions on Profile.status subresource is used to disallow Node-level
write access to Profile.spec.  Second, the Node resource is used to provide
status loops back to NTO using the kubelet credential to write an annotation
to the Node resource.

This change also simplifies the mechanism for accepting kernel
command-line parameters as calculated by the NTO operands.  Now, all NTO
operands must agree on the calculated kernel command-line parameters.

ClusterOperator/node-tuning now also reports operand version.  The
operand version changes only when all operand replicas have successfully
upgraded and are ready.  This information is used to block PPC
Reconcilation loop until the operator and operand RELEASE_VERSION agree.
This is a short-term measure to prevent from multiple node reboots
during upgrades.

Other fixes:
- Disallow MC updates during upgrades when kernel command-line
parameters of nodes within a MCP do not match.
- ClusterOperator/node-tuning object was sometimes giving false
information based on an old Metadata.Generation.

Resolves: OCPBUGS-17585


jmencak:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-gcp-ovn-rt-upgrade 10

jmencak:
/cc @MarSik 
As discussed on Monday, only block PPC reconciliation loop on upgrades.  I've intentionally split this PR into 2 commits.  The first one was already reviewed by @dagrayvid .  Please take a look at mostly [the parts affecting PPC](https://github.com/openshift/cluster-node-tuning-operator/pull/775/commits/38b7d9437a924449292958821517adcc4969af4d).

pkg/operator/mc.go
```diff
@@ -259,6 +259,36 @@ func (pc *ProfileCalculator) getPoolsForNode(node *corev1.Node) ([]*mcfgv1.Machi
...
 }
 
+// getNodesForPool returns a list of Nodes for MachineConfigPool 'pool'.
+func (pc *ProfileCalculator) getNodesForPool(pool *mcfgv1.MachineConfigPool) ([]*corev1.Node, error) {
+	selector, err := metav1.LabelSelectorAsSelector(pool.Spec.NodeSelector)
+	if err != nil {
+		return nil, fmt.Errorf("invalid label selector %s in MachineConfigPool %s: %v", util.ObjectInfo(selector), pool.ObjectMeta.Name, err)
+	}
+
+	initialNodes, err := pc.listers.Nodes.List(selector)
```
MarSik, line 269: I wonder if we should ignore nodes that are not ready or have scheduling disabled. Just to avoid a situation where a broken node (daemonset and/or kubelet) will prevent updates to the rest of the cluster.

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -370,10 +372,24 @@ func validateUpdateEvent(e *event.UpdateEvent) bool {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	co, err := r.getClusterOperator()
+	if err != nil {
+		klog.Errorf("failed to get ClusterOperator: %v", err)
+		return reconcile.Result{}, err
+	}
+
+	operatorReleaseVersion := os.Getenv("RELEASE_VERSION")
```
MarSik, line 381: Is this maybe too strict? Well at least it allows changing the release version via the env if needed.

pkg/tuned/controller.go
```diff
@@ -918,6 +942,42 @@ func getNodeName() string {
...
+		change = change || node.ObjectMeta.Annotations[k] != v
+		node.ObjectMeta.Annotations[k] = v
+	}
+
+	if !change {
+		// No Node annotations changed, no need to update
+		return nil
+	}
+
+	_, err = c.kubeclient.CoreV1().Nodes().Update(context.TODO(), node, metav1.UpdateOptions{})
```
MarSik, line 973: I think I need a better explanation about the security of this part. Is the annotation protected somehow? Is the kubelet client limited to updating only its own Node object?

In other words, please document the high level architecture and the rationale (the why) better here or in the commit message so we will not have to reread all the code to learn how it was supposed to work in the future.

pkg/performanceprofile/controller/performanceprofile_controller.go
jmencak: Not sure I know what you mean here, but I'm assuming this is just a comment without any action needed.

pkg/operator/mc.go
jmencak: I do not think I agree here, this would unnecessarily complicate the design and I"m assuming that at least during the upgrades, it is all hands on deck and people should deal with broken nodes.

pkg/tuned/controller.go
jmencak: Added a comment + the higher level architecture is already documented in the commit message.

pkg/operator/mc.go
MarSik: But what if someone updates the PerfProfile while there is a node with stopped kubelet / nto daemonset? That Node object will have old bootcmdline data and will therefore prevent all the other nodes from updating to the new tuning.

jmencak: Yes, but we're talking about a normal operation (not upgrades).  I agree such nodes need to be excluded and I'll adjust this PR to reflect it.

jmencak:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-gcp-ovn-rt-upgrade 10

jmencak:
@MarSik, all tests passed and the blocking job also succeeded successfully all 3 times I tried.  I belive I've hopefully also addressed your concerns.

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -370,10 +372,24 @@ func validateUpdateEvent(e *event.UpdateEvent) bool {
...
+	operandReleaseVersion := operatorv1helpers.FindOperandVersion(co.Status.Versions, tunedv1.TunedOperandName)
+	if operandReleaseVersion == nil || operatorReleaseVersion != operandReleaseVersion.Version {
```
yanirq, line 383: Are there scenarios where we can end up with a long period of operand and operator version trying to match ?

jmencak: Only during upgrades when we have failing nodes.  Once an upgrade is successful, operatorReleaseVersion and operandReleaseVersion will always match.

MarSik: Also we were testing this with jmencak and MCO itself will not do changes to an MCP when there are failing nodes. We should document or link to a doc that says the proper way to handle a failing node is to remove it from the pool.

MarSik (approved):
/lgtm

I like it, just make sure the user will be visibly informed when the nodes do not agree and update is blocked. The same about the version mismatch, we should report that in the status (probably as a condition).

pkg/operator/controller.go
```diff
@@ -775,25 +765,30 @@ func (c *Controller) getProviderName(nodeName string) (string, error) {
...
+
+	bootcmdline, bootcmdlineSet := c.pc.state.bootcmdline[profile.Name]
+	if !bootcmdlineSet {
+		klog.V(2).Infof("syncMachineConfig(): bootcmdline for %s not cached, sync cancelled", profile.Name)
 		return nil
 	}
 
-	bootcmdline := profile.Status.Bootcmdline
+	if ok := c.allNodesAgreeOnBootcmdline(nodes); !ok {
+		return fmt.Errorf("not all %d Nodes in MCP %v agree on bootcmdline: %s", len(nodes), pools[0].ObjectMeta.Name, bootcmdline)
```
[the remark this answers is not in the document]
jmencak: Not sure what you mean. Tuned CR/object doesn't really use status.

jmencak:
> I like it, just make sure the user will be visibly informed when the nodes do not agree and update is blocked. The same about the version mismatch, we should report that in the status (probably as a condition).

Let's do this as a follow up PR, this needs to get in.

PR #777, opened 2023-08-31
OCPBUGS-18392: Change the OVN trigger file name to adapt to OVN IC

MarSik:
OVN Interconnect feature was merged to CNO [1] and the manifests mounting /etc/openvswitch to the ovnkube-node containers changed.

This caused the trigger file to become invisible and so the dynamic OVS pinning was disabled.

This change moves the trigger file to the new directory used by OVN+CNO to re-enable the feature.

[1] https://github.com/openshift/cluster-network-operator/pull/1874

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -64,7 +64,7 @@ const (
 
 	ovsSliceName                     = "ovs.slice"
 	ovsDynamicPinningTriggerFile     = "ovs-enable-dynamic-cpu-affinity"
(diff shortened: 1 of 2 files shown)


zeeke:
LGTM

MarSik:
I just tested this with cluster bot and it seems to work.

@mrniranjan Do you have a way to validate pre-merge? Do not forget we still have the burstable affinity issue present.

/retest

MarSik:
/retest

@mrniranjan Have you managed to verify the change?

mrniranjan:
> /retest
> 
> @mrniranjan Have you managed to verify the change?

I managed to verify that this patch fixes the issue of ovs services  having the same cpu affinity as ovn kube node pods. I am executing all the tests, will update here with results.

PR #778, opened 2023-08-31
OCPBUGS-18052: feat: added logic to handle legacy sno install

eggfoobar:
added legacy check method to generate MCs for SNO installs using the old MC method 
updated documentation to correctly refer to feature as 4.14


eggfoobar:
/hold 

Validating fix

eggfoobar:
/unhold

Verified that the fix worked with a build from cluster bot.

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -210,6 +220,47 @@ func render(inputDir, outputDir string) error {
...
+	// If we can't determine SNO topology, we return.
+	if infra == nil {
+		return false
+	}
+
+	if infra.Status.ControlPlaneTopology != apicfgv1.SingleReplicaTopologyMode || (partitioningMode != nil && *partitioningMode == apicfgv1.CPUPartitioningAllNodes) {
+		return false
+	}
+
+	const kubernetesConfDir = "/etc/kubernetes/openshift-workload-pinning"
```
MarSik, line 243: How stable is this filename?

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -210,6 +220,47 @@ func render(inputDir, outputDir string) error {
...
+// isLegacySNOWorkloadPinningMethod provides a check for situations where the user is creating an SNO cluster with the
+// legacy method for CPU Partitioning. In order to make sure the bootstrap and running cluster MCs are synced up we check the MCs
+// provided by the user, if any one of them contain the file addition to `/etc/kubernetes/openshift-workload-pinning` and have not set
+// the API CPUPartitioningAllNodes at install time, we assume a legacy intention and alter the flag to generate the new bootstrap manifests.
+//
+// Note:
+//   - This will only trigger when Control plane is SNO or the CPU Partitioning API is NOT set to AllNodes
+//   - We do not alter the API flag here, when NTO starts up in cluster, it will notice the flag and
+//     update the flag and ignore the create error since the files already exist.
+func isLegacySNOWorkloadPinningMethod(mcs []*mcfgv1.MachineConfig, infra *apicfgv1.Infrastructure, partitioningMode *apicfgv1.CPUPartitioningMode) bool {
```
ffromani, line 232: can we add unit tests to this function?
as stretch goal, render e2e tests would be nicer, possibly using realistic manifests

ffromani (commented):
thanks @eggfoobar , makes sense to me. The only gap I see is that we should improve the testing side

pkg/performanceprofile/cmd/render/render.go
ffromani: Actually, let's think about _A_ form of e2e test coverage and let's try hard to add it.

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Unless there's a redesign in what how kubelet handles that, this file name should be rather static. It's currently hardcoded https://github.com/openshift/kubernetes/blob/ba1825544533d273d86b405195ee791e500b74c7/pkg/kubelet/managed/managed.go#L31

browsell: Stable. Path/filename is baked into kubelet

eggfoobar: Sure thing, updated with comment.

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Sounds good, adding them now

yanirq:
@eggfoobar  Thank you for the PR and the added e2e tests. 
/lgtm

PR #781, opened 2023-09-03
performanceprofile controller: requeue on intermittent errors/use correct context

akalenyu:
Just some nits I noticed going through the cgroups changes.

I think `getMachineConfigPoolByProfile` may also not requeue on errors, let me know if that is right and I can handle it,
Unless ofc that's desired


ffromani (commented):
/ok-to-test

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -448,16 +449,15 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
 		Version: "v1",
 		Kind:    "Node",
 	})
-	err = r.Client.Get(context.Background(), key, nodeCfg)
-	if err != nil {
+	if err := r.Client.Get(ctx, key, nodeCfg); err != nil {
```
ffromani, line 452: this style change (our style is not very consistent anyway) should go in its own commit

ffromani (commented):
I think this commit conflates different changes which would be better split in separate commits (even in separate PRs).

- passing the real context in production code (vs overusing `context.TODO()`) is a good change which can be merged anytime and should be split
- requieing changes look legit but they should be reviewed separately
- adding tests is fine but should (probably?) go with the requequing change

PR #784, opened 2023-09-05
OCPBUGS-18650: Memory manager E2E test fixes

mrniranjan:
Minor left over fixes of Memory manager E2E Test cases

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/memorymanager.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/memorymanager.go
@@ -1,7 +1,6 @@
 package __performance_update
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -23,6 +22,7 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/2_performance_update/memorymanager.go
```diff
@@ -105,13 +106,11 @@ var _ = Describe("[rfe_id: 43186][memorymanager] Memorymanager feature", func()
...
 				profile.Spec.RealTimeKernel = &performancev2.RealTimeKernel{
-					Enabled: pointer.BoolPtr(true),
+					Enabled: pointer.Bool(true),
 				}
 			}
-			if profile.Spec.WorkloadHints == nil {
-				profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
-					RealTime: pointer.Bool(true),
-				}
+			profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
```
ffromani, line 112: so we are always forcing the hints to this value. Is that what we really need to do? Is it a problem if we overwrite other hints?

test/e2e/performanceprofile/functests/2_performance_update/memorymanager.go
```diff
@@ -308,13 +304,11 @@ var _ = Describe("[rfe_id: 43186][memorymanager] Memorymanager feature", func()
...
```
ffromani, line 310: same as above

ffromani (commented):
/approve
/lgtm
/hold

LGTM bar a question inside

test/e2e/performanceprofile/functests/2_performance_update/memorymanager.go
mrniranjan: Right because if we set RealTime to true, then having a profile with Realtime performance hint being disabled would cause issues, So we wanted to make it default to have Realtime workload hint to be true, other workload hints are not an issue.

ffromani: Provided we restore the profile correctly if needed, and we don't pollute any global state (e.g. is the profile instance being changed unique to the testcase or shared among testcases?), fine for me.

ffromani:
/hold cancel

questions answered

PR #792, opened 2023-09-14
Improve co/node-tuning operand version reporting

jmencak:
This change improves reporting of operand version in the ClusterOperator/node-tuning object.  A total number of scheduled operands with operator's version is reported.  The change also fixes a potential brief period of false operand version reporting during upgrades.


MarSik (approved):
/lgtm

PR #794, opened 2023-09-15
Update required go version

jlojosnegros:
Some unit tests fail with go1.19 as some modules require 1.20. It seems that ci-lanes are using go1.20 already so just updating required version and Dockerfile base image to align all

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.20 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-node-tuning-operator
 COPY . .
 RUN make build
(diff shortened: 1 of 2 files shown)


ffromani:
/cc @jmencak 

I'd approve, but let's hear from Jiri first.

we may want (or likely, just _need_) to bump the build toolchain first, and only after that PR is merged, bump the requirements.

jmencak:
Thank you for the PR.  I agree with the golang bump, however I'm not sure I agree with the formatting changes.  Why, may I ask?  Also, see https://github.com/openshift/cluster-node-tuning-operator/pull/795

jlojosnegros:
> Thank you for the PR. I agree with the golang bump, however I'm not sure I agree with the formatting changes. Why, may I ask? Also, see #795

Formatting changes undone. 
Sorry IDE docker formatting tool did that automatically and I did not see it

#795 is updating OCP version but I cannot figure out the relation sorry.

jmencak:
Thank you for the changes.
/lgtm
/approve

PR #801, opened 2023-09-20
OCPVE-717: fix: add if check for no resource match error

eggfoobar:
OLM will be made optional in 4.15, when OLM resources are removed, the error check fails to be caught with `IsNotFound`  resulting in a false positive error. Updating to also check for `NoResourceMatchError`

### DIFF
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -225,13 +225,13 @@ func removePerformanceOLMOperator(cfg *rest.Config) error {
 	}
 
 	performanceOperatorCSVs, err := paocontroller.ListPerformanceOperatorCSVs(k8sclient, options, paginationLimit, performanceOperatorDeploymentName)
-	if err != nil {
+	if err != nil && !util.IsNoMatchError(err) {
 		return err
 	}
 
 	subscriptions := &olmv1alpha1.SubscriptionList{}
 	if err := k8sclient.List(context.TODO(), subscriptions); err != nil {
-		if !errors.IsNotFound(err) {
+		if !errors.IsNotFound(err) && !util.IsNoMatchError(err) {
 			return err
 		}
 	}
--- a/pkg/util/errors.go
+++ b/pkg/util/errors.go
@@ -1,0 +1,24 @@
+package util
+
+import (
+	"errors"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/discovery"
+)
+
+// IsNoMatchError checks if error is for a non existant resource, there is a meaningful difference between a
+// resource type not existing on the cluster as a whole versus an individual resource not being found.
+//
+// Example:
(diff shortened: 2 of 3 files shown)


ffromani (commented):
LGTM, but how can we test this change?

eggfoobar:
/hold

Had an error pop up in validation, I had validated this working with a small app with a release that had no OLM, however rebuilding the build with this fix seems to not resolve the error, not sure if cluster bot picked up the right SHA. Will validate tomorrow morning. 

@ffromani Unfortunately, There is no great way to run a pipeline of this working since we're in the early development of removing OLM and It requires changes in a few operators for the feature to work properly. I will run another cluster with this change in tomorrow and make sure it works as intended.

ffromani:
>@ffromani Unfortunately, There is no great way to run a pipeline of this working since we're in the early development of removing OLM and It requires changes in a few operators for the feature to work properly. I will run another cluster with this change in tomorrow and make sure it works as intended.

verification is surely welcome, but in general this flow is relatively complex and fragile (because it runs rarely and only in special cases). It would be great to have better test coverage to increase the confidence.

eggfoobar:
More tests are always better, I'm not entirely sure of the best way to test this part since I'm already poking around there I can try and take a stab at an e2e test if that's possible for and upgrade cycle. Unit test might be best here, what do you think?

Side note, the extra if checks won't change the behavior of that method so we're good there, this just adds an extra check for when the OLM goes optional.

eggfoobar:
/unhold

After some research, the issue in validation was around the error being wrapped and discovery helper func not being able to handle that scenario. I added a util helper to check. @ffromani Seems the unit test might be the best approach here, what do you think?

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -224,26 +224,24 @@ func removePerformanceOLMOperator(cfg *rest.Config) error {
...
 	if err := k8sclient.List(context.TODO(), csvs, client.MatchingLabelsSelector{Selector: csvSelector}); err != nil {
-		if !errors.IsNotFound(err) {
+		if !errors.IsNotFound(err) && !util.IsNoMatchError(err) {
 			return err
 		}
 	}
 	for i := range csvs.Items {
 		csv := &csvs.Items[i]
 		deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
-		if deploymentSpecs != nil {
```
shajmakh, line 234: was this "if" removed on purpose, if so why? if not perhaps it can be replaced with an "if dep.spec == nil" and then return with custom error.

eggfoobar: Oh yes, I removed it and left the reason in the squashed commit, it's a redundant check, the range in golang checks for nil and empty slices, so if it's ever nil or empty, the range will just not execute

eggfoobar:
@ffromani When you get a chance, this should be good now for a final review.

ffromani (commented):
/lgtm

PR #804, opened 2023-09-20
OCPBUGS-19459: check for object being nil

jmencak:
This change fixes a crash during error reporting when a returned object is nil.

Resolves: OCPBUGS-19459


kpouget:
thanks @jmencak , looks good to me.


/lgtm

PR #807, opened 2023-09-25
OCPBUGS-19785: Set non-default UserAgent for easier debugging

jmencak:
Add non-default UserAgent when communicating with the API server to ease debugging.  See OCPBUGS-19351 for more detail.

### DIFF
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -47,12 +47,10 @@ import (
 )
 
 const (
-	operandFilename  = "openshift-tuned"
-	operatorFilename = "cluster-node-tuning-operator"
-	webhookPort      = 4343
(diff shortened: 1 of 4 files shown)


jmencak:
@wking thank you for pointing out the issue with the UserAgent in OCPBUGS-19351, this should address it.

yanirq:
/lgtm
@jmencak  would it be useful to open a bug for this and downport it ? 
/hold in case other reviews are needed

jmencak:
> would it be useful to open a bug for this and downport it ?

That's probably a good idea.  Let me do that.

jmencak:
Looks like we're not getting any more reviews. Let's get this merged.
/hold cancel

jmencak:
Hypershift test succeded with this PR in the past.
/retest

PR #809, opened 2023-09-26
OCPBUGS-19791:[release-4.12] e2e: add expected max latency to hwlatde…

shajmakh:
…tec test & rename constant (#788) (#808)

* e2e: add expected max latancy to hwlatdetec test

The test started to fail frequently in CI. The reason for the failure is that there were measured latency samples that exceeds the maximal default threshold of hwlatdetec. Since we have enabled new CI machines, it seems that they are not well tuned hence causing in high latency.

To fix the test, just add a high threshold. Note that in these tests we do not care about latency measurments but about enabling the latency tools tests with differnt variables.

* latency: rename guaranteedLatency

Rename guaranteedLatency to untunedLatencyThreshold and point out that this constant is used for exercising the latency tools and when we set it as the expected latency we mean that we want the test to pass rather than fail on nonmeaningful measured latency in an untuned system, which is pointless.

* e2e latency: set untunedLatencyThreshold to 10s

untunedLatencyThreshold as 900000 is enough at the meantime to have all the tests passed in the suite. However, it doesn't hurt to have even more high value like 10s to avoid any future flakiness.

* makefile: set ginkgo timeout on latency testing

The 5_latency suite fails occasionally on the AfterSuite due to suite timeout. The default timeout of gingko test is 1 hour and when the suite passes one hour it fails. To change that we pass the parameter --timeout=5h.

---------

### DIFF
--- a/hack/run-latency-testing.sh
+++ b/hack/run-latency-testing.sh
@@ -18,8 +18,7 @@ fi
 echo "Running Functional Tests: ${GINKGO_SUITS}"
 # -v: print out the text and location for each spec before running it and flush output to stdout in realtime
 # -r: run suites recursively
-# --failFast: ginkgo will stop the suite right after the first spec failure
-# --flakeAttempts: rerun the test if it fails
-# -requireSuite: fail if tests are not executed because of missing suite
-GOFLAGS=-mod=vendor ginkgo $NO_COLOR --v -r -requireSuite ${GINKGO_SUITS} -- -junitDir /tmp/artifacts
-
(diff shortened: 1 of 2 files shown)


shajmakh:
manual cherry-pick of #808

shajmakh:
cc @swatisehgal @yanirq 
This is a manual cherrypick of #808, can you have a look and provide missing labels please? Thanks!

shajmakh:
we will need this back to 4.11
/cherry-pick release-4.11

PR #810, opened 2023-09-26
OCPBUGS-19795: Add kubeconfig path for IBM Managed OpenShift 

jmencak:
IBM Managed OpenShift seems to use `/etc/kubernetes/kubelet-kubeconfig` instead of the traditional `/var/lib/kubelet/kubeconfig` kubelet config path.

Other changes:
  - fix a crash when kubelet's kubeconfig is not found and in-cluster config is used instead

Resolves: OCPBUGS-19795


dagrayvid:
Thanks @jmencak!

/lgtm

PR #816, opened 2023-09-27
OCPBUGS-18662:rps: trigger udev even per queue

Tal-or:
Instead of trigger udev event per physical device creation,
triggers it per queue creation so in case of queue resize
(i.e. more queues are added) it applies the correct rps mask on the
new created queues as well.

### DIFF
--- a/assets/performanceprofile/configs/99-netdev-physical-rps.rules
+++ b/assets/performanceprofile/configs/99-netdev-physical-rps.rules
@@ -1,1 +1,1 @@
-SUBSYSTEM=="net", ACTION=="add", ENV{DEVPATH}!="/devices/virtual/net/*", TAG+="systemd", ENV{SYSTEMD_WANTS}="update-rps@%k.service"
+SUBSYSTEM=="queues", ACTION=="add", ENV{DEVPATH}=="/devices/pci*/queues/rx*", TAG+="systemd", ENV{SYSTEMD_WANTS}="update-rps@%p.service"
--- a/assets/performanceprofile/scripts/set-rps-mask.sh
+++ b/assets/performanceprofile/scripts/set-rps-mask.sh
@@ -1,38 +1,18 @@
 #!/usr/bin/env bash
 
-dev=$1
-[ -n "${dev}" ] || { echo "The device argument is missing" >&2 ; exit 1; }
+full_path=${1}
+[ -n "${full_path}" ] || { echo "The full device path argument is missing" >&2 ; exit 1; }
 
-mask=$2
+mask=${2}
 [ -n "${mask}" ] || { echo "The mask argument is missing" >&2 ; exit 1; }
 
-dev_dir="/sys/class/net/${dev}"
+# replace '-' with '/'
+de_escape_path="/sys${full_path//-//}"
 
-function find_dev_dir {
-  systemd_devs=$(systemctl list-units -t device | grep sys-subsystem-net-devices | cut -d' ' -f1)
+# get the path for the queues
+queues_path=${de_escape_path%/rx*}
 
-  for systemd_dev in ${systemd_devs}; do
-    dev_sysfs=$(systemctl show "${systemd_dev}" -p SysFSPath --value)
-
-    dev_orig_name="${dev_sysfs##*/}"
-    if [ "${dev_orig_name}" = "${dev}" ]; then
-      dev_name="${systemd_dev##*-}"
-      dev_name="${dev_name%%.device}"
-      if [ "${dev_name}" = "${dev}" ]; then # disregard the original device unit
-              continue
-      fi
-
-      echo "${dev} device was renamed to $dev_name"
-      dev_dir="/sys/class/net/${dev_name}"
-      break
-    fi
-  done
-}
-
-[ -d "${dev_dir}" ] || find_dev_dir                # the net device was renamed, find the new name
-[ -d "${dev_dir}" ] || { sleep 5; find_dev_dir; }  # search failed, wait a little and try again
-[ -d "${dev_dir}" ] || { echo "${dev_dir}" directory not found >&2 ; exit 0; } # the interface disappeared, not an error
-
-for i in "${dev_dir}"/queues/rx-*/rps_cpus; do
+# set rps affinity for all queues
+for i in "${queues_path}"/rx-*/rps_cpus; do
   echo "${mask}" > "${i}"
 done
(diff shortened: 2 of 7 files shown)


Tal-or:
/hold
concerns were raised about performance degradation in case this script would against many queues in short period of time (especially at boot time)
/cc @yanirq @MarSik @pabeni

assets/performanceprofile/configs/99-netdev-physical-rps.rules
```diff
@@ -1 +1 @@
-SUBSYSTEM=="net", ACTION=="add", ENV{DEVPATH}!="/devices/virtual/net/*", TAG+="systemd", ENV{SYSTEMD_WANTS}="update-rps@%k.service"
+SUBSYSTEM=="queues", ACTION=="add", ENV{DEVPATH}=="/devices/pci*/queues/rx*", TAG+="systemd", ENV{SYSTEMD_WANTS}="update-rps@%p.service"
```
pabeni, line 1: Here the devpath for patching events is almost full sysfs queue directory - just missing the '/sys' prefix...

assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -1,38 +1,18 @@
...
 [ -n "${mask}" ] || { echo "The mask argument is missing" >&2 ; exit 1; }
 
-dev_dir="/sys/class/net/${dev}"
+# replace '-' with '/'
+de_escape_path="/sys${full_path//-//}"
 
-function find_dev_dir {
-  systemd_devs=$(systemctl list-units -t device | grep sys-subsystem-net-devices | cut -d' ' -f1)
+# get the path for the queues
+queues_path=${de_escape_path%/rx*}
```
pabeni, line 13: ... so here you should be better off not trimming the trailing part and avoiding the loop below, with something alike:

queue_path=${de_escape_path//queues\/rx\//queues\/rx-}
echo "${mask}" > "${queue_path}"

Side question: why is this script invoked via a systemd unit and not directly by the udev rule?

pabeni: Additionally, is devpath escaping really needed here? Otherwise you can replace '%i' with '%I' in the unit file and avoid the de-escaping above.

assets/performanceprofile/configs/99-netdev-physical-rps.rules
Tal-or: AFAIK the path should start from `/devices` and not from `/sys` this is how it worked so far.

pabeni: @Tal-or: I was not clear in the sentence above, I'm sorry. What I mean is that since the devpath for patching events is almost full sysfs queue directory, we can exploit such fact to simplify the action, as hopefully better described in my comment on set-rps-mask.sh.
The above sentence and the comment there are supposed to be read one after another as part of a the same message

assets/performanceprofile/scripts/set-rps-mask.sh
Tal-or: > why is this script invoked via a systemd unit and not directly by the udev rule?

I think the reason is because then we can use it as `oneshot` service but maybe @MarSik knows better

Tal-or: The loop was just to make sure we're on the safe side and nothing get missed, but yes removing the loop should be suffice as well

test/e2e/performanceprofile/testdata/render-expected-output/manual_machineconfig.yaml
```diff
@@ -0,0 +1,193 @@
...
+        group: {}
+        mode: 448
+        path: /usr/local/bin/hugepages-allocation.sh
+        user: {}
+      - contents:
+          source: data:text/plain;charset=utf-8;base64,IyEvdXNyL2Jpbi9lbnYgYmFzaAoKZnVsbF9wYXRoPSR7MX0KWyAtbiAiJHtmdWxsX3BhdGh9IiBdIHx8IHsgZWNobyAiVGhlIGZ1bGwgZGV2aWNlIHBhdGggYXJndW1lbnQgaXMgbWlzc2luZyIgPiYyIDsgZXhpdCAxOyB9CgptYXNrPSR7Mn0KWyAtbiAiJHttYXNrfSIgXSB8fCB7IGVjaG8gIlRoZSBtYXNrIGFyZ3VtZW50IGlzIG1pc3NpbmciID4mMiA7IGV4aXQgMTsgfQoKIyByZXBsYWNlICctJyB3aXRoICcvJwpkZV9lc2NhcGVfcGF0aD0iL3N5cyR7ZnVsbF9wYXRoLy8tLy99IgoKIyBnZXQgdGhlIHBhdGggZm9yIHRoZSBxdWV1ZXMKcXVldWVzX3BhdGg9JHtkZV9lc2NhcGVfcGF0aCUvcngqfQoKIyBzZXQgcnBzIGFmZmluaXR5IGZvciBhbGwgcXVldWVzCmZvciBpIGluICIke3F1ZXVlc19wYXRofSIvcngtKi9ycHNfY3B1czsgZG8KICBlY2hvICIke21hc2t9IiA+ICIke2l9Igpkb25lCg==
+          verification: {}
+        group: {}
+        mode: 448
+        path: /usr/local/bin/set-rps-mask.sh
```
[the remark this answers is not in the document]
Tal-or: Yea, we have a make target named `redner-sync` for that, but it got break.
https://github.com/openshift/cluster-node-tuning-operator/pull/824 should fix it, but this PR not depends on it since I fixed it manually here.

assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -1,18 +1,16 @@
...
 
-# replace '-' with '/'
-de_escape_path="/sys${full_path//-//}"
+queue_path=${path%rx*}
 
-# get the path for the queues
-queues_path=${de_escape_path%/rx*}
+queue_num=${path#*queues/}
+# replace '/' with '-'
+queue_num="${queue_num/\//-}"
```
pabeni, line 13: minor nit: is this substitution required? If I read correctly $path is not escaped anymore and this is basically de-escaping the queue nr part

pabeni:
LGTM! some minor comments on patch 3, but really not blocking

yanirq:
/approve
@MarSik  PTAL for a final LGTM

MarSik (approved):
/lgtm

PR #819, opened 2023-10-05
NO-JIRA: PerformanceProfileController: Add relatedObjects status field

jlojosnegros:
Add relatedObjects support to better integrate with must-gather

relatedObjects is a list of objects that are "interesting" or related to this operator. 
must-gather uses this field to collect data from these object when data from the operator is collected.

### DIFF
--- a/manifests/20-performance-profile.crd.yaml
+++ b/manifests/20-performance-profile.crd.yaml
@@ -760,6 +760,32 @@ spec:
                   - type
                   type: object
                 type: array
+              relatedObjects:
+                description: 'relatedObjects is a list of objects that are "interesting"
+                  or related to this operator.  Common uses are: 1. the detailed resource
+                  driving the operator 2. operator namespaces 3. operand namespaces'
+                items:
+                  description: ObjectReference contains enough information to let
+                    you inspect or modify the referred object.
+                  properties:
+                    group:
+                      description: group of the referent.
+                      type: string
+                    name:
+                      description: name of the referent.
+                      type: string
+                    namespace:
+                      description: namespace of the referent.
+                      type: string
+                    resource:
+                      description: resource of the referent.
+                      type: string
+                  required:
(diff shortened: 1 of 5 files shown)


ffromani (commented):
/approve

I think we want this change

yanirq:
Looks good to me .  I personally would like 1 squashed commit for this PR with general description of the change (some can be taken from the comments in the code itself)

jlojosnegros:
> Looks good to me . I personally would like 1 squashed commit for this PR with general description of the change (some can be taken from the comments in the code itself)

Description changed and commits squashed

jlojosnegros:
/test e2e-gcp-pao
test timeout, not likely cause by this change

jmencak:
Thank you for the PR. It's been a while since I played with this. I believe we want this, but aren't all the object collected here already in must-gather? I'd say that at least Tuned CR and MC already are collected. Tuned CRs by NTO and MC's probably by MCO.  So I'd say my question is whether the aim of this PR is to be self-reliant and collect all objects PPC is creating?

jlojosnegros:
> Thank you for the PR. It's been a while since I played with this. I believe we want this, but aren't all the object collected here already in must-gather? I'd say that at least Tuned CR and MC already are collected. Tuned CRs by NTO and MC's probably by MCO. So I'd say my question is whether the aim of this PR is to be self-reliant and collect all objects PPC is creating?

I would say the aim for the PR is to be a "good citizen" and enhance observability.
As I see it `relatedObjects` is a way to explicitly declare which objects would be of interest to look at in case you want to get a better understanding of the `PerformanceProfile`. 

It is true that all the objects ( or almost all, not sure about `RuntimeClass`)  are already collected by must-gather cause `must-gather` itself does that, but with this we aim to ensure that all the objects are collected by `must-gather` and be "friendly" with any other observability tool.

PR #822, opened 2023-10-11
Makefile: Change `golangci-lint` cache folder

jlojosnegros:
golangci-lint uses a cache to improve performance. This cache is stored in the subdirectory `golangci-lint` inside the default user cache directory which, is under the HOME directory on most OSs.
This have two problems:
- Pollutes the user environment
- Makes this makefile target unusable for CI as it tries to write on a folder where it has no permissions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,8 @@ PAO_E2E_SUITES := $(shell hack/list-test-bin.sh)
 GOLANGCI_LINT_VERSION=1.54.2
 GOLANGCI_LINT_BIN=$(OUT_DIR)/golangci-lint
 GOLANGCI_LINT_VERSION_TAG=v${GOLANGCI_LINT_VERSION}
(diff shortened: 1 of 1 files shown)


jlojosnegros:
/cc @jmencak 
found a problem with `golangci-lint` cache while running rehearsal job

```
2023/10/11 08:52:11 failed to initialize build cache at /.cache/golangci-lint: mkdir /.cache: permission denied
```

jmencak:
Cannot we just pass HOME directory to e.g. /tmp in the CI?

jlojosnegros:
> Cannot we just pass HOME directory to e.g. /tmp in the CI?

Yes of course, I just thought it would be nice to not store things in the user home folder

jmencak:
> I just thought it would be nice to not store things in the user home folder

It depends, that's what the cache is for, isn't it?  To speed up things?

PR #823, opened 2023-10-11
Makefile: remote tmp folder on clean target

jlojosnegros:
Makefile target `pkg/generated` output some data on `tmp` folder. 
Lets delete it also on clean.


jmencak:
This looks good, however it won't help/work, unfortunately.  The clean target used to work at the beginning before we were forced to use stuff from `build-machinery-go`.  I'd say the Makefile definitely needs some cleanup.  So,
/lgtm
/approve
but as I say, this won't help during `make clean`.

PR #824, opened 2023-10-12
render: change directory path

Tal-or:
When https://github.com/openshift/cluster-node-tuning-operator/pull/778 got merged, the directory layout has changed and broke the `render-sync`  make target. This patch should fix it.


ffromani:
/approve
/lgtm

please backport if/as needed

PR #825, opened 2023-10-13
OCPBUGS-18640: Fix Racing Machine Configs and add Day 0 Support

rphillips:
This PR fixes a bug where two machine configs can race and sometimes we see incorrect kernel arguments on a Node.

Specifically, there is a problem with the controller updating the Node object to use Cgroupv1 and the controller writing the performance profile machine config. The patch addresses Day-0 and Day-2 operations.

## Day 0

Day 0 did not apply the Node object at bootstrap, so MCO configures the machine with Cgroupv1 at boot. The fix is to lay down the Node object to force cgroupv1 at bootstrap. MCO will pick up the Node object and configure the nodes with Cgroupv1 from the post-pivot boot.

## Day 2

The controller would originally update 2 MachineConfigs at the same time: 1) update the Node object (to switch to cgroupv1), and 2) create the perf profile machine config. These two configs are racing in certain scenarios; sometimes providing the node with duplicate (or wrong) kernel arguments.

The fix is to update the Node config to switch a cluster to Cgroupv1 and then watch for the Machine Config Pool generation to increment. The generation change signals that a new rendered-config has been created, and the NTO can then write the perfprofile machine config.

## Summary
Day 0 is the best approach and minimizes the amount of reboots necessary for PerformanceProfiles, since it does not require any extra reboots.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile_controller.go
+++ b/pkg/performanceprofile/controller/performanceprofile_controller.go
@@ -485,32 +485,6 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
 		if err != nil {
 			return reconcile.Result{}, err
 		}
-
-		klog.Info("waiting for cgroupv1 rollout")
-
-		// wait for Cgroupv1 roll out
-		err = wait.PollUntilContextCancel(context.Background(), 7*time.Second, true, func(ctx context.Context) (done bool, err error) {
-			profileMCP, err := r.getMachineConfigPoolByProfile(ctx, instance)
-			if err != nil {
-				return true, err
-			}
-			// check to see if all machines are updated
-			if profileMCP.Status.MachineCount == profileMCP.Status.ReadyMachineCount {
-				klog.Infof("MachineConfigPool %v and Node config has rolled out", profileMCP.Name)
-				return true, nil
-			}
-			// check to see if all machines are updated
-			if profileMCP.Status.DegradedMachineCount > 0 {
-				return true, fmt.Errorf("machines have become degraded")
-			}
-			klog.Infof("MachineConfigPool %v and Node config is still rolling out (%v/%v)", profileMCP.Name, profileMCP.Status.ReadyMachineCount, profileMCP.Status.MachineCount)
-			return false, nil
-		})
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-
-		klog.Infof("Cluster is using cgroupv1")
 	}
 
 	profileMCP, err := r.getMachineConfigPoolByProfile(ctx, instance)
--- a/pkg/performanceprofile/controller/performanceprofile_controller.go
+++ b/pkg/performanceprofile/controller/performanceprofile_controller.go
@@ -454,6 +454,8 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
 	}
 
 	if nodeCfg.Spec.CgroupMode != apiconfigv1.CgroupModeV1 {
+		klog.Infof("Switching cluster to cgroupv1")
+
 		// gather the MCP generation
 		profileMCP, err := r.getMachineConfigPoolByProfile(ctx, instance)
 		if err != nil {
@@ -474,12 +476,14 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
 			}
 			// check to see if machine rollout has happened
 			if startGeneration != profileMCP.Generation {
+				klog.Infof("MachineConfigPool %v has generation %v (startGeneration=%v)", profileMCP.Name, profileMCP.Generation, startGeneration)
 				return true, nil
 			}
 			// check to see if all machines are updating
 			if profileMCP.Status.DegradedMachineCount > 0 {
 				return true, fmt.Errorf("machines have become degraded")
 			}
+			klog.Infof("MachineConfigPool %v has not changed generations", profileMCP.Name)
 			return false, nil
 		})
 		if err != nil {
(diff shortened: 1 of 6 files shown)


Tal-or:
@rphillips thank you for the PR.
Please run `make render-sync` and commit the changes, this should help the `e2e-no-cluster` lane to pass.
Could you please elaborate in the PR description for what this change is needed and why?

rphillips:
@Tal-or @MarSik : Could you review?

ffromani:
the day-0 fix looks straightforwarded to me. It seems it should (brilliantly) sidestep the apiserver failure we're getting.
The day-2 fix makes me fuzzy about introducing an explicit wait during the reconcile pass. I wonder if there's a different approach here.

rphillips:
@ffromani It's possible we could just wait for the MCP generation change, and then proceed with the rest of function.

rphillips:
@ffromani I took out the blocking cgroupv1 rollout. The cluster successfully rolls out without it.

test/e2e/performanceprofile/testdata/render-expected-output/default/manual_node.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: config.openshift.io/v1
+kind: Node
+metadata:
```
MarSik, line 3: Will the MCO bootstrap be happy about the ownerReference? What if we remove the performance profile later? Will it delete the config as well?

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -449,16 +451,40 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		if err := r.Client.Update(ctx, nodeCfg); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		err = wait.PollUntilContextCancel(context.Background(), 7*time.Second, true, func(ctx context.Context) (done bool, err error) {
+			profileMCP, err := r.getMachineConfigPoolByProfile(ctx, instance)
+			if err != nil {
+				return true, err
+			}
+			klog.Infof("watching for generation change (startGeneration=%v) (mcpGeneration=%v)", startGeneration, profileMCP.Generation)
```
MarSik, line 476: Is the generation enough? I can actually still see a race here. The generation is set by the API server iirc, but MCO might not notice that immediately. What we should be watching is generation change + the observedGeneration if the MCP supports that. That way we can be sure MCO already noticed the update.

test/e2e/performanceprofile/testdata/render-expected-output/default/manual_node.yaml
rphillips: This is the test file, but the MCO is happy with the Node object. I would suggest not changing from cgroupv1 to cgroupv2 automatically since workloads may not be compatible with cgroupv2.

pkg/performanceprofile/controller/performanceprofile_controller.go
rphillips: Generation is enough, because the MCO is writing the rendered MachineConfig to the API, and has written to the MCP pool (causing the generation change). This code is watching on the pool generation change. MCO writes an update to the MCP to the API which causes this generation change.

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -449,16 +451,40 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-		klog.Errorf("failed to get config node object; name=%q err=%v", nodeCfg.GetName(), err)
-		nodeCfg.Name = nodeCfgName
-		nodeCfg.Spec.CgroupMode = apiconfigv1.CgroupModeV1
-		//nolint:errcheck
-		r.Client.Update(ctx, nodeCfg)
+		return reconcile.Result{}, err
 	}
+
 	if nodeCfg.Spec.CgroupMode != apiconfigv1.CgroupModeV1 {
+		klog.Infof("Switching cluster to cgroupv1")
```
yanirq, line 458: 4.15 wise,  we might need to perform the switch here back to cgroupv2 (after this PR is merged and cherry picked)

rphillips: Is 4.15 targeting cgroupv2 for NTO?

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -449,16 +450,61 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			}
+			// check to see if all machines are updated
+			if profileMCP.Status.DegradedMachineCount > 0 {
+				return true, fmt.Errorf("machines have become degraded")
+			}
+			return false, nil
+		})
+		if err != nil {
+			return reconcile.Result{}, err
+		}
```
ffromani, line 487: ok, reviewing again this is meant to wait at most 7+7 seconds, right?
if so, as it looks to me, it could be OK for the short term because a better fix would require more invasive changes, which we should plan and implement later (and backport).
The only missing thing is: how was 7 seconds determined and how it (if it all) interacts with the controller reconcile loop? e.g. there is any relationship between the two or not?

@MarSik @yanirq @jmencak any concern about accepting this wait as short-term solution, provided a wait-less solution is implemented as followup?

ffromani: a later commit removes the second wait, even better!

MarSik: Hmm, we are basically only waiting until MCP notices the cgroups change, right? That might be an acceptable evil here.

What happens when the 7+7 seconds pass? We return a reconcile error and try again later?

rphillips: The wait loop retries every 7 seconds to look for a MCP generation change.

MarSik: @ffromani I was thinking about the status solution. Can't we just return when the MCP is not ready here and rely on the watch that we already have (we watch MCP, right?) to rerun the Reconcile method once something changes?

ffromani:
experimenting with wait removal in #854

PR #827, opened 2023-10-15
OCPBUGS-20567: Remove obsolete protocols and weak ciphers

jmencak:
NTO metrics server was using the default tls.Config.  However, the default tls.Config supports obsolete/insecure protocols, such as TLS 1.0 and 1.1 and weak ciphers (3DES).  Drop support for TLS 1.0, 1.1 and 3DES. Also, drop CBC encryption method and RSA key exchanges.

Resolves: OCPBUGS-20567


jmencak:
/hold
possibly remove CBC too

jmencak:
/hold cancel
Based on the discussion with the reporter, removed also CBC and RSA key exchanges.  I've also verified the metrics are still successfully reported/scraped on the cluster.

jlojosnegros (approved):
/lgtm

PR #828, opened 2023-10-16
Tao mod

vitus133:
Draft for Tuned performance rendering

### DIFF
--- a/pkg/performanceprofile/cmd/render/render.go
+++ b/pkg/performanceprofile/cmd/render/render.go
@@ -26,12 +26,13 @@ import (
 
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	apicfgv1 "github.com/openshift/api/config/v1"
-	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	tunev1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/machineconfig"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/manifestset"
-
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -55,14 +56,18 @@ var (
 )
 
 func init() {
+	utilruntime.Must(corev1.AddToScheme(manifestScheme))
 	utilruntime.Must(performancev2.AddToScheme(manifestScheme))
 	utilruntime.Must(apicfgv1.Install(manifestScheme))
 	utilruntime.Must(mcfgv1.Install(manifestScheme))
+	utilruntime.Must(tunev1.AddToScheme(manifestScheme))
 	codecFactory = serializer.NewCodecFactory(manifestScheme)
 	runtimeDecoder = codecFactory.UniversalDecoder(
 		performancev2.GroupVersion,
 		apicfgv1.GroupVersion,
 		mcfgv1.GroupVersion,
+		corev1.SchemeGroupVersion,
+		tunev1.SchemeGroupVersion,
 	)
 }
 
--- a/pkg/operator/mc.go
+++ b/pkg/operator/mc.go
@@ -19,6 +19,11 @@ const (
 	MachineConfigPrefix                       string = "50-nto"
 )
 
+// NewMachineConfig exports newMachineConfig for offline rendering
+func NewMachineConfig(name string, annotations map[string]string, labels map[string]string, kernelArguments []string) *mcfgv1.MachineConfig {
+	return newMachineConfig(name, annotations, labels, kernelArguments)
+}
+
 func newMachineConfig(name string, annotations map[string]string, labels map[string]string, kernelArguments []string) *mcfgv1.MachineConfig {
 	if labels == nil {
 		labels = map[string]string{}
--- a/pkg/performanceprofile/cmd/render/cmd.go
+++ b/pkg/performanceprofile/cmd/render/cmd.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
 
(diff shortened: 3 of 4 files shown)


pkg/performanceprofile/cmd/render/render.go
```diff
@@ -58,14 +65,107 @@ func init() {
...
 	codecFactory = serializer.NewCodecFactory(manifestScheme)
 	runtimeDecoder = codecFactory.UniversalDecoder(
 		performancev2.GroupVersion,
 		apicfgv1.GroupVersion,
 		mcfgv1.GroupVersion,
+		tunev1.SchemeGroupVersion,
 	)
 }
 
+func expandCpuSet(cpuSet string) string {
```
ffromani, line 78: it's a draft, but please let's use the `cpuset` package (or one of its existing variants)

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -58,14 +65,107 @@ func init() {
...
+
+		for i := start; i <= end; i++ {
+			expandedCpus = append(expandedCpus, strconv.Itoa(i))
+		}
+	}
+
+	return strings.Join(expandedCpus, ",")
+}
+
+func coresToHexMask(coresExp string) string {
```
ffromani, line 99: likewise, I think we have existing utilities to cover this usecase

pkg/performanceprofile/cmd/render/render.go
vitus133: Thanks for the comment, fixed

pkg/performanceprofile/cmd/render/render.go
vitus133: Fixed

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -143,22 +209,102 @@ func render(inputDir, outputDir string) error {
...
-	for _, pool := range mcPools {
-		if err := genBootstrapWorkloadPinningManifests(partitioningMode, outputDir, pool.Name); err != nil {
+		if err := genBootstrapWorkloadPinningManifests(partitioningMode, outputDir, defaultMCPNames...); err != nil {
 			return err
 		}
-	}
 
+		// If the user supplies extra machine pools, we ingest them here
+		for _, pool := range mcPools {
+			if err := genBootstrapWorkloadPinningManifests(partitioningMode, outputDir, pool.Name); err != nil {
```
jlojosnegros, line 227: not really sure why this is needed.
Would not be enough to handle kernel arguments?

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -143,22 +209,102 @@ func render(inputDir, outputDir string) error {
...
+				continue
+			}
+
+			mainSection, err := cfg.GetSection("main")
+			if err != err {
+				klog.Infof("No main section in tune data (%s)", tunedData)
+				continue
+			}
+
+			profileName := mainSection.Key("include").String()
```
jlojosnegros, line 260: do we need to read included tuned profiles? I thought `tuned` would do that.

pkg/performanceprofile/cmd/render/render.go
vitus133: Right, other mcps are not needed. Remained from another try...
Will remove

pkg/performanceprofile/cmd/render/render.go
vitus133: The idea is to render performanceProfile and Tuned patch (both part of the DU profile). FOr example:
```yaml
kind: Tuned
metadata:
  name: performance-patch
spec:
  profile:
  - data: |
      [main]
      summary=Configuration changes profile inherited from performance created tuned
      include=openshift-node-performance-openshift-node-performance-profile
      [bootloader]
      cmdline_crash=nohz_full=4-47
      [sysctl]
      kernel.timer_migration=1
      [scheduler]
      group.ice-ptp=0:f:10:*:ice-ptp.*
      [service]
      service.stalld=start,enable
      service.chronyd=stop,disable
    name: performance-patch
  recommend:
  - machineConfigLabels:
      machineconfiguration.openshift.io/role: master
    priority: 19
    profile: performance-patch
---
kind: PerformanceProfile
metadata:
  name: openshift-node-performance-profile
spec:
  additionalKernelArgs:
  - rcupdate.rcu_normal_after_boot=0
  - efi=runtime
  - vfio_pci.enable_sriov=1
  - vfio_pci.disable_idle_d3=1
  - module_blacklist=irdma
  cpu:
    isolated: 4-47
    reserved: 0-3
  hugepages:
    defaultHugepagesSize: 1G
    pages:
    - count: 32
      size: 1G
  machineConfigPoolSelector:
    pools.operator.machineconfiguration.openshift.io/master: ""
  nodeSelector:
  numa:
    topologyPolicy: restricted
  realTimeKernel:
    enabled: true
  workloadHints:
    highPowerConsumption: false
    perPodPowerManagement: false
    realTime: true

```
I'm not sure how to do that with `Tuned` - do we need to extract this patch to some file before invoking the `Tuned`?

jlojosnegros: In my understanding the target is to create a MachineConfig to apply kernerl boot arguments.
To get those arguments we need to run tuned with the `tuned.conf` file generated by the PerformanceProfile and then read the `/etc/tuned/bootcmdline` file.

So, as far as I see it, we could use `PerformanceProfile` as "main" input and use [NewNodePerformance function](https://github.com/openshift/cluster-node-tuning-operator/blob/12b9a12c1cf6892b36464c26d6198d2211f28a14/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go#L55) to get the `tuned` from there, as the PerformanceProfile controller does.

Other inputs should be the MachineConfigPools for the SNO ( we already have them cause they are used in the previous step of the installer, the one for the PerformanceProfile render command [here](https://github.com/openshift/installer/blob/b36d7c201e0114aa0e9261cf8ae5b8d68c74bf07/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L297)

All those inputs will be in one folder (INPUT_DIR)

With all that in place `tuned` will be run and will generate the `/etc/tuned/bootcmdline` file that could be read using [getBootCmdline function](https://github.com/openshift/cluster-node-tuning-operator/blob/12b9a12c1cf6892b36464c26d6198d2211f28a14/pkg/tuned/controller.go#L748)
Here the challenge is to mount all the system folders `tuned` needs to work properly from the host machine but without letting `tuned` to modify them.  I am working with overlayfs and `podman-run` to get that.

With all that in place we just need to create the MachineConfig object and write the manifest to a folder which should be mounted from the host machine so it can be read after the container finish its execution.

This is just my understanding of the task, so maybe @MarSik or @yanirq can correct me.

A (WIP) implementation of this is https://github.com/openshift/cluster-node-tuning-operator/pull/844

MarSik: You probably need to read all extra Tuned objects too. Customers can use those to override pieces of our default one (https://access.redhat.com/solutions/5532341).

But the rest sounds right to me:

Prepare the (RHEL level) tuned profiles, give them to tuned, run the oneshot mode (in isolation somehow), collect the boot args, generate MC compatible with what NTO would do at Day 2.

vitus133:
not relevant anymore

PR #829, opened 2023-10-16
OCPBUGS-13330: Use RHEL9 as a base

jmencak:
`x86_energy_perf_policy` from RHEL8 causes a kernel taint on RHEL9 kernels.  This is a problem for CNF certifications.  Switch the NTO base to RHEL9 so that we use an up-to-date `x86_energy_perf_policy` aligned with RHEL9 kernels.

Resolves: OCPBUGS-13330

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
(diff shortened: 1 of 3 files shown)


jlojosnegros (approved):
/lgtm

PR #833, opened 2023-10-18
OCPBUGS-22095: Add default MCP objects for rendering

eggfoobar:
added ability to use the default rmaster worker mcp labels for rendering during boostrap

/hold

### DIFF
--- a/pkg/performanceprofile/cmd/render/render.go
+++ b/pkg/performanceprofile/cmd/render/render.go
@@ -161,6 +161,9 @@ func render(inputDir, outputDir string) error {
 		}
 	}
 
+	// Append any missing default manifests (i.e. `master`/`worker`)
+	mcPools = util.AppendMissingDefaultMCPManifests(mcPools)
+
 	for _, pp := range perfProfiles {
 		mcp, err := selectMachineConfigPool(mcPools, pp.Spec.NodeSelector)
 		if err != nil {
--- a/pkg/util/manifests.go
+++ b/pkg/util/manifests.go
@@ -24,6 +24,9 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
 )
 
@@ -96,4 +99,69 @@ func ListFilesFromMultiplePaths(dirPaths []string) ([]string, error) {
 		}
 	}
 	return results, nil
+}
+
+// AppendMissingDefaultMCPManifests When default MCPs are missing, it is desirable to still generate the relevant
+// files based off of the standard MCP labels and node selectors.
+//
+// Here we create the default `master` and `worker` MCP if they are missing with their respective base Labels and NodeSelector Labels,
+// this allows any resource such as PAO to utilize the default during bootstrap rendering.
+func AppendMissingDefaultMCPManifests(currentMCPs []*mcfgv1.MachineConfigPool) []*mcfgv1.MachineConfigPool {
+	const (
+		master             = "master"
+		worker             = "worker"
+		labelPrefix        = "pools.operator.machineconfiguration.openshift.io/"
+		masterLabels       = labelPrefix + master
+		workerLabels       = labelPrefix + worker
+		masterNodeSelector = components.NodeRoleLabelPrefix + master
+		workerNodeSelector = components.NodeRoleLabelPrefix + worker
+	)
+	var (
+		finalMCPList = []*mcfgv1.MachineConfigPool{}
+		defaultMCPs  = []*mcfgv1.MachineConfigPool{
+			{
+				ObjectMeta: v1.ObjectMeta{
+					Labels: map[string]string{
+						masterLabels: "",
+					},
+					Name: master,
+				},
+				Spec: mcfgv1.MachineConfigPoolSpec{
+					NodeSelector: v1.AddLabelToSelector(&v1.LabelSelector{}, masterNodeSelector, ""),
+				},
+			},
+			{
+				ObjectMeta: v1.ObjectMeta{
+					Labels: map[string]string{
+						workerLabels: "",
+					},
+					Name: worker,
+				},
+				Spec: mcfgv1.MachineConfigPoolSpec{
+					NodeSelector: v1.AddLabelToSelector(&v1.LabelSelector{}, workerNodeSelector, ""),
+				},
+			},
+		}
+	)
+
+	if len(currentMCPs) == 0 {
+		return defaultMCPs
+	}
+
+	for _, defaultMCP := range defaultMCPs {
+		missing := true
+		for _, mcp := range currentMCPs {
+			// Since users can supply MCP files and these will be raw files with out going through the API
+			// server validation, we normalize the file name so as to not mask any configuration errors.
+			if strings.ToLower(mcp.Name) == defaultMCP.Name {
+				missing = false
+				break
+			}
+		}
+		if missing {
+			finalMCPList = append(finalMCPList, defaultMCP)
+		}
+	}
+
+	return append(finalMCPList, currentMCPs...)
 }
--- a/pkg/tuned/cmd/render/render.go
+++ b/pkg/tuned/cmd/render/render.go
@@ -129,6 +129,9 @@ func render(inputDir []string, outputDir string, mcpName string) error {
 		}
 	}
 
+	// Append any missing default manifests (i.e. `master`/`worker`)
+	mcPools = util.AppendMissingDefaultMCPManifests(mcPools)
+
 	mcp := findMachineConfigPoolByName(mcPools, mcpName)
 	if mcp == nil {
 		klog.Errorf("Unable to find MachineConfigPool:%q in input folders", mcpName)
--- a/test/e2e/performanceprofile/cluster-setup/bootstrap-cluster/extra-mcp/machine_config_pool.yaml
+++ b/test/e2e/performanceprofile/cluster-setup/bootstrap-cluster/extra-mcp/machine_config_pool.yaml
@@ -1,0 +1,17 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfigPool
+metadata:
+  name: worker-cnf
+  namespace: openshift-machine-config-operator
+  labels:
+    machineconfiguration.openshift.io/role: worker-cnf
+spec:
+  paused: true
+  machineConfigSelector:
+    matchExpressions:
+      - key: machineconfiguration.openshift.io/role
+        operator: In
+        values: [worker,worker-cnf]
+  nodeSelector:
+    matchLabels:
+      node-role.kubernetes.io/worker-cnf: ""
--- a/test/e2e/performanceprofile/cluster-setup/bootstrap-cluster/performance/pao_master.yaml
+++ b/test/e2e/performanceprofile/cluster-setup/bootstrap-cluster/performance/pao_master.yaml
@@ -1,0 +1,12 @@
+apiVersion: performance.openshift.io/v2
+kind: PerformanceProfile
+metadata:
+  name: openshift-bootstrap-master
+spec:
+  cpu:
+    isolated: 0-1
+    reserved: 2-7
+  machineConfigPoolSelector:
+    pools.operator.machineconfiguration.openshift.io/master: ""
+  nodeSelector:
+    node-role.kubernetes.io/master: ""
--- a/test/e2e/performanceprofile/cluster-setup/bootstrap-cluster/performance/pao_worker.yaml
+++ b/test/e2e/performanceprofile/cluster-setup/bootstrap-cluster/performance/pao_worker.yaml
@@ -1,0 +1,12 @@
+apiVersion: performance.openshift.io/v2
+kind: PerformanceProfile
+metadata:
+  name: openshift-bootstrap-worker
+spec:
+  cpu:
+    isolated: 0-1
+    reserved: 2-3
+  machineConfigPoolSelector:
+    pools.operator.machineconfiguration.openshift.io/worker: ""
+  nodeSelector:
+    node-role.kubernetes.io/worker: ""
--- a/test/e2e/performanceprofile/functests-render-command/1_render_command/render_test.go
+++ b/test/e2e/performanceprofile/functests-render-command/1_render_command/render_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -12,8 +13,9 @@ import (
 )
 
 const (
-	defaultExpectedDir = "default"
-	pinnedExpectedDir  = "pinned"
+	defaultExpectedDir   = "default"
+	pinnedExpectedDir    = "pinned"
+	bootstrapExpectedDir = "bootstrap"
 )
 
 var (
@@ -23,6 +25,8 @@ var (
 	testDataPath       string
 	defaultPinnedDir   string
 	snoLegacyPinnedDir string
+	bootstrapPPDir     string
+	extraMCPDir        string
 )
 
 var _ = Describe("render command e2e test", func() {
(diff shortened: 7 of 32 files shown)


pkg/performanceprofile/controller/performanceprofile/components/consts.go
```diff
@@ -12,6 +12,8 @@ const (
 	ComponentNamePrefix = "performance"
 	// MachineConfigRoleLabelKey is the label key to use as label and in MachineConfigSelector of MCP which targets the performance profile
 	MachineConfigRoleLabelKey = "machineconfiguration.openshift.io/role"
+	// MachineConfigPoolRoleLabelPrefix
+	MachineConfigPoolRoleLabelPrefix = "pools.operator.machineconfiguration.openshift.io/"
```
MarSik, line 16: I would probably move this to the create function. Just to keep it close to the other hardcoded names.

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -159,6 +159,13 @@ func render(inputDir, outputDir string) error {
 		}
 	}
 
+	// If we find no MCPs present in the directory, then we assume the default
+	// Master, Worker ones and make sure that any PAO manifests that specify a Master or Worker label
+	// get correctly rendered. Specifying MCPs outside of the default will still cause a failure as is expected.
+	if len(mcPools) == 0 {
```
MarSik, line 163: There is a corner case here that I am not sure we want address however. An SNO with extra workers might in theory use master + eg. worker-cnf MCPs. The extra worker-cnf will make this condition return false and the master pool will be missing.

MarSik:
@eggfoobar I like the direction, but we should open a bug for this to allow backports and probably make the commit messages a bit nicer.

MarSik:
/retitle OCPBUGS-22095: [WIP] Add default MCP objects for rendering

pkg/performanceprofile/controller/performanceprofile/components/consts.go
eggfoobar: Sounds good, done

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -213,6 +213,11 @@ func render(inputDir, outputDir string) error {
 			componentObj.SetOwnerReferences(or)
 		}
 
+		// TODO: Resolve why kubelet ownership breaks bootstrap
+		components.KubeletConfig.SetOwnerReferences(nil)
```
ffromani, line 217: generalizing here: https://github.com/openshift/cluster-node-tuning-operator/pull/907

pkg/performanceprofile/cmd/render/render.go
MarSik: This is the only concern I have, I would rather create only the pools that are missing.

Eg. if no master is loaded from the folder, inject the master here. The same for worker.

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Nice! Should we keep this here for now and remove it in your PR?

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Ah my mistake, missed this, updated the code to account for this, please give it another look.

pkg/performanceprofile/cmd/render/render.go
ffromani: Either work. If we keep the fix in this PR, I'd avoid setting the OwnerReference in the first place rather than adding (lines 203-214) and then remove it right afterwards.
On top of that, I'm quite persuaded _all_ ownerReferences are wrong because autogenerating the UID is wrong in the first place, `kubeletconfig` is just the first failing here

pkg/util/manifests.go
```diff
@@ -140,3 +140,16 @@ func createLabeledDefaultMCPManifests() []*mcfgv1.MachineConfigPool {
 		},
 	}
 }
+
+func AddGeneratedByAnnotation(annotations map[string]string, profileName, profileNamespace string) map[string]string {
+	const generatedByAnnotationKey = "performanceprofile.openshift.io/generatedby"
```
ffromani: generated or rendered?

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Yeah same, in this latest commit, I'm removing those and triggering an update on ownerref diff to allow performance profile to add them after the fact, from what I can see that doesn't seem to trigger a change on the machine config side. Testing now

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -161,6 +159,9 @@ func render(inputDir, outputDir string) error {
 		}
 	}
 
+	// Append any missing default manifests (i.e. `master`/`worker`)
+	mcPools = util.AppendMissingDefaultMCPManifests(mcPools)
```
vitus133, line 163: Could it be possible to add this functionality to 
https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/tuned/cmd/render/render.go
?

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -191,29 +189,12 @@ func render(inputDir, outputDir string) error {
...
-				Kind:       pp.Kind,
-				Name:       pp.Name,
-				APIVersion: pp.APIVersion,
-				UID:        uid,
-			},
-		}
-
 		for _, componentObj := range components.ToObjects() {
-			componentObj.SetOwnerReferences(or)
+			componentObj.SetOwnerReferences(nil)
```
ffromani, line 193: Ideally this logic and this commit (which I think is getting ready to merge) should be split to its own PR because it's a related but different change wrt to "add default MCP objects for rendering"

eggfoobar: Sounds good, updated

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Per our conversation, ownership ref changes will be done in another PR

pkg/performanceprofile/cmd/render/render.go
eggfoobar: That's a good point, the logic seems the same, @jlojosnegros any concerns with adding this call to the tuned renderer?

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -204,6 +207,8 @@ func render(inputDir, outputDir string) error {
 
 		for _, componentObj := range components.ToObjects() {
 			componentObj.SetOwnerReferences(or)
+			anno := util.AddBootstrapRenderedByAnnotation(componentObj.GetAnnotations(), pp.Name, pp.Namespace)
```
MarSik, line 210: I think I would remove this too to make this PR solely about the MCPs.

eggfoobar: Sure thing, removed

pkg/performanceprofile/cmd/render/render.go
eggfoobar: Confirmed in Slack, added to tuned renderer

pkg/util/manifests.go
```diff
@@ -97,3 +100,68 @@ func ListFilesFromMultiplePaths(dirPaths []string) ([]string, error) {
 	}
 	return results, nil
 }
+
+// AppendMissingDefaultMCPManifests When default MCPs are missing, it is desirable to still generate the relevant
+// files based off of the standard MCP labels and node selectors.
+//
+// Here we create the default `master` and `worker` MCP if they are missing with their respective base Labels and NodeSelector Labels,
+// this allows any resource such as PAO's to utalize the deault during bootstrap rendering.
```
yanirq, line 108: nits:  s/PAO's/PAO       s/utalize/utilize      s/deault/default

vitus133:
@eggfoobar @jlojosnegros I think we now have a problem in [pkg/tuned/cmd/render/render.go](https://github.com/openshift/cluster-node-tuning-operator/pull/833/files#diff-f3655cb37779a7069442b7a63925fe91be42f16cafb43968dba654f44abf19c5) line 147:

```
if len(filteredPerformanceProfiles) == 0 {
		klog.Warningf("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
		return fmt.Errorf("Cannot find any PerformanceProfile complying with MachineConfigPool %s.", mcpName)
	}
```
Should probably become
```
if len(filteredPerformanceProfiles) == 0 {
		klog.Infof("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
		return nil
	}
```

eggfoobar:
> @eggfoobar @jlojosnegros I think we now have a problem in [pkg/tuned/cmd/render/render.go](https://github.com/openshift/cluster-node-tuning-operator/pull/833/files#diff-f3655cb37779a7069442b7a63925fe91be42f16cafb43968dba654f44abf19c5) line 147:
> 
> ```
> if len(filteredPerformanceProfiles) == 0 {
> 		klog.Warningf("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
> 		return fmt.Errorf("Cannot find any PerformanceProfile complying with MachineConfigPool %s.", mcpName)
> 	}
> ```
> 
> Should probably become
> 
> ```
> if len(filteredPerformanceProfiles) == 0 {
> 		klog.Infof("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
> 		return nil
> 	}
> ```

That makes sense to me, I'll change it

jlojosnegros:
> @eggfoobar @jlojosnegros I think we now have a problem in [pkg/tuned/cmd/render/render.go](https://github.com/openshift/cluster-node-tuning-operator/pull/833/files#diff-f3655cb37779a7069442b7a63925fe91be42f16cafb43968dba654f44abf19c5) line 147:
> 
> ```
> if len(filteredPerformanceProfiles) == 0 {
> 		klog.Warningf("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
> 		return fmt.Errorf("Cannot find any PerformanceProfile complying with MachineConfigPool %s.", mcpName)
> 	}
> ```
> 
> Should probably become
> 
> ```
> if len(filteredPerformanceProfiles) == 0 {
> 		klog.Infof("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
> 		return nil
> 	}
> ```

I am getting late to the party and maybe I have missed something but I do not understand this change. 
If there is no PerformanceProfile is an error so why returning `nil`?

vitus133:
> > @eggfoobar @jlojosnegros I think we now have a problem in [pkg/tuned/cmd/render/render.go](https://github.com/openshift/cluster-node-tuning-operator/pull/833/files#diff-f3655cb37779a7069442b7a63925fe91be42f16cafb43968dba654f44abf19c5) line 147:
> > ```
> > if len(filteredPerformanceProfiles) == 0 {
> > 		klog.Warningf("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
> > 		return fmt.Errorf("Cannot find any PerformanceProfile complying with MachineConfigPool %s.", mcpName)
> > 	}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should probably become
> > ```
> > if len(filteredPerformanceProfiles) == 0 {
> > 		klog.Infof("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
> > 		return nil
> > 	}
> > ```
> 
> I am getting late to the party and maybe I have missed something but I do not understand this change. If there is no PerformanceProfile is an error so why returning `nil`?

Because this PR adds default MCPs unconditionally. With the code as it was, NOT configuring a performance profile becomes an error.

jlojosnegros:
> > > @eggfoobar @jlojosnegros I think we now have a problem in [pkg/tuned/cmd/render/render.go](https://github.com/openshift/cluster-node-tuning-operator/pull/833/files#diff-f3655cb37779a7069442b7a63925fe91be42f16cafb43968dba654f44abf19c5) line 147:
> > > ```
> > > if len(filteredPerformanceProfiles) == 0 {
> > > 		klog.Warningf("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
> > > 		return fmt.Errorf("Cannot find any PerformanceProfile complying with MachineConfigPool %s.", mcpName)
> > > 	}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Should probably become
> > > ```
> > > if len(filteredPerformanceProfiles) == 0 {
> > > 		klog.Infof("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
> > > 		return nil
> > > 	}
> > > ```
> > 
> > 
> > I am getting late to the party and maybe I have missed something but I do not understand this change. If there is no PerformanceProfile is an error so why returning `nil`?
> 
> Because this PR adds default MCPs unconditionally. With the code as it was, NOT configuring a performance profile becomes an error.

Ok, this PR adds "master" and/or "worker" MCPs if they have not been defined in the input folder but, once that's done, the render process should remain the same, so if not having a PP was an error it should remain an error, right?

jlojosnegros:
> > > > I am getting late to the party and maybe I have missed something but I do not understand this change. If there is no PerformanceProfile is an error so why returning `nil`?
> > > 
> > > 
> > > Because this PR adds default MCPs unconditionally. With the code as it was, NOT configuring a performance profile becomes an error.
> > 
> > 
> > Ok, this PR adds "master" and/or "worker" MCPs if they have not been defined in the input folder but, once that's done, the render process should remain the same, so if not having a PP was an error it should remain an error, right?
> 
> I think not having a PP should be legit, it's a user's choice to configure or not configure a PP. We haven't reached the problem before because we have always configured / not configured PP and MCP simultaneously. But now MCP will be always there. I even think that logging a warning here is too far fetched, Info is probably more appropriate

But you are not rendering anything I mean, the main target of that command is to render a MC from a PP, if you return there you have not render anything so either it is an error or you need to continue

MarSik:
> > I think not having a PP should be legit, it's a user's choice to configure or not configure a PP. We haven't reached the problem before because we have always configured / not configured PP and MCP simultaneously. But now MCP will be always there. I even think that logging a warning here is too far fetched, Info is probably more appropriate
> 
> But you are not rendering anything I mean, the main target of that command is to render a MC from a PP, if you return there you have not render anything so either it is an error or you need to continue

I agree, when no PP is present, this must just be a no-op. And the machine config pool injection is virtual only, right? We never render the MCP to the destination folder, we just use it internally.

vitus133:
> > > > > I am getting late to the party and maybe I have missed something but I do not understand this change. If there is no PerformanceProfile is an error so why returning `nil`?
> > > > 
> > > > 
> > > > Because this PR adds default MCPs unconditionally. With the code as it was, NOT configuring a performance profile becomes an error.
> > > 
> > > 
> > > Ok, this PR adds "master" and/or "worker" MCPs if they have not been defined in the input folder but, once that's done, the render process should remain the same, so if not having a PP was an error it should remain an error, right?
> > 
> > 
> > I think not having a PP should be legit, it's a user's choice to configure or not configure a PP. We haven't reached the problem before because we have always configured / not configured PP and MCP simultaneously. But now MCP will be always there. I even think that logging a warning here is too far fetched, Info is probably more appropriate
> 
> But you are not rendering anything I mean, the main target of that command is to render a MC from a PP, if you return there you have not render anything so either it is an error or you need to continue

The main target of this command is to render PP from an MC. If a user have not provided a PP, we have nothing to render, so this step is not needed. By returning nil in this case we tell bootstrap process to continue peacefully to the next step. No PP - nothing happened, let's do other stuff. If we return error here, bootstrap fails, and it's all because user have not provided PP? Doesn't make sense, right?

jlojosnegros:
> > > > > > I am getting late to the party and maybe I have missed something but I do not understand this change. If there is no PerformanceProfile is an error so why returning `nil`?
> > > > > 
> > > > > 
> > > > > Because this PR adds default MCPs unconditionally. With the code as it was, NOT configuring a performance profile becomes an error.
> > > > 
> > > > 
> > > > Ok, this PR adds "master" and/or "worker" MCPs if they have not been defined in the input folder but, once that's done, the render process should remain the same, so if not having a PP was an error it should remain an error, right?
> > > 
> > > 
> > > I think not having a PP should be legit, it's a user's choice to configure or not configure a PP. We haven't reached the problem before because we have always configured / not configured PP and MCP simultaneously. But now MCP will be always there. I even think that logging a warning here is too far fetched, Info is probably more appropriate
> > 
> > 
> > But you are not rendering anything I mean, the main target of that command is to render a MC from a PP, if you return there you have not render anything so either it is an error or you need to continue
> 
> The main target of this command is to render PP from an MC. If a user have not provided a PP, we have nothing to render, so this step is not needed. By returning nil in this case we tell bootstrap process to continue peacefully to the next step. No PP - nothing happened, let's do other stuff. If we return error here, bootstrap fails, and it's all because user have not provided PP? Doesn't make sense, right?

Not the way other commands are done but ok, if that is what it's needed for bootstrap to work.

@MarSik Nope, default MCPs are just used internally but never rendered.

ffromani (commented):
I'm not convinced the better course of action is silently swallow errors and keep going, but this is a larger (and too late) concern. Everything else seems reasonnable.

LGTM.

jlojosnegros:
I will prefer a better error handling than returning "ok" when there is no PP but other than that LGTM.

MarSik (approved):
/lgtm

jmencak:
> /cc @jmencak
> 
> because it impacts non-perfprof code

Thank you Francesco. I don't have a problem with this change.
/hold cancel

PR #793, opened 2023-09-15
Add golangci-lint

jlojosnegros:
This commit adds a new golangci-lint target to the makefile and sets up a common config along the lines of other openshift operators.

- Address misspell violations
- Address staticheck violations
- Address deprecated libs used
- Address unused code

Signed-off-by: Daniel Mellado [dmellado@redhat.com](mailto:dmellado@redhat.com)

supersede: #680

PR #837, opened 2023-10-21
NO-JIRA: Remove most //nolint:* overrides

jmencak:
This change improves upon #793 by removing most linter overrides and replacing them with a proper fix to keep the linter happy.  Only overrides for one false positive (static analysis check), and two nolint:errcheck overrides due to failing unit tests (future work) in performanceprofile_controller.go are kept.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -90,8 +90,6 @@ type wqKey struct {
 	kind      string // object kind
 	namespace string // object namespace
 	name      string // object name
-	//nolint:unused
-	event string // object event type (add/update/delete) or pass the full object on delete
 }
 
 func NewController() (*Controller, error) {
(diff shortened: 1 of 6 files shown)


pkg/operator/controller.go
```diff
@@ -90,8 +90,6 @@ type wqKey struct {
 	kind      string // object kind
 	namespace string // object namespace
 	name      string // object name
-	//nolint:unused
-	event string // object event type (add/update/delete) or pass the full object on delete
```
ffromani, line 94: Unless there are backward compatibility constraints (e.g. public facing API), I prefer to just removed unused code/fields. I won't block this PR for this reason, though.

pkg/operator/hypershift.go
```diff
@@ -240,8 +240,9 @@ func parseNamespacedName(namespacedName string) string {
 
 func getMachineConfigFromConfigMap(config *corev1.ConfigMap) (*mcfgv1.MachineConfig, error) {
 	scheme := runtime.NewScheme()
-	//nolint:errcheck
-	mcfgv1.Install(scheme)
+	if err := mcfgv1.Install(scheme); err != nil {
+		return nil, err
+	}
```
ffromani, line 245: while this change per se is correct, I prefer to set up the schema once, not per each function call. Even if the function is expected to be called once (which doesn't seem the case?). This probably just means this code need further work besides this PR

pkg/operator/hypershift.go
```diff
@@ -290,8 +291,9 @@ func newConfigMapForMachineConfig(configMapName string, nodePoolName string, mc
 
 func serializeMachineConfig(mc *mcfgv1.MachineConfig) ([]byte, error) {
 	scheme := runtime.NewScheme()
-	//nolint:errcheck
-	mcfgv1.Install(scheme)
+	if err := mcfgv1.Install(scheme); err != nil {
+		return nil, err
+	}
```
ffromani, line 296: ditto

pkg/tuned/controller.go
```diff
@@ -147,8 +147,6 @@ type Controller struct {
 type wqKey struct {
 	kind string // object kind
 	name string // object name
-	//nolint:unused
-	event string // object event type (add/update/delete) or pass the full object on delete
```
ffromani, line 151: ditto about unused fields

ffromani (commented):
/lgtm
/hold

nonblocking comments, holding to let other reviewers chime in

jmencak:
Thank you for the review, Francesco. I've done more cleanup and also removed more unused code as suggested.  As for the `mcfgv1.Install(scheme)`, I totally agree this should be setup only once.  If you agree, I'd leave that as future work.

ffromani:
> Thank you for the review, Francesco. I've done more cleanup and also removed more unused code as suggested. As for the `mcfgv1.Install(scheme)`, I totally agree this should be setup only once. If you agree, I'd leave that as future work.

I totally agree that improving how we handle schema is better done in a future, separate PR.

ffromani:
/lgtm

let's do possible followups later

jmencak:
> /lgtm
> 
> let's do possible followups later

Absolutely. Adding to my TODO unless @dagrayvid want's do adjust his own code.

PR #838, opened 2023-10-25
[release-4.12] OCPBUGS-21837: nto: pao avoid timeout when there are too many CSV

jlojosnegros:
This is a manual cherry-pick of https://github.com/openshift/cluster-node-tuning-operator/pull/818 

supersedes #830

### DIFF
--- a/test/e2e/basic/csvs.go
+++ b/test/e2e/basic/csvs.go
@@ -8,7 +8,7 @@ import (
 	"strconv"
 	"time"
 
-	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


yanirq:
/payload-job periodic-ci-openshift-release-master-ci-4.12-upgrade-from-stable-4.11-e2e-gcp-ovn-upgrade

test/e2e/basic/csvs.go
```diff
@@ -30,35 +30,43 @@ const (
...
 	var baseOptions []client.ListOption
-	BeforeAll(func() {
-		var err error
+	var count int
 
-		By("Getting new client")
-		cli, err = newClient()
-		Expect(err).To(Succeed())
+	BeforeEach(func() {
+		if count == 0 {
```
ffromani, line 41: this works iff the test are strictly serial, are them guaranteed serial? otherwise it seems racy (or wrong)

ffromani (commented):
LGTM except for part of the ginkgo v1 adaptation

test/e2e/basic/csvs.go
jlojosnegros: your are totally right.
Thanks for the catch. 

I can't find a way to run the test sequentially using ginkgo v1, so I am thinking about creating the `client` in `BeforeSuite` and let the namespace be created/destroyed on `BeforeEach/AfterEach`.
I am gonna "merge" the `It` clauses so the code is only executed once to save execution time.

PR #844, opened 2023-11-03
CNF-10238: NTO render command for SNO boot arguments

jlojosnegros:
We need to speed up bootstrap.
To do so we need to apply kernel boot arguments without restarting node.

Add a new render command to prepare all tuned profiles, run `tuneD`, read `bootcmdline`  and render a MachineConfig to apply those kernel arguments.

Warning: running `tuneD` would modify some system files so this command should be executed in a properly isolated environment.

see: https://github.com/openshift/installer/pull/7692

### DIFF
--- a/pkg/tuned/cmd/render/cmd.go
+++ b/pkg/tuned/cmd/render/cmd.go
@@ -1,0 +1,85 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package render
+
+import (
+	"flag"
+	"fmt"
+
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"k8s.io/klog"
+)
+
+type renderOpts struct {
+	assetsInDir  []string
+	assetsOutDir string
+}
+
+func NewRenderBootCmdMCCommand() *cobra.Command {
+	renderOpts := renderOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "render-bootcmd-mc",
+		Short: "Render MC with kernel args",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := renderOpts.Validate(); err != nil {
+				klog.Fatal(err)
+			}
+
+			if err := renderOpts.Run(); err != nil {
+				klog.Fatal(err)
+			}
+		},
+	}
+
+	addKlogFlags(cmd)
+	renderOpts.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.StringArrayVar(&r.assetsInDir, "asset-input-dir", []string{components.AssetsDir}, "Input path for the assets directory. (Can use it more than one to define multiple directories)")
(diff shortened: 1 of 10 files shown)


pkg/tuned/cmd/render/render.go
```diff
@@ -0,0 +1,251 @@
...
+		//REVIEW - Not sure about the "recommended"profile.
+		_, _, _, err = tunedpkg.ProfilesExtract(tt.Spec.Profile, *tunedFromPP.Spec.Recommend[0].Profile)
+		if err != nil {
+			klog.Error("error extracting tuned profiles : %v", err)
+			return fmt.Errorf("error extracting tuned profiles: %w", err)
+		}
+	}
+
+	//Should run tuned
+	tunedrun.TunedOperandRun()
```
jlojosnegros, line 155: Not fully sure about this. 
Need to just run `/var/lib/tuned/bin/run start` so maybe something like [execCmd](https://github.com/openshift/cluster-node-tuning-operator/blob/12b9a12c1cf6892b36464c26d6198d2211f28a14/pkg/tuned/tuned_parser.go#L215) could be enough

pkg/tuned/cmd/render/cmd.go
```diff
@@ -0,0 +1,85 @@
...
+		},
+	}
+
+	addKlogFlags(cmd)
+	renderOpts.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&r.assetsInDir, "asset-input-dir", components.AssetsDir, "Input path for the assets directory. (Can be a comma separated list of directories.)")
```
MarSik, line 58: Comma is a valid (although seldom) used character for paths. What about allowing multiple occurences of asset-input-dir with one path each?

pkg/tuned/cmd/render/render.go
```diff
@@ -0,0 +1,251 @@
...
+				mcConfigs = append(mcConfigs, obj)
+			case *tunedv1.Tuned:
+				tuneD = append(tuneD, obj)
+			default:
+				klog.Infof("skipping %q [%d] manifest because of unhandled %T", file.Name(), idx+1, obji)
+			}
+		}
+	}
+
+	if len(perfProfiles) != 1 {
```
MarSik, line 125: Are you sure about this? It would work for SNO, but can this logic help with bigger clusters too? There can be more PerfProfiles with associated MCPs I think.

pkg/tuned/cmd/render/cmd.go
jlojosnegros: Just following the same "interface" as the PerformanceProfile render command 
https://github.com/openshift/cluster-node-tuning-operator/blob/cd53f65cdbc8e64210a632f7739410fd7697ef7f/pkg/performanceprofile/cmd/render/cmd.go#L63

MarSik: The interface there was arbitrary, there is no standard way of writing the render commands. I would feel better if we used the typical shell pattern of repeated arguments (at both places really...).

jlojosnegros: Changed to allow multiple instances of the `asset-input-dir` flag.

pkg/tuned/cmd/render/render.go
```diff
@@ -0,0 +1,264 @@
...
+		if err != nil {
+			klog.Errorf("Unable to get tuned. error : %v", err)
+			return fmt.Errorf("unable to get tuned from profile: %w", err)
+		}
+		perfProfilesTuneDProfiles[profile] = tunedFromPP
+		// add tuned from PP to the list
+		tuneD = append(tuneD, tunedFromPP)
+	}
+
+	tuneDrecommended := operator.TunedRecommend(tuneD)
```
MarSik, line 141: I wonder if this should go to an MCP loop in case there are multiple pools with multiple profiles.

jlojosnegros: So ... still need to load all the PerformanceProfiles? or just those which will match with MCP master?

MarSik: I think master only and only on SNO.

MarSik:
/retitle CNF-10238: NTO render command for SNO boot arguments
/approve
/lgtm

We want to merge the current working solution (tested) and iterate from there to avoid unnecessary duplication of work wrt OCP branching.

PR #848, opened 2023-11-13
CNF-10344: NTO: Avoid depending on kubernetes cpuset 

jlojosnegros:
Change k8s.io/kubernetes/pkg/kubelet/cm/cpuset to k8s.io/utils/cpuset

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -16,7 +16,7 @@ import (
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+	"k8s.io/utils/cpuset"
 	"k8s.io/utils/pointer"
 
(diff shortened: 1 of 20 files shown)


jlojosnegros:
> is it time yet to bump to kube 1.28?

Not really sure about that. 
I just did it cause I found some problems while rebasing https://github.com/openshift/hypershift/pull/1782 as hypershift uses v0.28.3 for some of the dependencies and from v0.28.0 onwards `k8s.io/kubernetes/pkg/kubelet/cm/cpuset` does not exists any more and `k8s.io/utils/cpuset` should be used instead [here](https://github.com/kubernetes/kubernetes/commit/cede96336a809a67546ca08df0748e4253ec270d)

ffromani:
> > is it time yet to bump to kube 1.28?
> 
> Not really sure about that. I just did it cause I found some problems while rebasing [openshift/hypershift#1782](https://github.com/openshift/hypershift/pull/1782) as hypershift uses v0.28.3 for some of the dependencies and from v0.28.0 onwards `k8s.io/kubernetes/pkg/kubelet/cm/cpuset` does not exists any more and `k8s.io/utils/cpuset` should be used instead [here](https://github.com/kubernetes/kubernetes/commit/cede96336a809a67546ca08df0748e4253ec270d)

`k8s.io/utils/cpuset` is pretty self contained so you can just add that dep in theory. I'd try first bumping _only_ the `k8s.io/utils` dep to 1.28 and see how it goes. I'm confident this should work - and surely it's a way less invasive change.

PR #851, opened 2023-11-14
WIP: e2e: wait for update only when needed

shajmakh:
The test was failing because the updates on MCP were not starting. This is a correct behavior when the workload hints settings are set explicitly to the defaults in the performance profile. so completely removing the hints field shouldn't trigger an update.

Modify the test code to wait for MCP updates only when it is expected to get updated (initial settings are different than the intended ones in essence)

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -593,8 +593,13 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				By("Updating the performance profile")
 				profiles.UpdateWithRetry(profile)
 
-				By("Applying changes in performance profile and waiting until mcp will start updating")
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -593,8 +593,13 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				By("Updating the performance profile")
 				profiles.UpdateWithRetry(profile)
 
-				By("Applying changes in performance profile and waiting until mcp will start updating")
-				mcps.WaitForCondition(performanceMCP, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+				expectedMCPUpdatingStatus := corev1.ConditionTrue
+				if isDefaultWorkloadHints(*initialProfile.Spec.WorkloadHints) {
```
ffromani, line 597: can this be `nil`?

shajmakh: yes, I'll address this. thanks for pointing it out

PR #852, opened 2023-11-15
refactor:components: pass all arguments under `Options`

Tal-or:
There are different variable that are needed,
in order to create the different components that
are handled by the controller.

The number of this variables keeps growing.
This require us to change the functions' signatures for every new variable that is being added.

It also make the code less readable as more and more variables are getting added.

In order to overcome the above,
we suggest to house all the different variables
under a single struct named `Options`

This changes should not  (and mustn't) affect the code functionality


ffromani:
/lgtm
/hold

I like the idea

PR #854, opened 2023-11-16
  OCPBUGS-18640: Fix Racing Machine Configs and add Day 0 Support

ffromani:
Take 2: removing the explicit wait inside the reconcile loop

This PR fixes a bug where two machine configs can race and sometimes we see incorrect kernel arguments on a Node.

Specifically, there is a problem with the controller updating the Node object to use Cgroupv1 and the controller writing the performance profile machine config. The patch addresses Day-0 and Day-2 operations.

## Day 0
Day 0 did not apply the Node object at bootstrap, so MCO configures the machine with Cgroupv1 at boot. The fix is to lay down the Node object to force cgroupv1 at bootstrap. MCO will pick up the Node object and configure the nodes with Cgroupv1 from the post-pivot boot.

## Day 2
The controller would originally update 2 MachineConfigs at the same time: 1) update the Node object (to switch to cgroupv1), and 2) create the perf profile machine config. These two configs are racing in certain scenarios; sometimes providing the node with duplicate (or wrong) kernel arguments.

The fix is to update the Node config to switch a cluster to Cgroupv1 and then watch for the Machine Config Pool generation to increment. The generation change signals that a new rendered-config has been created, and the NTO can then write the perfprofile machine config.

## Summary
Day 0 is the best approach and minimizes the amount of reboots necessary for PerformanceProfiles, since it does not require any extra reboots.


pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -802,3 +827,27 @@ func validateProfileMachineConfigPool(profile *performancev2.PerformanceProfile,
 
 	return nil
 }
+
+func validateCurrentMachineConfig(mc *mcov1.MachineConfig) error {
```
MarSik, line 831: I would use a more descriptive name like validateCgroupVersionKArgs or something myself.

ffromani: done!

ffromani:
/test e2e-gcp-pao

last job failed setting up

yanirq:
failure seems to be consistent for` override system-reserved memory` test .   might be an issue with the nightly  but we can check that.

yanirq:
checking for flake another attempt : 
/retest

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -482,6 +484,24 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return reconcile.Result{}, nil
 	}
 
+	// (TODO) This code can be removed in the future when the cgroupsv2 is supported
+	currentMC, err := r.getCurrentMachineConfigByProfile(ctx, profileMCP)
```
MarSik, line 488: A nit: getCurrentMachineConfigByProfile -> getCurrentMachineConfigByMCP

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -802,3 +823,27 @@ func validateProfileMachineConfigPool(profile *performancev2.PerformanceProfile,
...
+func validateCurrentMachineConfigCgroupV1(mc *mcov1.MachineConfig) error {
+	// taken from MCO pkg/controller/kubelet-config/helpers.go @ d469addcb
+	kernelArgsv1 := []string{
+		"systemd.unified_cgroup_hierarchy=0",
+		"systemd.legacy_systemd_cgroup_controller=1",
+	}
+
+	for _, karg := range kernelArgsv1 {
+		if !containString(mc.Spec.KernelArguments, karg) {
+			return fmt.Errorf("missing required cgroups V1 kernel argument: %q", karg)
```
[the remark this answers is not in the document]
ffromani: It goes in the degraded condition, yes. I'm reluctant to be so detailed in this text though. The condition doesn't seem a great place to give user tips and I'd rather settle to clarify the condition reason. I'll rework a bit the error text though.

pkg/performanceprofile/controller/performanceprofile_controller.go
ffromani: done, thanks.

ffromani:
/hold

to enable ppl to review

MarSik:
/lgtm

@rphillips Ryan, Francesco rewrote the MCP race patch to avoid waiting in the reconcile loop. Could you take a look?

@ffromani I think you should update the commit / PR description, it still describes Ryan's approach

yanirq:
@ffromani  would it make sense to squash the commits here ?  they are divided logically so its not really a must have.

ffromani:
/hold cancel
reached consensus offline (cc @yanirq)

PR #855, opened 2023-11-17
NO-JIRA: perfprof: fix condition reason

ffromani:
Per Condition type description,
`Reason: one-word CamelCase reason for the condition's last transition`

So let's fix existing conditions to make sure they are all indeed CamelCase one-words.

It must be noted that even though the code was wrong, this is a breaking change.


jmencak:
Thank you for the fix.
/lgtm

yanirq:
failure seems to be consistent for` override system-reserved memory` test .   might be an issue with the nightly  but we can check that.

ffromani:
will unhold after the branching

ffromani:
like it or not, we're breaking backward compat with this fix, so better not have it :\

PR #858, opened 2023-11-28
CNF-10473:featuregates: main: add feature gates support for PAO controller

Tal-or:
In order to protect the oprator (and the cluster in general)
Tech preview (TP) features, we should add feature gates support.

This commit bootstraps the feature gate api under
performance-addon-controller.

While developing TP features it shall be used in the following manner:
`if featuregate.Enabled(myfeature) { do thing }`


cmd/cluster-node-tuning-operator/main.go
```diff
@@ -321,6 +334,45 @@ func migratePinnedSingleNodeInfraStatus(cfg *rest.Config, scheme *apiruntime.Sch
...
+	}
+	k8sClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	controllerRef, err := events.GetControllerReferenceForCurrentPod(ctx, k8sClient, operatorNamespace, nil)
+	if err != nil {
+		return nil, err
+	}
+	eventRecorder := events.NewKubeRecorder(k8sClient.CoreV1().Events(metav1.NamespaceNone), "performance-profile-controller", controllerRef)
```
ffromani, line 354: can we get this from controller-runtime? (line 188 above)

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -169,11 +176,17 @@ func operatorRun() {
...
+
 	if !config.InHyperShift() {
 		if err = (&paocontroller.PerformanceProfileReconciler{
-			Client:   mgr.GetClient(),
-			Scheme:   mgr.GetScheme(),
-			Recorder: mgr.GetEventRecorderFor("performance-profile-controller"),
+			Client:      mgr.GetClient(),
+			Scheme:      mgr.GetScheme(),
+			Recorder:    mgr.GetEventRecorderFor("performance-profile-controller"),
+			FeatureGate: fg,
```
ffromani, line 189: and how do we use this value?

ffromani (commented):
I don't know of a better way than using a separate client and pulling in `library-go`
few questions inside

cmd/cluster-node-tuning-operator/main.go
Tal-or: I tried that, but the interfaces are completely different

cmd/cluster-node-tuning-operator/main.go
Tal-or: Specified in the PR and in the comment.
```
if fg.Enabled(<featureGateName>) {
do something
}
```
featureGateName are lister under the `openshift/api`
Here is the PR that adds the mixedCPUs feature gate: https://github.com/openshift/api/pull/1684

ffromani:
/cc @jmencak 
because we're 1. using library-go and 2. using a new and different client

cmd/cluster-node-tuning-operator/main.go
ffromani: right, silly me. So is going to be used in a followup PR. Works for me.

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -321,6 +334,45 @@ func migratePinnedSingleNodeInfraStatus(cfg *rest.Config, scheme *apiruntime.Sch
 	return nil
 }
 
+func setupFeatureGates(ctx context.Context, config *rest.Config, operatorNamespace string) (featuregates.FeatureGate, error) {
+	// TODO is this version correct?
+	desiredVersion := version.Version
```
ffromani, line 339: this is the last thing we need to check. Other than this, LGTM - dunno of better ways to achieve the same result (in the required time frame)

cmd/cluster-node-tuning-operator/main.go
Tal-or: > right, silly me. So is going to be used in a followup PR. Works for me.

Exactly

Tal-or:
> > because we're 1. using library-go and 2. using a new and different client
> 
> Yeah, this is becoming a monster. client-go, library-go, controller-runtime .

@jmencak I agree, maybe we should aim for merging all our clients under the client-go since it has the wider support as of today.
But that will requires some preparation work and huge changes especially in performance-profile-controller

ffromani:
> > > because we're 1. using library-go and 2. using a new and different client
> > 
> > 
> > Yeah, this is becoming a monster. client-go, library-go, controller-runtime .
> 
> @jmencak I agree, maybe we should aim for merging all our clients under the client-go since it has the wider support as of today. But that will requires some preparation work and huge changes especially in performance-profile-controller

it's even more complex than that. When we merged ppc, we pivoted to controller-runtime, because at time it seemed the best fit. Now the landscape changed again and we have new requirements.
Realistically speaking I don't think we can do much better in the remaining dev time (I'd love to be proven wrong though) but we should make a point to clean up. The best I can offer, besides actual fixes, is to file a OCPBUG and followup in the next cycle. WDYT @Tal-or @jmencak ?

Tal-or:
> When we merged ppc, we pivoted to controller-runtime, because at time it seemed the best fit. Now the landscape changed again and we have new requirements.

Yes. but replacing the runtime-client with go-client won't require any user-facing changes. so it's more of internal factoring I would say.

> The best I can offer, besides actual fixes, is to file a OCPBUG and followup in the next cycle. WDYT @Tal-or @jmencak ?

+1 you can assign that on me if you like since I already begun with the mess anyway :)

ffromani:
For the record and for the sake of completeness, I think that reimplementing this functionality on top of controller-runtime is also a completely valid option which I'd be in favor of, so let's keep this option open as well for the future.

ffromani:
I'll LGTM once we have full clarity about the version per inline comment above.
let's make sure to add _some_ e2e coverage, which will turn handy for the planned future work. This likely will require merging more PRs to leverage the functionality added here.

ffromani:
> > For the record and for the sake of completeness, I think that reimplementing this functionality on top of controller-runtime is also a completely valid option which I'd be in favor of, so let's keep this option open as well for the future.
> 
> I think it depends how much in a rush are we to get this in. If we are not, I vote for reimplementing this for controller-runtime. If we're in a rush to merge, let's postpone.

unfortunately AFAICT we're under pressure and @Tal-or has (quite a few) other PRs pending depending on this one.

Tal-or:
> > For the record and for the sake of completeness, I think that reimplementing this functionality on top of controller-runtime is also a completely valid option which I'd be in favor of, so let's keep this option open as well for the future.
> 
> I think it depends how much in a rush are we to get this in. If we are not, I vote for reimplementing this for controller-runtime. If we're in a rush to merge, let's postpone.

This means adding support to create feature gate accessor using runtime-client here:
https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/featuregates/simple_featuregate_reader.go#L108

ffromani:
the version seems ok but let's make sure we have the correct (and minimal) RBAC permissions to access `clusterversion` objects

ffromani:
this should be it, PTAL
```
$ git diff
diff --git a/manifests/40-rbac.yaml b/manifests/40-rbac.yaml
index 1deb2ef5..7a306dca 100644
--- a/manifests/40-rbac.yaml
+++ b/manifests/40-rbac.yaml
@@ -59,6 +59,10 @@ rules:
 - apiGroups: [""]
   resources: ["nodes/metrics","nodes/specs"]
   verbs: ["get"]
+# Needed to check the feature gates
+- apiGroups: ["config.openshift.io"]
+  resources: ["clusterversions"]
+  verbs: ["list"]
 # Needed by every CVO-managed operator.
 # infrastructures needed for the leader election code to get infra topology
 - apiGroups: ["config.openshift.io"]
```

ffromani:
> this should be it, PTAL
> 
> ```
> $ git diff
> diff --git a/manifests/40-rbac.yaml b/manifests/40-rbac.yaml
> index 1deb2ef5..7a306dca 100644
> --- a/manifests/40-rbac.yaml
> +++ b/manifests/40-rbac.yaml
> @@ -59,6 +59,10 @@ rules:
>  - apiGroups: [""]
>    resources: ["nodes/metrics","nodes/specs"]
>    verbs: ["get"]
> +# Needed to check the feature gates
> +- apiGroups: ["config.openshift.io"]
> +  resources: ["clusterversions"]
> +  verbs: ["list"]
>  # Needed by every CVO-managed operator.
>  # infrastructures needed for the leader election code to get infra topology
>  - apiGroups: ["config.openshift.io"]
> ```

actually also for `featuregates`. We possibly need to also add `watch` and `get` permissions.

Tal-or:
/hold
Tested locally, operator is crashing.

Tal-or:
/hold cancel
Fixed the rbac

cmd/cluster-node-tuning-operator/main.go
```diff
@@ -321,6 +334,47 @@ func migratePinnedSingleNodeInfraStatus(cfg *rest.Config, scheme *apiruntime.Sch
 	return nil
 }
 
+func setupFeatureGates(ctx context.Context, config *rest.Config, operatorNamespace string) (featuregates.FeatureGate, error) {
+	// TODO is this version correct?
+	missingVersion := "0.0.1-snapshot"
```
mrunalp, line 339: Looks fine comparing to https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/operator.go#L100-L131

mrunalp:
Let us get this merged to unblock the other PRs behind the feature gate. We can follow up to refactor this as needed.

ffromani:
> Let us get this merged to unblock the other PRs behind the feature gate. We can follow up to refactor this as needed.

yes, we reached agreement already with offline discussion.

manifests/40-rbac.yaml
```diff
@@ -99,7 +99,14 @@ rules:
...
   resources: ["clusterserviceversions","operatorgroups","subscriptions"]
-  verbs: ["get","delete","list","update","watch"] 
+  verbs: ["get","delete","list","update","watch"]
+# Needed to check the feature gates
+- apiGroups: ["config.openshift.io"]
+  resources: ["clusterversions", "featuregates"]
+  verbs: ["list", "watch"]
+- apiGroups: ["apps"]
+  resources: ["replicasets"]
+  verbs: ["get"]
```
ffromani, line 109: what do we need replicasets for?

ffromani (commented):
/lgtm

for reasons explained during the review

manifests/40-rbac.yaml
Tal-or: > what do we need replicasets for?

For the event recorder. It detect the underline object automatically and I presume it's a replicaset because I got an error about prohibited access to that type of resource

ffromani: ok, this needs further investigation but it's not blocking now. NTO watches access already a bunch of objects and allowing `get` of replicasets is not too concerning.

Tal-or:
/hold 
@joselsegura please take a look from hypershift perspective

Tal-or:
@jlojosnegros tagged the wrong jose

ffromani:
assuming CI still fails, which seems likely, we can't ignore the fact the NOOP PR (https://github.com/openshift/cluster-node-tuning-operator/pull/862) passed.
Please either:
- rebase on top of current head
- do query for FGs only if not running on hypershift (we have code to detect this)
(we should do both IMO, and then we can have a bug to enable support for hypershift, that is acceptable IMO)

I commit to be on top of this PR and review again quickly once updates are uploaded.

Tal-or:
> assuming CI still fails, which seems likely, we can't ignore the fact the NOOP PR (#862) passed. Please either:
> 
> * rebase on top of current head
> * do query for FGs only if not running on hypershift (we have code to detect this)
>   (we should do both IMO, and then we can have a bug to enable support for hypershift, that is acceptable IMO)
> 
> I commit to be on top of this PR and review again quickly once updates are uploaded.

Before CI runs, rebase is always done automatically against HEAD (unless there's a conflict)
About the second bullet, sounds fair. Done

ffromani:
/lgtm

we now don't do anything on hypershift flow, besides adding trivial RBAC permissions which can't cause lane failures

PR #864, opened 2023-12-01
NO-JIRA: Report duplicate profiles as ERRORs

jmencak:
In the past, duplicate profiles were being report as WARNINGs only. Increase seriousness of this fact, so that it helps the users to fix their configuration issues.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -133,7 +133,7 @@ func tunedRenderedProfiles(tuned *tunedv1.Tuned, m map[string]tunedv1.TunedProfi
 					if *v.Data == *existingProfile.Data {
 						klog.Infof("duplicate profiles names %s but they have the same contents", *v.Name)
 					} else {
-						klog.Warningf("WARNING: duplicate profiles named %s with different contents", *v.Name)
+						klog.Errorf("ERROR: duplicate profiles named %s with different contents found in Tuned CR %q", *v.Name, tuned.Name)
 					}
 				}
 				m[*v.Name] = v


pkg/manifests/manifests.go
```diff
@@ -133,7 +133,7 @@ func tunedRenderedProfiles(tuned *tunedv1.Tuned, m map[string]tunedv1.TunedProfi
 					if *v.Data == *existingProfile.Data {
 						klog.Infof("duplicate profiles names %s but they have the same contents", *v.Name)
 					} else {
-						klog.Warningf("WARNING: duplicate profiles named %s with different contents", *v.Name)
+						klog.Errorf("ERROR: duplicate profiles named %s with different contents!", *v.Name)
```
MarSik, line 136: Is here a way to print the Tuned object name the profile comes from too? And maybe the profile that was selected (+ its parent Tuned) at the end of the loop?

jmencak: > Is here a way to print the Tuned object name the profile comes from too? 
Yes. I added it to the PR.

> And maybe the profile that was selected (+ its parent Tuned) at the end of the loop?
Not sure what exactly do you mean, but I'm sensing it would require larger code changes and hopefully the above is sufficient.

jmencak:
> I'm fine with the idea here. Do we report or want to report this condition as `Degraded` condition in the tuned object? If we don't do already, it's totally new PR material.

Thank you for the review. 'Degraded' is reported in profiles.tuned.openshift.io, but not for this particular situation. We might in the future, but as you say, not in this PR (added to my TODO list).

@MarSik , please see if the changes are sufficient, so that we can get this in 4.15.0. Thank you!

ffromani:
/lgtm
/hold

waiting for second opinion from @MarSik

MarSik:
/hold cancel

I would also like to see a Tuned.Status.Conditions improvement, but this is a good step forward.

PR #866, opened 2023-12-04
CNF-9664: cpuset-configure: exit gracefully on cgroup v2

haircommander:


MarSik:
@haircommander We rely on systemd to create the system.slice now, right? The old code was just because the cpuset controller was not managed by systemd.

haircommander:
> We rely on systemd to create the system.slice now, right? The old code was just because the cpuset controller was not managed by systemd

yeah, systemd will enable the cpuset controller now

MarSik:
@haircommander You need to update the tests. There is a Makefile target for that (render-sync)

MarSik:
/retitle CNF-9664: cpuset-configure: exit gracefully on cgroup v2

PR #868, opened 2023-12-11
OCPBUGS-24638: Do not set default RPS sysctl twice

MarSik:
tuned complains that the sysctl value was already set via a sysctl.d file:

```
2023-12-05T22:23:07.343622915Z 2023-12-05 22:23:07,343 INFO     tuned.plugins.plugin_sysctl: Overriding sysctl parameter 'net.core.rps_default_mask' from '${not_isolated_cpumask}' to '30000003'
```

The comparison in tuned is broken as ${not_isolated_cpumask} should actually expand to the same value. But this INFO causes NTO to mark the profile as Degraded and we do not want that.

Commenting out the sysctl in the tuned profile is not going to break anything, because the sysctl.d version of it is applied anyway.

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -109,9 +109,11 @@ vm.dirty_background_ratio=3
 vm.swappiness=10
 
 # also configured via a sysctl.d file
-# placed here for better reflection of the change
+# placed here for documentation purposes and commented out due
+# to a tuned logging bug complaining about duplicate sysctl:
+#   https://issues.redhat.com/browse/RHEL-18972
 #> rps configuration
-net.core.rps_default_mask=${not_isolated_cpumask}
+# net.core.rps_default_mask=${not_isolated_cpumask}
 
 
 [selinux]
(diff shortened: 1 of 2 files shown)


assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -109,9 +109,10 @@ vm.dirty_background_ratio=3
 vm.swappiness=10
 
 # also configured via a sysctl.d file
-# placed here for better reflection of the change
+# placed here for documentation purposes and commented out due
+# to a tuned logging bug complaining about duplicate sysctl
```
jmencak, line 113: Please reference the exact TuneD bug here.  Otherwise, LGTM.

jmencak:
Actually, it needs more changes.  ci/prow/e2e-no-cluster fails.

assets/performanceprofile/tuned/openshift-node-performance
MarSik: Done

ffromani:
/lgtm

because this seems to be fixed now

> Actually, it needs more changes. ci/prow/e2e-no-cluster fails.

PR #870, opened 2023-12-11
  OCPBUGS-18640: Fix Racing Machine Configs and add Day 0 Support (#854)

rbaturov:
* render nodeconfig to cgroup v1

then, wait for cgroupv1 rollout; add logging and update tests and lint. add interceptor to increase generation counter.
do render sync.

* perfprof: controller: implicit wait for cgroups v1

rather than inject an explicit wait in the reconcile loop, we check that the MachineConfig includes the required cgroups V1 settings, exiting from the reconcile loop if not.

In addition, all the tuning features depend on cgroups v1. Hence, we set the perfprofile status to Degraded until the node(s) are configured to v1.

Last but not least, it must be noted that this approach depends on implementation details pertaining to MCO. We are thus adding an implicit dependency between the two components.

* perfprof: controller: e2e: dissolve the annotation test

Dissolve the early e2e annotation test in the new more modern 7_peformance_kubelet_node suite, which had already a better test running - more comprehensive, and add a controller test for extra safety.

Overall this change increases test coverage, not reduces it.

* Add annotation to disable cgroups v2 downgrade

This adds performance.openshift.io/ignore-cgroups-version that can be used during development of cgroups v2 related code.

* perfprof: e2e: test shortcut for second profile

We have (few) tests which want to add a second profile, but in CI we can hardly afford, if at all, a second MCP. So the tests takes shortcut and assume the manifests are created even without a matching MCP.

This assumption doesn't hold anymore when we apply the cgroups V1 revert, because in this flow we need to check the matching MCP picked up before we move on, intentionally.

To move forward, we disable the cgroup v1 switch in these tests which add a second profile. This admittedly creates a gap - we don't test how the two features interact, but it seem the only practical option without overhauling the CI flows.

---------

### DIFF
--- a/pkg/apis/performanceprofile/v2/performanceprofile_types.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_types.go
@@ -34,6 +34,10 @@ const PerformanceProfileEnablePhysicalRpsAnnotation = "performance.openshift.io/
 // that ignores the removal of all RPS settings when realtime workload hint is explicitly set to false.
 const PerformanceProfileEnableRpsAnnotation = "performance.openshift.io/enable-rps"
 
+// PerformanceProfileIgnoreCgroupsVersion allows an admin to suspend the operator's
+// automatic downgrade of Cgroups version to V1 for development purposes.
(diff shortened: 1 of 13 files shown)


ffromani (commented):
need to fix some rebase/merge issues

PR #872, opened 2023-12-11
NO-JIRA: Update OWNERS file

yanirq:


jmencak:
/lgtm
but rbaturov needs to be a member of openshift.org

PR #875, opened 2023-12-12
OCPBUGS-24792: Make MC names deterministic

jmencak:
OpenShift Nodes can be a part of only one custom MachineConfigPool. However, MCP configuration allows this not to be the case.  This is caught by the machine-config-controller and reported as an error (<node> belongs to <N> custom roles, cannot proceed with this Node).

In order to target an MCP with a configuration, NTO uses machineConfigLabels.  However, one or more MCPs can select a particular single MC.  This is due to the MCP's machineConfigSelector.  This is another challenge scenario.

In the above two scenarios, it was possible for NTO to generate a randomly-named MC based on the membership of one of the matching MCPs. Then, a pruning function would mistakenly remove the other MCs, seemingly unused.  This could result in a flip between the rendered MCs and cause a Node reboot.

This PR makes the process of establishing the names of the MC for the purposes of MachineConfigPool based matching deterministic.

Other changes/fixes:
 - Synced with MCO's latest getPoolsForNode() changes.
 - Logging in syncMachineConfigHyperShift().

Resolves: OCPBUGS-24792

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -42,7 +42,6 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/pkg/util"
 	"github.com/openshift/cluster-node-tuning-operator/version"
 
-	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	mcfgclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned"
 	mcfginformers "github.com/openshift/machine-config-operator/pkg/generated/informers/externalversions"
 )
@@ -620,7 +619,6 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 	var (
 		tunedProfileName string
 		mcLabels         map[string]string
-		pools            []*mcfgv1.MachineConfigPool
 		operand          tunedv1.OperandConfig
 		nodePoolName     string
 	)
@@ -659,7 +657,7 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 			return err
 		}
 	} else {
-		tunedProfileName, mcLabels, pools, operand, err = c.pc.calculateProfile(nodeName)
+		tunedProfileName, mcLabels, operand, err = c.pc.calculateProfile(nodeName)
 		if err != nil {
 			return err
 		}
@@ -730,12 +728,12 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 	} else {
 		if mcLabels != nil {
 			// The TuneD daemon profile 'tunedProfileName' for nodeName matched with MachineConfig
-			// labels set for additional machine configuration.  Sync the operator-created
-			// MachineConfig for MachineConfigPools 'pools'.
+			// labels 'mcLabels' set for additional machine configuration.  Sync the operator-created
+			// MachineConfig based on 'mcLabels'.
 			if profile.Status.TunedProfile == tunedProfileName && profileApplied(profile) {
 				// Synchronize MachineConfig only once the (calculated) TuneD profile 'tunedProfileName'
 				// has been successfully applied.
-				err := c.syncMachineConfig(pools, mcLabels, profile)
+				err := c.syncMachineConfig(mcLabels, profile)
 				if err != nil {
 					return fmt.Errorf("failed to update Profile %s: %v", profile.Name, err)
 				}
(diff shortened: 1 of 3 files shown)


jmencak:
@yanirq if you have any free cycles, I'd appreciate a review. Also asked @liqcui for a review and premerge testing. I'd like to get this in and backport ASAP.

pkg/operator/controller.go
```diff
@@ -777,25 +775,36 @@ func (c *Controller) getProviderName(nodeName string) (string, error) {
...
-	if ok := c.allNodesAgreeOnBootcmdline(nodes); !ok {
-		return fmt.Errorf("not all %d Nodes in MCP %v agree on bootcmdline: %s", len(nodes), pools[0].ObjectMeta.Name, bootcmdline)
+	for _, pool := range pools {
+		var bootcmdlineSet bool
+		nodes, err := c.pc.getNodesForPool(pool)
+		if err != nil {
+			return err
+		}
+
+		bootcmdline, bootcmdlineSet = c.pc.state.bootcmdline[profile.Name]
```
yanirq, line 799: could state be nil ?

yanirq:
Code wise this looks good to me.
In terms of functionality,  this kindof introduce a new ability to have more that one MCP for nodes with picking an MCP for applying tuned by an alphabetical rule.
The approach is not harmful but should be probably documented .

Would we like to have functional tests for this deterministic behavior ?  Its a heavy test to perform so Im not saying Im in favor, just thinking out loud.

cc @MarSik  In case you want to weigh in.

pkg/operator/controller.go
jmencak: No. And we've been running with that code for a long time.

pkg/operator/controller.go
```diff
@@ -777,25 +775,36 @@ func (c *Controller) getProviderName(nodeName string) (string, error) {
...
+		if !bootcmdlineSet {
+			klog.V(2).Infof("syncMachineConfig(): bootcmdline for %s not cached, sync canceled", profile.Name)
+			return nil
+		}
+
+		if ok := c.allNodesAgreeOnBootcmdline(nodes); !ok {
```
MarSik, line 805: If you want to support multiple pools matching the recommendation section of a single Tuned, then I think this loop has to extend further below this. Because you fail the whole sync on two MCPs not having the same kargs. Which they are not required to have, that condition only applies to each MCP separately.

MarSik:
Hmm, we can either improve the support for multiple MCPs (like I commented in the code) or block the approach by either a check or by generating a sufficiently unique label set for the new MC so it is matched by a single MCP only.

pkg/operator/controller.go
jmencak: > If you want to support multiple pools matching the recommendation section of a single Tuned

@MarSik , I believe there is a fundamental misunderstanding what this change does.  We are not adding support, we are fixing a potential cyclic reboot issue that was not fixed by previous code.

So my aim is not to support multiple MCPs, but we need to check all the matching MCPs affected by this.  Would logging an error or warning when we see multiple matching MCPs be sufficient?

jmencak:
> Hmm, we can either improve the support for multiple MCPs (like I commented in the code) or block the approach by either a check or by generating a sufficiently unique label set for the new MC so it is matched by a single MCP only.

As mentioned above, we are not adding support for multipe MCPs.  Also, we cannot generate a MCP label, because that's what customers/PPC specifies.

jmencak:
Liquan found an issue when manually running one test scenario. Let's see if it is real first.
/hold

pkg/operator/controller.go
MarSik: But it essentially does add the support for multiple MCPs.

You can block the sync when multiple MCPs are detected instead to make it clear that is is not supported. Not need to validate kargs in such case.

liqcui:
/lgtm
I reproduced the bug on ocp4.14 cluster, then tested the same steps using new pr code, the bug fixed using the new pr code. I also did a full regression testing using pr code, pre-merge testing passed

liqcui:
> Liquan found an issue when manually running one test scenario. Cannot reproduce it, but let's see if it is real first. /hold

It's not a real issue. re-test is ok

jmencak:
> It's not a real issue. re-test is ok

Thank you, still keeping the hold for now.

pkg/operator/controller.go
jmencak: > But it essentially does add the support for multiple MCPs.

Let's agree to disagree.  It fixes a bug.

> You can block the sync when multiple MCPs are detected instead to make it clear that is is not supported. Not need to validate kargs in such case.

No, we still need to check the kargs.  You can have machines with different topology within a single MCP.

jmencak: @MarSik , please see the "[Enforce per-pool machineConfigLabels selectors](https://github.com/openshift/cluster-node-tuning-operator/pull/875/commits/2963f6671d6c6f7672dc20b7e114e6abab9af1e9)" commit.  Hopefully it resolves the issues you see with this PR.

jmencak:
Keeping a hold for now as I want to do more manual testing on SNO.  Will unhold after that.

jmencak:
Manual tests on SNO worked fine.
/hold cancel

pkg/operator/controller.go
```diff
@@ -777,25 +775,44 @@ func (c *Controller) getProviderName(nodeName string) (string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	// The following enforces per-pool machineConfigLabels selectors.
+	if len(pools) > 1 {
+		// Log an error and do not requeue, this is a configuration issue.
+		klog.Errorf("profile %v uses machineConfigLabels that match across multiple MCPs (%v); this is not supported",
+			profile.Name, printMachineConfigPoolsNames(pools))
+		return nil
```
MarSik, line 794: Should this really return success? What was wrong on returning the error? (the other log update below on line 798)

MarSik: Ack, so the only place we will report this at will be the NTO log?

jmencak: Yes, for now.

MarSik:
/lgtm

But we should get together and figure out a better error reporting scheme. Right now we only have the information about errors in logs only. Users do not notice those from the cluster management level.

jmencak:
> /lgtm
> 
> But we should get together and figure out a better error reporting scheme. Right now we only have the information about errors in logs only. Users do not notice those from the cluster management level.

I propose doing a similar thing MCO does when it detects issues with its MCPs, it raises alerts.

PR #877, opened 2023-12-12
OCPBUGS-24581: rps: fix mask update for SR-IOV devices

Tal-or:
SR-IOV devices get moved (renamed) upon their creation.
This causes the `set-rps-mask.sh` to fail since the queue's path has changed.

We should add additional udev rule to act upon the move of a physical device
and set the rps mask correctly.

The `set-rps-mask.sh` script has modified to act upon those two
different scenarios:
1. when queues are being added.
2. when net device is moved.

In addition we fail silently (exit 0) when queues are failed to get
updated due to the mentioned above.
The queues that were failed to get updated, are expected to be updated
by the instance that get trrigered after the device move (renaming).


assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -1,16 +1,37 @@
...
 queue_num="${queue_num/\//-}"
 
 # set rps affinity for the queue
-echo "${mask}" > "/sys${queue_path}${queue_num}/rps_cpus"
\ No newline at end of file
+echo "${mask}"  2> /dev/null > "/sys${queue_path}${queue_num}/rps_cpus"
+
+# the 'echo' command might failed if the device path which the queue belongs to has changes
+# this can happen in case of SRI-OV devices renaming
+exit 0
```
ffromani, line 11: The idea is ok but forcing an exit 0 will make the function non-composable. The calling program will always exit here, won't it? If we want to swallow the error from echo, I'd either use `return 0` or `echo "${mask}" 2> /dev/null > "/sys${queue_path}${queue_num}/rps_cpus" || :`

assets/performanceprofile/configs/99-netdev-physical-rps.rules
```diff
@@ -1 +1 @@
-SUBSYSTEM=="queues", ACTION=="add", ENV{DEVPATH}=="/devices/pci*/queues/rx*", TAG+="systemd", ENV{SYSTEMD_WANTS}="update-rps@%p.service"
\ No newline at end of file
+SUBSYSTEM=="queues", ACTION=="add", ENV{DEVPATH}=="/devices/pci*/queues/rx*", TAG+="systemd", PROGRAM="/bin/systemd-escape --path --template=update-rps@.service $env{DEVPATH}", ENV{SYSTEMD_WANTS}="%c"
```
ffromani, line 1: this looks clean and it's probably the cleanest approach, but we're spawning a new process for each invocation. Can we avoid that? Or is it a performance concern at all, actually?

assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -1,24 +1,20 @@
...
 
 function set_queue_rps_mask() {
-queue_path=${path%rx*}
-
-queue_num=${path#*queues/}
-
-# replace '/' with '-'
-queue_num="${queue_num/\//-}"
-
+# replace x2d with hyphen (-) which is an escaped character
```
ffromani, line 4: does `systemd-escape` support reverse functionality (un-escape) and if so, do we want to use it?

ffromani (commented):
inline comments about implementation

assets/performanceprofile/configs/99-netdev-physical-rps.rules
MarSik: Yeah, that is unfortunate, however we could not avoid that. Also we only do that for physical devices, so hopefully just once per boot.

assets/performanceprofile/scripts/set-rps-mask.sh
MarSik: We should return 0 in the main flow context (at the end of the file) instead of the function I think

assets/performanceprofile/configs/99-netdev-physical-rps.rules
ffromani: fair enough

assets/performanceprofile/scripts/set-rps-mask.sh
Tal-or: It supports reverse.
The original systemd-escape output is:
`\x2d` but the backslash gets removed when we pass the argument to the script. Due to that fact it failed to revert this part.
Besides that, when we call the script we invoked it under the systemd-unit and passing the path argument with `%I` which basically perform the unesacping for us and which is cheaper than fork another process for the systemd-escape call here

assets/performanceprofile/scripts/set-rps-mask.sh
Tal-or: We don't want to exit 0 in the main flow because we still want to catch errors if the second function is being called

Tal-or: The `set_net_dev_rps` one

assets/performanceprofile/scripts/set-rps-mask.sh
ffromani: it surely is, but it's also asymmetric. We do escaping using systemd-escape, we do unescaping ourselves. We save resources, we need to ensure the consistency ourselves.
Now: this can very much be a path we want to choose, but let's make sure it's intentional, agreed, explicit.

Tal-or: Yes I agree it's kind of odd and if the de-escaping was perfect I would definitely choose this path.
But I really didn't understand why the backslash get omitted. It's probably related to the way of how the systemd-unit passes the argument

ffromani: possibly true. Considering the nature of this bug I won't block for this, even though I think it would deserve a followup

assets/performanceprofile/scripts/set-rps-mask.sh
ffromani: ok, we do want swallow error of `echo` on line 12 though, right?

Tal-or: Yes and you're right we should return 0 and not exit. it would basically have the same affect but will keep the function composable

assets/performanceprofile/configs/99-netdev-physical-rps.rules
```diff
@@ -1 +1,4 @@
 SUBSYSTEM=="queues", ACTION=="add", ENV{DEVPATH}=="/devices/pci*/queues/rx*", TAG+="systemd", PROGRAM="/bin/systemd-escape --path --template=update-rps@.service $env{DEVPATH}", ENV{SYSTEMD_WANTS}="%c"
+
+# SRI-OV devices are moved (renamed), hence we want to catch this event as well
```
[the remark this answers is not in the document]
Tal-or: Done. Thanks

ffromani:
/approve

we want this fix

assets/performanceprofile/scripts/set-rps-mask.sh
```diff
@@ -1,24 +1,20 @@
...
 # this can happen in case of SRI-OV devices renaming
-exit 0
+return 0
 }
 
 function set_net_dev_rps_mask() {
   # in case of device we want to iterate through all queues
-   for i in /sys"${path}"/queues/rx-*; do
+   for i in /sys/"${path}"/queues/rx-*; do
+     echo "${i}/rps_cpus"
```
ffromani, line 17: what is this echo for?

ffromani:
getting there. A comment inside, and need to update `manual_machineconfig.yaml`.

assets/performanceprofile/scripts/set-rps-mask.sh
Tal-or: Debugging left overs

ffromani:
/lgtm
/hold

to give time to other reviewers

the indentation was lost, but not blocking for now

ffromani:
/hold cancel

we have second review now

ffromani:
/retest-required

flakes + infra issues (crio overloaded?!?)

ffromani:
ok, let's hold the retests. It's either infra issue or the rules are actually causing excessive load. Investigating.

ffromani:
/retest-required

we have strong suspects it's a flake because the lane passed previously without relevant updates being done

ffromani:
> /retest-required
> 
> we have strong suspects it's a flake because the lane passed previously without relevant updates being done

```
 Adding transient rw bind mount for /run/secrets/rhsm
[1/2] STEP 1/5: FROM image-registry.openshift-image-registry.svc:5000/ci-op-fjyqjw1h/pipeline@sha256:d336e16fa089260e51b888c7acbd69728c1eac3a7e858f0e1755e396d81302ec AS builder
Trying to pull image-registry.openshift-image-registry.svc:5000/ci-op-fjyqjw1h/pipeline@sha256:d336e16fa089260e51b888c7acbd69728c1eac3a7e858f0e1755e396d81302ec...
error: build error: creating build container: parsing image configuration: fetching blob: StatusCode: 400, 
```

ffromani:
> /test e2e-gcp-pao-workloadhints

```
 Adding transient rw bind mount for /run/secrets/rhsm
[1/2] STEP 1/5: FROM image-registry.openshift-image-registry.svc:5000/ci-op-fjyqjw1h/pipeline@sha256:d336e16fa089260e51b888c7acbd69728c1eac3a7e858f0e1755e396d81302ec AS builder
Trying to pull image-registry.openshift-image-registry.svc:5000/ci-op-fjyqjw1h/pipeline@sha256:d336e16fa089260e51b888c7acbd69728c1eac3a7e858f0e1755e396d81302ec...
error: build error: creating build container: parsing image configuration: fetching blob: StatusCode: 400, 
INFO[2023-12-15T13:11:12Z] Building cluster-node-tuning-operator-us-tuned 
INFO[2023-12-15T13:11:12Z] Found existing build "cluster-node-tuning-operator-us-tuned-amd64" 
INFO[2023-12-15T13:11:12Z] Build cluster-node-tuning-operator-us-tuned-amd64 succeeded after 4m10s 
INFO[2023-12-15T13:11:12Z] Image ci-op-fjyqjw1h/pipeline:cluster-node-tuning-operator-us-tuned created  for-build=cluster-node-tuning-operator-us-tuned
INFO[2023-12-15T13:11:12Z] Tagging cluster-node-tuning-operator-us-tuned into stable 
INFO[2023-12-15T13:11:12Z] Ran for 3s                                   
ERRO[2023-12-15T13:11:12Z] Some steps failed:                           
ERRO[2023-12-15T13:11:12Z] 
  * could not run steps: step cluster-node-tuning-operator failed: error occurred handling build cluster-node-tuning-operator-amd64: the build cluster-node-tuning-operator-amd64 failed after 1m9s with reason DockerBuildFailed: Dockerfile build strategy has failed.
Writing manifest to image destination
Adding transient rw bind mount for /run/secrets/rhsm
[1/2] STEP 1/5: FROM image-registry.openshift-image-regist...51b888c7acbd69728c1eac3a7e858f0e1755e396d81302ec AS builder
Trying to pull image-registry.openshift-image-registry.svc...a089260e51b888c7acbd69728c1eac3a7e858f0e1755e396d81302ec...
error: build error: creating build container: parsing image configuration: fetching blob: StatusCode: 400, 
INFO[2023-12-15T13:11:12Z] Reporting job state 'failed' with reason 'executing_graph:step_failed:building_project_image' 
```

PR #879, opened 2023-12-12
OCPBUGS-24929: Use go 1.21

jmencak:
Changed go.mod and ran "go mod tidy".

Also see: https://github.com/openshift/cluster-node-tuning-operator/pull/873

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift/cluster-node-tuning-operator
 
-go 1.20
+go 1.21
 
 require (
 	github.com/RHsyseng/operator-utils v0.0.0-20200213165520-1a022eb07a43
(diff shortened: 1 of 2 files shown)


ffromani:
do we require matching changes in `openshift/release`?

jmencak:
> do we require matching changes in `openshift/release`?

I already [did some](https://github.com/openshift/release/pull/46678), but to be honest, not sure that's all.

ffromani:
> > do we require matching changes in `openshift/release`?
> 
> I already [did some](https://github.com/openshift/release/pull/46678), but to be honest, not sure that's all.

fair enough. Let's see what CI thinks and then we'll see how to move forward.

ffromani:
/lgtm

CI likes it, and seems fine to me.

PR #881, opened 2023-12-13
NO-JIRA: e2e-test: Call SetLogger to avoid controller-runtime error.

jlojosnegros:
You must set a logger for controller-runtime, otherwise it complains.

### DIFF
--- a/test/e2e/basic/operator_test.go
+++ b/test/e2e/basic/operator_test.go
@@ -1,10 +1,15 @@
 package e2e
 
 import (
+	"log"
+	"os"
 	"testing"
 
+	"github.com/go-logr/stdr"
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
+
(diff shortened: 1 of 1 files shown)


jmencak:
Thank you for the change.
/approve
/lgtm

ffromani:
/lgtm

what about the other suites? are them good?

jmencak:
> /lgtm
> 
> what about the other suites? are them good?

They don't use controller-runtime, at least not in core NTO.

jlojosnegros:
> /lgtm
> 
> what about the other suites? are them good?

It seems so. 
I managed to reproduce the issue using a time.Sleep(45 seconds) on the affected test suite, so I did the same with all the other e2e test suites and no one got affected, so I assume they are ok.

ffromani:
I suspect the perfprof suites may hit the same issue, but since I don't have any hard data to back my claim, I guess it's fine like it is now.

PR #884, opened 2023-12-14
WIP: performance: e2e: configure adaptive cpu in profile

shajmakh:
We've been observing lately that some tests that involve disabling load balancing are failing (like 32646) because the expected result does not have specific anticipated CPUs. After investigation, it turns out that one factor is the profile configuration of the CPU distribution.

PAO functional tests configure fixed CPU values under the PP. This is considered misconfiguration, especially when the system has more than 4 CPUs, and there is no guarantee that the functionality of the performance profile controller will work adequately with not all cpus reflected in the CPU section in the PP.

To fix this, we check the actual CPU capacity on a worker node and divide that between reserved and isolated (required cpu fields in PP).

### DIFF
--- a/test/e2e/performanceprofile/functests/0_config/config.go
+++ b/test/e2e/performanceprofile/functests/0_config/config.go
@@ -5,7 +5,11 @@ import (
 	"fmt"
 	"github.com/jaypipes/ghw/pkg/cpu"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/profilecreator"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
 	"os"
+	"regexp"
+	"strconv"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -253,10 +257,15 @@ func testProfile() (*performancev2.PerformanceProfile, error) {
 	return profile, nil
 }
 
-func getSysInfoForPPC() profilecreator.SystemInfo {
+func getSysInfoForPPC() (profilecreator.SystemInfo, error) {
+	totalCores, totalThreads, err := getTotalCPUInfo()
+	if err != nil {
+		return profilecreator.SystemInfo{}, err
+	}
+
 	cpuInfo := cpu.Info{
-		TotalCores:   getTotalcores(),
-		TotalThreads: getTotalThreads(),
+		TotalCores:   uint32(totalCores),
+		TotalThreads: uint32(totalThreads),
 		Processors:   processors,
 	}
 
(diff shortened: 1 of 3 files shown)


test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -3,6 +3,7 @@ package __performance_config
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
```
Tal-or, line 6: Please move it down along with the other relevant imports

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -158,9 +164,18 @@ func externalPerformanceProfile(performanceManifest string) (*performancev2.Perf
...
+	workerNodes, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
+	capacityCPU, _ := workerNodes[0].Status.Capacity.Cpu().AsInt64()
+	//consider cpu siblings - avoid noisy neighbor case
+	if capacityCPU < 4 {
+		//should never reach but still
+		return nil, fmt.Errorf("insufficient cpus on a node to create a valid performanceprofile for the tests, found %d cpus, expects at least 4", capacityCPU)
+	}
+	reserved := performancev2.CPUSet(fmt.Sprintf("0,%d", capacityCPU/2))
+	isolated := performancev2.CPUSet(fmt.Sprintf("1-%d,%d-%d", capacityCPU/2-1, capacityCPU/2+1, capacityCPU-1))
```
Tal-or, line 178: Are 2 reserved CPUs are enough for machine with a big amount of cpus?

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -158,9 +164,18 @@ func externalPerformanceProfile(performanceManifest string) (*performancev2.Perf
...
 }
 
-func testProfile() *performancev2.PerformanceProfile {
-	reserved := performancev2.CPUSet("0")
-	isolated := performancev2.CPUSet("1-3")
+func testProfile() (*performancev2.PerformanceProfile, error) {
+	//Get cpus capacity on a node; typically it is same for all nodes
+	workerNodes, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
+	capacityCPU, _ := workerNodes[0].Status.Capacity.Cpu().AsInt64()
```
Tal-or, line 171: I would validate that all workers are homogeneous and failed other wise.

test/e2e/performanceprofile/functests/0_config/config.go
mrniranjan: This will cause issues on BM because on HT enabled if we don't pick cpus from the same core.

mrniranjan: Also the same for isolated, In terms of upstream where there are only 4 cpus, No Numa, this works , but as we use this on multinuma systems with HT enabled, this gets complicated. I strongly suggest to use PPC . Otherwise you will have to duplicate what ppc does here.

test/e2e/performanceprofile/functests/0_config/config.go
ffromani: yes please; also, as unlikely it could be, let's handle the error here

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -158,9 +164,18 @@ func externalPerformanceProfile(performanceManifest string) (*performancev2.Perf
 	return profile, nil
 }
 
-func testProfile() *performancev2.PerformanceProfile {
-	reserved := performancev2.CPUSet("0")
-	isolated := performancev2.CPUSet("1-3")
+func testProfile() (*performancev2.PerformanceProfile, error) {
+	//Get cpus capacity on a node; typically it is same for all nodes
+	workerNodes, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
+	Expect(err).ToNot(HaveOccurred())
```
ffromani, line 170: please either use `Expect` in the code (but please use `ExpectWithOffset` OR return error and let the caller handle - but let's not do both

test/e2e/performanceprofile/functests/0_config/config.go
ffromani: Using PPC can work here, even though it will make the lane setup significantly more complex. But I agree we're reimplementing PPC.
The best IMO would be reuse part of PPC code, but we would need more detailed HW information to enable PPC.

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -158,9 +164,18 @@ func externalPerformanceProfile(performanceManifest string) (*performancev2.Perf
...
+	capacityCPU, _ := workerNodes[0].Status.Capacity.Cpu().AsInt64()
+	//consider cpu siblings - avoid noisy neighbor case
+	if capacityCPU < 4 {
+		//should never reach but still
+		return nil, fmt.Errorf("insufficient cpus on a node to create a valid performanceprofile for the tests, found %d cpus, expects at least 4", capacityCPU)
+	}
+	reserved := performancev2.CPUSet(fmt.Sprintf("0,%d", capacityCPU/2))
```
ffromani, line 177: I strongly recommend to use CPUSets and do logic with that instead of using just numbers

ffromani (commented):
I side with @mrniranjan here. The main problem is we're reimplementing PPC.
Best would be to either reuse (part of) PPC code or just use PPC as prerequiste before the testsuite starts.

I don't have strong opinions

ffromani:
another option we could explore is somehow a middleground:

1. get the cpu topology (same format as PPC) as parameter of the config lane.
2. iff the cpu topology is given, THEN the code will assume all the workers have the same topology. It will be responsability of the lane maintainer to enforce this and to ship up to date topology data
3. reuse (possibly refactoring the code) the PPC code to consume the given data to optimally compute (parts of) the PPC profile.

It is not fully automated (e.g. just run must-gather and let PPC do its magic) but would be a step in the right direction, would be more automated than it is now and should be deliverable quickly as fix.

ffromani:
> another option we could explore is somehow a middleground:
> 
>     1. get the cpu topology (same format as PPC) as parameter of the config lane.
> 
>     2. iff the cpu topology is given, THEN the code will assume all the workers have the same topology. It will be responsability of the lane maintainer to enforce this and to ship up to date topology data
> 
>     3. reuse (possibly refactoring the code) the PPC code to consume the given data to optimally compute (parts of) the PPC profile.
> 
> 
> It is not fully automated (e.g. just run must-gather and let PPC do its magic) but would be a step in the right direction, would be more automated than it is now and should be deliverable quickly as fix.

however we will need to check how to actually ship the data and make it accessible to the 0_config suite

shajmakh:
@Tal-or @mrniranjan @ffromani thank you for the valuable reviews and comments. you highlighted legit points about using PPC. however, this PR is intended to address the blocker CI failure on cnf-feature-deploy, considering that d/s CI it is in fact using PPC to generate the profile. I'm researching to see if we can enhance this even more to satisfy all needs taking into account the urgency of this for u/s.

ffromani:
> @Tal-or @mrniranjan @ffromani thank you for the valuable reviews and comments. you highlighted legit points about using PPC. however, this PR is intended to address the blocker CI failure on cnf-feature-deploy, considering that d/s CI it is in fact using PPC to generate the profile. I'm researching to see if we can enhance this even more to satisfy all needs taking into account the urgency of this for u/s.

The urgency IS a factor indeed, but the main blocker for this approach is that we need to take into account HT and NUMA cpu affinity when deciding the reserved cpus. Otherwise the perfprof will be incorrect for other reasons, perhaps lesser reasons than the current state but not much better; the NUMA allocation is perhaps debatablem but HT is a requirement for which we have e2e tests for.

I understand both camps and I don't have strong opinions yet.

shajmakh:
After reassessing the cost of the approach described here, I am closing this in favor of https://github.com/openshift/cluster-node-tuning-operator/pull/909

PR #889, opened 2023-12-18
PP: render add UID to PP if needed.

jlojosnegros:
When rendering manifests from PerformanceProfile we add an OwnerReference to them.
To do so we need the PerformanceProfile's UID, and if there is none we create a new one.
We need to add that new UID to PerformanceProfile to avoid problems when creating those manifests in a cluster.

IMPORTANT: Now we write ALL the input PerformanceProfiles handled to the output folder.

### DIFF
--- a/pkg/performanceprofile/cmd/render/render.go
+++ b/pkg/performanceprofile/cmd/render/render.go
@@ -87,11 +87,14 @@ func render(inputDir, outputDir string) error {
 	}
 
 	var (
-		perfProfiles []*performancev2.PerformanceProfile
-		mcPools      []*mcfgv1.MachineConfigPool
-		mcConfigs    []*mcfgv1.MachineConfig
-		infra        *apicfgv1.Infrastructure
-		ctrcfgs      []*mcfgv1.ContainerRuntimeConfig
+		perfProfiles []struct {
+			string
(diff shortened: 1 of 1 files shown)


pkg/performanceprofile/cmd/render/render.go
```diff
@@ -222,6 +223,21 @@ func render(inputDir, outputDir string) error {
 			}
 			klog.Info(fileName)
 		}
+
+		//write PerformanceProfile to output
+		ppBin, err := yaml.Marshal(pp)
+		if err != nil {
+			return fmt.Errorf("unable to marshal PerformanceProfile %s. error: %w", pp.Name, err)
+		}
+		fileName := fmt.Sprintf("%s_pp.yaml", pp.Name)
```
MarSik, line 232: We should use the same name we got from the input. That will allow clean overwrite in the installer.

jlojosnegros: yeah makes sense.
thanks for the comment

jlojosnegros:
This approach is not good so this PR is no longer needed.
Check https://github.com/openshift/cluster-node-tuning-operator/pull/907 for the way to go.

PR #890, opened 2023-12-18
OCPBUGS-24636: Report duplicate priority only for multiple matching profiles

rbaturov:
This fix addresses a false warning that is dumped to logs in the NTO tuned profile selection process.
Deploying 2 performance profiles on two different nodes, resulting in false priority warnings that were dumped to logs. As part of the profile selection process, NTO checks if there are profiles that share the same priority, regardless of their associated node!
The current solution fixes this and priority warnings are raised only when multiple profiles with matching priorities are associated with the same node.
Note that the fixes address both the Openshift and hypershift platforms separately.

### DIFF
--- a/pkg/operator/profilecalculator.go
+++ b/pkg/operator/profilecalculator.go
@@ -167,55 +167,100 @@ func (pc *ProfileCalculator) calculateProfile(nodeName string) (string, map[stri
 		return "", nil, operand, fmt.Errorf("failed to list Tuned: %v", err)
 	}
 
-	for _, recommend := range TunedRecommend(tunedList) {
-		var (
-			pools []*mcfgv1.MachineConfigPool
-			node  *corev1.Node
-		)
+	recommendAll := TunedRecommend(tunedList)
+	recommendProfile := func(nodeName string, iStart int) (int, string, map[string]string, tunedv1.OperandConfig, error) {
+		var i int
+		for i = iStart; i < len(recommendAll); i++ {
+			var (
+				pools []*mcfgv1.MachineConfigPool
+				node  *corev1.Node
+			)
+			recommend := recommendAll[i]
 
-		// Start with node/pod label based matching to MachineConfig matching when
-		// both the match section and MachineConfigLabels are specified.
-		// Also note the catch-all functionality when "recommend.Match == nil",
-		// we do not want to call profileMatches() in that case unless machineConfigLabels
-		// is undefined.
-		if (recommend.Match != nil || recommend.MachineConfigLabels == nil) && pc.profileMatches(recommend.Match, nodeName) {
-			return *recommend.Profile, nil, recommend.Operand, nil
+			// Start with node/pod label based matching to MachineConfig matching when
+			// both the match section and MachineConfigLabels are specified.
+			// Also note the catch-all functionality when "recommend.Match == nil",
+			// we do not want to call profileMatches() in that case unless machineConfigLabels
+			// is undefined.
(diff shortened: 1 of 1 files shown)


jmencak:
/cc @jmencak 
@rbaturov please see the "make verify" target, it fails. You need code formatting adjustments, it should be as simple as running gofmt on the code.

As for the HyperShift testing. If this proves too tricky, we could potentially do this after the fact and rollback if we see issues.

pkg/operator/profilecalculator.go
```diff
@@ -158,7 +158,9 @@ func (pc *ProfileCalculator) nodeChangeHandler(nodeName string) (bool, error) {
 // * whether to run the Tuned daemon in debug mode on node nodeName
 // * an error if any
 func (pc *ProfileCalculator) calculateProfile(nodeName string) (string, map[string]string, tunedv1.OperandConfig, error) {
-	var operand tunedv1.OperandConfig
+	var (
+		operand tunedv1.OperandConfig
+	)
```
ffromani, line 163: why do we need this change?

pkg/operator/profilecalculator.go
```diff
@@ -167,55 +169,100 @@ func (pc *ProfileCalculator) calculateProfile(nodeName string) (string, map[stri
 		return "", nil, operand, fmt.Errorf("failed to list Tuned: %v", err)
 	}
 
-	for _, recommend := range TunedRecommend(tunedList) {
-		var (
-			pools []*mcfgv1.MachineConfigPool
-			node  *corev1.Node
-		)
+	recommendAll := TunedRecommend(tunedList)
+	r := func(nodeName string, iStart int) (int, string, map[string]string, tunedv1.OperandConfig, error) {
```
ffromani, line 173: let's use a more descriptive name - and in any case let's try hard to avoid single-letter identifiers except for some loop variables

ffromani (commented):
initial review, the logic is surprisingly complex so I'll need more time to comment the actual logic

pkg/operator/profilecalculator.go
jmencak: We do not. :) I think it is an artifact of me messing with the code, adding variables, then removing them and forgetting to rollback to the original.

jmencak:
/approve

I cannot say LGTM, because this is largely based on my code.  Offering a commit message and a description I had:


    Report duplicate priority only for multiple matching profiles
    
    Originally, any profile matching rules with the same priority would be
    reported as warnings.  We do not necessarily need to issue a warning
    unless a Node has multiple matching profiles with the same priority.
    
    Resolves: OCPBUGS-24636

pkg/operator/profilecalculator.go
MarSik: The lambda is actually rather long, I think giving it a proper top level name might be more readable.

jmencak: FWIR, it would require bigger code changes to remove the lambda and move it outside of the function, but leaving this to Ronny. My suggestion is to just rename the `r()` to something like `recommendProfile()` and perhaps document the parameters and values returned.

rbaturov: @MarSik I was thinking of moving the lambda to the top level but it does require passing more objects as parameters and a little more effort. Note this would be required to happen for both `calculateProfile` and `calculateProfileHyperShift` implementations (As their `r` function is a bit different). If you think this would be more readable to create these two outer functions I will make the changes

pkg/operator/profilecalculator.go
rbaturov: @ffromani Changed this back to one line.

PR #892, opened 2023-12-19
CNF-9173: e2e: mixedcpus test

Tal-or:
This is a first PR (more to come) that tests the mixed-cpus e2e.
It contains the following tests:
1. Basic test to verify that pod which asks for shared cpus, has the shared cpus under
its cgroup cpuset.

2. Check that `cpu-quota.crio.io: disable` annotation works along with mixedcpus. 
3. Check that `OPENSHIFT_ISOLATED_CPUS` and `OPENSHIFT_SHARED_CPUS` env variables are under the container and configured properly.
4. Verify that updates to shared CPUs under the profile are reflect under the containers correctly.

### DIFF
--- a/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
+++ b/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
@@ -84,11 +84,10 @@ var _ = Describe("Mixedcpus", Ordered, func() {
 			Expect(testclient.Client.Get(ctx, key, kc)).ToNot(HaveOccurred())
 			k8sKc := &kubeletconfig.KubeletConfiguration{}
 			Expect(json.Unmarshal(kc.Spec.KubeletConfig.Raw, k8sKc)).ToNot(HaveOccurred())
-			// don't need to check the error, the values were already validated.
-			reserved, _ := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
-			shared, _ := cpuset.Parse(string(*profile.Spec.CPU.Shared))
-			reservedSystemCpus, _ := cpuset.Parse(k8sKc.ReservedSystemCPUs)
-			Expect(reservedSystemCpus.Equals(reserved.Union(shared))).To(BeTrue(), "reservedSystemCPUs should contain the shared cpus; reservedSystemCPUs=%q reserved=%q shared=%q",
+			reserved := mustParse(string(*profile.Spec.CPU.Reserved))
+			shared := mustParse(string(*profile.Spec.CPU.Shared))
+			reservedSystemCpus := mustParse(k8sKc.ReservedSystemCPUs)
+			Expect(reservedSystemCpus.Equals(reserved.Union(*shared))).To(BeTrue(), "reservedSystemCPUs should contain the shared cpus; reservedSystemCPUs=%q reserved=%q shared=%q",
 				reservedSystemCpus.String(), reserved.String(), shared.String())
 		})
 
@@ -107,11 +106,10 @@ var _ = Describe("Mixedcpus", Ordered, func() {
 			cpus, err := nodes.ExecCommandOnNode(ctx, cmd, worker)
 			Expect(err).ToNot(HaveOccurred(), "failed to execute command on node; cmd=%q node=%q", cmd, worker)
 			cpus = strings.Trim(cpus, "\n")
-			crioShared, err := cpuset.Parse(cpus)
-			Expect(err).ToNot(HaveOccurred())
+			crioShared := mustParse(cpus)
 			// don't need to check the error, the values were already validated.
-			shared, _ := cpuset.Parse(string(*profile.Spec.CPU.Shared))
-			Expect(shared.Equals(crioShared)).To(BeTrue(), "crio config file does not contain the expected shared cpuset; shared=%q crioShared=%q",
+			shared := mustParse(string(*profile.Spec.CPU.Shared))
+			Expect(shared.Equals(*crioShared)).To(BeTrue(), "crio config file does not contain the expected shared cpuset; shared=%q crioShared=%q",
 				shared.String(), crioShared.String())
 		})
 	})
@@ -149,10 +147,10 @@ var _ = Describe("Mixedcpus", Ordered, func() {
 				cfg := &controller.CpuSet{}
 				err = getter.Container(ctx, p, p.Spec.Containers[0].Name, cfg)
 				Expect(err).ToNot(HaveOccurred())
-				cgroupCpuSet, err := cpuset.Parse(cfg.Cpus)
+				cgroupCpuSet := mustParse(cfg.Cpus)
 				Expect(err).ToNot(HaveOccurred())
-				shared, _ := cpuset.Parse(string(*profile.Spec.CPU.Shared))
-				Expect(cgroupCpuSet.Intersection(shared).List()).ToNot(BeEmpty(), "shared cpus are not in the pod cgroups; pod=%q, cgroupscpuset=%q sharedcpuset=%q",
+				shared := mustParse(string(*profile.Spec.CPU.Shared))
+				Expect(cgroupCpuSet.Intersection(*shared).List()).ToNot(BeEmpty(), "shared cpus are not in the pod cgroups; pod=%q, cgroupscpuset=%q sharedcpuset=%q",
 					fmt.Sprintf("%s/%s", p.Namespace, p.Name), cgroupCpuSet.String(), shared.String())
 			})
 			It("should be able to disable cfs_quota", func() {
@@ -187,13 +185,12 @@ var _ = Describe("Mixedcpus", Ordered, func() {
 				output, err := pods.ExecCommandOnPod(testclient.K8sClient, p, "", cmd)
 				Expect(err).ToNot(HaveOccurred(), "failed to execute command on pod; cmd=%q pod=%q", cmd, client.ObjectKeyFromObject(p).String())
 				isolatedAndShared := strings.Split(string(output), "\r\n")
-				isolated, err := cpuset.Parse(isolatedAndShared[0])
-				Expect(err).ToNot(HaveOccurred())
+				isolated := mustParse(isolatedAndShared[0])
 				Expect(isolated.IsEmpty()).ToNot(BeTrue())
-				shared, err := cpuset.Parse(isolatedAndShared[1])
-				ppShared, _ := cpuset.Parse(string(*profile.Spec.CPU.Shared))
+				shared := mustParse(isolatedAndShared[1])
+				ppShared := mustParse(string(*profile.Spec.CPU.Shared))
 				Expect(err).ToNot(HaveOccurred())
-				Expect(shared.Equals(ppShared)).To(BeTrue(), "OPENSHIFT_SHARED_CPUS value not equal to what configure in the performance profile."+
+				Expect(shared.Equals(*ppShared)).To(BeTrue(), "OPENSHIFT_SHARED_CPUS value not equal to what configure in the performance profile."+
 					"OPENSHIFT_SHARED_CPUS=%s spec.cpu.shared=%s", shared.String(), ppShared.String())
 			})
 		})
@@ -268,8 +265,7 @@ func setup(ctx context.Context) func(ctx2 context.Context) {
 	initialProfile := profile.DeepCopy()
 
 	if !profileutil.IsMixedCPUsEnabled(profile) {
-		isolated, err := cpuset.Parse(string(*profile.Spec.CPU.Isolated))
-		Expect(err).ToNot(HaveOccurred())
+		isolated := mustParse(string(*profile.Spec.CPU.Isolated))
 
 		// arbitrary take the first one
 		sharedcpu := cpuset.New(isolated.List()[0])
--- a/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
+++ b/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
@@ -175,6 +175,30 @@ var _ = Describe("Mixedcpus", Ordered, func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(cfg.Quota).To(Or(Equal("max"), Equal("-1")))
 		})
+		It("should have OPENSHIFT_ISOLATED_CPUS and OPENSHIFT_SHARED_CPUS env variables under the container", func() {
+			rl := &corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("1"),
+				corev1.ResourceMemory: resource.MustParse("100Mi"),
+				sharedCpusResource:    resource.MustParse("1"),
+			}
+			p, err := createPod(ctx, testclient.Client, testutils.NamespaceTesting,
+				withRequests(rl),
+				withLimits(rl),
+				withRuntime(components.GetComponentName(profile.Name, components.ComponentNamePrefix)))
+			Expect(err).ToNot(HaveOccurred())
+			cmd := printMixedCPUsEnvCmd()
+			output, err := pods.ExecCommandOnPod(testclient.K8sClient, p, "", cmd)
+			Expect(err).ToNot(HaveOccurred(), "failed to execute command on pod; cmd=%q pod=%q", cmd, client.ObjectKeyFromObject(p).String())
+			isolatedAndShared := strings.Split(string(output), "\r\n")
+			isolated, err := cpuset.Parse(isolatedAndShared[0])
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isolated.IsEmpty()).ToNot(BeTrue())
+			shared, err := cpuset.Parse(isolatedAndShared[1])
+			ppShared, _ := cpuset.Parse(string(*profile.Spec.CPU.Shared))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(shared.Equals(ppShared)).To(BeTrue(), "OPENSHIFT_SHARED_CPUS value not equal to what configure in the performance profile."+
+				"OPENSHIFT_SHARED_CPUS=%s spec.cpu.shared=%s", shared.String(), ppShared.String())
+		})
 	})
 })
 
@@ -216,6 +240,14 @@ func setup(ctx context.Context, profile *performancev2.PerformanceProfile) func(
 func cpuSetToPerformanceCPUSet(set *cpuset.CPUSet) *performancev2.CPUSet {
 	c := performancev2.CPUSet(set.String())
 	return &c
+}
+
+func printMixedCPUsEnvCmd() []string {
+	return []string{
+		"/bin/printenv",
+		"OPENSHIFT_ISOLATED_CPUS",
+		"OPENSHIFT_SHARED_CPUS",
+	}
 }
 
 // checks whether file exists and not empty
--- a/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
+++ b/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -115,7 +116,7 @@ var _ = Describe("Mixedcpus", Ordered, func() {
 		})
 	})
 
-	When("workloads requests access for shared cpus", func() {
+	Context("single workload - request validation", func() {
 		var profile *performancev2.PerformanceProfile
 		var getter cgroup.ControllersGetter
 		BeforeEach(func() {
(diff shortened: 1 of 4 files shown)


Tal-or:
depends on https://github.com/cri-o/cri-o/pull/7608

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (m *Manager) GetConfig(ctx context.Context, c *kubernetes.Clientset, pod *corev1.Pod, containerName string) (*config.Config, error) {
+	var cmd []string
+	cfg := &config.Config{}
+	cmd = []string{
+		"/bin/cat",
+		config.CgroupMountPoint + "/cpuset/cpuset.cpus",
+		config.CgroupMountPoint + "/cpu/cpu.cfs_quota_us",
+		config.CgroupMountPoint + "/cpu/cpu.cfs_period_us",
+	}
```
mrniranjan, line 28: when crun is enabled , this creates a subdirectory called container which contains all the cpuset interface files,  Can we get those interface files too ?

Tal-or: Yes. As discussed offline we'll have separate PR for it

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -0,0 +1,38 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/config"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
+)
+
+type Manager struct {
+}
+
+func (m *Manager) GetConfig(ctx context.Context, c *kubernetes.Clientset, pod *corev1.Pod, containerName string) (*config.Config, error) {
```
mrniranjan, line 20: Similary we can get another Get method which will return interface files from the Node , specifically we need to see if these interface files have right values in pods scope

Tal-or: @mrniranjan What do you mean by interface files?

mrniranjan: ah sorry i commented here,  any how by interface files i mean cpuset.cpus (is an interface file ) for the cpuset controller , just a technical name. 
cgroup-> Controllers(cpuset, cpu, memory, hugetlb, pids) and files inside these controller files are called interface files

Tal-or:
/hold
depends on #906

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -200,18 +194,46 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+		testlog.Infof("enable mixed cpus for profile %q", profile.Name)
+		updateNeeded = true
+	} else {
+		testlog.Infof("mixed cpus already enabled for profile %q", profile.Name)
+	}
+
+	workers, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
 	Expect(err).ToNot(HaveOccurred())
+	for _, worker := range workers {
+		if worker.Status.Capacity.Cpu().Size() == numberOfCoresThatRequiredCancelingSMTAlignment {
```
ffromani, line 325: maybe you want to have  a `>=` check rather than exact equality? exact equality is oftentimes fragile

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -200,18 +194,46 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+			profile.Annotations = map[string]string{
+				"kubeletconfig.experimental": `{"cpuManagerPolicyOptions": {"full-pcpus-only": "false"}}`,
+			}
+			testlog.Infof("cancaling SMT alignment for nodes under profile %q", profile.Name)
```
ffromani, line 329: typo: `canceling`

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: That was my thought process as well, but less than `numberOfCoresThatRequiredCancelingSMTAlignment` (4) the test would failed anyway.
But yes we can change that, no biggie

test/e2e/performanceprofile/functests/utils/namespaces/namespaces.go
```diff
@@ -30,6 +30,9 @@ func init() {
 var TestingNamespace = &corev1.Namespace{
 	ObjectMeta: metav1.ObjectMeta{
 		Name: testutils.NamespaceTesting,
+		Annotations: map[string]string{
```
shajmakh, line 33: since this is mixed-cpus feature related only and is not needed by default. do we verify it somewhere in the mixedcpus suite? I understand the functional tests that require workload will fail, but I am trying t find out if this specific detail is covered in testing

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -162,3 +228,52 @@ func isFileExistCmd(absoluteFileName string) []string {
 		fmt.Sprintf("if [[ -s %s ]]; then echo true; else echo false; fi", absoluteFileName),
 	}
 }
+
+func createPod(ctx context.Context, c client.Client, ns string, opts ...func(pod *corev1.Pod)) (*corev1.Pod, error) {
```
shajmakh, line 351: I see this a smart generic solution that deserves to be more public so it can be used also outside this package. maybe the below is a good place:
https://github.com/openshift/cluster-node-tuning-operator/blob/a3b866e0ec5576a8278fa1329e146494a7a4b8e9/test/e2e/performanceprofile/functests/utils/pods/pods.go

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -175,6 +175,30 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+			p, err := createPod(ctx, testclient.Client, testutils.NamespaceTesting,
+				withRequests(rl),
+				withLimits(rl),
+				withRuntime(components.GetComponentName(profile.Name, components.ComponentNamePrefix)))
+			Expect(err).ToNot(HaveOccurred())
+			cmd := printMixedCPUsEnvCmd()
+			output, err := pods.ExecCommandOnPod(testclient.K8sClient, p, "", cmd)
+			Expect(err).ToNot(HaveOccurred(), "failed to execute command on pod; cmd=%q pod=%q", cmd, client.ObjectKeyFromObject(p).String())
+			isolatedAndShared := strings.Split(string(output), "\r\n")
+			isolated, err := cpuset.Parse(isolatedAndShared[0])
```
shajmakh: addressing unknown elements can hit nil pointer, I believe we need to verify (expect()) that the length of isolatedAndShared is 2 and fail if not.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -224,12 +248,10 @@ func setup(ctx context.Context, profile *performancev2.PerformanceProfile) func(
...
 		By(fmt.Sprintf("executing teardown - revert profile %q back to its intial state", profile.Name))
 		Expect(testclient.Client.Get(ctx2, client.ObjectKeyFromObject(initialProfile), profile))
-		profile.Spec = initialProfile.Spec
-		resourceVersion := profile.ResourceVersion
-		testprofiles.UpdateWithRetry(profile)
+		testprofiles.UpdateWithRetry(initialProfile)
 
 		// do not wait if nothing has changed
-		if resourceVersion != profile.ResourceVersion {
+		if initialProfile.ResourceVersion != profile.ResourceVersion {
```
shajmakh, line 319: I understand this should always be true considering `if !updateNeeded`  at line 236, but I think we need to fetch the new PP again (instead initialProfile) before this comparison to compare the correct values.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -132,66 +133,130 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				p, err := createPod(ctx, testclient.Client, testutils.NamespaceTesting,
+					withRequests(rl),
+					withLimits(rl),
+					withRuntime(components.GetComponentName(profile.Name, components.ComponentNamePrefix)))
+				Expect(err).ToNot(HaveOccurred())
+				cmd := printMixedCPUsEnvCmd()
+				output, err := pods.ExecCommandOnPod(testclient.K8sClient, p, "", cmd)
+				Expect(err).ToNot(HaveOccurred(), "failed to execute command on pod; cmd=%q pod=%q", cmd, client.ObjectKeyFromObject(p).String())
+				isolatedAndShared := strings.Split(string(output), "\r\n")
+				isolated, err := cpuset.Parse(isolatedAndShared[0])
```
shajmakh, line 190: new place with same comment about verifying the length (am reviewing by commits)

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -132,66 +133,130 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
-			isolated, err := cpuset.Parse(isolatedAndShared[0])
-			Expect(err).ToNot(HaveOccurred())
-			Expect(isolated.IsEmpty()).ToNot(BeTrue())
-			shared, err := cpuset.Parse(isolatedAndShared[1])
-			ppShared, _ := cpuset.Parse(string(*profile.Spec.CPU.Shared))
-			Expect(err).ToNot(HaveOccurred())
-			Expect(shared.Equals(ppShared)).To(BeTrue(), "OPENSHIFT_SHARED_CPUS value not equal to what configure in the performance profile."+
-				"OPENSHIFT_SHARED_CPUS=%s spec.cpu.shared=%s", shared.String(), ppShared.String())
+
+		When("Modifying the shared CPUs in the performance profile", func() {
```
shajmakh, line 201: Can we add tags/labels to differentiate between spec blocks that check mixed-CPUs configuration and those that apply updates (slow and require reboot)?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -132,66 +133,130 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				rl := &corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("1"),
+					corev1.ResourceMemory: resource.MustParse("100Mi"),
+					sharedCpusResource:    resource.MustParse("1"),
+				}
+				p, err := createPod(ctx, testclient.Client, testutils.NamespaceTesting,
+					withRequests(rl),
+					withLimits(rl),
+					withRuntime(components.GetComponentName(profile.Name, components.ComponentNamePrefix)))
+				Expect(err).ToNot(HaveOccurred())
```
shajmakh, line 237: this is a good check but it is similar to the feature first configuration and verification. I think what even adds more value in the update test is to verify that a controller-related pod (e.g. deployment pod) gets updated with the new shared cpuset, given that an orphan pod will not survive node reboot.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -132,66 +133,130 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+
+				By("checking cgroups under the container")
+				ppShared := mustParse(string(*profile.Spec.CPU.Shared))
```
shajmakh, line 240: here we are comparing actual ppShared (because of [LOC 222](https://github.com/openshift/cluster-node-tuning-operator/pull/892/commits/78f8ae65fa117c188803f8dd0f2003251d5e676a#diff-a8e5e2d088216d6686138de966b7d90e5f2895bfa8e8b7fc6b9bcdd5b0397d69R222)) with actual pod values. This can hide (the test will pass) the bug of not having the PP shared cpus updated as expected which is the main goal of these set of tests.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -132,66 +133,130 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				cgroupCpuSet := mustParse(cfg.Cpus)
+				Expect(cgroupCpuSet.Intersection(*ppShared).List()).ToNot(BeEmpty(), "shared cpus are not in the pod cgroups; pod=%q, cgroupscpuset=%q sharedcpuset=%q",
+					fmt.Sprintf("%s/%s", p.Namespace, p.Name), cgroupCpuSet.String(), ppShared.String())
+
+				By("checking environment variable under the container")
+				cmd := printMixedCPUsEnvCmd()
+				output, err := pods.ExecCommandOnPod(testclient.K8sClient, p, "", cmd)
+				Expect(err).ToNot(HaveOccurred(), "failed to execute command on pod; cmd=%q pod=%q", cmd, client.ObjectKeyFromObject(p).String())
+				isolatedAndShared := strings.Split(string(output), "\r\n")
+				isolatedFromEnv := mustParse(isolatedAndShared[0])
```
shajmakh, line 253: same note regarding validating the length.

shajmakh (commented):
Thank you for this PR.
added few comments below.

test/e2e/performanceprofile/functests/utils/namespaces/namespaces.go
Tal-or: We can add negative test - deploy a pod in a namespace without the annotation and expect admission error, is that ok?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: Sure. Is there a list of standardize labels that we should use?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: What about the following flow:
1. create deployment with pod asks for shared cpus
2. check shared cpus reflected under the pod
3. change the shared cpus under the performance profile
4. check that the pod from step 1 reflect the new shared cpus

Is this type of test work for you?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: Not sure I follow, we are comparing the new shared cpus set from the profile with the test pod.
Could you please elaborate a little bit?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: I should replace it with `mustParse` which should add assertion if it failed to parse the cpuset

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: `mustParse` should assert in case of failure

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: We should use `mustParse` here, i'll fix that

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: `profile` contains the most updated profile and `initialProfile` contains the profile values before the call to `Update()` if something has changed we expect  
 `initialProfile.ResourceVersion != profile.ResourceVersion` to be `true` 
But, if we'll fetch the latest profile into `initialProfile` then `initialProfile.ResourceVersion != profile.ResourceVersion` will *always* be false even if something has changed.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: We might, but I suggest doing that in a separate PR because if we want it to be an "offical" API for creating tests pods in our functests, it requires more attention and additional discussion.

test/e2e/performanceprofile/functests/utils/namespaces/namespaces.go
shajmakh: will the feature still need this annotation when in GA?
unless this annotation will affect non mixed-cpus tests, I think no harm to keep it here by default. otherwise, I'd just preserve the default ns and adjust it in the suite when needed.
adding the negative test sounds good and is part of the TP which can be added later, no rush.

Tal-or: AFAIK the annotation would stay.
And it won't affect non mixed CPUs workload

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -102,28 +106,199 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				profile.Spec.CPU.Isolated = cpuSetToPerformanceCPUSet(&newIsolated)
+				profile.Spec.CPU.Shared = cpuSetToPerformanceCPUSet(&newShared)
+				By("applying new performanceProfile")
+				testprofiles.UpdateWithRetry(profile)
+				mcp, err := mcps.GetByProfile(profile)
+				Expect(err).ToNot(HaveOccurred())
+				By("waiting for mcp to catch up")
+				mcps.WaitForCondition(mcp, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
+				mcps.WaitForCondition(mcp, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
+				Expect(testclient.Client.Get(ctx, client.ObjectKeyFromObject(profile), profile))
```
shajmakh, line 219: can we make perform Expect() without condition? what is the default in this case?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
shajmakh: ah I see this is addressed in line 219 (but with expect with no conditions, left a comment there)

ffromani (commented):
I'm generally OK to merge this once other comments are addressed. I have mostly minor comment, but I think we really should add tests as mentioned in the inline comments

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
shajmakh: we are about to compare the updated value from actual PP instead of the expected. 
what will happen in a case updating the shared cpus wasn't actually reflected in the new profile? 
we need to compare the expected value with the actual one.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
shajmakh: right but that is a different error and it hides the detail of the actual root cause

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -110,6 +120,62 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			cfg := &controller.CpuSet{}
+			err = getter.Container(ctx, p, p.Spec.Containers[0].Name, cfg)
+			Expect(err).ToNot(HaveOccurred())
+			cgroupCpuSet, err := cpuset.Parse(cfg.Cpus)
+			Expect(err).ToNot(HaveOccurred())
+			shared, _ := cpuset.Parse(string(*profile.Spec.CPU.Shared))
+			Expect(cgroupCpuSet.Intersection(shared).List()).ToNot(BeEmpty(), "shared cpus are not in the pod cgroups; pod=%q, cgroupscpuset=%q sharedcpuset=%q",
+				fmt.Sprintf("%s/%s", p.Namespace, p.Name), cgroupCpuSet.String(), shared.String())
+		})
```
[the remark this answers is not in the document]
Tal-or: Yes, it's in my backlog, but at this point in time the test won't pass because https://github.com/cri-o/cri-o/pull/7716 not merged yet.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: Yes. it will fail the test in case the error returned from the called function (`Get` in this case) is not `nil`

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
shajmakh: right, I misread. thanks for elaborating

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: We can continue the discussion here: https://github.com/openshift/cluster-node-tuning-operator/pull/936

mrniranjan:
@Tal-or @shajmakh if these tests are important, we need to add test case id's and add it to our test case management systems.  Its not something very urgent can be addressed in another PR

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -331,3 +396,9 @@ func withRuntime(name string) func(p *corev1.Pod) {
 func getTestingNamespace() corev1.Namespace {
 	return *namespaces.TestingNamespace
 }
+
+func mustParse(cpus string) *cpuset.CPUSet {
+	set, err := cpuset.Parse(cpus)
+	ExpectWithOffset(1, err).ToNot(HaveOccurred(), "failed to parse cpuset; cpus=%q", cpus)
```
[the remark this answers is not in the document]
Tal-or: Done. Thanks.

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -30,8 +30,8 @@ func (cm *ControllersManager) CpuSet(ctx context.Context, pod *corev1.Pod, conta
 	dirPath := path.Join(controller.CgroupMountPoint, "cpuset", childName)
 	store := map[string]*string{
 		"cpuset.cpus":               &cfg.Cpus,
-		"cpuset.cpus.exclusive":     &cfg.Exclusive,
-		"cpuset.cpus.effective":     &cfg.Effective,
+		"cpuset.cpu_exclusive":      &cfg.Exclusive,
```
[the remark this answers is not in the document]
Tal-or: Done. Thanks

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: Fixed. Thanks

Tal-or:
> @Tal-or @shajmakh if these tests are important, we need to add test case id's and add it to our test case management systems. Its not something very urgent can be addressed in another PR

Yes we can add id's later one, you're welcome to provide list of valid ids and we shall add them.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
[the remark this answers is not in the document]
ffromani: fair enough, let's defer to a later PR

ffromani:
LGTM

shajmakh:
> > @Tal-or @shajmakh if these tests are important, we need to add test case id's and add it to our test case management systems. Its not something very urgent can be addressed in another PR
> 
> Yes we can add id's later on, you're welcome to provide list of valid ids and we shall add them.

indeed that are, and they're part of the TP so when imported to Polarion the ids will get generated and I'll consider to add them accordingly.

/lgtm
/hold 
for other reviewers 
@Tal-or feel free to unhold as you see fit.

Tal-or:
Thanks @shajmakh I'm doing some local tests on 4.15 and will remove the hold if they are passing

Tal-or:
/hold cancel
Tests are passing on 4.15 cgroupv1

PR #894, opened 2023-12-21
CNF-10750: Adding env var to enable feautregates ignore versions differentiation

rbaturov:
This modification has been added to enable manually upgrading NTO for developers who want to upgrade and test a custom NTO image.
In the current design, upgrading NTO to a specific version by disabling CVO and manually adjusting RELEASE_VERSION in the operator manifest results in a mismatch of the RELEASE_VERSION with the cluster version.
Therefore, feature gates will force NTO to exit.
The proposed solution bypasses this and will enable feature-gate logic to ignore the differentiation of the versions by setting OVERRIDE_RELEASE_VERSION to "true" in the NTO deployment manifest.

### DIFF
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -338,9 +338,11 @@ func migratePinnedSingleNodeInfraStatus(cfg *rest.Config, scheme *apiruntime.Sch
 func setupFeatureGates(ctx context.Context, config *rest.Config, operatorNamespace string) (featuregates.FeatureGate, error) {
 	missingVersion := "0.0.1-snapshot"
 	desiredVersion := missingVersion
-	if v, ok := os.LookupEnv(version.ReleaseVersionEnvVarName); ok {
-		desiredVersion = v
-	}
+	if val, ok := os.LookupEnv(version.OverrideReleaseVersionVarName); !ok || (ok && val != "true") {
+		if v, ok := os.LookupEnv(version.ReleaseVersionEnvVarName); ok {
+			desiredVersion = v
+		}
+	}	
 	configClient, err := configclient.NewForConfig(config)
 	if err != nil {
 		return nil, err
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -338,11 +338,12 @@ func migratePinnedSingleNodeInfraStatus(cfg *rest.Config, scheme *apiruntime.Sch
 func setupFeatureGates(ctx context.Context, config *rest.Config, operatorNamespace string) (featuregates.FeatureGate, error) {
 	missingVersion := "0.0.1-snapshot"
 	desiredVersion := missingVersion
-	if val, ok := os.LookupEnv(version.OverrideReleaseVersionVarName); !ok || (ok && val != "true") {
-		if v, ok := os.LookupEnv(version.ReleaseVersionEnvVarName); ok {
-			desiredVersion = v
-		}
-	}	
+	if v, ok := os.LookupEnv(version.ReleaseVersionEnvVarName); ok {
+		desiredVersion = v
+	}
+	if v, ok := os.LookupEnv(version.OverrideReleaseVersionVarName); ok && v == "true" {
+		desiredVersion = missingVersion
+	}
 	configClient, err := configclient.NewForConfig(config)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


cmd/cluster-node-tuning-operator/main.go
```diff
@@ -338,9 +338,11 @@ func migratePinnedSingleNodeInfraStatus(cfg *rest.Config, scheme *apiruntime.Sch
 func setupFeatureGates(ctx context.Context, config *rest.Config, operatorNamespace string) (featuregates.FeatureGate, error) {
 	missingVersion := "0.0.1-snapshot"
 	desiredVersion := missingVersion
-	if v, ok := os.LookupEnv(version.ReleaseVersionEnvVarName); ok {
-		desiredVersion = v
-	}
+	if val, ok := os.LookupEnv(version.OverrideReleaseVersionVarName); !ok || (ok && val != "true") {
```
Tal-or, line 341: let's simplify a bit:
```
if v, ok := os.LookupEnv(version.ReleaseVersionEnvVarName); ok {
    desiredVersion = v
}
if v, ok := os.LookupEnv(version.OverrideReleaseVersionVarName); ok && v == "true" {
    desiredVersion = missingVersion
}

```

Tal-or:
@rbaturov Please add the jira ticket number to the title

cmd/cluster-node-tuning-operator/main.go
rbaturov: Done

jmencak:
> In the current design, upgrading NTO to a specific version by disabling CVO and manually adjusting RELEASE_VERSION in the operator manifest results in a mismatch of the RELEASE_VERSION with the cluster version.

Why don't we use the "cluster version" for feature gates then? Is this because it doesn't change until all the operators have upgraded?

Tal-or:
> > In the current design, upgrading NTO to a specific version by disabling CVO and manually adjusting RELEASE_VERSION in the operator manifest results in a mismatch of the RELEASE_VERSION with the cluster version.
> 
> Why don't we use the "cluster version" for feature gates then? Is this because it doesn't change until all the operators have upgraded?

We use the "cluster version" ,which is the value in the `clusterversions.config.openshift.io` object, as a fallback in case the `desiredVersion` not mentioned specifically

jmencak:
> We use the "cluster version" ,which is the value in the `clusterversions.config.openshift.io` object, as a fallback in case the `desiredVersion` not mentioned specifically

What is the reason for using `RELEASE_VERSION` then?  I think it would be worth documenting the flow and use cases before we introduce another variable into the mix.  Let's cover all the use cases while keeping things simple.

jmencak:
> > We use the "cluster version" ,which is the value in the `clusterversions.config.openshift.io` object, as a fallback in case the `desiredVersion` not mentioned specifically
> 
> What is the reason for using `RELEASE_VERSION` by the featuregate then? I think it would be worth documenting the flow and use cases before we introduce another variable into the mix. Let's cover all the use cases while keeping things simple.

Any thoughts on this, @Tal-or / @rbaturov ?

jmencak:
After thinking about this a bit more, I believe this PR adds unnecessary complexity.  Manual/Quick NTO upgrade can still be done when a valid RELEASE_VERSION is specified.  Then corresponding feature gates are returned and the operator does not exit and I do not see OVERRIDE_RELEASE_VERSION bringing much value.

If you still want to follow this PR, the sentence in the description:
```
In the current design, upgrading NTO to a specific version by disabling CVO and manually adjusting RELEASE_VERSION in the operator manifest results in a mismatch of the RELEASE_VERSION with the cluster version. Therefore, feature gates will force NTO to exit.
```
needs to be corrected, because I do not believe it is true.  RELEASE_VERSION does not have to match the "clusterversion".  If you supply a valid RELEASE_VERSION, NTO will continue to work without exitting. 

If you have better ideas to bring back the old behaviour where any RELEASE_VERSION will work, happy to hear them.  If not, this PR can probably be closed.

Tal-or:
> After thinking about this a bit more, I believe this PR adds unnecessary complexity. Manual/Quick NTO upgrade can still be done when a valid RELEASE_VERSION is specified. Then corresponding feature gates are returned and the operator does not exit and I do not see OVERRIDE_RELEASE_VERSION bringing much value.
> 
> If you still want to follow this PR, the sentence in the description:
> 
> ```
> In the current design, upgrading NTO to a specific version by disabling CVO and manually adjusting RELEASE_VERSION in the operator manifest results in a mismatch of the RELEASE_VERSION with the cluster version. Therefore, feature gates will force NTO to exit.
> ```
> 
> needs to be corrected, because I do not believe it is true. RELEASE_VERSION does not have to match the "clusterversion". If you supply a valid RELEASE_VERSION, NTO will continue to work without exitting.
> 
> If you have better ideas to bring back the old behaviour where any RELEASE_VERSION will work, happy to hear them. If not, this PR can probably be closed.

@rbaturov Your call. From my side I'm ok with either approaches suggested by Jiri

PR #898, opened 2023-12-25
CNF-10786:e2e: general improvements and cleanup 

Tal-or:
Set of API updates, best practices adoption and cleanup to the e2e suite:
* Replacing pointer package API
* Passing contexts from the caller functions
* Removing unused functions

### DIFF
--- a/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
+++ b/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
@@ -66,7 +66,7 @@ var _ = Describe("Mixedcpus", Ordered, func() {
 			// test arbitrary one should be good enough
 			worker := &workers[0]
 			cmd := isFileExistCmd(kubeletMixedCPUsConfigFile)
-			found, err := nodes.ExecCommandOnNode(cmd, worker)
+			found, err := nodes.ExecCommandOnNode(ctx, cmd, worker)
 			Expect(err).ToNot(HaveOccurred(), "failed to execute command on node; cmd=%q node=%q", cmd, worker)
 			Expect(found).To(Equal("true"), "file not found; file=%q", kubeletMixedCPUsConfigFile)
 		})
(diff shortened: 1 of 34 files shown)


test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -131,7 +131,7 @@ func setup(ctx context.Context, profile *performancev2.PerformanceProfile) func(
 	mcps.WaitForCondition(mcp, machineconfigv1.MachineConfigPoolUpdating, corev1.ConditionTrue)
 	mcps.WaitForCondition(mcp, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
 
-	teardown := func(ctx context.Context) {
+	teardown := func() {
```
ffromani, line 134: why  we want to remove the `ctx` argument here?

ffromani (commented):
using `PollUntilContextTimeout` is a step in the right direction but it won't help us much until we complete the wiring, passing `ctx` from the top level down to the latest helper. I'm not sure we're there, even including this change. Nevertheless, let's do this step.

Tal-or:
> using `PollUntilContextTimeout` is a step in the right direction but it won't help us much until we complete the wiring, passing `ctx` from the top level down to the latest helper. I'm not sure we're there, even including this change. Nevertheless, let's do this step.

This is done on later commits in this PR.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: The `teardown` function is a value returned by the `setup` function and will be using the context that was passed initially to `setup` instead of accepting new one.

ffromani: this works, but it means that the same context (so possibly the same deadline) will cover setup + actual run (which spans between setup and teardown) + teardown. Previously we had a context for setup, a different context for teardown, and the actual run was not guarded by a context. Reducing to just one seems too much of a stretch.

Tal-or: Ack. Fixed that, thank you for your input

ffromani (commented):
/lgtm

ffromani:
/cc @jmencak 

because this PR is changing mostly perfprof-related code, but also some generic NTO tests

jmencak:
Thank you for the PR
/approve

PR #899, opened 2023-12-26
OCPBUGS-25699: irqbalance: set banned cpus list to '0'

Tal-or:
Empty irqbalance ban list means that irqbalance will be calculated by default.
The default will exclude all isolated and nohz_full cpus from the mask resulting in the irq’s being balanced over the reserved cpus only,
breaking the user intent.

CPU numbers which have their corresponding bits set to one in the ban mask
will not have any irq's assigned to them on rebalance.
Then we should set the mask to zero, so all cpus will participate in load balancing.


test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -313,7 +312,11 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 
 			origBannedCPUsFile := "/etc/sysconfig/orig_irq_banned_cpus"
 			By(fmt.Sprintf("Checking content of %q on node %q", origBannedCPUsFile, node.Name))
-			expectFileEmpty(node, origBannedCPUsFile)
+			fullPath := filepath.Join("/", "rootfs", origBannedCPUsFile)
```
yanirq, line 315: is the a reason to expand the expectFileEmpty function here ?     If so then you can delete the function itself since its only used here  (unless we need to keep this function extracted for other uses)

Tal-or: > since its only used here
Exactly the reason for inline the function

i'll delete the function forgot about it.

PR #906, opened 2024-01-04
CNF-9173: e2e: cgroups: introduce cgroup package

Tal-or:
This package provides a Getter which is cgroup version agnostic.
It provides a unified api for testing the mixedcpus feature for both v1 and v2.
    
The package can be expand in the future for testing additional features that requires cgroups inspection. (for example runc vs crun)

How to use:
```
getter := BuildGetter(context.TODO(), controllerRuntimeClient, k8sClient)
cpusetCfg := &controller.CpuSet{}
err := getter.Container(context.TODO(), pod, containerName, cpusetCfg)
// error handling
fmt.Println(cpusetCfg.Cpus)
```

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/hugepages.go
+++ b/test/e2e/performanceprofile/functests/1_performance/hugepages.go
@@ -148,7 +148,7 @@ var _ = Describe("[performance]Hugepages", Ordered, func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			cmd2 := []string{"/bin/bash", "-c", "tmux new -d 'LD_PRELOAD=libhugetlbfs.so HUGETLB_MORECORE=yes top -b > /dev/null'"}
-			_, err = pods.ExecCommandOnPod(testclient.K8sClient, testpod, cmd2)
+			_, err = pods.ExecCommandOnPod(testclient.K8sClient, testpod, "", cmd2)
 			Expect(err).ToNot(HaveOccurred())
 
 			By("checking free hugepages - one should be used by pod")
--- a/test/e2e/performanceprofile/functests/utils/cgroup/cgroup.go
+++ b/test/e2e/performanceprofile/functests/utils/cgroup/cgroup.go
@@ -1,0 +1,53 @@
+package cgroup
+
+import (
+	"context"
+	"fmt"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	apiconfigv1 "github.com/openshift/api/config/v1"
+	cgroupv1 "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/v1"
+	cgroupv2 "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/v2"
+)
+
+type ControllersGetter interface {
+	// Pod is for getting controller config at the pod level
+	Pod(ctx context.Context, pod *corev1.Pod, controllerConfig interface{}) error
+	// Container is for getting controller config at the container level
+	Container(ctx context.Context, pod *corev1.Pod, containerName string, controllerConfig interface{}) error
+	// Child is for getting controller config at the container's child level
+	Child(ctx context.Context, pod *corev1.Pod, containerName, childName string, controllerConfig interface{}) error
+}
+
+func IsVersion2(ctx context.Context, c client.Client) (bool, error) {
+	nodecfg := &apiconfigv1.Node{}
+	key := client.ObjectKey{
+		Name: "cluster",
+	}
+	err := c.Get(ctx, key, nodecfg)
+	if err != nil {
+		return false, fmt.Errorf("failed to get configs.node object. name=%q; %w", key.Name, err)
+	}
+	if nodecfg.Spec.CgroupMode == apiconfigv1.CgroupModeV1 {
+		return false, nil
+	}
+	// in case `apiconfigv1.CgroupMode` not set, it's returned true since
+	// the platform's default is cgroupV2
+	return true, nil
+}
+
+// BuildGetter return a cgroup information getter that complies with
+// the cgroup version that is currently used by the nodes on the cluster
+func BuildGetter(ctx context.Context, c client.Client, k8sClient *kubernetes.Clientset) (ControllersGetter, error) {
+	isV2, versionErr := IsVersion2(ctx, c)
+	if versionErr != nil {
+		return nil, versionErr
+	}
+	if isV2 {
+		return cgroupv2.NewManager(c, k8sClient), nil
+	} else {
+		return cgroupv1.NewManager(c, k8sClient), nil
+	}
+}
--- a/test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
+++ b/test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/controller"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/runtime"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
 )
 
@@ -26,35 +27,28 @@ func NewManager(c client.Client, k8sClient *kubernetes.Clientset) *ControllersMa
 
 func (cm *ControllersManager) CpuSet(ctx context.Context, pod *corev1.Pod, containerName, childName, runtimeType string) (*controller.CpuSet, error) {
 	cfg := &controller.CpuSet{}
-	dirPath := path.Join(controller.CgroupMountPoint, childName)
-	cmd := []string{
-		"/bin/cat",
-		dirPath + "/cpuset/cpuset.cpus",
-		dirPath + "/cpuset/cpuset.cpu_exclusive",
-		dirPath + "/cpuset/cpuset.effective_cpus",
-		dirPath + "/cpuset/cpuset.mems",
-		dirPath + "/cpuset/cpuset.sched_load_balance",
+	dirPath := path.Join(controller.CgroupMountPoint, "cpuset", childName)
+	store := map[string]*string{
+		"cpuset.cpus":               &cfg.Cpus,
+		"cpuset.cpus.exclusive":     &cfg.Exclusive,
+		"cpuset.cpus.effective":     &cfg.Effective,
+		"cpuset.sched_load_balance": &cfg.SchedLoadBalance,
+		"cpuset.mems":               &cfg.Mems,
 	}
-	b, err := pods.ExecCommandOnPod(cm.k8sClient, pod, containerName, cmd)
+	err := cm.execAndStore(pod, containerName, dirPath, store)
 	if err != nil {
 		return nil, fmt.Errorf("failed to retrieve cgroup config for pod. pod=%q, container=%q; %w", client.ObjectKeyFromObject(pod).String(), containerName, err)
 	}
-	output := strings.Split(string(b), "\r\n")
-	cfg.Cpus = output[0]
-	cfg.Exclusive = output[1]
-	cfg.Effective = output[2]
-	cfg.Mems = output[3]
-	cfg.SchedLoadBalance = output[4] == "1"
 	return cfg, nil
 }
 
 func (cm *ControllersManager) Cpu(ctx context.Context, pod *corev1.Pod, containerName, childName, runtimeType string) (*controller.Cpu, error) {
 	cfg := &controller.Cpu{}
-	dirPath := path.Join(controller.CgroupMountPoint, childName)
+	dirPath := path.Join(controller.CgroupMountPoint, "cpu", childName)
 	cmd := []string{
 		"/bin/cat",
-		dirPath + "/cpu/cpu.cfs_quota_us",
-		dirPath + "/cpu/cpu.cfs_period_us",
+		dirPath + "/cpu.cfs_quota_us",
+		dirPath + "/cpu.cfs_period_us",
 	}
 	b, err := pods.ExecCommandOnPod(cm.k8sClient, pod, containerName, cmd)
 	if err != nil {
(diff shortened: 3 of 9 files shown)


Tal-or:
This PR was part of #892 which we decide to extract in order merge it more quickly since it needed for the cgroupv2 testing

test/e2e/performanceprofile/functests/utils/cgroup/config/config.go
```diff
@@ -0,0 +1,20 @@
...
+
+const CgroupMountPoint = "/sys/fs/cgroup"
+
+type Config struct {
+	CPUSet string
+	Quota  string
+	Period string
+}
+
+type Getter interface {
```
ffromani, line 18: I'd actually get rid of this interface entirely.
Plus, in golang better to define interface on the consumer side: https://www.thoughtworks.com/insights/blog/programming-languages/mistakes-to-avoid-when-coming-from-an-object-oriented-language

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -0,0 +1,38 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/config"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
+)
+
+type Manager struct {
+}
+
+func (m *Manager) GetConfig(ctx context.Context, c *kubernetes.Clientset, pod *corev1.Pod, containerName string) (*config.Config, error) {
```
ffromani, line 20: single function interface suggests you may want to return... a function rather than an object. Unless there's a very clear path about adding more methods

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -0,0 +1,38 @@
...
+		config.CgroupMountPoint + "/cpuset/cpuset.cpus",
+		config.CgroupMountPoint + "/cpu/cpu.cfs_quota_us",
+		config.CgroupMountPoint + "/cpu/cpu.cfs_period_us",
+	}
+	b, err := pods.ExecCommandOnPod(c, pod, containerName, cmd)
+	output := strings.Split(string(b), "\r\n")
+	cfg.CPUSet = output[0]
+	cfg.Quota = output[1]
+	cfg.Period = output[2]
+	if err != nil {
```
ffromani, line 34: you may want to do this earlier, because if `Split` doesn't yield 3 values, we will hit an error. OTOH `Split` likely fails only iff `err != nil`.

test/e2e/performanceprofile/functests/utils/cgroup/v2/v2.go
```diff
@@ -0,0 +1,37 @@
...
+		config.CgroupMountPoint + "/cpuset.cpus",
+		config.CgroupMountPoint + "/cpu.max",
+	}
+	b, err := pods.ExecCommandOnPod(c, pod, containerName, cmd)
+	output := strings.Split(string(b), "\r\n")
+	cfg.CPUSet = output[0]
+	quotaAndPeriod := strings.Split(output[1], " ")
+	cfg.Quota = quotaAndPeriod[0]
+	cfg.Period = quotaAndPeriod[1]
+	if err != nil {
```
ffromani, line 33: ditto

ffromani (commented):
looks nice, but we need some form of testing, at least partial

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
Tal-or: Yes we do.
It should have additional interfaces to retrieve pod config and container's child config (we have childs in crun and mixedCpus)

ffromani: fair enough, so using a type (vs a function) has its merits

Tal-or: Now I noticed I wrote additional interfaces. I meant I want to add additional functions to the same interface, where each manager implement it own logic

ffromani: NP, somehow parsed the statement as intended rather than as written :)

test/e2e/performanceprofile/functests/utils/cgroup/config/config.go
Tal-or: Thank you for the useful information, it really helps understand the principles better. 

So in our case it means the interface would be implemented in the test files themselves, i.e. `7_performance_kubelet_node/cgroups.go` or `11_mixedcpus/mixedcpus.go` . 
But what if both packages want to use the exact interface? should both declare it?

And my second questions is how do you suggest the `BuildGetter` to return a struct, while it suppose to return an abstraction for getting the cgroup information.
 
Of course we can implement the `BuildGetter` on the consumer side, but again it's a duplication of code, since at least two packages (from what I know of) would need to implement it to retrieve the existing cgroup configuration on the node.

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
Tal-or: Thanks. fixed.

test/e2e/performanceprofile/functests/utils/cgroup/config/config.go
```diff
@@ -0,0 +1,20 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+const CgroupMountPoint = "/sys/fs/cgroup"
+
+type Config struct {
+	CPUSet string
+	Quota  string
+	Period string
+}
```
mrniranjan, line 16: Can we define structure in a more hierarchial fashion i.e: Controllers->Controller-interfacefiles

```
type CgroupController struct {
          //suported cgroup Controllers
           cpuset
           cpu
           // any future cgroup controllers 
   }
```
   Then created methods that return interface files of that particular controller ?

test/e2e/performanceprofile/functests/utils/cgroup/v2/v2.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/config"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
+)
+
+type Manager struct {
+}
+
+func (m *Manager) GetConfig(ctx context.Context, c *kubernetes.Clientset, pod *corev1.Pod, containerName string) (*config.Config, error) {
+	var cmd []string
+	cfg := &config.Config{}
```
mrniranjan, line 21: GetConfig here supports getting here only cpuset.cpus, cpu.max, but there are other interface files that equally important and we do need them in future like 
cpuset.cpu.partition, cpuset.cpus.effective, cpuset.mems (this is specifically required for memory manager to know which numa interface the pod has been assigned).

mrniranjan:
This PR is a  good starting point, i think we can improve it in further iteration

Tal-or:
@mrniranjan IMO the existing implementation  is good enough for now and we can expand it later

ffromani:
it's probabably wasteful to retry until we can consume https://github.com/cri-o/cri-o/issues/7643 in our CI clusters

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, fmt.Errorf("failed to retrieve cgroup config for pod. pod=%q, container=%q; %w", client.ObjectKeyFromObject(pod).String(), containerName, err)
+	}
+	output := strings.Split(string(b), "\r\n")
+	cfg.Cpus = output[0]
+	return cfg, nil
+}
+
+func (cm *ControllersManager) Cpu(ctx context.Context, pod *corev1.Pod, containerName, runtimeType string) (*controller.Cpu, error) {
+	cfg := &controller.Cpu{}
```
mrniranjan, line 43: We are passing runtimeType but we are not using this variable in Cpu function

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -0,0 +1,90 @@
...
+type ControllersManager struct {
+	client    client.Client
+	k8sClient *kubernetes.Clientset
+}
+
+func NewManager(c client.Client, k8sClient *kubernetes.Clientset) *ControllersManager {
+	return &ControllersManager{client: c, k8sClient: k8sClient}
+}
+
+func (cm *ControllersManager) CpuSet(ctx context.Context, pod *corev1.Pod, containerName, runtimeType string) (*controller.CpuSet, error) {
```
mrniranjan, line 27: We are passing runtimeType but we are not using this variable in CpuSet function

mrniranjan:
LGTM from my side, we can improve this further in future iterations.

Tal-or:
> it's probabably wasteful to retry until we can consume [cri-o/cri-o#7643](https://github.com/cri-o/cri-o/issues/7643) in our CI clusters

This is why the updating lane is falling?

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
Tal-or: Yes I need to learn the details, i just passed the argument for now

Tal-or:
/hold
There's a bug in the implementation that needs to be addressed

test/e2e/performanceprofile/functests/utils/cgroup/cgroup.go
```diff
@@ -0,0 +1,48 @@
...
+func checkVersionV2(ctx context.Context, c client.Client) (bool, error) {
+	nodecfg := &apiconfigv1.Node{}
+	key := client.ObjectKey{
+		Name: "cluster",
+	}
+	err := c.Get(ctx, key, nodecfg)
+	if err != nil {
+		return false, fmt.Errorf("failed to get configs.node object. name=%q; %w", key.Name, err)
+	}
+	return nodecfg.Spec.CgroupMode == apiconfigv1.CgroupModeV2, nil
```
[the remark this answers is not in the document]
Tal-or: Fixed, Thanks!

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -0,0 +1,101 @@
...
+		dirPath + "/cpuset/cpuset.cpus",
+		dirPath + "/cpuset/cpuset.cpu_exclusive",
+		dirPath + "/cpuset/cpuset.effective_cpus",
+		dirPath + "/cpuset/cpuset.mems",
+		dirPath + "/cpuset/cpuset.sched_load_balance",
+	}
+	b, err := pods.ExecCommandOnPod(cm.k8sClient, pod, containerName, cmd)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve cgroup config for pod. pod=%q, container=%q; %w", client.ObjectKeyFromObject(pod).String(), containerName, err)
+	}
```
mrniranjan, line 41: This method of passing multiple interface files at once and reading them is problematic because there is no guarantee that all the files will be read, there have been cases where only the first file is read and others are skipped and no error is returned.  Or some times one of the files are read. 

because "cat"  when run on normal system and when run inside a container , the behavior changes and it can skip files. We had multiple instances in past where we have faced these issues. 

Also we are just using split to differentiate the outputs between different interface files , though the files are read in the same order as given, if for some reason one of files are not read or the values are empty due to some errors, the values can get mixed up causing the interface values in cfg struct  fields to be jumbled.

shajmakh: +1. perhaps is better to divide this into submethods to fetch each value and reporting errors/warnings accordingly.

Tal-or: I just did please review the last commit

mrniranjan: the latest changes look good . We can improve iteratively.

mrniranjan: LGTM from my side

test/e2e/performanceprofile/functests/1_performance/hugepages.go
```diff
@@ -148,7 +148,7 @@ var _ = Describe("[performance]Hugepages", Ordered, func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			cmd2 := []string{"/bin/bash", "-c", "tmux new -d 'LD_PRELOAD=libhugetlbfs.so HUGETLB_MORECORE=yes top -b > /dev/null'"}
-			_, err = pods.ExecCommandOnPod(testclient.K8sClient, testpod, cmd2)
+			_, err = pods.ExecCommandOnPod(testclient.K8sClient, testpod, "", cmd2)
```
ffromani, line 151: this works. An alternative option could have been to add a variant function which accepts a container name and reimplement the existing ExecCommandOnPod (with the current signature) on top of it. But we can keep this approach.

test/e2e/performanceprofile/functests/utils/cgroup/v1/v1.go
```diff
@@ -99,3 +93,25 @@ func (cm *ControllersManager) Child(ctx context.Context, pod *corev1.Pod, contai
 	// TODO
 	return nil
 }
+
+func (cm *ControllersManager) execAndStore(pod *corev1.Pod, containerName, dirPath string, store map[string]*string) error {
```
ffromani, line 97: not a big fan of mutating arguments (and of maps whose values are pointers) but I see why you are doing like this and I don't have compelling suggestions

test/e2e/performanceprofile/functests/utils/cgroup/v2/v2.go
```diff
@@ -98,3 +92,24 @@ func (cm *ControllersManager) Child(ctx context.Context, pod *corev1.Pod, contai
 	}
 	return nil
 }
+func (cm *ControllersManager) execAndStore(pod *corev1.Pod, containerName, dirPath string, store map[string]*string) error {
```
ffromani, line 95: this seems one of the instances on which seems better to generalize the function. But I think you did not because the v1 and v2 package have no common package to share library functions, and adding one doesn't seem so great. If that's the case I agree to keep it like this for starters.

ffromani (commented):
LGTM
nice addition

ffromani:
/approve

we want this change in

test/e2e/performanceprofile/functests/1_performance/hugepages.go
Tal-or: Yes this is an option as well.

test/e2e/performanceprofile/functests/utils/cgroup/v2/v2.go
Tal-or: Thanks. We may consider shared library in the future in case of additional common functions.

Tal-or:
/cherry-pick release-4.15

Needed for mixed-cpus e2e testing

PR #907, opened 2024-01-08
perfprof: render: make owneReferenc option

ffromani:
WIP TBD


pkg/performanceprofile/cmd/render/render.go
```diff
@@ -188,45 +190,96 @@ func render(inputDir, outputDir string) error {
...
+	weakOwnerReferenceAnnotation = "performance.openshift.io/weak-owner-reference"
+)
+
+func addWeakOwnerReferenceAnnotation(components *manifestset.ManifestResultSet, pp *performancev2.PerformanceProfile, addAnnRef string) {
+	ref := makeWeakOwnerRef(addAnnRef, pp.UID)
+
+	if pp.Annotations == nil {
+		pp.Annotations = make(map[string]string)
+	}
+	pp.Annotations[weakOwnerReferenceAnnotation] = ref
```
MarSik, line 261: I wonder if label is better here. As it will allow searching later when we re-link the objects together.

ffromani: fair point, I'll have a look

jlojosnegros:
Aside Martin's comment about labels LGTM

ffromani:
/retest

changes here don't affect the updating-profile lane

pkg/performanceprofile/cmd/render/cmd.go
```diff
@@ -31,6 +32,8 @@ import (
 type renderOpts struct {
 	assetsInDir  string
 	assetsOutDir string
+	addOwnerRef  bool
```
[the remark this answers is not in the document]
ffromani: the default value is `false`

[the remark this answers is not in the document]
ffromani: I misread. let me retry. By default the code DOES NOT add anymore the owner reference.
Likewise, by default the code won't add the new label.

yanirq:
Looks good to me.  @MarSik   @jlojosnegros   since you had some iterations here please have the final ack

pkg/performanceprofile/cmd/render/cmd.go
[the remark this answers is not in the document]
MarSik: And that is exactly my point. The generated content should be equivalent to whatever the operator will do later. So either with ownership references or with the weak references and with code in the operator that converts the weak link to the real thing as the first thing during reconciliation.

Or maybe in from a different angle:

What happens when you generate the content without any references and post it to the cluster?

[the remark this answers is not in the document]
ffromani: With the current defaults the cluster is expected to accept the objects without conflicts, because the bogus ownerReference is now gone; however, the operator reconciliation has no helps or hints to rebuild the correct ownerReference during reconciliation. 

The reason why the default about adding the weak reference is disabled is because we either need:
1. to generate a UID during the rendering _and annotate the performance profile to set the root reference_, so the reconciliation loop can trivially rebuild the chain
2. to let the reconciliation loop tolerate the lack of such root (= all dependant object will have the label with the same UID, but there's no way to link back that UI to the owning perf-prof, so some inference is needed)

I didn't figure out a third way yet to link the autogenerated uid to the owning performance profile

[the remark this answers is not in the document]
MarSik: What about using just the PP name? I know it is less secure, but in this case it should be unique enough unless you as an user actively work to exploit it somehow.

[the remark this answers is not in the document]
ffromani: implemented option to use the pp name

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -191,45 +194,101 @@ func render(inputDir, outputDir string) error {
...
+
+func addWeakOwnerReferenceLabel(components *manifestset.ManifestResultSet, pp *performancev2.PerformanceProfile, labelWeakRef string) {
+	ref := makeWeakOwnerRef(labelWeakRef, pp)
+
+	klog.Infof("using weak ref label %q -> %q", labelWeakRef, ref)
+
+	if pp.Labels == nil {
+		pp.Labels = make(map[string]string)
+	}
+	pp.Labels[weakOwnerReferenceLabel] = ref
```
MarSik, line 267: How will the relink implementation later distinguish name from uuid? It can just try, that is fine, but lets make sure the plan is documented.

ffromani: thing is, we can't (easily/cleanly) because we mix types in the label values: can either be a string (name) or a uuid, thuse the field we use of a key depends.

I think the way forward is to either always use the name or the uuid. Both have their drawbacks.

ffromani: or to use different labels for name vs uuid?

jlojosnegros: Why do we want to give the option to choose between name and uid?

I thought the "weak reference" was a way to rebuild the owner ref structure as we cannot do it directly in the rendered objects, so what we use to do it seems like an implementation detail to me. All we need is a unique identifier shared by both the PP and the rendered objects so we could rebuild the "link" in the controller, right?

ffromani: > Why do we want to give the option to choose between name and uid?

because uid is the strongest we can achieve, but is not predictable; we can make predictable, but then it becomes hard to consume (because we need to mutate the performance profile and send to the cluster the _mutated_ performance profile).
Name is predictable, so we're considering which one we should use.

> I thought the "weak reference" was a way to rebuild the owner ref structure as we cannot do it directly in the rendered objects, so what we use to do it seems like an implementation detail to me. All we need is a unique identifier shared by both the PP and the rendered objects so we could rebuild the "link" in the controller, right?

Yes, that's correct, this is the intent.

pkg/performanceprofile/cmd/render/render.go
```diff
@@ -191,45 +194,110 @@ func render(inputDir, outputDir string) error {
...
+		if addOwnerRef {
+			err = addOwnerReference(components, pp)
+			if err != nil {
+				return err
+			}
 		}
-
-		for _, componentObj := range components.ToObjects() {
-			componentObj.SetOwnerReferences(or)
+		if len(weakRef) > 0 {
```
MarSik, line 203: Hmm, I wonder if this should be `else if`. The weak ref makes not much sense when the real reference is set.

MarSik: Yeah, I am just trying to make the arguments as comfortable as possible (= prevent mistakes). And considering the default (weakref with name) we actually do not have a way to unset it when a full ref is requested.

ffromani: sure, fixed in the latest upload.

pkg/performanceprofile/cmd/render/cmd.go
```diff
@@ -74,6 +77,11 @@ func (r *renderOpts) readFlagsFromEnv() {
 	if assetsOutDir := os.Getenv("ASSET_OUTPUT_DIR"); len(assetsOutDir) > 0 {
 		r.assetsOutDir = assetsOutDir
 	}
+	if addOwnerRef := os.Getenv("ADD_OWNER_REF"); len(addOwnerRef) > 0 {
```
Tal-or: nit: you probably followed the old code but we should use `os.LookupEnv`

Tal-or (commented):
nit comment from my side.
besides that LGTM but i'll leave it for @MarSik since he had additional comments offline.

pkg/performanceprofile/cmd/render/cmd.go
ffromani: I did. For the options we're using here, there's not much difference though, because empty string is not a legal value so we can use empty string to umambibuously mean "value not set"

MarSik (approved):
/lgtm

I like this now.

PR #908, opened 2024-01-09
   CNF-10163: Extend PerformanceProfile to include CPU frequencies

sabbir-47:
## What?
Configuring CPU frequency for reserved and isolated cpus through PerformanceProfile.

## Why?

- For FlexRAN (and FlexRAN-like applications), the hardware vendors require that the default cpu frequency be set to a specific frequency for running workloads which is actually lower than the default running frequency.
- Users constantly ask for reducing the platform core footprint. By setting higher frequencies for reserved cpus can  translates the effort into reducing the number of cpu cores reserved for running Openshift components.

## Additional resources:
- EPIC link: [Reserved Core Frequency Tuning](https://issues.redhat.com/browse/CNF-9112)
- [Design discussion](https://docs.google.com/document/d/1KW3Wy14L2w3hAF4DF6LIwhT7jdprlutTPPuGIgWf1Xw/edit)

 ## Characteristics
 - CPU frequency setting is optional
 - Requires **intel_pstate to be active**
 - This low level parameters should only be used based on the vendors recommendation on frequencies.
 - Feature is hardware agnostic, but must follow frequency guideline depending on hw generation.
 
 ## How?
By introducing a `hardwareTuning` parameter in the spec of PerformanceProfile. The onus is on user to provide correct frequencies.

```
hardwareTuning:
  isolatedCpuFreq: 2500000
  reservedCpuFreq: 2800000
```
It will update the `assets/performanceprofile/tuned/openshift-node-performance` and cpu plugin will apply the frequencies using sysfs directive
## Changes
- Extended CRD definition
- Modified static openshift-node-performance profile
- intel_pstate=active is changed as default
- When realTime workloadHints is set to true, previously it would disable intel_pstate and would inactive cpufreq driver. With this proposal, `intel_pstate=active` is set for `realTime=true`
- Extended performance profile creator by introducing a new flag i.e.; `--enable-hardware-tuning`. When passed, commented lines get added at the end with an example on how to set the frequency values. For example:
```
kind: PerformanceProfile
metadata:
  name: performance
spec:
  cpu:
    isolated: 4-15
    reserved: 0-3
    .......................
    .......................
  workloadHints:
    highPowerConsumption: false
    perPodPowerManagement: false
    realTime: true
#HardwareTuning is an advanced feature, and only intended to be used if 
#user is aware of the vendor recommendation on maximum cpu frequency.
#The structure must follow
#
# hardwareTuning:
#   isolatedCpuFreq: <Maximum frequency for applications running on isolated cpus>
#   reservedCpuFreq: <Maximum frequency for platform software running on reserved cpus>
```
## When it will not work?
- For RAN cases, we recommend below workloadHints:
```
    highPowerConsumption: false
    perPodPowerManagement: false
    realTime: true
```
  For, non-RAN cases, if perPodPowerManagement is set to true, the default intel_pstate is set to passive. For this feature to take in effect, P-state needs to be active. Therefore, cpu frequencies will not be set if perPodPowerManagement is true
- Providing incorrect frequencies will not take in effect, and it is strongly recommended to follow hardware vendors guideline.

## Can it be done other ways?
Changing cpu frequencies can be done through the tuned performance-patch, using the sysfs plugin. This would look something like this (using 2.5 GHz as the default frequency and 2.8 GHz for reserved CPUs on a system with 64 CPUs and CPUS 0,1,32,33 reserved for the platform):
```
[sysfs]
/sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq=2800000
/sys/devices/system/cpu/cpu1/cpufreq/scaling_max_freq=2800000
/sys/devices/system/cpu/cpu2/cpufreq/scaling_max_freq=2500000
/sys/devices/system/cpu/cpu3/cpufreq/scaling_max_freq=2500000
<...>
/sys/devices/system/cpu/cpu31/cpufreq/scaling_max_freq=2500000
/sys/devices/system/cpu/cpu32/cpufreq/scaling_max_freq=2800000
/sys/devices/system/cpu/cpu33/cpufreq/scaling_max_freq=2800000
/sys/devices/system/cpu/cpu34/cpufreq/scaling_max_freq=2500000
/sys/devices/system/cpu/cpu35/cpufreq/scaling_max_freq=2500000
<...>
/sys/devices/system/cpu/cpu63/cpufreq/scaling_max_freq=2500000
```
This requires 64 lines of configuration (one per logical CPU). The proposed approach reduces the complexity of this configuration, makes it less prone to mistake from users side.


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -81,6 +81,13 @@ var (
 	validTMPolicyValues        = []string{kubeletconfig.SingleNumaNodeTopologyManagerPolicy, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
 	validInfoModes             = []string{infoModeLog, infoModeJSON}
 	validPowerConsumptionModes = []string{defaultLatency, lowLatency, ultraLowLatency}
+	hardwareTuningMessage      = `#HardwareTuning is an advanced feature, and only intended to be used if 
+#customer is aware of the vendor recommendation on maximum cpu frequency.
```
bartwensley, line 85: Change customer to user.

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -487,6 +502,7 @@ func getProfileData(args ProfileCreatorArgs, cluster ClusterData) (*ProfileData,
 	}
 	log.Infof("%d reserved CPUs allocated: %v ", reservedCPUs.Size(), reservedCPUs.String())
 	log.Infof("%d isolated CPUs allocated: %v", isolatedCPUs.Size(), isolatedCPUs.String())
+
```
bartwensley, line 505: Undo

examples/performanceprofile/samples/performance_v2_performanceprofile.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   cpu:
       isolated: "2"
       reserved: "0-1"
-      offlined: "3"
+      offlined: "3"
```
bartwensley, line 16: Undo

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -344,3 +345,25 @@ func (r *PerformanceProfile) validateWorkloadHints() field.ErrorList {
 	}
 	return allErrs
 }
+
+func (r *PerformanceProfile) validateCpuFrequency() field.ErrorList {
```
bartwensley, line 349: I think this should also verify that the user doesn't set one of the new fields but not both. That should be treated as an error as both are required for the new feature to work.

sabbir-47: Added that verification

examples/performanceprofile/samples/performance_v2_performanceprofile.yaml
sabbir-47: Done

cmd/performance-profile-creator/cmd/root.go
sabbir-47: Done

cmd/performance-profile-creator/cmd/root.go
sabbir-47: Done

ffromani:
just a passing, non-blocking comment (note I'm not in the reviewer/approver list here)
```
- Requires intel_pstate to be active
- This low level parameters should only be used based on the vendors recommendation on frequencies.
- **Feature is hardware agnostic**, but must follow frequency guideline depending on hw generation.
```
Perhaps a "not" is missing here? because this feature seems very much _not_ hardware agnostic and on the contrary strongly hardware dependent - on type and family of CPU.

sabbir-47:
Hi @ffromani, its a good catch. The message i wanted to convey is, this feature would work whether it is sapphire rapids/ice lake/cascade lake servers, but the cpu frequencies will definitely be different based on the CPU profile. In short, most of the underlying hardware's cpu frequency can be managed with this work, it is not only limited to, for say, sapphire rapids servers.

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
bartwensley: thanks - looks good

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -160,8 +160,19 @@ cmdline_additionalArg=+{{.AdditionalArgs}}
 
 {{if .PerPodPowerManagement}}
 cmdline_pstate=+intel_pstate=passive
-{{else if .RealTimeHint}}
-cmdline_pstate=+intel_pstate=disable
+{{else}}
+cmdline_pstate=+intel_pstate=active
```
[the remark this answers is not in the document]
bartwensley: Yes - the intention is to change the default setting to intel_pstate=active (unless perPodPowerManagement is enabled, which will use intel_pstate=passive).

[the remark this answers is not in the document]
sabbir-47: One of the RAN customer wanted to have `intel_pstate=true` by default with the recommended workloadHints for RAN cases:
```
highPowerConsumption: false
perPodPowerManagement: false
realTime: true
```    
I do not have a strong opinion on what i have proposed above with changing the pstate, but we can change it to something like below as well: 
```
{{else if and .RealTimeHint .HardwareTuning}}
cmdline_pstate=+intel_pstate=active
```
Tagging @bartwensley for providing more context on this.

jmencak:
Please see the `make verify` target and linting issues.  Looks like the manifests were not auto-generated?

assets/performanceprofile/tuned/openshift-node-performance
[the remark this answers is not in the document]
bartwensley: Hey @sabbir-47 - see my previous comment - we want to go with intel_pstate=active as the default - regardless of whether hardwareTuning is being used.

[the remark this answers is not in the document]
sabbir-47: Sorry, i didn't see your message earlier. Thanks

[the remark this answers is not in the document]
MarSik: @bartwensley @sabbir-47 But I think in that case I would prefer that as a separate PR with its own rationale.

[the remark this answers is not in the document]
sabbir-47: Hi @MarSik, as suggested, i will create a separate PR for intel_pstate=active to be the default and modify this PR.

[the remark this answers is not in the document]
sabbir-47: Hi @MarSik with the commit 793c1e24aac30acd81080017016284c791526aef the current behavior is restored. Once the PR gets merged, i will create separate PR to make intel_pstate=active.

sabbir-47:
> Please see the `make verify` target and linting issues. Looks like the manifests were not auto-generated?

Hi @jmencak Those are fixed. Thank you.

jmencak:
Hi @sabbir-47, thank you for the changes.  Again, LGTM from my side, but you want a review from the cnf-team, as this PR changes mostly their code.

sabbir-47:
> Hi @sabbir-47, thank you for the changes. Again, LGTM from my side, but you want a review from the cnf-team, as this PR changes mostly their code.

Hi @jmencak Can you please point me towards whom should i tag for a review?

ffromani:
> > Hi @sabbir-47, thank you for the changes. Again, LGTM from my side, but you want a review from the cnf-team, as this PR changes mostly their code.
> 
> Hi @jmencak Can you please point me towards whom should i tag for a review?

/cc @MarSik @yanirq

yanirq:
Would it be reasonable to add an e2e under https://github.com/openshift/cluster-node-tuning-operator/blob/1e3a67e1d54caa4aa82b55a106c1a862035786f0/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go#L76 in this PR ?   or is it planned to be added separately ?

sabbir-47:
> Would it be reasonable to add an e2e under
> 
> https://github.com/openshift/cluster-node-tuning-operator/blob/1e3a67e1d54caa4aa82b55a106c1a862035786f0/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go#L76
> 
> in this PR ? or is it planned to be added separately ?

Hi @yanirq, there was no behavior change happened in the workloadHints, this PR still keeps the current behavior:
- intel_pstate disabled for `realtimeHint:true` and passive for `perPodPowerManagement:true`

But changes:
- the pstate to active when cpu frequencies are defined in the performanceProfile. Do i need to still add an e2E test over there?

My next separate PR will target the intel_pstate to be active at all cases except when `perPodPowerManagement:true`. Would that make more sense to modify/add tests in that case?

yanirq:
> > Would it be reasonable to add an e2e under
> > https://github.com/openshift/cluster-node-tuning-operator/blob/1e3a67e1d54caa4aa82b55a106c1a862035786f0/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go#L76
> > 
> > in this PR ? or is it planned to be added separately ?
> 
> Hi @yanirq, there was no behavior change happened in the workloadHints, this PR still keeps the current behavior:
> 
>     * intel_pstate disabled for `realtimeHint:true` and passive for `perPodPowerManagement:true`
> 
> 
> But changes:
> 
>     * the pstate to active when cpu frequencies are defined in the performanceProfile. Do i need to still add an e2E test over there?

Sorry I mixed things up a bit.   What I am missing is the actual API e2e test  that could potentially reside under [/functests/1_performance](https://github.com/openshift/cluster-node-tuning-operator/tree/master/test/e2e/performanceprofile/functests/1_performance)  or [functests/2_performance_update](https://github.com/openshift/cluster-node-tuning-operator/tree/master/test/e2e/performanceprofile/functests/2_performance_update) and check that the frequencies which where set via the performance profile were properly applied on the node/s.

As I mentioned , this can be done in a separate PR and we do need to make sure these tests are not disruptive.  

> My next separate PR will target the intel_pstate to be active at all cases except when `perPodPowerManagement:true`. Would that make more sense to modify/add tests in that case?

only if the combination already exists then we will need to avoid test regression.

sabbir-47:
> Sorry I mixed things up a bit. What I am missing is the actual API e2e test that could potentially reside under [/functests/1_performance](https://github.com/openshift/cluster-node-tuning-operator/tree/master/test/e2e/performanceprofile/functests/1_performance) or [functests/2_performance_update](https://github.com/openshift/cluster-node-tuning-operator/tree/master/test/e2e/performanceprofile/functests/2_performance_update) and check that the frequencies which where set via the performance profile were properly applied on the node/s.
> 
> As I mentioned , this can be done in a separate PR and we do need to make sure these tests are not disruptive.

Sure, i can add those test in a separate PR. The only caveat is we need to know the underlying HW where this functional tests will run. For example, the frequencies should be different from SPR-EE HW to Icelake/cascade lake HW based on the vendors(I//) suggestion.  And invalid frequencies will not take into effect.

> > My next separate PR will target the intel_pstate to be active at all cases except when `perPodPowerManagement:true`. Would that make more sense to modify/add tests in that case?
> 
> only if the combination already exists then we will need to avoid test regression.

Yes, i have verified already, it will not need to have duplicated tests, rather some modification would suffice in some places.

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -160,8 +160,21 @@ cmdline_additionalArg=+{{.AdditionalArgs}}
...
 {{else if .RealTimeHint}}
 cmdline_pstate=+intel_pstate=disable
 {{end}}
 
 [rtentsk]
+
+{{ if .HardwareTuning }}
+[sysfs]
+# sets provided frequencies to isolated and reserved cpus
+{{ range .IsolatedCpuList }}
```
yanirq, line 174: It would be nice to add a unit test for the ranges testing different formats (comma separated lists , range lists, combined)

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -62,6 +67,30 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
...
 	}
 
+	if profile.Spec.HardwareTuning != nil {
+		isolatedCpuSet, err := cpuset.Parse(string(*profile.Spec.CPU.Isolated))
+		if err != nil {
+			return nil, err
+		}
+		isolatedCpuString := components.ListToString(isolatedCpuSet.List())
+		// converts a string to a string slice
+		isolatedCpuList := strings.SplitN(isolatedCpuString, ",", len(isolatedCpuString))
```
sabbir-47, line 77: Hi @yanirq, based on your comment on unit test for `isolatedCpuList`, the final `isolatedCpuList` will always be an array. Do you want me to create a new generic function which returns a cpulist based on the input provided, and add unit tests for that? Also the `range` go template function can only range over an array but not over lists separated by comma.

assets/performanceprofile/tuned/openshift-node-performance
sabbir-47: Hi @yanirq, i have replied to your comment in the following link:
https://github.com/openshift/cluster-node-tuning-operator/pull/908/files#r1477824265

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
yanirq: In that case I think we are good

PR #909, opened 2024-01-09
NO-JIRA: e2e: set performance profile cpus using env vars

shajmakh:
We've been observing lately that some tests that involve disabling load balancing are failing (like 32646) because the expected result does not have specific anticipated CPUs. After investigation, it turns out that one factor is the profile configuration of the CPU distribution.

PAO functional tests configure fixed CPU values under the PP. This is considered misconfiguration, especially when the system has more than 4 CPUs, and there is no guarantee that the functionality of the performance profile controller will work adequately with not all cpus reflected in the CPU section in the PP.

To resolve this complication, we are introducing new environment variables RESERVED_CPU_SET, ISOLATED_CPU_SET, OFFLINED_CPU_SET, should be set the profile would use them instead of the defaults.

### DIFF
--- a/test/e2e/performanceprofile/functests/0_config/config.go
+++ b/test/e2e/performanceprofile/functests/0_config/config.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/cpuset"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -65,11 +66,11 @@ var _ = Describe("[performance][config] Performance configuration", Ordered, fun
 
 	It("Should successfully deploy the performance profile", func() {
 
-		performanceProfile := testProfile()
+		performanceProfile, err := testProfile()
+		Expect(err).ToNot(HaveOccurred(), "failed to build performance profile: %v", err)
 		profileAlreadyExists := false
 
 		performanceManifest, foundOverride := os.LookupEnv("PERFORMANCE_PROFILE_MANIFEST_OVERRIDE")
-		var err error
 		if foundOverride {
 			performanceProfile, err = externalPerformanceProfile(performanceManifest)
 			Expect(err).ToNot(HaveOccurred(), "Failed overriding performance profile", performanceManifest)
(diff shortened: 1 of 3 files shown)


shajmakh:
Note: The README.md file is generally outdated and needs to be updated with the new suites and instructions on adding and running them, yet this can be addressed later as it is outside of this PR scope.

report.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ffromani, line 1: this should probably not be committed, should it?

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -3,6 +3,7 @@ package __performance_config
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/cpuset"
```
ffromani, line 6: minor: wrong grouping of imports. This group should only contain stdlib imports

ffromani (commented):
at first I was a bit taken aback by the approach here but after a more careful review and some thinking I'm growing to like it.
Do you have already a sketch about how the changes will be used e.g in `openshift/release` flows? (e.g. which changes to prow config will be needed?)

test/e2e/performanceprofile/functests/0_config/config.go
shajmakh: I keep missing that, will fix it thank you for the note.

report.xml
shajmakh: right, thanks for catching it

shajmakh:
> at first I was a bit taken aback by the approach here but after a more careful review and some thinking I'm growing to like it. Do you have already a sketch about how the changes will be used e.g in `openshift/release` flows? (e.g. which changes to prow config will be needed?)

Thank you for reviewing this. 
yes here is the changes for cnf-features-deploy u/s CI:
https://github.com/openshift/release/pull/47439
another similar change is needed for NTO lane.

ffromani (commented):
/approve

so this is the minimal solution to adapt the builtin hardcoded profile to different environments setting up the cpu setup correctly and consuming all CPUs.
Many other tradeoffs are possible ranging from pre-building a profile and letting the tests consume it (which should already be possible today) to fully automatically compute the right baseline CPU allocation autodetecting from the cluster at each run

where to draw the line is exactly one of the many cases about finding the correct tradeoff.

I think this PR can unblock us and enable us to further explore the solution space if we can and want (we really should FWIW) so I'm fine with this _initial_ step.

ffromani (commented):
/lgtm

shajmakh:
/hold
failure is legit, working on a fix

shajmakh:
timeout failures look flaky
/retest

ffromani:
It's wasteful to retry until we can consume https://github.com/cri-o/cri-o/issues/7643 in our CI clusters

shajmakh:
the last failure doesn't relate to crun (suite didn't manage to run)
retest to verify if latest nightly has the fixed crun version.
/retest

ffromani:
the failure is well known and we hit it recently. let's discuss a fix offline (and let's not retry until then)

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -158,9 +159,40 @@ func externalPerformanceProfile(performanceManifest string) (*performancev2.Perf
...
 	isolated := performancev2.CPUSet("1-3")
+	profileCpus := &performancev2.CPU{
+		Reserved: &reserved,
+		Isolated: &isolated,
+	}
+
+	customReserved, foundReserved := os.LookupEnv("RESERVED_CPU_SET")
+	customIsolated, foundIsolated := os.LookupEnv("ISOLATED_CPU_SET")
+	if foundReserved != foundIsolated {
+		return nil, fmt.Errorf("both reserved and isolated cpusets are required.")
```
ffromani, line 173: we need to either provide both or none - let's reflect this in the error message

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -158,9 +159,40 @@ func externalPerformanceProfile(performanceManifest string) (*performancev2.Perf
...
+	}
+	if foundReserved {
```
ffromani, line 175: this is a bit more oscure than it can. I'd split the if block in `foundReserved` and `foundIsolated). Yes, its redundant but it is also IMO clearer

ffromani (commented):
actually minor comments

ffromani:
other than actually minor inline comments, LGTM

test/e2e/performanceprofile/functests/0_config/config.go
shajmakh: yeah this depended on the fact that this condition:  `foundReserved != foundIsolated`  is checked before. fixing.

ffromani:
/lgtm

thanks fro the updates!

shajmakh:
infra issue
/retest ci/prow/e2e-gcp-pao-updating-profile

shajmakh:
ci/prow/e2e-gcp-pao passing is already a good sign.
/retest e2e-gcp-pao-updating-profile

ffromani:
/retitle NO-JIRA: e2e: set performance profile cpus using env vars

PR #912, opened 2024-01-10
CNF-9173: e2e: mixedcpus: kubelet restart test

Tal-or:
Add a test to check if shared CPUs survive Kubelet restart.

### DIFF
--- a/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
+++ b/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
@@ -45,6 +45,7 @@ const (
 	// reserved = one core, shared = one core, infra workload = one core, test pod = one core - 4 in total
 	// smt alignment won't allow us to run the test pod with a single core, hence we should cancel it.
 	numberOfCoresThatRequiredCancelingSMTAlignment = 4
+	restartCooldownTime                            = 1 * time.Minute
 )
 
 var _ = Describe("Mixedcpus", Ordered, func() {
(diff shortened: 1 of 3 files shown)


Tal-or:
/hold
Moved this specific test into separate PR since it depends on CRIO patch while the others appear under #892 are not
depends on https://github.com/cri-o/cri-o/pull/7648, #906

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -193,6 +194,50 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				// The command would fail since it aborts all the pods during restart
+				_, _ = nodes.ExecCommandOnNode(ctx, cmd, node)
+				// check that the node is ready after we restart Kubelet
+				nodes.WaitForReadyOrFail("post restart", node.Name, 20*time.Minute, 3*time.Second)
+
+				// giving kubelet more time to stabilize and initialize itself before
+				// moving on with the testing.
+				testlog.Infof("post restart: entering cooldown time: %v", restartCooldownTime)
+				time.Sleep(restartCooldownTime)
+				testlog.Infof("post restart: finished cooldown time: %v", restartCooldownTime)
```
ffromani, line 231: I understand why you do that, the problem this is inherently fragile and racy. I guess we can't do much better for the time being

ffromani (commented):
/lgtm

but with the usual caveat: this part is ok, but we need tests which check the shared cpus from the perspective of the running container (vs execing into it) because running a container and execing into it can get different pictures. And it actually happened not so long ago.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: > but with the usual caveat: this part is ok, but we need tests which check the shared cpus from the perspective of the running container (vs execing into it) because running a container and execing into it can get different pictures. And it actually happened not so long ago.

Yes, this type of check should be replaced everywhere in the testing framework but it requires some extra work from my side.

Tal-or: > I understand why you do that, the problem this is inherently fragile and racy. I guess we can't do much better for the time being

I Used the exact same logic from:
https://github.com/openshift/cluster-node-tuning-operator/blob/b185119c5d5d100a4a38342989e8aefa0e5428f6/test/e2e/performanceprofile/functests/9_reboot/devices.go#L212-L211

ffromani: yes, it's debatable even if I did it first :)

PR #914, opened 2024-01-11
NO-JIRA:e2e: `UpdateWithRetry` only update spec

Tal-or:
The function contains a subtle bug on which the update is done only for fileds under the `Spec`

This means that changes to annotations, labels and any other field outside the `Spec` won't take affect.

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/profiles/profiles.go
+++ b/test/e2e/performanceprofile/functests/utils/profiles/profiles.go
@@ -100,27 +100,18 @@ func All() (*performancev2.PerformanceProfileList, error) {
 
 func UpdateWithRetry(profile *performancev2.PerformanceProfile) {
 	EventuallyWithOffset(1, func() error {
-		updatedProfile := &performancev2.PerformanceProfile{}
-		key := types.NamespacedName{
-			Name:      profile.Name,
-			Namespace: profile.Namespace,
-		}
-		// We should get the updated version of the performance profile.
-		// Otherwise, we will always try to update the profile with the old resource version
-		// and will always get the conflict error
-		if err := testclient.Client.Get(context.TODO(), key, updatedProfile); err != nil {
+		profileFromAPIServer := &performancev2.PerformanceProfile{}
+		// get the current resourceVersion
+		if err := testclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(profile), profileFromAPIServer); err != nil {
 			return err
 		}
-
-		updatedProfile.Spec = *profile.Spec.DeepCopy()
-		if err := testclient.Client.Update(context.TODO(), updatedProfile); err != nil {
+		prepared := prepareForUpdate(profile, profileFromAPIServer)
+		if err := testclient.Client.Update(context.TODO(), prepared); err != nil {
 			if !errors.IsConflict(err) {
 				testlog.Errorf("failed to update the profile %q: %v", profile.Name, err)
 			}
-
 			return err
 		}
-
 		return nil
 	}, time.Minute, 5*time.Second).Should(BeNil())
 }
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/utils/profiles/profiles.go
```diff
@@ -100,27 +100,18 @@ func All() (*performancev2.PerformanceProfileList, error) {
...
-		if err := testclient.Client.Get(context.TODO(), key, updatedProfile); err != nil {
+		profileFromAPIServer := &performancev2.PerformanceProfile{}
+		// get the current resourceVersion
+		if err := testclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(profile), profileFromAPIServer); err != nil {
 			return err
 		}
-
-		updatedProfile.Spec = *profile.Spec.DeepCopy()
-		if err := testclient.Client.Update(context.TODO(), updatedProfile); err != nil {
+		profile.ResourceVersion = profileFromAPIServer.ResourceVersion
```
ffromani, line 108: The concept is good but almost always we write `ResourceVersion` then we're doing it wrong.
In this case, we should update Spec from the new data AND likely merge annotations and labels.

IOW, something like
```
updateProfile.Annotations = mergeMap(updateProfile.Annotations, profile.Annotations)
updateProfile.Labels = mergeMap(updateProfile.Labels, profile.Labels)
```
(`mergeMap` to be written)

Tal-or: What is wrong with almost write resource version? If no other changes then resource version the updat call will return with doing nothing.

And I'm not sure the merge approach is right because then we forced to keep the old labels/annotation

Tal-or: Actually the latter depends on how we want to implement the merge but still

ffromani: > What is wrong with almost write resource version? If no other changes then resource version the updat call will return with doing nothing.
> 
> And I'm not sure the merge approach is right because then we forced to keep the old labels/annotation

because we should not. ResourceVersion is managed by the API server and we're doing borderline wrong. OK, it's a big simplification (intended like copying everything from old object from new object), but is conceptually incorrect. What we want to do here is to keep overriding the spec and possibly overriding annotations (and labels)

test/e2e/performanceprofile/functests/utils/profiles/profiles.go
```diff
@@ -100,27 +100,18 @@ func All() (*performancev2.PerformanceProfileList, error) {
...
-		if err := testclient.Client.Get(context.TODO(), key, updatedProfile); err != nil {
+		profileFromAPIServer := &performancev2.PerformanceProfile{}
+		// get the current resourceVersion
+		if err := testclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(profile), profileFromAPIServer); err != nil {
 			return err
 		}
-
-		updatedProfile.Spec = *profile.Spec.DeepCopy()
-		if err := testclient.Client.Update(context.TODO(), updatedProfile); err != nil {
+		profile = prepareForUpdate(profile, profileFromAPIServer)
```
ffromani, line 108: I'd still use DeepCopy() and reduce the variable aliasing:
```
prepared := prepareForUpdate(profile.DeepCopy(), profileFromAPIServer)
if err := testclient.Client.Update(context.TODO(), prepared); err != nil {
...
```

ffromani (commented):
almost there, suggesting the last change and I think we're good

test/e2e/performanceprofile/functests/utils/profiles/profiles.go
Tal-or: Thanks. Done

ffromani:
/lgtm

it seems my suggestion to use DeepCopy() wasn't accepted, but that's no big deal

PR #916, opened 2024-01-15
CNF-6517: Export ValidateBasicFields for Hypershift

jlojosnegros:
In Hypershift we need to validate PerformanceProfile from NodePool controller.
As a one-to-one relationship between NodePool and Performance Profile is already enforced there is no need ( and no easy way ) to check for PerformanceProfiles with duplicated selectors, so we just need to check the fields.


jlojosnegros:
/retest

timeout unlikely caused by this change

PR #922, opened 2024-01-19
Improve error handling in tuned render cmd

jlojosnegros:
As per [this conversation](https://github.com/openshift/cluster-node-tuning-operator/pull/833#issuecomment-1891793520) we need to be able to differentiate when the command is returning an error because there is no PerformanceProfile from the rest of errors [in the installer](https://github.com/openshift/installer/pull/7692).

To do that change command to pop errors up and return a different error code so scripts could handle this error differently

### DIFF
--- a/pkg/tuned/cmd/render/render.go
+++ b/pkg/tuned/cmd/render/render.go
@@ -14,6 +14,7 @@ limitations under the License.
 package render
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


pkg/tuned/cmd/render/render.go
```diff
@@ -146,7 +148,7 @@ func render(inputDir []string, outputDir string, mcpName string) error {
 
 	if len(filteredPerformanceProfiles) == 0 {
 		klog.Warningf("Cannot find any PerformanceProfile applicable to MachineConfigPool %s.", mcpName)
-		return fmt.Errorf("Cannot find any PerformanceProfile complying with MachineConfigPool %s.", mcpName)
```
MarSik, line 149: I still do not see why this would be an error. In case no PP is present, there is still the openshift-node profile. The code from line 167 and below is going to work just fine.

This is a tuned renderer, why should it require a PP?

jlojosnegros:
/close 
No need after #924

PR #924, opened 2024-01-22
CNF-10170: Render MC without additional PP

jlojosnegros:
This command should render a MC from the available sources. PerformanceProfile is a source, but it is not mandatory.

So, as there are scenarios where this command should work with no additional PP let's the command proceed even if there is no PP


yanirq:
The meaning of "additional" is a bit obscured here ?    Can we use the render command without a performance profile at all ?

MarSik:
@yanirq Yep, it is a tuned render after all. We only need to know the perf profiles to be able to collect all the generated tuned objects. But when there are no extra tuneds, NTO still ships the stock openshift-node Tuned with the proper recommendation selectors.

So the logic still works fine, even though there are no extra kernel args atm.

MarSik (approved):
/lgtm

jlojosnegros:
/hold 
Need to rebase once #929 is merged.

PR #926, opened 2024-01-23
CNF-10616: e2e: Added hwlatdetect default window size

SargunNarula:
The hwlatdetect tests were executed without a default window size. We need to make sure the window size is always consistent and same.

### DIFF
--- a/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
+++ b/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
@@ -73,6 +73,7 @@ const (
 	// we do not care about the actual system latency because CI systems are not tuned well enough to be an example for
 	// latency measuring, besides this suite only cares about testing the test executable with different values of env vars.
 	untunedLatencyThreshold = "10000000" //10s
+	windowSize              = "1000000"
 	negativeTesting         = false
 	positiveTesting         = true
 )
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -135,7 +135,8 @@ var _ = DescribeTable("Test latency measurement tools tests", func(testGroup []l
 					if test.hwlatdetectMaxLatency != "" {
 						thr = test.hwlatdetectMaxLatency
 					}
-					passedArgs = []string{"--duration " + test.testRuntime, "--threshold " + thr, "--hardlimit " + thr, "--window ", "--width "}
+					windowSize := "1000000"
```
shajmakh, line 138: let's make this a constant and use it in passedArgs.

shajmakh (approved):
Thanks for this. 
Please don't modify 4_latency in this context, two comments below, and we should be good to go.

shajmakh:
/lgtm cancel
label added by mistake.

shajmakh:
/lgtm
thanks for this

shajmakh:
/retitle CNF-10616: e2e: Added hwlatdetect default window size

shajmakh:
@SargunNarula I think this needs a rebase, may you handle that?

ffromani:
/override e2e-hypershift
/override e2e-hypershift-pao

these lanes passed previously, and this is a trivial test-only fix which was in the workins since January.
I'll watch it post-merge and will fix any (unlikely) breakage caused.

PR #929, opened 2024-01-25
NO-JIRA: Skip PerformanceProfile update test

jlojosnegros:
This tests is failing because of a known issue that is already being handled.
In the meantime skip it to avoid PRs to pile.


shajmakh:
thanks for this!
but why the job is still failing?

Tal-or:
Looks like there's another test related to ovs which is falling

PR #931, opened 2024-01-29
set realtimekernel to false only if it is true

mrniranjan:
workload hints test assume that profile will
always have realtimekernel to true,thats not the case with Downstream CI causing tests to timeout

This change fixes the test code by setting realtimekernel to false only if it's true.

### DIFF
--- a/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
+++ b/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
@@ -84,10 +84,12 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 				By("Modifying profile")
 				profile.Spec.WorkloadHints = nil
 
-				profile.Spec.RealTimeKernel = &performancev2.RealTimeKernel{
-					Enabled: pointer.Bool(false),
+				// set realtimekernel to false only if its enabled
+				if profile.Spec.RealTimeKernel.Enabled == pointer.Bool(true) {
+					profile.Spec.RealTimeKernel = &performancev2.RealTimeKernel{
+						Enabled: pointer.Bool(false),
+					}
 				}
-
 				By("Updating the performance profile")
 				profiles.UpdateWithRetry(profile)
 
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
```diff
@@ -84,10 +84,12 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 				By("Modifying profile")
 				profile.Spec.WorkloadHints = nil
 
-				profile.Spec.RealTimeKernel = &performancev2.RealTimeKernel{
```
MarSik, line 87: I do not see how this can fix the timeout. I see two cases:

1) the RealTimeKernel was false -> both new and old code just keep or set false, no change in the value
2) the RealTimeKernel was true -> both new and old code set false

Then both old and new code call UpdateWithRetry with a profile that asks for regular kernel.

What am I missing here?

mrniranjan: @MarSik , the timeout was noticed in D/s after we switched the Realtime kernel to false as default in Performance profile.

MarSik: Realtimekernel false was always the default.

But what I do not understand in this test change is the mechanism by which it should fix it. In other words, the added if makes no difference in what gets passed to the updatewithretry method.

`If true: set false` is simply nothing else than `set false`, just mode complicated. It does not change the outcome.

mrniranjan: According to 0_config (https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/0_config/config.go#L195) we set the RealTimeKernel to true when the profile is applied. 

As to why does it timeout when the defaultProfile is false , i agree it should not timeout, this needs to be investigated.

MarSik: Ok, so it was a test default.

But the `if` still makes no difference. Pure write we had there before has the same effect.

mrniranjan: Agree, the problem of timeout is else where. i will close this PR.

PR #932, opened 2024-01-29
CNF-11145: Enhance render sync to include bootstrap rendering tests

MarSik:


hack/render-sync.sh
```diff
@@ -1,16 +1,35 @@
...
+  cd "${WORKDIR}" || { echo "failed to change dir to ${WORKDIR}"; exit; }
+  _output/cluster-node-tuning-operator render \
+  --asset-input-dir "${INPUT_DIRS}" \
+  --asset-output-dir "${ARTIFACT_DIR}"
+
+  cp "${ARTIFACT_DIR}"/*  "${WORKDIR}"/test/e2e/performanceprofile/testdata/render-expected-output/${OUTPUT_DIR}
+  for f in  "${WORKDIR}"/test/e2e/performanceprofile/testdata/render-expected-output/${OUTPUT_DIR}/*
+  do
+    sed -i "s/uid:.*/uid: \"\"/" "${f}"
+  done
```
ffromani, line 29: After extra thinking, this should probably be done once to reconcile the current state with the expected state, and then not enforced at every sync.
Other than this LGTM

ffromani: Extra: I'd remove the entry entirely rather than set to empty.

MarSik: I actually did not change this part, just moved it into a function.

But the sync is the reconciliation of current state with expected state. The uid reset only happend on make render-sync to make sure the tests themselves do not have to do that anymore.

IOW: I do not understand what do you mean by "once" here.

MarSik:
/retest-required

This PR only affects the e2e-no-cluster lane directly, but lets see if the CI is healthier than yesterday.

PR #933, opened 2024-01-29
CNF-11091: Change webhook failure policy to ignore

vitus133:
Change performance profile validating webhook failure policy to "Ignore". This is needed as a temporary workaround for bootstrap-in-place, where webhooks can not be served. This PR should be reverted after a permanent solution for BiP is implemented

/cc @MarSik @yanirq


MarSik:
/approved
/lgtm

This needs to be followed by a Perf profile controller change to do the validation (for the second time) in the operator as well.

MarSik (approved):

PR #935, opened 2024-02-05
CNF-10170: Avoid command to fail if there is no PP on input folders

jlojosnegros:
When there is no enough information to choose a recommended TuneD profile from the input folder manifests we load the `default-cr-tuned` manifests.

modifies: #844 
see: https://github.com/openshift/installer/pull/8007

### DIFF
--- a/pkg/tuned/cmd/render/render.go
+++ b/pkg/tuned/cmd/render/render.go
@@ -15,10 +15,12 @@ package render
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
 
+	assets "github.com/openshift/cluster-node-tuning-operator/assets/tuned"
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/manifests"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/operator"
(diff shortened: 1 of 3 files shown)


pkg/util/manifests.go
```diff
@@ -127,7 +128,8 @@ func AppendMissingDefaultMCPManifests(currentMCPs []*mcfgv1.MachineConfigPool) [
 					Name: master,
 				},
 				Spec: mcfgv1.MachineConfigPoolSpec{
-					NodeSelector: v1.AddLabelToSelector(&v1.LabelSelector{}, masterNodeSelector, ""),
+					NodeSelector:          v1.AddLabelToSelector(&v1.LabelSelector{}, masterNodeSelector, ""),
```
jlojosnegros, line 131: @eggfoobar please take a look at this part of the change as I think this MCPs lack `MachineConfigSelector`

pkg/tuned/cmd/render/render.go
```diff
@@ -144,33 +169,55 @@ func render(inputDir []string, outputDir string, mcpName string) error {
...
 
+	klog.Infof("working with %d additional tuneD profiles", len(tuneD))
+
 	tuneDrecommended := operator.TunedRecommend(tuneD)
 	if len(tuneDrecommended) == 0 {
-		klog.Error("Unable to get tuned recommended profile.")
-		return fmt.Errorf("Unable to get tuned recommended profile.")
+		klog.Infof("Unable to get tuned recommended profile with current info. Adding default tuneD")
+
+		fileName := "default-cr-tuned.yaml"
```
MarSik, line 192: You should always load this CR. Then you will not have to run the recommendation logic twice.

jlojosnegros: thanks for the comment.
Done

pkg/util/manifests.go
MarSik: I wonder if we should keep this fix as part of the PR. I know it is needed, but it should link to the proper bug it is fixing.

MarSik (approved):

pkg/util/manifests.go
jlojosnegros: Moved to its own PR #956

PR #936, opened 2024-02-07
NO-JIRA: e2e: classifying tests using labels

Tal-or:
The label package contains a list of labels that can be used in the functests to indicate and point certain behavior or characteristics of the various tests.

Those can be filtered/focused by ginkgo before the test runs.

Example:
`ginkgo --label-filter="!slow"` means avoid any specs labelled `slow`

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/label/label.go
+++ b/test/e2e/performanceprofile/functests/utils/label/label.go
@@ -1,0 +1,81 @@
+package label
+
+// The label package contains a list of labels that can be used in
+// the functests to indicate and point certain behavior or characteristics
+// of the various tests.
+// Those can be filtered/focused by ginkgo before the test runs.
+
+// Kind is a label indicates specific criteria that classify the test.
+// A Mixture of kinds can be used for the same test
+type Kind string
+
+const (
+	// Slow means test that usually requires reboot or takes a long time to finish.
+	Slow Kind = "slow"
+
+	// ReleaseCritical is for tests that are critical for a successful release of the component that is being tested.
+	ReleaseCritical Kind = "release-critical"
+
+	// SpecializedHardware is for tests that need special Hardware like SR-IOV devices, specific NICs, etc.
+	SpecializedHardware Kind = "specialized-hardware"
+)
+
+// Feature is a label indicates the feature that being tested.
+type Feature string
+
+const (
+	// WorkloadHints should be added in tests that are validating/verifying workload-hints behavior.
+	WorkloadHints Feature = "workload-hints"
+
+	// MustGather should be added in tests that are validating/verifying must-gather tool functionality.
+	MustGather Feature = "must-gather"
+
+	// MixedCPUs should be added in tests that are validating/verifying mixed-cpus feature functionality.
+	MixedCPUs Feature = "mixed-cpus"
+
+	// Latency should be added in tests that are meant to test latency sensitivity.
+	Latency Feature = "latency"
+
+	// PerformanceProfileCreator should be added in tests that are validating/verifying performance-profile-creator tool
+	// functionally.
+	PerformanceProfileCreator Feature = "performance-profile-creator"
+)
+
+// Tire is a label to classify tests under specific grade/level
+// that should roughly describe the execution complexity, maintainer identity and processing criteria.
+type Tire string
+
+const (
+	// Tier0 are automated unit tests
+	// Minimal time needed to execute (minutes to 1 hour)
+	// Process criteria:
+	// 100% automated, must-pass 100%
+	// Development maintains tests, and reviews result
+	Tier0 Tire = "tier-0"
+
+	// Tier1 are component level functional tests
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/utils/label/label.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// ReleaseCritical is for tests that are critical for a successful release of the component that is being tested.
+	ReleaseCritical = "release-critical"
+
+	// SpecializedHardware is for tests that need special Hardware like SR-IOV devices, specific NICs, etc.
+	SpecializedHardware = "specialized-hardware"
+)
+
+// FeatureTag is a label indicates the feature that being tested.
+type FeatureTag string
```
Tal-or, line 20: I'm not sure about the new type since it requires casting to string when using under `ginkgo.label` which accepts `string`

ffromani, line 20: why a type for feature tags and not for the consts (e.g. `Slow`) above?

Tal-or: I couldn't find a good name to classify the set about :)

ffromani:
/approve

I very much like the idea

test/e2e/performanceprofile/functests/utils/label/label.go
ffromani: the new type definitions still help somehow because the golang compiler can ensure we don't mix the types - e.g we can only assign one of the defined `FeatureTag`s and not a random string.

Going to be a tradeoff anyway

ffromani: maybe just `Label` ? it's a overused name though.

ffromani: maybe `Kind for those` and `Feature` instead of `FeatureTag`?

Tal-or: Sounds good

mrniranjan: May be we can classify them as below

**Tiered** Tests  Tests covering different levels (tier0,tier1) 
**Featured **  Tests for specific features  (memorymanager, mixedcpus, workloadhints)
**Release Critical** tests Tests which are critical for tests 
**HardwareCompatibility**  Tests for testing hardware compatibility (sriov, GPU, hugepages) ?

All the tests should be under some label by default

Tal-or: @mrniranjan 
Do we have a concrete division for the different tiers?

mrniranjan: Yes we do have:
**tier0:** 
Definition :
Automated unit tests
Minimal time needed to execute (minutes to 1 hour)
Process criteria :
100% automated, must pass 100%
Development maintains tests and reviews results

**tier1** (component)
Definition :
Component level functional tests
Minimal time needed to execute (minutes to hours)
Process criteria:
Executed after tier 0 passing
100% automated and must pass 100%
QE / Development maintains tests and reviews results

**tier2** (integration)
Definition:
Integration level functional tests
May include basic non-functional tests (security, performance regression, install, compose validation)
Runs during nightly time frame
Process criteria :
Executed after components pass tier 1 testing
100% automated and must pass 100%
QE  maintains tests and reviews result

**tier3**  (system)
Definition:
System, scenario and  non-functional tests (which includes Fault Tolerance / Recovery / Failover)
Test that don't fit in tier 2 due to time, complexity, and other  factors
Process criteria :
Executed after components pass tier 2 testing
System or scenario testing doesn't block Non-functional tests. They can be executed in parallel.
100% automated
QE maintains tests and reviews results

yanirq: @mrniranjan  @Tal-or  can we add this information at least in one of our u/s docs ?  maybe even create a testing u/s doc under https://github.com/openshift/cluster-node-tuning-operator/tree/master/docs/performanceprofile ?

mrniranjan: yes , the above is more standard definitions, but we can make our own definitions of tiered tests and document them.

Tal-or: @mrniranjan 
Thank you for the valuable information, i'll add the tiers then.

As @ffromani the advantage of having specific types and not plain `strings` is that we can force specific labels and not random strings.


>All the tests should be under some label by default

We need to develop a way to force that. probably some GH action

Tal-or:
> since we wanted to backport this, i think we need either NO-JIRA or a bug.

Backport means we should also add labels to all test and backport that as well

mrniranjan:
> > since we wanted to backport this, i think we need either NO-JIRA or a bug.
> 
> Backport means we should also add labels to all test and backport that as well

Right , should we do that in separate PR's because we need to decide which tests gets what labels, Some tests may have multiple labels.

PR #937, opened 2024-02-07
OCPBUGS-29183: Enable crun as runtime only if missing

mrniranjan:
In certain situation crun is configured prior to running the tests causing the test to fails.

In this PR we check if crun is already enabled , in which case we skip the configuration part.

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -1178,7 +1178,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 		})
 
 		When("updates the default runtime to crun", func() {
-			It("should run high-performance runtimes class with crun as container-runtime", func() {
+			BeforeAll(func() {
 				const ContainerRuntimeConfigName = "ctrcfg-test"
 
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -1190,26 +1190,26 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				By("checking if ContainerRuntimeConfig object already exists")
 				ctrcfg, err := getContainerRuntimeConfigFrom(context.TODO(), profile, mcp)
 				Expect(err).ToNot(HaveOccurred(), "failed to get ContainerRuntimeConfig from profile %q mcp %q", profile.Name, mcp.Name)
+				if ctrcfg == nil {
```
Tal-or, line 1193: Can we set the `When` to `Ordered` and add this code in `BeforeAll()` closure?

mrniranjan: addressed in the latest commit

Tal-or:
/lgtm
Thanks

ffromani (commented):
/approve

PR #940, opened 2024-02-08
OCPBUGS-29595: Reduce number of reboots in offline tests

mrniranjan:
Currently in offline tests we revert to original profile after each tests, this is very costly and causing tests to run for long time.

This patch addresses by reverting to the profile after all offline tests are run.

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"sort"
 	"strconv"
 	"strings"
@@ -551,8 +552,9 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 
 	})
 
-	Context("Offlined CPU API", func() {
-		BeforeEach(func() {
+	Context("Offlined CPU API", Ordered, func() {
+		var numaCoreSiblings map[int]map[int][]int
+		BeforeAll(func() {
 			//Saving the old performance profile
 			initialProfile = profile.DeepCopy()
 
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -989,18 +959,18 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
 							allCpus = append(allCpus, cpus)
 						}
 					}
 				}
+				// we don't need to check all the cpus as this may be costly
+				// on BM system which have more than 100 cpus, so we pick any 10 cpus randomly.
+				cpusToCheck := randomCpus(allCpus)
 				for _, node := range workerRTNodes {
-					for _, v := range allCpus {
+					for _, v := range cpusToCheck {
```
Tal-or, line 973: You can still check all cpus and even reduce it to single call using:
`find /sys/devices/system/cpu/cpu* -type f -name online -exec cat {} \;`
The check that all lines are equal to `1`

mrniranjan: addressed it in the latest commit.

ffromani: Or cat `/sys/devices/system/cpu/online` and check the corresponding cpuset. The current method looks good enough, no need for further changes now.

ffromani (commented):
/lgtm

I don't like creating implicit dependencies between tests, but the current state is barely tenable and I fully agree that having tests taking so much time is a different and real issue.

Tal-or:
> /lgtm
> 
> I don't like creating implicit dependencies between tests, but the current state is barely tenable and I fully agree that having tests taking so much time is a different and real issue.

Another option is to use `BeforeSuite` and `AfterSuite` but these tests are running with many other tests so in this case it's not such a good idea and having the test `Ordered` + using `BeforeAll` seems like the most balanced option

mrniranjan:
> > /lgtm
> > I don't like creating implicit dependencies between tests, but the current state is barely tenable and I fully agree that having tests taking so much time is a different and real issue.
> 
> Another option is to use `BeforeSuite` and `AfterSuite` but these tests are running with many other tests so in this case it's not such a good idea and having the test `Ordered` + using `BeforeAll` seems like the most balanced option

Right , but there is another downside to this patch,  if one of the tests fails, Since tests are ordered, ginkgo skips all the other tests in that spec.

Tal-or:
> Right , but there is another downside to this patch, if one of the tests fails, Since tests are ordered, ginkgo skips all the other tests in that spec.

Masking failures is bad indeed

PR #948, opened 2024-02-12
OCPBUGS-29183: e2e: when crun is enabled by default skip checking runc config

mrniranjan:
This patch fixes runc configuration test which
was failing when crun is configured prior to running the test

In this patch we move BeforeAll at the context level to check if any container runtime is pre configured

if crun is preconfigured: skip runc config test
if crun is not preconfigured: check runc configuration and in the subsequent test, we enable crun and test crun and delete crun configuration

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -1163,9 +1163,25 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 		})
 	})
 
-	Context("ContainerRuntimeConfig", func() {
+	Context("ContainerRuntimeConfig", Ordered, func() {
+		var ctrcfg *machineconfigv1.ContainerRuntimeConfig
+		const ContainerRuntimeConfigName = "ctrcfg-test"
+		mcp := &machineconfigv1.MachineConfigPool{}
+		BeforeAll(func() {
+			key := types.NamespacedName{
+				Name: performanceMCP,
+			}
+			Expect(testclient.Client.Get(context.TODO(), key, mcp)).ToNot(HaveOccurred(), "cannot get MCP %q", performanceMCP)
+			By("checking if ContainerRuntimeConfig object already exists")
+			ctrcfg, err = getContainerRuntimeConfigFrom(context.TODO(), profile, mcp)
+			Expect(err).ToNot(HaveOccurred(), "failed to get ContainerRuntimeConfig from profile %q mcp %q", profile.Name, mcp.Name)
+		})
+
 		When("is not given", func() {
 			It("should run high-performance runtimes class with runc as container-runtime", func() {
+				if ctrcfg != nil {
+					Skip("runc is not the default runtime configuration")
+				}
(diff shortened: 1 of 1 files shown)


ffromani (commented):
it seems to me this works only if a specific test order is used. Is that the case?

mrniranjan:
> it seems to me this works only if a specific test order is used. Is that the case?

Yes tests are ordered.

ffromani:
/approve

let's keep looking for balance between test isolation and wall clock execution time

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -1163,9 +1163,25 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
+			Expect(testclient.Client.Get(context.TODO(), key, mcp)).ToNot(HaveOccurred(), "cannot get MCP %q", performanceMCP)
+			By("checking if ContainerRuntimeConfig object already exists")
+			ctrcfg, err = getContainerRuntimeConfigFrom(context.TODO(), profile, mcp)
+			Expect(err).ToNot(HaveOccurred(), "failed to get ContainerRuntimeConfig from profile %q mcp %q", profile.Name, mcp.Name)
+		})
+
 		When("is not given", func() {
 			It("should run high-performance runtimes class with runc as container-runtime", func() {
+				if ctrcfg != nil {
+					Skip("runc is not the default runtime configuration")
```
Tal-or, line 1183: Do we actually want to skip the test?
Why not configuring the system to use `runc`, run the test and then revert back?

mrniranjan: So in our D/S ci, we have enabled to to execute the whole tests against runc and crun.  So this test gets skipped when crun is enabled, but runs when runc is enabled. 

Doing both complicates the process, because then we have to do the following:
1. Delete the existing configure runtime 
2. Execute the test 
3. The next test itself configures crun, and then deletes 
4. Re-create the crun runtime.

Tal-or: If this is problematic due to runtime constrains, and we need to support both runtime versions for the long run, why not having a lane dedicated for crun and lane dedicated to runc

mrniranjan: Yes that's what we have in D/S, But the whole tests suite can run only 1 runtime at a time,

Tal-or: I know but you can have separate lanes with separate configuration

mrniranjan: Yes thats how we do it, we have one lane for crun and one for runc.

yanirq: @Tal-or  crun  in 4.16 + should always be tested  while we can reduce the runc cadence

shajmakh:
Thanks for this. 
/lgtm
/hold
for other reviewers, please unhold as you see fit.

ffromani:
/hold cancel

LGTM as well, but I concur with @Tal-or : we should have more lanes configured accordingly. Capacity is a concern, though.

PR #950, opened 2024-02-12
CNF-11099: set intel_pstate driver to automatic as default

sabbir-47:
## What?
Set CPUFreq driver mode based on hardware generation which will set intel_pstate=active for IceLake and newer processors while it will disable the pstate for older generation of processors.

## Why?

- For FlexRAN (and FlexRAN-like applications), the hardware vendor(I//) recommends to use the intel_pstate CPUFreq driver in active mode with HWP enabled on Ice Lake and later generations. The majority of Telco RAN DU deployments will be on the Ice Lake or newer generation hardware.
- We have One RAN customer who wants the intel pstate to be active as default. [RFE link](https://issues.redhat.com/browse/RFE-4138)

 ## How?
We introduced a function in [tuned](https://github.com/redhat-performance/tuned/blob/master/tuned/profiles/functions/function_intel_recommended_pstate.py) which returns appropriate CPUFreq driver mode based on hardware generation. Then we invoke the function in the custom tuned profile for Openshift in the NTO operator. We still keep pstate to active for HardwareTuning case,because user may want to tune cpu frequencies in older generation processors.

```
[variables]

automatic_pstate=${f:intel_recommended_pstate}
.........
.........

{{if .PerPodPowerManagement}}
{{else if .HardwareTuning}}
cmdline_pstate=+intel_pstate=active
{{else}}
cmdline_pstate=+intel_pstate=${automatic_pstate}
{{end}}
```
It will update the `assets/performanceprofile/tuned/openshift-node-performance` and render the profile with appropriate intel_pstate

## Performance impact on the system
We internally ran KPI tests, i.e. oslat, cyclicTest, cpu utilization and RFC2544 to identify if activating pstate in IceLake and Sapphire Rapids processor servers cause any performance variance. We found no indication of performance degradation.

- IceLake: [KPI test Result](http://ocp-far-edge-vran-deployment-kpi.hosts.prod.psi.rdu2.redhat.com/backend/api/v1/reports/file/html/5a6d9b92-141b-4e78-97f8-53c6fc9b7c65)
- Sapphire Rapids: [KPI test Result](http://ocp-far-edge-vran-deployment-kpi.hosts.prod.psi.rdu2.redhat.com/backend/api/v1/reports/file/html/c8ad5389-6772-4d4a-9860-3d786fa1ee5b)

## Can it be overridden by user?
User can always override this kernel configuration with tuned, for example if they want to disable intel_pstate:

```
kind: Tuned
........
spec:
    profile:
    - data: |
        ....
        [bootloader]
        cmdline_pstate=intel_pstate=disable

```

/cc @MarSik @yanirq @jmencak @bartwensley

### DIFF
--- a/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
+++ b/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
@@ -284,7 +284,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 					"vm.stat_interval":              "10",
 				}
 				kernelParameters := []string{noHzParam, "tsc=reliable", "nosoftlockup", "nmi_watchdog=0", "mce=off", "skew_tick=1",
-					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate=disable", "idle=poll"}
+					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate= {f:intel_recommended_pstate}", "idle=poll"}
 
 				wg := sync.WaitGroup{}
 				By("Waiting for TuneD to start on nodes")
@@ -343,7 +343,8 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 				cmdline, err := nodes.ExecCommandOnMachineConfigDaemon(context.TODO(), &workerRTNodes[0], []string{"cat", "/proc/cmdline"})
 				Expect(err).ToNot(HaveOccurred())
 				Expect(cmdline).To(ContainSubstring("intel_pstate=passive"))
-				Expect(cmdline).ToNot(ContainSubstring("intel_pstate=disable"))
+				Expect(cmdline).ToNot(ContainSubstring("intel_pstate=active"))
+				Expect(cmdline).ToNot(ContainSubstring("intel_pstate= {f:intel_recommended_pstate}"))
 
 				By("Verifying tuned profile")
 				key := types.NamespacedName{
@@ -404,7 +405,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 					"vm.stat_interval":              "10",
 				}
 				kernelParameters := []string{noHzParam, "tsc=reliable", "nosoftlockup", "nmi_watchdog=0", "mce=off", "skew_tick=1",
-					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate=disable", "idle=poll"}
+					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate= {f:intel_recommended_pstate}", "idle=poll"}
 
 				wg := sync.WaitGroup{}
 				By("Waiting for TuneD to start on nodes")
@@ -567,7 +568,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 				}
 				wg.Wait()
 
-				//Update the profile to disable HighPowerConsumption and enable PerPodPowerManagment
+				//Update the profile to enable HighPowerConsumption and disable PerPodPowerManagment
 				profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
 					HighPowerConsumption:  pointer.Bool(true),
 					RealTime:              pointer.Bool(true),
@@ -605,7 +606,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 					"vm.stat_interval":              "10",
 				}
 				kernelParameters = []string{noHzParam, "tsc=reliable", "nosoftlockup", "nmi_watchdog=0", "mce=off", "skew_tick=1",
-					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate=disable", "idle=poll"}
+					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate= {f:intel_recommended_pstate}", "idle=poll"}
 
 				wg = sync.WaitGroup{}
 				By("Waiting for TuneD to start on nodes")
--- a/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
+++ b/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
@@ -284,7 +284,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 					"vm.stat_interval":              "10",
 				}
 				kernelParameters := []string{noHzParam, "tsc=reliable", "nosoftlockup", "nmi_watchdog=0", "mce=off", "skew_tick=1",
-					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate= {f:intel_recommended_pstate}", "idle=poll"}
+					"processor.max_cstate=1", "intel_idle.max_cstate=0", "idle=poll"}
 
 				wg := sync.WaitGroup{}
 				By("Waiting for TuneD to start on nodes")
@@ -306,6 +306,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 							fmt.Sprintf("stalld is not running on %q when it should", node.Name))
 
 						By(fmt.Sprintf("Checking TuneD parameters on %q", node.Name))
+						kernelParameters = append(kernelParameters, utilstuned.AddPstateParameter(context.TODO(), node))
 						utilstuned.CheckParameters(context.TODO(), node, sysctlMap, kernelParameters, stalldEnabled, rtKernel)
 					}()
 				}
@@ -344,7 +345,6 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 				Expect(err).ToNot(HaveOccurred())
 				Expect(cmdline).To(ContainSubstring("intel_pstate=passive"))
 				Expect(cmdline).ToNot(ContainSubstring("intel_pstate=active"))
-				Expect(cmdline).ToNot(ContainSubstring("intel_pstate= {f:intel_recommended_pstate}"))
 
 				By("Verifying tuned profile")
 				key := types.NamespacedName{
@@ -405,7 +405,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 					"vm.stat_interval":              "10",
 				}
 				kernelParameters := []string{noHzParam, "tsc=reliable", "nosoftlockup", "nmi_watchdog=0", "mce=off", "skew_tick=1",
-					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate= {f:intel_recommended_pstate}", "idle=poll"}
+					"processor.max_cstate=1", "intel_idle.max_cstate=0", "idle=poll"}
 
 				wg := sync.WaitGroup{}
 				By("Waiting for TuneD to start on nodes")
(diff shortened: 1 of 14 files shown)


MarSik:
I am a bit worried about the assumptions here, because it is a backwards incompatible change.

> The majority of Telco RAN DU deployments will be on the Ice Lake or newer generation hardware.

I still see many reports and labs where older generations are in use.

Also I seem to remember we disabled p-states and c-states for the most sensitive real time cases in the past as the power management features were introducing latency. Is that fixed on Ice Lake and newer cpus?

Switching the knob is fine, but lets make sure all the pieces fit together for all cases and old deployments we support here.

Maybe we should just use the cpu matching tuned provides and limit this to new enough cpus? I wonder if https://github.com/redhat-performance/tuned/commit/8d9cd00387426ed0cf220f920be4f9e185f61e12 lets us do what we need or some other functionality should be introduced.

sabbir-47:
> I am a bit worried about the assumptions here, because it is a backwards incompatible change.
> 
> > The majority of Telco RAN DU deployments will be on the Ice Lake or newer generation hardware.
> 
> I still see many reports and labs where older generations are in use.

 Yes in many places, we still use older servers.
 
> Also I seem to remember we disabled p-states and c-states for the most sensitive real time cases in the past as the power management features were introducing latency. Is that fixed on Ice Lake and newer cpus?

According to @joemario from the email chain where you are included as well with subject: `Using intel_pstate CPUFreq driver for Telco RAN`:
```
Some historical context.  Back when the cpu-partitioning and realtime profiles got developed, which was around RHEL-7.4, the intel_pstate driver was notorious for inducing jitter.  It was common practice to disable it in any latency sensitive application.
In recent years, Intel has fixed the jitter issues in the intel_pstate driver.
I know the Intel SST_BF feature requires the intel_pstate driver be enabled, and I've seen customers get good latency results with it enabled.
```
> Switching the knob is fine, but lets make sure all the pieces fit together for all cases and old deployments we support here.
> 
> Maybe we should just use the cpu matching tuned provides and limit this to new enough cpus? I wonder if [redhat-performance/tuned@8d9cd00](https://github.com/redhat-performance/tuned/commit/8d9cd00387426ed0cf220f920be4f9e185f61e12) lets us do what we need or some other functionality should be introduced.

Do we have a way to understand the processors generation from `/proc/cpuinfo` or `lscpu` or other places. My understanding is cascade lake, ice lake, sapphire rapids - these terms are used as marketing purposes. Please let me know if there is a way to distinguish between processors generation.
Adding @bartwensley for more information.

bartwensley:
Good questions @MarSik - I'll try to summarize:

> Also I seem to remember we disabled p-states and c-states for the most sensitive real time cases in the past as the power management features were introducing latency. Is that fixed on Ice Lake and newer cpus?

The latency issues have been fixed on Intel processors of the Ice Lake and newer generations. The recommendation from Intel is to use `intel_pstate=active` on these generations.

> Maybe we should just use the cpu matching tuned provides and limit this to new enough cpus? I wonder if [redhat-performance/tuned@8d9cd00](https://github.com/redhat-performance/tuned/commit/8d9cd00387426ed0cf220f920be4f9e185f61e12) lets us do what we need or some other functionality should be introduced.

My understanding is that there is no reliable way of determining the processor generation at runtime, so I do not believe this is an option. The above change only looks at /dev/cpuinfo which only contains the model number, which would require an unmaintainable regex to match all current and future Intel processor model names. 

The proposed change would only be an issue for a user that:
1. Has a cascade lake (or older) processor.
2. Is applying the openshift-node-performance performance profile.
3. Is running a latency sensitive application.

In that case, the user would be required to override the intel_pstate to set it to disabled as described in the PR description above. This would have to be done as part of new deployments or when upgrading to the version of the NTO that contains this change. We would want to add a release not for this. 

The alternative is to withdraw this PR and require all users running latency sensitive applications on Ice Lake (or newer) processors to override the intel_pstate to set it to active.

Given that the Ice Lake generation was launched in 2019 and that most users will be using these processors to run low latency applications, I would prefer to make this change to the default intel_pstate configuration, but I agree this could be a backwards incompatible change in some cases and would require the user to make a configuration change.

We need consensus here before making this change.

/cc @browsell

sabbir-47:
@MarSik Actually i fixed that test with my last commit, it seemed to pass now! I needed to rebase to get newer changes.

joemario:
If this change is important to OCP for newer Intel cpus, would it make sense to make the change in TuneD?    Non-OCP environments would be able to benefit as well if it were in TuneD.

test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
```diff
@@ -289,7 +289,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 					"vm.stat_interval":              "10",
 				}
 				kernelParameters := []string{noHzParam, "tsc=reliable", "nosoftlockup", "nmi_watchdog=0", "mce=off", "skew_tick=1",
-					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate=disable", "idle=poll"}
+					"processor.max_cstate=1", "intel_idle.max_cstate=0", "intel_pstate=${automatic_pstate}", "idle=poll"}
```
sabbir-47, line 292: @MarSik Do we know where or in what kind of machine this `e2e-gcp-pao-workloadhints` test runs? If tuned hasn't released a new version, the `intel_recommended_pstate` function will not be evaluated. In addition, without knowing underlying hw, we also don't know whether `${automatic_pstate}` will be resolved to active or disable.

sabbir-47: @MarSik So i found out that in this prow job, functional tests run, and from the log i can see that the kernel parameters set in the node are
`<string>: BOOT_IMAGE=(hd0,gpt3)/ostree/rhcos-653f1642375587f2cc93c3a0d201773ed3fd73c0567f2cc8fea333e1f2d236c4/vmlinuz-5.14.0-427.el9.x86_64+rt ostree=/ostree/boot.0/rhcos/653f1642375587f2cc93c3a0d201773ed3fd73c0567f2cc8fea333e1f2d236c4/0 ignition.platform.id=gcp ......intel_pstate= {f:intel_recommended_pstate}`

I think what happened here is 
- tuned doesn't have this function yet, so it can't resolve. i.e. **intel_pstate= {f:intel_recommended_pstate}**
- when the function will be available, do we know whether this test will run in pre-IceLake or IceLake+ servers?

Also, i think my proposed changes in the workloadhints.go is wrong, because we will either have `disable|active` in those cases where i put `intel_pstate=${automatic_pstate}`

I can change `intel_pstate=${automatic_pstate}` to `intel_pstate= {f:intel_recommended_pstate}` in the kernelParams to pass the test, but this test will fail once the function becomes available in the tuned side.

or may be, i'm mistaken, these tests probably never rely on tuned at all?

How do we proceed?

MarSik: Hmm, this is an interesting issue. We actually do not guarantee a specific platform. @mrniranjan Any idea how to make the test hardware aware?

mrniranjan: Currently to execute tests only hardware we verify it the worker nodes  have  2 numa nodes and number of cpus is more than 32.

MarSik: The kernel argument test will have to be improved to read the information about what CPU was used.

sabbir-47:
/hold
needs TuneD FDP 24.C release first

sabbir-47:
@MarSik @yanirq @jmencak @bartwensley can the PR get the approval since the tuned 24.C FDP release happened 2 days back?

test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
```diff
@@ -628,6 +629,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 							fmt.Sprintf("stalld is not running on %q when it should", node.Name))
 
 						By(fmt.Sprintf("Checking TuneD parameters on %q", node.Name))
+						kernelParameters = append(kernelParameters, utilstuned.AddPstateParameter(context.TODO(), node))
```
sabbir-47, line 632: Hi @MarSik, so after the recent tuned merge, the var `intel_pstate= {f:intel_recommended_pstate}` is expanding to either `active` or `disable`.  It is not causing any issue for render test cases, but for `e2e-gcp-pao-workloadhints` a real cluster is deployed so the variable is expanding. The problem is, i don't think we can anticipate where these tests are running, therefore i added a check with `func AddPstateParameter()` to check the pmu_name and add the corresponding pstate mode in the kernel Parameters. But it seems the file doesn't exist on the node and the test fails with:

```
[FAILED] failed to execute command ["cat" "/sys/devices/cpu/caps/pmu_name"] on node: "ci-op-lrm0rmvm-36360-nncwg-worker-a-5nnq6"
  Unexpected error:
      <*fmt.wrapError | 0xc00059e160>: 
      failed to run command [cat /sys/devices/cpu/caps/pmu_name]: output "cat: /sys/devices/cpu/caps/pmu_name: No such file or directory\r\n"; error ""; command terminated with exit code 1
      {
          msg: "failed to run command [cat /sys/devices/cpu/caps/pmu_name]: output \"cat: /sys/devices/cpu/caps/pmu_name: No such file or directory\\r\\n\"; error \"\"; command terminated with exit code 1",
          err: <exec.CodeExitError>{
              Err: <*errors.errorString | 0xc0005120a0>{
                  s: "command terminated with exit code 1",
              },
              Code: 1,
          },
      }
      ```
      
      @mrniranjan Do these tests get executed on baremetal nodes or vms?

mrniranjan: They get executed on VM's and also on BM"s (but manually)

MarSik: Yeah, the test should skip gracefully on VMs (or figure out it is a VM and test for the proper value = disable?)

sabbir-47: I think, i found it how to verify if it is a [VM](https://github.com/openshift/cluster-node-tuning-operator/blob/dd2698c43517191d17dc5c4ac842e6d4dc3aab32/test/e2e/performanceprofile/functests/utils/infrastructure/vm.go#L13)

mrniranjan: @sabbir-47 That code is broken, since the systemd-detect-vm is executed from a container, it will always return podman/container and the test will skip (even on BM)

sabbir-47: Well, is there any better way to verify then? I can hardcode the pstate to disable  in the kernel parameters and the test will always pass if the tests are executed on VM. It will fail if the tests are run on BM(IceLake+ servers), where it expects to set the pstate to active.

mrniranjan: we have to fix that code, for now just check the number of cpus or execute lscpu and check  Bios Vendor id , if it's QEMU it's a VM

sabbir-47: All tests passed with recent changes.

sabbir-47:
@MarSik @yanirq @jmencak @bartwensley all tests passed. Kindly consider reviewing it.

MarSik (approved):
/lgtm

PR #954, opened 2024-02-14
OCPBUGS-23167: Add performance real time tuned template

rbaturov:
Kernel parameters that are not supported on RT kernel systems are being applied, causing errors to be logged to the tuned daemon, resulting in a degraded profile state.
Therefore, I added the openshift-node-performance-rt profile that would be included if an RT kernel is detected, thereby dropping the unsupported kernel parameters before they are applied.

### DIFF
--- a/hack/render-sync.sh
+++ b/hack/render-sync.sh
@@ -41,3 +41,5 @@ function rendersync() {
 rendersync manual-cluster/performance base/performance default
 rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap/no-mcp
 rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap-cluster/extra-mcp bootstrap/extra-mcp
+rendersync --owner-ref none -- base/performance manual-cluster/performance no-ref 
+rendersync --owner-ref none -- base/performance manual-cluster/cpuFrequency default/cpuFrequency 
--- a/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
@@ -37,6 +37,7 @@ const (
 	templateReservedCpuMaxFreq              = "ReservedCpuMaxFreq"
 	templateIsolatedCpuList                 = "IsolatedCpuList"
 	templateReservedCpuList                 = "ReservedCpuList"
+	templatePerformanceProfileName          = "PerformanceProfileName"
 )
 
 func new(name string, profiles []tunedv1.TunedProfile, recommends []tunedv1.TunedRecommend) *tunedv1.Tuned {
@@ -59,6 +60,8 @@ func new(name string, profiles []tunedv1.TunedProfile, recommends []tunedv1.Tune
 // NewNodePerformance returns tuned profile for performance sensitive workflows
 func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tuned, error) {
 	templateArgs := make(map[string]interface{})
+
+	templateArgs[templatePerformanceProfileName] = profile.Name
 
 	if profile.Spec.CPU.Isolated != nil {
 		templateArgs[templateIsolatedCpus] = string(*profile.Spec.CPU.Isolated)
@@ -215,11 +218,20 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 		return nil, err
 	}
 
+	RealTimeKernelProfileData, err := getProfileData(filepath.Join("tuned", components.ProfileNamePerformanceRT), templateArgs)
+	if err != nil {
+		return nil, err
+	}
 	name := components.GetComponentName(profile.Name, components.ProfileNamePerformance)
+	RealTimeKernelProfileName := components.GetComponentName(profile.Name, components.ProfileNamePerformanceRT)
 	profiles := []tunedv1.TunedProfile{
 		{
 			Name: &name,
 			Data: &profileData,
+		},
+		{
+			Name: &RealTimeKernelProfileName,
+			Data: &RealTimeKernelProfileData,
 		},
 	}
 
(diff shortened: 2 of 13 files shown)


pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -214,13 +214,22 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 	if err != nil {
 		return nil, err
 	}
-
+	RealTimeKernelProfileData, err := getProfileData(filepath.Join("tuned", components.ProfileNamePerformanceRT), templateArgs)
+	if err != nil {
+		return nil, err
+	}
 	name := components.GetComponentName(profile.Name, components.ProfileNamePerformance)
+	RealTimeKernelProfileName := components.ProfileNamePerformanceRT
```
MarSik, line 222: You need to use the GetComponentName method here too, otherwise two perf profiles will try to create the same profile.

rbaturov: I don't get why two perf profiles will try to create the same profile.
Anyway I tried to do that before, what happend was that the tuned profile got posted to the cluster with the name of the profile, i. e "-performance"  suffix (openshift-node-performance-rt-performance) while the include section is looking for hardcoded "openshift-node-performance-rt"
`2024-02-14 10:49:19,013 ERROR    tuned.daemon.daemon: Cannot set initial profile. No tunings will be enabled: Cannot load profile(s) 'openshift-node-performance-performance': Cannot find profile 'openshift-node-performance-rt' in '['/etc/tuned', '/usr/lib/tuned']'.
`

MarSik: Think about the flow when you have two PerformanceProfiles in a cluster:

Each performance profile generates a new Tuned object that contains the tuned profiles. So two performance profiles result in two Tuned objects that will both contain the same openshift-node-performance-rt tuned profile. This results in NTO trying to save both of the defined profiles into the same filename.

You are right the include will have to be improved to refer to the proper name. Template variable with the perf profile name can do that.

rbaturov: Fixed that.

MarSik (approved):
/lgtm

hack/render-sync.sh
```diff
@@ -32,4 +32,8 @@ function rendersync() {
 
 rendersync manual-cluster/performance base/performance default
 rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap/no-mcp
-rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap-cluster/extra-mcp bootstrap/extra-mcp
\ No newline at end of file
+rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap-cluster/extra-mcp bootstrap/extra-mcp
+rendersync base/performance manual-cluster/performance no-ref
```
ffromani, line 36: the `no-ref` set should not include any reference label, we need to parametrize the render to deal with this.
I'm not sure why we are adding the render-sync invocations in this PR btw.

rbaturov: I will issue a different PR that would parametrize the render to deal with this.

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -114,6 +114,21 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 				Expect(strings.TrimSpace(activeProfileName)).To(Equal(tunedExpectedName), "active profile name mismatch got %q expected %q", activeProfileName, tunedExpectedName)
 			}
 		})
+
+		It("Tuned profile shouldn't be degraded", func() {
```
ffromani, line 118: should not be degraded after what? in this test you check a condition (this part LGTM) but there is no obvious link to how to get to this state. IOW, does this test depend on a specific order? which part of the system creates the conditions we're checking?

hack/render-sync.sh
```diff
@@ -40,4 +40,6 @@ function rendersync() {
 
 rendersync manual-cluster/performance base/performance default
 rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap/no-mcp
-rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap-cluster/extra-mcp bootstrap/extra-mcp
\ No newline at end of file
+rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap-cluster/extra-mcp bootstrap/extra-mcp
+rendersync --owner-ref none -- base/performance manual-cluster/performance no-ref 
+rendersync --owner-ref none -- base/performance manual-cluster/cpuFrequency default/cpuFrequency
```
ffromani, line 45: don't we want owner reference in this case? it should be only `no-ref` without the reference (hence the name)

ffromani (commented):
this seems like a good direction, few comments inside

test/e2e/performanceprofile/functests/1_performance/performance.go
rbaturov: should not be degraded after tuned profile has been applied to the system.
The object will be degraded if error messages are observed by the tuned daemon when applying the TuneD daemon profile.

MarSik: @ffromani Any error in the tuned log causes the state to switch to degraded. This fix removes a typical case - a sysctl that does not exist for the given kernel.

hack/render-sync.sh
MarSik: This syncs the manifests for this test https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests-render-command/1_render_command/render_test.go#L115 

And it uses no owner references. Probably copy&paste. I believe we should default to having the references most of the time, but I do not think we should fix that as part of this PR.

ffromani: fair enough

test/e2e/performanceprofile/functests/1_performance/performance.go
ffromani: sure, and what in the test causes any change in the system which could lead to Degraded state?
After reading the surrounding tests (something that could have pointed out earlier) I see other tests do in this debatable way, so this is a more general issue which is out of the scope here.

ffromani (commented):
/lgtm

PR #956, opened 2024-02-14
OCPBUGS-22095: Render: Add MCSelector to missing default MCPs

jlojosnegros:
Newly added MCPs lack MachineConfigSelector field.


MarSik (approved):
/lgtm

PR #957, opened 2024-02-15
OCPBUGS-29594: e2e: fix comparing incompatible types

mrniranjan:
This patch fixes comparing incompatible types,
profile.Spec.RealTimeKernel.Enabled == pointer.Bool(false)

Instead of using pointer.Bool, directly compare the boolean value.

### DIFF
--- a/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
+++ b/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
@@ -84,7 +84,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 				By("Modifying profile")
 				profile.Spec.WorkloadHints = nil
 
-				if profile.Spec.RealTimeKernel.Enabled == pointer.Bool(false) {
+				if *profile.Spec.RealTimeKernel.Enabled != true {
 					testutils.KnownIssueJira("OCPBUGS-28828")
 				}
 				profile.Spec.RealTimeKernel = &performancev2.RealTimeKernel{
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
```diff
@@ -84,7 +84,7 @@ var _ = Describe("[rfe_id:49062][workloadHints] Telco friendly workload specific
 				By("Modifying profile")
 				profile.Spec.WorkloadHints = nil
 
-				if profile.Spec.RealTimeKernel.Enabled == pointer.Bool(false) {
+				if *profile.Spec.RealTimeKernel.Enabled != true {
```
jmencak, line 87: So I'm assuming we want to skip the test when 

```
*profile.Spec.RealTimeKernel.Enabled == false 
```

right? Would it be a bit more readable? Similarly below.

jmencak (approved):
/approve

test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go
mrniranjan: Right == false would have made more readable. updating the patch

jmencak:
The only concern I have is a potential null-pointer dereference, but if you're sure this cannot happen:
/lgtm

mrniranjan:
> The only concern I have is a potential null-pointer dereference, but if you're sure this cannot happen: /lgtm

@jmencak If for some reason profile is nil, it will fail very early in the test.

jmencak:
> > The only concern I have is a potential null-pointer dereference, but if you're sure this cannot happen: /lgtm
> 
> @jmencak If for some reason profile is nil, it will fail very early in the test.

I'm not worried about profile == nil.  Can `profile.Spec.RealTimeKernel` be nil though?

jmencak:
/hold
please unhold when we're sure this cannot happen ^

mrniranjan:
@jmencak 

Profile.Spec.RealTimeKernel cannot be nil because ,  realTimeKernel attribute of Performance Profile is a mandatory field to be given, It take 2 values true or false. 

With this value performance profile cannot be applied.

PR #960, opened 2024-02-19
NO-JIRA: Add support to inject owner-ref argument to render command

rbaturov:
We need this change because we also want to use also non default values for the owner-ref like the value none.
  The work on [this PR depends](https://github.com/openshift/cluster-node-tuning-operator/pull/954) on these modifications to be merged.

### DIFF
--- a/hack/render-sync.sh
+++ b/hack/render-sync.sh
@@ -5,6 +5,14 @@ function join_by { local IFS="$1"; shift; echo "$*"; }
 
 function rendersync() {
   INPUT_DIRS=()
+  EXTRA_ARGS=()
+  if [[ "$1" =~ ^-.* ]]; then
+      while [[ "x$1" != "x--" ]]; do
+          EXTRA_ARGS+=("$1")
+          shift
+      done
+      shift
+  fi
   while (( $# > 1 )); do
     INPUT_DIRS+=("${WORKDIR}/test/e2e/performanceprofile/cluster-setup/$1")
     shift
@@ -18,7 +26,7 @@ function rendersync() {
   ARTIFACT_DIR=$(mktemp -d)
 
   cd "${WORKDIR}" || { echo "failed to change dir to ${WORKDIR}"; exit; }
-  _output/cluster-node-tuning-operator render \
+  _output/cluster-node-tuning-operator render ${EXTRA_ARGS[@]} \
   --asset-input-dir "${INPUT_DIRS}" \
   --asset-output-dir "${ARTIFACT_DIR}"
 


rbaturov:
@Tal-or @ffromani

ffromani:
this works, but IMO it should work the other way around: the owner-ref should be on by default, and opted-out only in the few (currently, exactly one and only)

hack/render-sync.sh
```diff
@@ -30,6 +32,6 @@ function rendersync() {
   rm -r "${ARTIFACT_DIR}"
 }
 
-rendersync manual-cluster/performance base/performance default
-rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap/no-mcp
-rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap-cluster/extra-mcp bootstrap/extra-mcp
\ No newline at end of file
+rendersync manual-cluster/performance base/performance default label-name
```
MarSik, line 35: I think mixing directories and ownership reference arguments is confusing a bit. The traditional unix way would be something like:

`rendersync --owner-ref label-name -- manual-cluster/performance base/performance default`

In other words, extra arguments followed by -- separator and then the rendering arguments.

The code to implement this is a bit more complicated than what you have here, but I believe it would make the code less error prone.

ffromani: This would be the best option indeed. I was about to settle for a little less, but if we can get this far that would be great.

Tal-or: The only thing that bothers me with this approach is that we basically implement `render` command line again but in the lesser convenient way which is using bash for parsing the arguments.

Tal-or: Maybe we can enhance render command to read its configuration from yaml file something like:
``` yaml
config:
  - inputDir: "foo"
    outputDir: "bar"
    ownerRef: none
 -  inputDir: "foo2"
    outputDir: "bar2"
    ownerRef: k8s
 -  inputDir: "foo2"
    outputDir: "bar2"
    // use default owner-ref
```

MarSik: I think we can implement my proposal like this. Just below line 7 include

```
EXTRA_ARGS=()
if [[ "$1" =~ ^-.* ]]; then
    while (( "x$1" != "x--" )); do
        EXTRA_ARGS+=("$1")
        shift
    done
    shift
fi
```

And then in the execution of render use `"${EXTRA_ARGS[@]}"`

We should make the right value the default one in the render command so we do not have to use this method too often.

ffromani:
LGTM
how do we test this change?

Tal-or:
> LGTM how do we test this change?

There was a test which was failing and exposed the issue. if the test will pass then it mean we're out of the woods

ffromani:
/approve
/lgtm

per https://github.com/openshift/cluster-node-tuning-operator/pull/960#issuecomment-1954175713

ffromani:
this PR won't affect the payload

ffromani:
/retitle NO-JIRA: Add support to inject owner-ref argument to render command

PR #961, opened 2024-02-19
NO-JIRA: e2e:mixedcpus: check env variables of init process

Tal-or:
We had a bug (https://github.com/cri-o/cri-o/pull/7710) in CRI-O on which the Mixed-CPUs environment variables were injected to all processes but the init process.

The test has been modified to verify that the env variables are injected to the init process as well. From this point all child processes would inherit the Mixed-CPUs environment variables, unless removed explicitly, but we double check the child process anyway.

### DIFF
--- a/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
+++ b/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
@@ -13,6 +13,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/client-go/kubernetes"
 	kubeletconfig "k8s.io/kubelet/config/v1beta1"
 	"k8s.io/utils/cpuset"
 	"k8s.io/utils/pointer"
@@ -46,6 +47,8 @@ const (
 	// smt alignment won't allow us to run the test pod with a single core, hence we should cancel it.
 	numberOfCoresThatRequiredCancelingSMTAlignment = 4
 	restartCooldownTime                            = 1 * time.Minute
+	isolatedCpusEnv                                = "OPENSHIFT_ISOLATED_CPUS"
+	sharedCpusEnv                                  = "OPENSHIFT_SHARED_CPUS"
 )
 
 var _ = Describe("Mixedcpus", Ordered, func() {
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -457,3 +470,24 @@ func mustParse(cpus string) *cpuset.CPUSet {
 	Expect(err).ToNot(HaveOccurred(), "failed to parse cpuset; cpus=%q", cpus)
 	return &set
 }
+
+func fetchSharedCPUsFromEnv(c *kubernetes.Clientset, p *corev1.Pod, containerName string) *cpuset.CPUSet {
+	GinkgoHelper()
+	// we should check the environment variable under the init process (pid 1)
+	cmd := []string{
+		"/bin/cat",
+		"/proc/1/environ",
```
ffromani, line 479: nice idea. I was thinking about changing the base image to dump the env, but this is even better

ffromani (commented):
/lgtm

very nice idea

ffromani:
for my understanding, why we moved away from `BeEmpty()` ?

Tal-or:
> for my understanding, why we moved away from `BeEmpty()` ?

The variables we checked were `cpuset.CPU` struct. you cannot check `BeEmpty()` for struct (only for string, array, map, chan, or slice)
It was done by me hurry up :) 
My actually meaning was to use `IsEmpty()` implemented by the   `cpuset.CPU` struct

ffromani:
thanks, makes sense to me.
I would still add a better explanation around here: 

```
Expect(shared.IsEmpty()).ToNot(BeTrue())
```
because this is likely to generate not-so-helpful error message. I'd use something like

```
Expect(shared.IsEmpty()).ToNot(BeTrue(), "shared cpuset is empty")
```

Tal-or:
> thanks, makes sense to me. I would still add a better explanation around here:
> 
> ```
> Expect(shared.IsEmpty()).ToNot(BeTrue())
> ```
> 
> because this is likely to generate not-so-helpful error message. I'd use something like
> 
> ```
> Expect(shared.IsEmpty()).ToNot(BeTrue(), "shared cpuset is empty")
> ```

Np will update in the close hour

yanirq:
/label acknowledge-critical-fixes-only label
very low risk - not affecting payload

yanirq:

very low risk - not affecting payload

PR #964, opened 2024-02-21
NO-JIRA: Add positive tests with ginkgo timeout

swatisehgal:
Since ginkgo default timeout was updated from 24h to 1hr we add additional tests to ensure that the tests are executed taking ginkgo timeout into account and condering the values specified for LATENCY_TEST_RUNTIME and LATENCY_TEST_DELAY by the user.

### DIFF
--- a/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
+++ b/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
@@ -231,8 +231,8 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 
 	successRuntime := "30"
 	// Using a timeout value such that ginkgo timeout > runtime + latency to ensure successfull runs
-	successGinkgoTimeout := "400s"
-	testSet = append(testSet, latencyTest{testDelay: "200", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
+	successGinkgoTimeout := "200s"
+	testSet = append(testSet, latencyTest{testDelay: "140", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 	testSet = append(testSet, latencyTest{testDelay: "0", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 	testSet = append(testSet, latencyTest{testDelay: "0", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "6", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 	testSet = append(testSet, latencyTest{testDelay: "1", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
--- a/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
+++ b/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
@@ -88,15 +88,26 @@ type latencyTest struct {
 	testCpus              string
 	outputMsgs            []string
 	toolToTest            string
+	ginkgoTimeout         string
 }
 
 var _ = DescribeTable("Test latency measurement tools tests", func(testGroup []latencyTest, isPositiveTest bool) {
 	format.MaxLength = 0
+	var output []byte
+	var err error
 	for _, test := range testGroup {
 		clearEnv()
 		testDescription := setEnvAndGetDescription(test)
 		By(testDescription)
-		output, err := exec.Command(testExecutablePath, "-ginkgo.v", "-ginkgo.focus", test.toolToTest).Output()
+
+		if isPositiveTest {
+			// We only override ginkgo default value for positive tests
+			output, err = exec.Command(testExecutablePath, "-ginkgo.v", "-ginkgo.focus", test.toolToTest, "-ginkgo.timeout", test.ginkgoTimeout).Output()
+		} else {
+			// default ginkgo timeout value for negative tests
+			output, err = exec.Command(testExecutablePath, "-ginkgo.v", "-ginkgo.focus", test.toolToTest).Output()
+		}
+
 		if err != nil {
 			//we don't log Error level here because the test might be a negative check
 			testlog.Info(err.Error())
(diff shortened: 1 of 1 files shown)


ffromani (commented):
LGTM besides an inline question. I have limited experience here, so I'll leave the final approval to @shajmakh

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -88,6 +88,7 @@ type latencyTest struct {
 	testCpus              string
 	outputMsgs            []string
 	toolToTest            string
+	ginkgoTimeout         string
```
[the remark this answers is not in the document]
swatisehgal: In real world when the user runs latency tests, if the field is not set it defaults to ginkgo timeout which is now 1 hr. The positive tests should run and pass as expected but I was trying to be more explicit here with the timeout value. The next step would be to add similar negative tests where ginkgo.timeout < LATENCY_TEST_RUNTIME where this field will be necessary.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -96,7 +97,8 @@ var _ = DescribeTable("Test latency measurement tools tests", func(testGroup []l
 		clearEnv()
 		testDescription := setEnvAndGetDescription(test)
 		By(testDescription)
-		output, err := exec.Command(testExecutablePath, "-ginkgo.v", "-ginkgo.focus", test.toolToTest).Output()
+		output, err := exec.Command(testExecutablePath, "-ginkgo.v", "-ginkgo.focus", test.toolToTest, "-ginkgo.timeout", test.ginkgoTimeout).Output()
```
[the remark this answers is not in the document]
swatisehgal: Nope, not incorrect, just more explicit.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -224,8 +224,8 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 
 	successRuntime := "30"
 	// Using a timeout value such that ginkgo timeout > runtime + latency to ensure successfull runs
-	successGinkgoTimeout := "400s"
-	testSet = append(testSet, latencyTest{testDelay: "200", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
+	successGinkgoTimeout := "200s"
+	testSet = append(testSet, latencyTest{testDelay: "20", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
```
shajmakh, line 228: I believe we should keep one test with 200 seconds of delay, because we had a bug earlier that failed to run the tests with more than 120s seconds:
https://issues.redhat.com/browse/OCPBUGS-5219 
maybe you can reduce it a bit but better to keep it above 120s.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -154,6 +156,9 @@ var _ = DescribeTable("Test latency measurement tools tests", func(testGroup []l
 	Entry("[test_id:42853] Oslat - Verify that the latency tool test should print an expected error message when passing invalid environment variables values", getNegativeTests(oslat), negativeTesting),
 	Entry("[test_id:42115] Cyclictest - Verify that the tool is working properly with valid environment variables values", getValidValuesTests(cyclictest), positiveTesting),
 	Entry("[test_id:42852] Cyclictest - Verify that the latency tool test should print an expected error message when passing invalid environment variables values", getNegativeTests(cyclictest), negativeTesting),
+	Entry("[test_id:xxxxx] Oslat - Verify that the tool is working properly when latency test runtime < ginkgo timeout", getValidValuesTests(oslat), positiveTesting),
+	Entry("[test_id:xxxxx] Cyclictest - Verify that the tool is working properly when latency test runtime < ginkgo timeout", getValidValuesTests(cyclictest), positiveTesting),
+	Entry("[test_id:xxxxx] Hwlatdetect - Verify that the tool is working properly when latency test runtime < ginkgo timeout", getValidValuesTests(hwlatdetect), positiveTesting),
```
shajmakh, line 161: this way we are running same test sets twice (positiveTesting for the three tools are called twice for each tool). I believe it is simply enough to keep the ginkgo.timeout appended by default to each test in the positive set, and the current positive entries will cover that.

swatisehgal: Please take a look at the most recent changes to confirm that I understand this comment correctly and have addressed it properly.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -254,34 +259,34 @@ func getNegativeTests(toolToTest string) []latencyTest {
 		latencyFailureMsg = hwlatdetectFail
 	}
 
-	testSet = append(testSet, latencyTest{testDelay: "0", testRuntime: "5", testMaxLatency: "1", outputMsgs: []string{latencyFailureMsg, fail}, toolToTest: toolToTest})
```
shajmakh, line 257: adding ginkgo.timeout value to the negative sets looks risky. The goal of these tests is to test that the latency tests fail on the expected reason. If for any reason a negative test failed for timing out because of ginkgo 10s timeout, this will not give us any clue on whether the check passed or not. 
implementing a negative test for the original bug should be written a bit differently than the positive. for now, what I'd go with is running negative tests with default ginkgo.timeout. let me know what you think.

swatisehgal: Thanks for the feedback and apologies for the delay in getting back to this. I think your suggestion makes sense. My intention was to keep the ginkgo timeout high enough compared to the runtime to not interfere with the original intent of the test I will update now so that the negative tests are executed with the default ginkgo timeout.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -218,31 +230,33 @@ func getValidValuesTests(toolToTest string) []latencyTest {
...
-	testSet = append(testSet, latencyTest{testDelay: "0", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "0", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "6", outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "1", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testDelay: "60", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{success}, toolToTest: toolToTest})
-	testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "5", testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{skip, skipOddCpuNumber}, toolToTest: toolToTest})
+	// Using a timeout value such that ginkgo timeout > runtime + latency to ensure successfull runs
```
shajmakh, line 233: " runtime + latency" better to be " runtime + delay" , but no need to resubmit

shajmakh (approved):
Thank you for addressing this.
/lgtm

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
swatisehgal: Ah yeah, can fix it in conjunction with a future PR.

ffromani:

test fixes are beneficial for the CI signal quality and stability which in turn drives the quality of the product

yanirq:
/retitle NO-JIRA: Add positive tests with ginkgo timeout

PR #967, opened 2024-02-22
CNF-11213: added mixedCpus e2e test

rbaturov:
Added a test for mixedCpus feature along with some util functions.
This test checks that pods using shared CPUs in a deployment won't
be able to schedule after the mixedCpus feature is disabled and the node reboots.

### DIFF
--- a/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
+++ b/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
@@ -13,6 +13,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	kubeletconfig "k8s.io/kubelet/config/v1beta1"
 	"k8s.io/utils/cpuset"
@@ -28,6 +29,7 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/controller"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/runtime"
 	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/deployments"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
 	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/mcps"
@@ -49,6 +51,8 @@ const (
 	restartCooldownTime                            = 1 * time.Minute
 	isolatedCpusEnv                                = "OPENSHIFT_ISOLATED_CPUS"
 	sharedCpusEnv                                  = "OPENSHIFT_SHARED_CPUS"
+	// DeploymentName contains the name of the deployment
+	DeploymentName = "test-deployment"
 )
 
 var _ = Describe("Mixedcpus", Ordered, func() {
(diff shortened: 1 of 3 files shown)


test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -303,6 +305,72 @@ var _ = Describe("Mixedcpus", Ordered, func() {
 
 			})
 		})
+		When("Disabling mixedCpus in the performance profile", func() {
+			It("should results with deployment pods that used shared cpus to failing to be scheduled", func() {
```
Tal-or, line 309: `should causes all pods that uses shared CPUs to failed`

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -403,11 +471,7 @@ func kubeletRestartCmd() []string {
 }
 
 func createPod(ctx context.Context, c client.Client, ns string, opts ...func(pod *corev1.Pod)) (*corev1.Pod, error) {
-	p := pods.GetTestPod()
-	p.Namespace = ns
-	for _, opt := range opts {
-		opt(p)
-	}
+	p := createPodTemplate(ctx, c, ns, opts...)
```
Tal-or, line 474: Why do we need another function here?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -303,6 +305,72 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				Expect(err).ToNot(HaveOccurred(), "failed to execute command on pod; cmd=%q pod=%q", cmd, client.ObjectKeyFromObject(p).String())
+				isolatedAndShared := strings.Split(string(output), "\r\n")
+				shared := mustParse(isolatedAndShared[1])
+				ppShared := mustParse(string(*profile.Spec.CPU.Shared))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(shared.Equals(*ppShared)).To(BeTrue(), "OPENSHIFT_SHARED_CPUS value not equal to what configure in the performance profile."+
+					"OPENSHIFT_SHARED_CPUS=%s spec.cpu.shared=%s", shared.String(), ppShared.String())
+				testlog.Infof("shared CPU set=%q", shared.String())
+
+				By("Editing performanceProfile to have empty shared cpus and setting mixedCpus to false")
```
Tal-or, line 347: Keep consistency in the `By` format - sometime it starts with capital letter and sometimes it's not

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -303,6 +305,72 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				emptySet := mustParse("")
+				profile.Spec.CPU.Shared = cpuSetToPerformanceCPUSet(emptySet)
```
Tal-or, line 349: Just pass `&cpuset.CPUSet{}`

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -303,6 +305,72 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				profile.Spec.WorkloadHints.MixedCpus = pointer.Bool(false)
+
+				By("applying new performanceProfile")
+				testprofiles.UpdateWithRetry(profile)
```
Tal-or, line 353: Make sure to revert the changes at the end of the test.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
ffromani: or even better `should causes all pods that uses shared CPUs to fail`

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
rbaturov: The thing was that the previous createPod function not only constructed the pod template it also created it on the cluster by sending a request to the API server. I wanted to decouple these two functions because I need only the first part of this (creating the template) for the deployment.
I could create this function separately and not change anything in the original `createPod` function but this would result in some code duplication.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
Tal-or: 1. This is happening at the end of all tests, not in between each test.
2. If the setup didn't run, the cleanup won't be running either

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -303,6 +306,86 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				shared := mustParse(isolatedAndShared[1])
+				ppShared := mustParse(string(*profile.Spec.CPU.Shared))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(shared.Equals(*ppShared)).To(BeTrue(), "OPENSHIFT_SHARED_CPUS value not equal to what configure in the performance profile."+
+					"OPENSHIFT_SHARED_CPUS=%s spec.cpu.shared=%s", shared.String(), ppShared.String())
+				testlog.Infof("shared CPU set=%q", shared.String())
+
+				By("Editing performanceProfile to have empty shared cpus and setting mixedCpus to false")
+				emptySet := &cpuset.CPUSet{}
+				profile.Spec.CPU.Shared = cpuSetToPerformanceCPUSet(emptySet)
```
Tal-or, line 352: `profile.Spec.CPU.Shared = cpuSetToPerformanceCPUSet(&cpuset.CPUSet{})`

Tal-or (commented):
Almost there

test/e2e/performanceprofile/functests/utils/deployments/deployments.go
```diff
@@ -0,0 +1,63 @@
...
+	"time"
+
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewTestDeployment(name, namespace string, podTemplate *corev1.Pod, opts ...func(dp *appsv1.Deployment)) *appsv1.Deployment {
```
Tal-or, line 14: podTemplate can also be a option:
`WithPodTemplate`

Tal-or, line 14: lets follow the convention from before and call it `Make` so from outside it called:
`deployments.Make(With...)`

rbaturov: Done

test/e2e/performanceprofile/functests/utils/deployments/deployments.go
```diff
@@ -0,0 +1,91 @@
+package deployments
+
+import (
+	"context"
+	"time"
+
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/images"
```
Tal-or, line 7: Please rearrange imports in the following order:

standarads libs

k8s libs

third-part libs

To be more specific  Openshift imports should go under the k8s imports

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -303,6 +305,85 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+				rl := &corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("1"),
+					corev1.ResourceMemory: resource.MustParse("100Mi"),
+					sharedCpusResource:    resource.MustParse("1"),
+				}
+				p := makePod(ctx, testclient.Client, testutils.NamespaceTesting,
+					withRequests(rl),
+					withLimits(rl),
+					withRuntime(components.GetComponentName(profile.Name, components.ComponentNamePrefix)))
+				dp := deployments.Make(testutils.DeploymentTesting, testutils.NamespaceTesting,
```
Tal-or, line 321: Why 2 namespaces? the namespace of the deployment should be the namespace of the pod

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -303,6 +305,85 @@ var _ = Describe("Mixedcpus", Ordered, func() {
...
+
+				Expect(testclient.Client.Create(ctx, dp)).ToNot(HaveOccurred())
+				podList := &corev1.PodList{}
+				listOptions := &client.ListOptions{Namespace: testutils.NamespaceTesting, LabelSelector: labels.SelectorFromSet(dp.Spec.Selector.MatchLabels)}
+				Eventually(func() bool {
+					isReady, err := deployments.IsReady(ctx, testclient.Client, listOptions, podList, dp)
+					Expect(err).ToNot(HaveOccurred())
+					return isReady
+				}, time.Minute, time.Second).Should(BeTrue())
+				Expect(testclient.Client.List(ctx, podList, listOptions)).To(Succeed())
```
Tal-or, line 333: Is `To(Succeed)` is the same as `ToNot(HaveOccured)`?

test/e2e/performanceprofile/functests/utils/consts.go
```diff
@@ -101,6 +101,8 @@ const (
 	NamespaceMachineConfigOperator = "openshift-machine-config-operator"
 	// NamespaceTesting contains the name of the testing namespace
 	NamespaceTesting = "performance-addon-operators-testing"
+	// DeploymentTesting contains the name of the deployment namespace
+	DeploymentTesting = "test-deployment"
```
Tal-or, line 105: Can't we use the `NamespaceTesting`?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
rbaturov: These are not 2 namespaces. The first is the deployment name, second is the namespace.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
rbaturov:
```
func Succeed() types.GomegaMatcher
Succeed passes if actual is a nil error Succeed is intended to be used with functions that return a single error value. Instead of

err := SomethingThatMightFail()
Expect(err).ShouldNot(HaveOccurred())
You can write:

Expect(SomethingThatMightFail()).Should(Succeed())
```
According to the docs, in this case it should have the same effect.

test/e2e/performanceprofile/functests/utils/consts.go
rbaturov: Theoretically we can. I though we could use a name for the deployment that is more accurate rather than reusing the namespace of the deployment to its name.

Tal-or: We have a logic to delete the everything under the `NamespaceTesting` at the end of the test so it save us the burden of making sure we delete the pods/deployments. 
Having another namespace that we should take care of and do the same, is a bit tedious in my opinion.

rbaturov: but it is not another namespace. so that wouldn't be a problem right?

Tal-or: The comment on line 104 says it is

rbaturov: Correct it.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
rbaturov: Done.

test/e2e/performanceprofile/functests/utils/consts.go
```diff
@@ -101,6 +101,8 @@ const (
 	NamespaceMachineConfigOperator = "openshift-machine-config-operator"
 	// NamespaceTesting contains the name of the testing namespace
 	NamespaceTesting = "performance-addon-operators-testing"
+	// DeploymentName contains the name of the deployment
+	DeploymentName = "test-deployment"
```
Tal-or, line 105: Let's just move this constant to test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go and use it there and we're good to go

I should noticed the location of the constant earlier - my bad

rbaturov: Done

Tal-or:
/lgtm

Great work here, thank you @rbaturov

PR #969, opened 2024-02-23
NO-JIRA: Remove RELEASE_VERSION from test-upstream-tuned.sh

jmencak:
Starting with PR858, we can no longer use a random `RELEASE_VERSION` to trigger daemonset/operand updates.  In the context of testing upstream TuneD this is causes an issue.  Not changing `RELEASE_VERSION` at all is a simple fix, because we are also adjusting `CLUSTER_NODE_TUNED_IMAGE` and that will cause a daemonset/operand update.

### DIFF
--- a/hack/test-upstream-tuned.sh
+++ b/hack/test-upstream-tuned.sh
@@ -31,7 +31,7 @@ cvo_scale() {
 
 nto_deploy_custom() {
   oc project openshift-cluster-node-tuning-operator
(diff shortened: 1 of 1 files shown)


jmencak:
Tested with the latest TuneD 2.22.1 which will be the next FDP release.

jmencak:
There's no way this can break `ci/prow/e2e-gcp-pao`
/retest

ffromani:
/lgtm
/hold

holding to let other reviewers chime in, feel free to remove anytime

PR #970, opened 2024-02-26
Jira PSAP-1288: NTO operand cleanup

jmencak:
In preparation for running `TuneD` prior to kubelet, make things simpler and cleaner.

Notable changes:
  * Even though there is currently no namespace collision with `TuneD` using in /var/lib/tuned, change this path to /var/lib/ocp-tuned
  * Do not inherit `--enable-leader-election` and `--version` NTO flags as they are not handled by subcommands anyway (yet)
  * Clean up after PR844


jmencak:
> Until when do we need to maintain [Dockerfile.rhel8](https://github.com/openshift/cluster-node-tuning-operator/pull/970/files#diff-9da637863f1a7b402a22724a788049d8571eed68b5c023ad2a1314008791caed) ?

As long as 4.12 is supported, because it is the last one, where NTO uses RHEL8 as the base.  I'd say we can safely drop it after January 17, 2025.

MarSik (approved):
This looks good to me. I will let Yanir issue the last lgtm as he had a question.

yanirq:
> This looks good to me. I will let Yanir issue the last lgtm as he had a question.

lgtm as well (although the bot already has interpreted the comment as lgtm)

MarSik:
Heh, I thought you have to use the command form of lgtm. The bot is too smart it seems :)

PR #972, opened 2024-02-29
OCPBUGS-29641: fix rendering extra ctrcfgs

vitus133:
Fix bootstrap error  of "more than one ContainerRuntimeConfig found that matches MCP labels" when an additional ContainerRuntimeConfig is provided by user.

/cc @yanirq @MarSik @Tal-or @ffromani


MarSik (approved):
/lgtm

MarSik:
The only failing test is "fail [github.com/openshift/origin/test/extended/operators/images.go:53]: unable to decode release payload with error: invalid character 'w' looking for beginning of value" and that is for sure not related to this change.

yanirq:
@vitus133  does this require a backport ? (if so then to what version?)

MarSik:
Yes it does. I would say we have the bug in 4.14 for sure https://github.com/openshift/cluster-node-tuning-operator/blob/release-4.14/pkg/performanceprofile/cmd/render/render.go#L337

MarSik:
As well as in 4.13., 4.12 does not have it.

PR #976, opened 2024-03-03
OCPBUGS-26400: scheduler plugin: ignore IRQs

yanirq:
Set tuned scheduler plugin not to change any IRQ affinities. This overrides the default(true) which is set via
the parent cpu partitioning profile.

Utilized https://github.com/redhat-performance/tuned/pull/595


MarSik (approved):
/lgtm

PR #980, opened 2024-03-04
NO-JIRA: Add sync for rendering extra ctrcfgs

yanirq:
Add missing sync for rendering extra ctrcfgs under the render-sync target

This was missing from https://github.com/openshift/cluster-node-tuning-operator/pull/972 

(no backport needed)


yanirq:
no risk, local tool change only.

PR #983, opened 2024-03-06
 [release-4.15] OCPBUGS-30306: scheduler plugin: ignore IRQs

yanirq:
Manual Cherry pick of https://github.com/openshift/cluster-node-tuning-operator/pull/976 + https://github.com/openshift/cluster-node-tuning-operator/pull/987

Set tuned scheduler plugin not to change any IRQ affinities. This overrides the default(true) which is set via
the parent cpu partitioning profile.

Utilized https://github.com/redhat-performance/tuned/pull/595


MarSik (approved):
/lgtm

PR #991, opened 2024-03-13
NO-JIRA: E2E: Add  systemd package to fetch properties of cgroup slice 

mrniranjan:
* systemd package to fetch container's cgroup properties like AllowedCpus (cpuset.cpus), AllowedMemoryNodes (cpuset.mems) etc. T his will help us to fetch information irrespective of runtimes where the directories differ

* Add PidParser  Parser returns the container's or any Pid cgroup information which we can then traverse

* add util function to fetch Pid of a container process

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/cgroup/cgroup.go
+++ b/test/e2e/performanceprofile/functests/utils/cgroup/cgroup.go
@@ -1,8 +1,11 @@
 package cgroup
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"strings"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


shajmakh:
Thanks for this. 
I'd just comment to add a description for the first 2 commits and an overall description for the PR. 
/lgtm
/hold
for more reviews, unhold as you see better.

test/e2e/performanceprofile/functests/utils/cgroup/cgroup.go
```diff
@@ -51,3 +54,26 @@ func BuildGetter(ctx context.Context, c client.Client, k8sClient *kubernetes.Cli
 		return cgroupv1.NewManager(c, k8sClient), nil
 	}
 }
+
+// PidCgroupParser parsing /proc/pid/cgroup
+func PidCgroupParser(lines []byte) (string, error) {
```
Tal-or, line 59: My only suggestion would be to change the name to `PidParser` and outside the package it will be called as followed:
`cgroup.PidParser`

mrniranjan: fixed in the latest commit

yanirq:
@mrniranjan  Can you please squash all the commits into one ?

PR #992, opened 2024-03-14
OCPBUGS-30569: Systemd processes not being moved to cpuset/systemd.slice fix

rbaturov:
The script cpuset-configure.sh is responsible to move the
systemd processes to the cpuset/systemd.slice cgroup and  is executed
in a form of a service (cpuset-configure.service).

In the current implementation, the script is executed too early - some
system processes are yet to be created.
This in turn leads to them not being moved to the custom system slice.

Moreover, in the current implementation, the script is executed before
the network-online.target. The intention was to execute the script before
kubelet and crio services are initialized (by the fact network-online.target
is a common parent) in order to make sure that no workload pods are starting
before we are making this transition.

The fix I'm proposing consist of the following changes:
1. Adding an After statement - The script will start once crio service is initialized,
due to the fact it's initialized in the very end of the boot process,
just a bit before kubelet.
Thereby we can ensure late starting processes do not fall between the cracks.
2. Narrowing down the Before statement to a more accurate
one, reflecting its original intention. (Running the script before kubelet
only would be enough guarantee no workload pods are started at that time).

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -80,6 +80,7 @@ const (
 	systemdSectionInstall  = "Install"
 	systemdDescription     = "Description"
 	systemdBefore          = "Before"
+	systemdAfter           = "After"
 	systemdEnvironment     = "Environment"
 	systemdType            = "Type"
 	systemdRemainAfterExit = "RemainAfterExit"
@@ -88,13 +89,12 @@ const (
 )
 
 const (
-	systemdServiceIRQBalance   = "irqbalance.service"
-	systemdServiceKubelet      = "kubelet.service"
-	systemdServiceCrio         = "crio.service"
-	systemdServiceTypeOneshot  = "oneshot"
-	systemdTargetMultiUser     = "multi-user.target"
-	systemdTargetNetworkOnline = "network-online.target"
-	systemdTrue                = "true"
+	systemdServiceIRQBalance  = "irqbalance.service"
+	systemdServiceKubelet     = "kubelet.service"
+	systemdServiceCrio        = "crio.service"
+	systemdServiceTypeOneshot = "oneshot"
+	systemdTargetMultiUser    = "multi-user.target"
+	systemdTrue               = "true"
 )
 
 const (
@@ -451,7 +451,9 @@ func getCpusetConfigureServiceOptions() []*unit.UnitOption {
 		// Description
 		unit.NewUnitOption(systemdSectionUnit, systemdDescription, "Move services to reserved cpuset"),
 		// Before
-		unit.NewUnitOption(systemdSectionUnit, systemdBefore, systemdTargetNetworkOnline),
+		unit.NewUnitOption(systemdSectionUnit, systemdBefore, systemdServiceKubelet),
+		// After
+		unit.NewUnitOption(systemdSectionUnit, systemdAfter, systemdServiceCrio),
 		// Type
 		unit.NewUnitOption(systemdSectionService, systemdType, systemdServiceTypeOneshot),
 		// ExecStart
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -32,6 +32,7 @@ import (
 	componentprofile "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
 	profileutil "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
 	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup"
 	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cluster"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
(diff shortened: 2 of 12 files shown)


rbaturov:
/hold 
This is not final, need to rerun render-sync and push updated manifests.

rbaturov:
System services boot order without the fix
Screenshot from 2024-03-14 14-17-34
System services boot order with the fix
Screenshot from 2024-03-14 16-43-33

haircommander:
general approach LGTM

one question but it doesn't necessarily need to be addressed

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -451,7 +456,14 @@ func getCpusetConfigureServiceOptions() []*unit.UnitOption {
...
 		// Before
-		unit.NewUnitOption(systemdSectionUnit, systemdBefore, systemdTargetNetworkOnline),
+		unit.NewUnitOption(systemdSectionUnit, systemdBefore, systemdServiceKubelet),
+		unit.NewUnitOption(systemdSectionUnit, systemdBefore, systemdServiceCrio),
+		// After
+		unit.NewUnitOption(systemdSectionUnit, systemdAfter, systemdServiceNetworkManager),
+		unit.NewUnitOption(systemdSectionUnit, systemdAfter, systemdServiceSSHD),
+		unit.NewUnitOption(systemdSectionUnit, systemdAfter, systemdServiceDnsmasq),
+		unit.NewUnitOption(systemdSectionUnit, systemdAfter, systemdServiceDbusBroker),
+		unit.NewUnitOption(systemdSectionUnit, systemdAfter, systemdServiceGetty),
```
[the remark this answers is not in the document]
rbaturov: You can view this list in the Jira ticket - commented by Brent.
You can reproduce it by running - `cat /sys/fs/cgroup/cpuset/cgroup.procs | sort -n | xargs ps -o cmd  | grep -v "\["` in the node.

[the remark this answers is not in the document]
rbaturov: > Is there anything we can do other than hope new units aren't added that find their way through?


This is why I opened a [discussion](https://redhat-internal.slack.com/archives/CQNBUEVM2/p1710765617823529) in the channel. I ca'nt think of a better one.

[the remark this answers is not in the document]
ffromani: This is a concern of mine as well. Probably not blocking, but still.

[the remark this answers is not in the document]
haircommander: do we have an e2e suite that runs in this setup? we could add a test that checks this list and it'll fail if a new one comes up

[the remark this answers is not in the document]
haircommander: alternatively there could be a test added to QE's suite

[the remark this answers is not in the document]
rbaturov: @haircommander Added a test here.

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -282,6 +283,39 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+				if isV2 {
+					rootCgroupPath = "/rootfs/sys/fs/cgroup/cgroup.procs"
+				}
+				// Getting the list of PIDS that are running on the root cgroup.
+				output, err := nodes.ExecCommandOnNode(context.TODO(), []string{"cat", rootCgroupPath}, &node)
+				Expect(err).ToNot(HaveOccurred())
+				pids := strings.Split(output, "\n")
+				// Retrieving the processes represented by these PIDS.
+				output, err = nodes.ExecCommandOnNode(context.TODO(), append([]string{"ps", "-o", "cmd"}, pids...), &node)
+				Expect(err).ToNot(HaveOccurred())
```
ffromani, line 304: can we run a pipe to get the final result in one go?

rbaturov: Unfortunately no. I learned it the hard way.

ffromani: and what's the problem? why doesn't it work?

ffromani: > That's because the underlying ExecPCommandOnPod function expect the [] string command to be array of argv. It takes commad[0] as the name of the cmd, and the latter slice as params. When I tried to change the line to `output, err := nodes.ExecCommandOnNode(context.TODO(), []string{"cat", rootCgroupPath, "|", "xargs", "ps", "-o", "cmd"}, &node)` and run the test this is what I get:
> 
> ```
>   [FAILED] Unexpected error:
>       <*fmt.wrapError | 0xc000242420>: 
>       failed to run command [cat /rootfs/sys/fs/cgroup/cgroup.procs | xargs ps -o cmd]: output "cat: invalid option -- 'o'\r\nTry 'cat --help' for more information.\r\n"; error ""; command terminated with exit code 1
>       {
>           msg: "failed to run command [cat /rootfs/sys/fs/cgroup/cgroup.procs | xargs ps -o cmd]: output \"cat: invalid option -- 'o'\\r\\nTry 'cat --help' for more information.\\r\\n\"; error \"\"; command terminated with exit code 1",
>           err: <exec.CodeExitError>{
>               Err: <*errors.errorString | 0xc0003085c0>{
>                   s: "command terminated with exit code 1",
>               },
>               Code: 1,
>           },
>       }
> ```

And what about something like
```
/bin/bash -c "/bin/cat $rootCgroupPath | /bin/xargs /bin/ps -o cmd"
```

rbaturov: That sounds like a nice workaround indeed!

rbaturov: @ffromani Updated the code according to your suggestion.

ffromani: cool, if you verified it works let's keep it

rbaturov: @ffromani Yes, I have verified it (both cgroups v1 and v2).
BTW now I noticed, if we can use pipe with your workaround, maybe I also pipe a cmd to filter the kernel threads (`| grep -v "\["`) in the node instead of filtering the output.
WDYT? I will also add a more descriptive comment to explain the complicated command.

ffromani: note this is not a workaround but a fully legit way to run commands
that said yes, you can move the filtering in the pipeline, and please comment it.
The `ps` output includes both userspace processes and kernel threads, but kernel threads are always presented in `[square brackets]`.

rbaturov: Done.

rbaturov:
/hold
Please don't retest.
We have legit errors with this solution. 

```
   [FAILED] The node ci-op-xb6r1606-13bd6-r6bt4-worker-a-fdtrd has the following processes on the root cgroup: [/usr/bin/rpcbind -w -f /usr/sbin/rpc.statd /usr/lib/polkit-1/polkitd --no-debug]
  Expected
      <[]string | len:3, cap:3>: [
          "/usr/bin/rpcbind -w -f",
          "/usr/sbin/rpc.statd",
          "/usr/lib/polkit-1/polkitd --no-debug",
      ]
  to be empty
  In [It] at: /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/performance.go:304 @ 03/24/24 20:34:35.789
```

ffromani:
/approve

the test seems ok - as simple as can be, and the dependency logic seems to be the sweet spot

PR #998, opened 2024-03-18
OCPBUGS-30647: Remove TuneD timeout code and reload on ERRORs

jmencak:
In the past, there were reports of TuneD being stuck during profile application (e.g. rhbz#2013940).  As a workaround a timeout with exponential backoff was implemented to restart TuneD if it was ever stuck not finishing profile application.  However, with some TuneD profiles and hardware configurations, it can take very long for a TuneD profile to be applied.  In environments, such as latency-sensitive deployments a repeated profile application can make things matters worse. Therefore, remove the timeout functionality.

Also, remove the reload on TuneD ERRORs.  This is an openshift-tuned bug.

### DIFF
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -42,7 +42,6 @@ const (
 	scWarn
 	scError
 	scSysctlOverride
-	scTimeout
 	scUnknown
 )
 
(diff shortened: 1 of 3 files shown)


yanirq:
> In the past, there were reports of TuneD being stuck during profile application (e.g. rhbz#2013940). As a workaround a timeout with exponential backoff was implemented to restart TuneD if it was ever stuck not finishing profile application. 

Can this still be the case here ?  what is the alternative for the old workaround ?

jmencak:
> > In the past, there were reports of TuneD being stuck during profile application (e.g. rhbz#2013940). As a workaround a timeout with exponential backoff was implemented to restart TuneD if it was ever stuck not finishing profile application.
> 
> Can this still be the case here ? what is the alternative for the old workaround ?


We never root caused the issue.  This could have been something kernel-related.  The alternative would be finding the root cause and fix it.  If we are to block kubelet to TuneD completion in the future, we'll need a timeout mechanism of some sort, but likely systemd-based.

jmencak:
Cherry-pick to 4.15 will fail (https://patch-diff.githubusercontent.com/raw/openshift/cluster-node-tuning-operator/pull/970 is also needed), but do it anyway so that the automation creates the necessary Jira bugs.
/cherry-pick release-4.15

PR #1000, opened 2024-03-21
OCPBUGS-26162: Fixed .snyk format

rbaturov:
Previous format was wrong, and in turn had no effect on ignoring code issues.


ffromani (approved):
/lgtm

snyk docs are surprising at times

PR #1004, opened 2024-03-25
CNF-11817: e2e: control and data planes clients

Tal-or:
On Hypershift we should be able to access both the management and hosted clusters.
For that, we should now have two clients, each communicates with a different cluster.

ControlPlaneClient - defines the API client to run CRUD operations on the control plane that will be used for testing.

DataPlaneClient - defines the API client to run CRUD operations on the data plane that will be used for testing.

On non-hypershift both client would point the same the cluster under the hood.

We also added Ginkgo assertions that prevent the user from using the wrong clients for the wrong objects.
For example using the `DataPlaneClient` for querying the `PerformanceProfile` object is completley fine on non-hypershift cluster, but will fail badly on hypershift (since `PerformanceProfile` objects, are not presented on the hosted clusters.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
+++ b/pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
@@ -1,0 +1,115 @@
+package hypershift
+
+import (
+	"bytes"
+	"context"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	serializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	machineconfigv1 "github.com/openshift/api/machineconfiguration/v1"
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
+)
+
+// a set of keys which used to classify the encapsulated objects in the ConfigMap
+const (
+	TuningKey = "tuning"
+	ConfigKey = "config"
+)
+
+type ControlPlaneClientImpl struct {
+	// A client with access to the management cluster
+	client.Client
+
+	// managementClusterNamespaceName is the namespace name on the management cluster
+	// on which the control-plane objects reside
+	managementClusterNamespaceName string
+}
+
+func (ci *ControlPlaneClientImpl) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	if ObjIsEncapsulatedInConfigMap(obj) {
+		return ci.getFromConfigMap(ctx, key, obj, opts...)
+	}
+	return ci.Client.Get(ctx, key, obj, opts...)
+}
+
+func (ci *ControlPlaneClientImpl) Create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error {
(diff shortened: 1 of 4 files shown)


test/e2e/performanceprofile/functests/utils/client/clients.go
```diff
@@ -67,29 +75,62 @@ func init() {
...
 	if err != nil {
 		return nil, err
 	}
+	return client.New(cfg, client.Options{})
+}
 
-	c, err := client.New(cfg, client.Options{})
+// NewControlPlane returns a new controller-runtime client for ControlPlaneClient cluster.
+func NewControlPlane() (client.Client, error) {
+	c, err := New()
```
ffromani, line 109: nit: outside hypershift this is wasted, we should do like `NewDataPlane()` and create this client only if needed - or document why doing like this is fine

test/e2e/performanceprofile/functests/utils/client/clients.go
```diff
@@ -14,19 +14,27 @@ import (
...
 	configv1 "github.com/openshift/api/config/v1"
-	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
-	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
-	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 
 	performancev1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v1"
 	performancev1alpha1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v1alpha1"
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
+	config2 "github.com/openshift/cluster-node-tuning-operator/pkg/config"
```
ffromani, line 22: please rename `ntoconfig` because `config2` is quite obscure

ffromani (commented):
initial review, so far so good

test/e2e/performanceprofile/functests/utils/client/clients.go
Tal-or: Done

test/e2e/performanceprofile/functests/utils/client/clients.go
Tal-or: Right. Thanks!

MarSik (approved):
The initial unused implementation looks good to me.

yanirq:

This is required infrastructure preparation to be included in 4.16 . currently not wired to existing functionality

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -0,0 +1,115 @@
...
+	cm := &corev1.ConfigMap{}
+	if key.Namespace == metav1.NamespaceNone {
+		key.Namespace = ci.managementClusterNamespaceName
+	}
+	err := ci.Client.Get(ctx, key, cm, opts...)
+	if err != nil {
+		return err
+	}
+	var objAsYAML string
+	// can't have both
```
ffromani, line 76: but we don't check this, and also we don't check we have _either_.
Well, the decoder does implicitely I guess

ffromani (commented):
overall LGTM as well

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: There's a room for improvement indeed. I'll add it in a follow PR. Anyhow this code is not wired yet

ffromani: sure, no big deal

PR #1008, opened 2024-04-01
CNF-11815: e2e: Added node inspector for inspecting nodes configuration

rbaturov:
On hypershift there is no MCO, hence there are no machine-config-daemon pods.
A different resolution is needed for accessing the underlying node for inspecting configurations.
This commit introduces a node inspector implemented as a daemonset.
Upon execution of test suites, a pod with elevated privileges and host filesystem mounted will be deployed on every node.
Also I have added Z-deconfig suite ('Z' prefix, will guarantee that it will be the last suite run) that will be used for cleanup.
This API will be used for both hypershift and non-hypershift systems.

### DIFF
--- a/hack/run-test.sh
+++ b/hack/run-test.sh
@@ -92,6 +92,7 @@ main() {
     if [ "$ONLY_CLI_PRINT" = true ]; then
         print "Skipping execution as requested by user ... "
     else
+        trap '{ echo "Running cleanup test suite"; ginkgo test/e2e/performanceprofile/functests/Z_deconfig --flake-attempts=2 --junit-report=report.xml; }' EXIT SIGINT SIGTERM SIGSTOP
         print "Executing test suites ... "
         GOFLAGS=-mod=vendor ginkgo ${GINKGO_FLAGS}
     fi
--- a/test/e2e/performanceprofile/functests/0_config/test_suite_performance_config_test.go
+++ b/test/e2e/performanceprofile/functests/0_config/test_suite_performance_config_test.go
@@ -4,6 +4,7 @@
 package __performance_config_test
 
 import (
+	"context"
 	"flag"
 	"log"
 	"os"
@@ -14,6 +15,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/ginkgo/v2/reporters"
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -22,7 +24,10 @@ import (
 
 	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
 	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/images"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/k8sreporter"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/namespaces"
+	nodeInspector "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/node_inspector"
 )
 
 var (
@@ -59,6 +64,13 @@ func TestPerformanceConfig(t *testing.T) {
 
 var _ = BeforeSuite(func() {
 	Expect(testclient.ClientsEnabled).To(BeTrue())
+	// create test namespace
+	if err := testclient.DataPlaneClient.Create(context.TODO(), namespaces.NodeInspectorNamespace); !errors.IsAlreadyExists(err) {
+		Expect(err).ToNot(HaveOccurred())
+	}
+
+	err := nodeInspector.Create(testclient.DataPlaneClient, testutils.NodeInspectorNamespace, testutils.NodeInspectorName, images.Test())
+	Expect(err).ToNot(HaveOccurred())
 
 })
 
--- a/test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go
+++ b/test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go
@@ -1,0 +1,24 @@
+package Z_deconfig
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/namespaces"
+	nodeInspector "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/node_inspector"
+)
+
+// This test suite is designed to perform cleanup actions that should occur after all test suites have been executed.
+
+var _ = Describe("Deconfig", func() {
+	It("should delete the node inspector and its namespace", func() {
+		err := nodeInspector.Delete(testclient.DataPlaneClient, testutils.NodeInspectorNamespace, testutils.NodeInspectorName)
+		Expect(err).ToNot(HaveOccurred())
+		err = namespaces.WaitForDeletion(testutils.NodeInspectorNamespace, 5*time.Minute)
+		Expect(err).ToNot(HaveOccurred())
+	})
+})
--- a/test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
+++ b/test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
@@ -2,6 +2,7 @@ package node_inspector
 
 import (
 	"context"
+	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -9,10 +10,17 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/daemonset"
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	testpods "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
 )
 
 const serviceAccountSuffix = "sa"
@@ -36,8 +44,11 @@ func Create(cli client.Client, namespace, name, image string) error {
 		return err
 	}
 	ds := createDaemonSet(name, namespace, serviceAccountName, image)
-	if err := cli.Create(context.Background(), ds); err != nil && !errors.IsAlreadyExists(err) {
-		return err
+	if err := cli.Create(context.Background(), ds); err != nil {
+		if !errors.IsAlreadyExists(err) {
+			return err
+		}
+		klog.Infof("The node inspector daemonset was not expected to be running")
 	}
 	if err := daemonset.WaitToBeRunning(cli, namespace, name); err != nil {
 		return err
@@ -58,8 +69,42 @@ func Delete(cli client.Client, namespace, name string) error {
 	return nil
 }
 
-func IsRunning(cli client.Client, namespace, name string) (bool,error){
+func isRunning(cli client.Client, namespace, name string) (bool, error) {
 	return daemonset.IsRunning(cli, namespace, name)
+}
+
+// getDaemonPodByNode returns the daemon pod that runs on the specified node
+func getDaemonPodByNode(node *corev1.Node) (*corev1.Pod, error) {
+	listOptions := &client.ListOptions{
+		Namespace:     testutils.NodeInspectorNamespace,
+		FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": node.Name}),
+		LabelSelector: labels.SelectorFromSet(labels.Set{"name": testutils.NodeInspectorName}),
+	}
+
+	pods := &corev1.PodList{}
+	if err := testclient.DataPlaneClient.List(context.TODO(), pods, listOptions); err != nil {
+		return nil, err
+	}
+	if len(pods.Items) < 1 {
+		return nil, fmt.Errorf("failed to get daemon pod for the node %q", node.Name)
+	}
+	return &pods.Items[0], nil
+}
+
+// ExecCommand executing the command on a daemon pod of the given node
+func ExecCommand(ctx context.Context, node *corev1.Node, command []string) ([]byte, error) {
+	// Ensure the node inspector is running
+	ok, err := isRunning(testclient.DataPlaneClient, testutils.NodeInspectorNamespace, testutils.NodeInspectorName)
+	if err != nil || !ok {
+		return nil, err
+	}
+	pod, err := getDaemonPodByNode(node)
+	if err != nil {
+		return nil, err
+	}
+	testlog.Infof("found daemon pod %s for node %s", pod.Name, node.Name)
+
+	return testpods.WaitForPodOutput(ctx, testclient.K8sClient, pod, command)
 }
 
 func createDaemonSet(name, namespace, serviceAccountName, image string) *appsv1.DaemonSet {
--- a/test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
+++ b/test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
@@ -1,0 +1,186 @@
+package node_inspector
+
+import (
+	"context"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/daemonset"
+)
+
+const serviceAccountSuffix = "sa"
+const clusterRoleSuffix = "cr"
+const clusterRoleBindingSuffix = "crb"
+
+func Create(cli client.Client, namespace, name, image string) error {
+	serviceAccountName := fmt.Sprintf("%s-%s", name, serviceAccountSuffix)
+	sa := createServiceAccount(serviceAccountName, namespace)
+	if err := cli.Create(context.Background(), sa); err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+	clusterRoleName := fmt.Sprintf("%s-%s", name, clusterRoleSuffix)
+	cr := createClusterRole(clusterRoleName)
+	if err := cli.Create(context.Background(), cr); err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+	clusterRoleBindingName := fmt.Sprintf("%s-%s", name, clusterRoleBindingSuffix)
+	rb := createClusterRoleBinding(clusterRoleBindingName, namespace, serviceAccountName, clusterRoleName)
+	if err := cli.Create(context.Background(), rb); err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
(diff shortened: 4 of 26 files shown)


rbaturov:
In this PR, I have created a distinct namespace for the testing daemonset that will be created in the 0_config test suites.
I'm not sure how we can clean up them in the end.

test/e2e/performanceprofile/functests/utils/daemonset/daemonset.go
```diff
@@ -46,3 +55,164 @@ func IsRunning(cli client.Client, namespace, name string) (bool, error) {
 	testlog.Infof("daemonset %q %q desired %d scheduled %d ready %d", namespace, name, ds.Status.DesiredNumberScheduled, ds.Status.CurrentNumberScheduled, ds.Status.NumberReady)
 	return (ds.Status.DesiredNumberScheduled > 0 && ds.Status.DesiredNumberScheduled == ds.Status.NumberReady), nil
 }
+
+func WaitForDeletion(cli client.Client, namespace, name string, timeout time.Duration) error {
+	return wait.PollUntilContextTimeout(context.TODO(), time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+		_, err := GetByName(cli, namespace, name)
+		if err != nil {
```
ffromani, line 62: you likely want to check this is error is exactly `IsNotFound` (the apierror variant) before to declare success

test/e2e/performanceprofile/functests/utils/daemonset/daemonset.go
```diff
@@ -46,3 +55,164 @@ func IsRunning(cli client.Client, namespace, name string) (bool, error) {
...
+		return err
+	}
+	if err = WaitForDeletion(cli, namespace, name, 5*time.Minute); err != nil {
+		return err
+	}
+	return nil
+}
+
+func CreateForTesting(cli client.Client, namespace, name, image string) error {
+	serviceAccountName := name + "-sa"
```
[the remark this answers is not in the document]
rbaturov: Creating a func like this is the serviceaccount package is that your intention?
```
func ConstructName(name string) string {
	return name + "-sa"
}
```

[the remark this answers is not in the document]
ffromani: Yes, pretty much. We have two options IMO: either we expose the well-known suffixes as package constants and we let the consumers trivially construct the name with string manipulation, or we expose functions to do so. Either work for me (the function could be simply called `Name()`)

rbaturov:
/hold 
Need to test on hypershift

test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go
```diff
@@ -0,0 +1,27 @@
+package Z_deconfig
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
```
Tal-or, line 7: Please reorder imports:

`std libs`

`third party`

`openshift libs`

test/e2e/performanceprofile/functests/0_config/test_suite_performance_config_test.go
```diff
@@ -59,6 +64,13 @@ func TestPerformanceConfig(t *testing.T) {
 
 var _ = BeforeSuite(func() {
 	Expect(testclient.ClientsEnabled).To(BeTrue())
+	// create test namespace
+	if err := testclient.Client.Create(context.TODO(), namespaces.TestingDaemonsetNamespace); !errors.IsAlreadyExists(err) {
+		Expect(err).ToNot(HaveOccurred())
+	}
+
+	err := daemonset.CreateForTesting(testclient.DataPlaneClient, testutils.TestingDaemonsetNamespace, testutils.TestingDaemonsetName, images.Test())
```
Tal-or, line 72: let's try be more explicit about what `daemonset.CreateForTesting` is doing.
what about `CreateNodeInspector`? and adding a short comment that explains what it's doing

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -113,39 +114,43 @@ func GetNonPerformancesWorkers(nodeSelectorLabels map[string]string) ([]corev1.N
...
-		return nil, fmt.Errorf("failed to get machine-config-daemon pod for the node %q", node.Name)
+	if len(pods.Items) < 1 {
+		return nil, fmt.Errorf("failed to get testing daemon pod for the node %q", node.Name)
 	}
-	return &mcds.Items[0], nil
+	return &pods.Items[0], nil
 }
 
-// ExecCommandOnMachineConfigDaemon returns the output of the command execution on the machine-config-daemon pod that runs on the specified node
-func ExecCommandOnMachineConfigDaemon(ctx context.Context, node *corev1.Node, command []string) ([]byte, error) {
```
Tal-or, line 136: Both `ExecCommandOnTestingDaemonPod` and `ExecCommandOnNode` have the same purpose.
Unless there is a very good reason to have both functions, I would merge them into one named:
`ExecCommandOnNode` which declare explicitly where the command is about to be running.

rbaturov: @Tal-or 
`ExecCommandOnTestingDaemonPod` is being used as an underlying function for `ExecCommandOnNode`.
The difference between the two is that `ExecCommandOnNode` returns a string ([after some formatting](https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/utils/nodes/nodes.go#L174)) rather than a bytes slice (`ExecCommandOnTestingDaemonPod`).
As throughout the repo,  developers have chosen to use both of them, I think we should keep both of these functions and just rename them appropriately. (Another option is to unify them to one function that returns []bytes and change all calls to target this func, following a call if needed to a util function that will do the string format that is left). But again the latter option would require some work for refactoring. 
WDYT?

rbaturov: @Tal-or  I'll need to rename `ExecCommandOnNode` to `ExecCommandOnNodeToString` and `ExecCommandOnTestingDaemonPod` which returns a []byte to `ExecCommandOnNode`.
This however, may cause a slight confusion for developers that got used that  `ExecCommandOnNode` returns a string. but I think this is not a big deal. 
Another option could be to change `ExecCommandOnTestingDaemonPod` to `ExecCommandOnNodeToByteSlice`. I personally think we should go with the first option, even if it might cause a bit of confusion.

Tal-or: since the name of the package is `nodes` I would rename `ExecCommandOnTestingDaemonPod` to `ExecCommand` and call it a day.
From the caller side it looks like:
`nodes.ExecCommand()` The package name provides a clear indication about where this command is about to be running.

rbaturov: But what will be the name of the second func?
Anyways I was thinking that we should move these functions to node inspector package, as they rely on it running.
So it's going to be more of `nodeInspector.ExecCommand...`

Tal-or: The node inspector is an implementation detail. you can call `nodeInspector.ExecCommand` from `nodes.ExecCommand`

rbaturov:
@Tal-or ready for another review iteration

Makefile
```diff
@@ -221,7 +221,7 @@ pao-functests: cluster-label-worker-cnf pao-functests-only
 pao-functests-only:
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/1_performance test/e2e/performanceprofile/functests/6_mustgather_testing test/e2e/performanceprofile/functests/10_performance_ppc" -p "-v -r --fail-fast  --flake-attempts=2 --junit-report=report.xml" -m "Running Functional Tests"
+	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config test/e2e/performanceprofile/functests/1_performance test/e2e/performanceprofile/functests/6_mustgather_testing test/e2e/performanceprofile/functests/10_performance_ppc test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go" -p "-v -r --fail-fast  --flake-attempts=2 --junit-report=report.xml" -m "Running Functional Tests"
```
yanirq, line 224: why is this addition of folder (very weird name) + direct mention of file ?  is this for local tests ?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -140,39 +140,14 @@ func GetNonPerformancesWorkers(nodeSelectorLabels map[string]string) ([]corev1.N
...
+func ExecCommand(ctx context.Context, node *corev1.Node, command []string) ([]byte, error) {
+	return nodeInspector.ExecCommand(ctx, node, command)
 }
 
-// ExecCommandOnNode executes given command on given node and returns the result
-func ExecCommandOnNode(ctx context.Context, cmd []string, node *corev1.Node) (string, error) {
-	out, err := ExecCommandOnMachineConfigDaemon(ctx, node, cmd)
+// ExecCommandToString returns the output of the command execution in a string format
+func ExecCommandToString(ctx context.Context, cmd []string, node *corev1.Node) (string, error) {
+	out, err := ExecCommand(ctx, node, cmd)
```
ffromani, line 150: Shallow functions usually add very little value, and basically the main reason to have them is backward compatibility. What's the benefit of this function? Is it really helpful?

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/daemonset"
+)
+
+const serviceAccountSuffix = "sa"
+const clusterRoleSuffix = "cr"
+const clusterRoleBindingSuffix = "crb"
+
+func Create(cli client.Client, namespace, name, image string) error {
+	serviceAccountName := name + serviceAccountSuffix
```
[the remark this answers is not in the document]
Tal-or: +1

test/e2e/performanceprofile/functests/utils/daemonset/daemonset.go
```diff
@@ -46,3 +45,27 @@ func IsRunning(cli client.Client, namespace, name string) (bool, error) {
...
+		_, err := GetByName(cli, namespace, name)
+		if err != nil && k8serrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, nil
+	})
+}
+
+func Delete(cli client.Client, namespace, name string) error {
+	daemonset, err := GetByName(cli, namespace, name)
```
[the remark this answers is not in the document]
Tal-or: I would inline this function and call `WaitForDeletion` from the caller side

test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go
```diff
@@ -0,0 +1,27 @@
...
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/namespaces"
+	nodeInspector "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/node_inspector"
+)
+
+// This test suite is designed to perform cleanup actions that should occur after all test suites have been executed.
+
+var _ = Describe("Deconfig", func() {
+	It("Should delete the node inspector and its namespace", func() {
```
Tal-or, line 19: nit: small case, no capital

test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go
```diff
@@ -0,0 +1,27 @@
...
+		err := nodeInspector.Delete(testclient.Client, testutils.NodeInspectorNamespace, testutils.NodeInspectorName)
```
Tal-or, line 20: The deletion of the namespace two line below, would remove the pod inspector automatically

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -0,0 +1,199 @@
...
+	if err := cli.Create(context.Background(), cr); err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+	clusterRoleBindingName := name + clusterRoleBindingSuffix
+	rb := createClusterRoleBinding(clusterRoleBindingName, namespace, serviceAccountName, clusterRoleName)
+	if err := cli.Create(context.Background(), rb); err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+	ds := createDaemonSet(name, namespace, serviceAccountName, image)
+	if err := cli.Create(context.Background(), ds); err != nil && !errors.IsAlreadyExists(err) {
```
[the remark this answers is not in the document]
Tal-or: I think the NS should be deleted in the cleanup isn't it @rbaturov ?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
Tal-or: @ffromani IIRC it's for backward compatibility, because some caller expects `string` and no `[]byte`

Tal-or: Maybe a cleaner solution would be adding and `Options` argument to and merge both function, but this is out of the scope of this PR IMO

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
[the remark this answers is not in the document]
rbaturov: Certainly, we should, have missed that.

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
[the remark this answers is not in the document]
rbaturov: We have special [Z_deconfig suite](https://github.com/rbaturov/cluster-node-tuning-operator/blob/add-daemonset-for-tests/test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go#L20), that should handle [deletion](https://github.com/rbaturov/cluster-node-tuning-operator/blob/add-daemonset-for-tests/test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go#L56) of the daemon set and the rest of the objects if found in the cluster.

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -0,0 +1,199 @@
...
+func Delete(cli client.Client, namespace, name string) error {
+	ds := &appsv1.DaemonSet{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace}}
+	if err := cli.Delete(context.Background(), ds); err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	sa := &corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: name + serviceAccountSuffix, Namespace: namespace}}
+	if err := cli.Delete(context.Background(), sa); err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	crb := &rbacv1.ClusterRoleBinding{ObjectMeta: metav1.ObjectMeta{Name: name + clusterRoleBindingSuffix, Namespace: namespace}}
```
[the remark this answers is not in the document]
rbaturov: I don't know how we could ensure this. Do you have a suggestion?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
rbaturov: What Talor said. also, I agree we should merge the functions, and do so in a separate PR.

test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go
rbaturov: This is what I did previously. I counted on the namespace to delete the node inspector and its related object.
I found out that the ClusterRole was not deleted and lived past the cleanup, since it is cluster scoped.
I could instead call a delete function solely for the ClusterRole, but I don't see any big difference here.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
ffromani: A middleground could be add a separate `ToString()` helper and use it wherever needed

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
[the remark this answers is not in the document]
ffromani: ok, so let's say `Create` fails mid-way. Would call a second time `Create` fix it (assuming there's no error from the client of course)? If not, let's ensure `Delete` can handle partial creation just fine.

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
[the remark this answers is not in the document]
ffromani: A way is to make sure to tolerate `IsNotFound` error, so we can call `Delete` multiple times, and each time the function will do as much as work as possible.

test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go
ffromani: Deleting more objects, which is what `node_inspector.Delete` does, is inherently more fragile though, because we need to handle more error paths robustly. However we need `node_inspector.Delete` anyway, so we may very much call it here.

rbaturov:
Did some modifications:
1. Calling Z_deconfig as part of a trap command in the run-test.sh script will ensure cleanup occurs, even if we run through failures mid-way.
2. Changed the node-inspector delete function, to delete namespace (which will implicitly delete service account, cluster-role-binding), thereby reducing API calls. Clusterrole requires an additional delete request as it is not namespaced.

Makefile
```diff
@@ -303,4 +303,4 @@ pao-build-e2e:
 
 .PHONY: pao-clean-e2e
 pao-clean-e2e:
-	@rm -f _output/e2e-pao*.test
+	@rm -f _output/e2e-pao*.test
```
Tal-or, line 306: Please add new line

test/e2e/performanceprofile/functests/Z_deconfig/deconfig.go
Tal-or: Small case, `should`

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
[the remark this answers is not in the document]
Tal-or: Is it expected to have the DaemonSet already exist?

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -0,0 +1,191 @@
...
+func createClusterRole(name string) *rbacv1.ClusterRole {
+	return &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Resources: []string{"nodes"},
+				Verbs:     []string{"get", "list", "watch"},
```
Tal-or, line 201: We discuss about it offline, are these needed?

hack/run-test.sh
```diff
@@ -92,6 +92,7 @@ main() {
     if [ "$ONLY_CLI_PRINT" = true ]; then
         print "Skipping execution as requested by user ... "
     else
+        trap '{ echo "Running cleanup test suite"; ginkgo test/e2e/performanceprofile/functests/Z_deconfig --flake-attempts=2 --junit-report=report.xml; }' EXIT SIGINT SIGTERM SIGSTOP
```
Tal-or, line 95: Please be more specific:
`Deleting node inspector pod`

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
Tal-or: That sounds like a good idea and since it does not requires new struct definitions (like `Options`) I would recommend to implement that as part of this PR

hack/run-test.sh
rbaturov: @Tal-or I chose to write this in a general way, as in the future Z_deconfig might be used to do more than cleaning node inspector.

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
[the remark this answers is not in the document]
rbaturov: @Tal-or  it is not expected. but if is already there (which should never happen) do you think we should throw an error?

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
rbaturov: Checked that now on a certain unit test, seems like it is redundant indeed.

hack/run-test.sh
Tal-or: When it does, we can always rephrase

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
Tal-or: Please also check the privileged permissions, since we're only reading, i'm no sure it's needed

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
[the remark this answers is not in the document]
Tal-or: At least print a warning

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
rbaturov: I don't think we are only reading from the filesystem, we are executing commands.

rbaturov: for example here:
https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go#L271

Tal-or:
/lgtm
Thank you for all your work on this

ffromani:
test-only code and this is an improvement anyway. Plus, we depend on this work for a critical feature

PR #1010, opened 2024-04-01
CNF-10873: set Cgroups V2 as default

yanirq:
On fresh installations cgroups v2 will be the default as Openshift.

Remove helper annotation that sets the cluster to V2 since its not needed anymore.

Reverts  https://github.com/openshift/cluster-node-tuning-operator/pull/737 , https://github.com/openshift/cluster-node-tuning-operator/pull/854 (Partially)


pkg/performanceprofile/controller/performanceprofile_controller_test.go
```diff
@@ -799,182 +787,6 @@ var _ = Describe("Controller", func() {
...
-
-			kc := &mcov1.KubeletConfig{}
-			err := r.Get(context.TODO(), key, kc)
-			Expect(err).ToNot(HaveOccurred())
-
-			kubeletConfigString := string(kc.Spec.KubeletConfig.Raw)
-			Expect(kubeletConfigString).To(ContainSubstring(`"kubeReserved":{"memory":"256Mi"}`))
-			Expect(kubeletConfigString).To(ContainSubstring(`"systemReserved":{"memory":"256Mi"}`))
-		})
-	})
```
ffromani, line 825: I think we should not delete this

pkg/performanceprofile/controller/performanceprofile_controller_test.go
```diff
@@ -799,182 +787,6 @@ var _ = Describe("Controller", func() {
...
-
-	Context("with profile with deletion timestamp", func() {
-		BeforeEach(func() {
-			profile.DeletionTimestamp = &metav1.Time{
-				Time: time.Now(),
-			}
-			profile.Finalizers = append(profile.Finalizers, finalizer)
-		})
-
-		It("should remove all components and remove the finalizer on first reconcile loop", func() {
```
ffromani, line 835: I think we should not delete this

MarSik (approved):
/lgtm

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -472,6 +463,16 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	if err := validateProfileMachineConfigPool(instance, profileMCP); err != nil {
+		conditions := r.getDegradedConditions(conditionBadMachineConfigLabels, err.Error())
+		if err := r.updateStatus(instance, conditions); err != nil {
+			klog.Errorf("failed to update performance profile %q status: %v", instance.Name, err)
+			return reconcile.Result{}, err
+		}
+
+		return reconcile.Result{}, nil
+	}
+
```
ffromani, line 475: from where this comes from?

ffromani (commented):
almost LGTM, question inside

pkg/performanceprofile/controller/performanceprofile_controller.go
yanirq: @ffromani   this is a good question ,   I saw it was removed (or moved) in [#854](https://github.com/openshift/cluster-node-tuning-operator/pull/854/files#diff-b26e8713b0dcfb154cd6a3430e574a73f0b601dc1a206b4fcd732e587d4fa38aL475)   to the[ v1 reconciler](https://github.com/openshift/cluster-node-tuning-operator/pull/854/files#diff-b26e8713b0dcfb154cd6a3430e574a73f0b601dc1a206b4fcd732e587d4fa38aR572) in that same PR. 
It looks like it was missing for the non v1 reconcile loop in #854 . Would that statement be correct ?

MarSik: Yep. Moved from line 599 in this patch.

ffromani: ok, LGTM then

PR #1013, opened 2024-04-03
OCPBUGS-29340: [manual] [release-4.15]e2e: when crun is enabled by default skip checking runc config

mrniranjan:
this patch fixes runc configuration test which
was failing when crun is configured prior to running the test

In this patch we move BeforeAll at the context level to check if any container runtime is pre configured

if crun is preconfigured: skip runc config test
if crun is not preconfigured: check runc configuration and in the subsequent test, we enable crun and test crun and delete crun configuration


ffromani (commented):
/approve
/lgtm

PR #1014, opened 2024-04-03
OCPBUGS-31690: [manual][release-4.15] E2E: Reduce number of reboots in offline tests

mrniranjan:
OCPBUGS-29594: Reduce number of reboots in offline tests (#940)

* Reduce number of reboots in offline tests

Currently in offline tests we rever to original profile after each tests, this is very costly and causing tests to run for long time.

This patch addresses by reverting to the profile after all offline tests are run.

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -551,8 +551,9 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 
 	})
 
-	Context("Offlined CPU API", func() {
-		BeforeEach(func() {
+	Context("Offlined CPU API", Ordered, func() {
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -607,39 +613,33 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 
 		It("[test_id:50964] Offline Higher CPUID's", func() {
 			var reserved, isolated, offline []string
-			// This map is of the form numaNode[core][cpu-siblings]
```
ffromani, line 610: this comment should have been preserved when moving the var up the flow. But we can't fix into a backport

ffromani (commented):
/approve
/lgtm

let's keep looking for a good balance between textbook correctness and manageable execution time

PR #1016, opened 2024-04-03
OCPBUGS-31692:  [release-4.15] [manual]: Systemd processes not being moved to cpuset/systemd.slice fix

rbaturov:
This is a manual backport for #992

* Systemd processes not being moved to cpuset/systemd.slice fix

The script cpuset-configure.sh is responsible to move the systemd processes to the cpuset/systemd.slice cgroup and  is executed in a form of a service (cpuset-configure.service).

In the current implementation, the script is executed too early - some system processes are yet to be created.
This in turn leads to them not being moved to the custom system slice.

Moreover, in the current implementation, the script is executed before the network-online.target. The intention was to execute the script before kubelet and crio services are initialized (by the fact network-online.target is a common parent) in order to make sure that no workload pods are starting before we are making this transition.

The fix I'm proposing consist of the following changes:
1. Adding an After statements - The script will start once crio service is initialized, due to the fact it's initialized in the very end of the boot process, just a bit before kubelet.
Thereby we can ensure late starting processes do not fall between the cracks.
2. Narrowing down the Before statement to a more accurate one, reflecting its original intention. (Running the script before kubelet only would be enough guarantee no workload pods are started at that time).

* Added a test to verify system processes are in the correct cgroup

When we are using cgroups v1 we are counting on the cpuset-configure.service to move all the system services to the custom system.slice. This test ensures the service indeed moved them.

It is also a good practice to check for similar errors on cgroup v2 systems.

---------

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -32,6 +32,7 @@ import (
 	componentprofile "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
 	profileutil "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
 	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup"
 	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cluster"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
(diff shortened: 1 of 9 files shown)


jmencak:
/approved
/label backport-risk-assessed
Any reason some of the e2e tests are excluded from the manual backport?
/hold

rbaturov:
> /approved
> 
> /label backport-risk-assessed
> 
> Any reason some of the e2e tests are excluded from the manual backport?
> 
> /hold
> 
> 

Which e2e tests are missing?

jmencak:
> > /approved
> > /label backport-risk-assessed
> > Any reason some of the e2e tests are excluded from the manual backport?
> > /hold
> 
> Which e2e tests are missing?


I've noticed the patch difference between 4.16/4.15 in these "testdata" files and wonder if this is intentional:
```
test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-ctrcfg/openshift-bootstrap-master_machineconfig.yaml
test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-ctrcfg/openshift-bootstrap-worker_machineconfig.yaml
test/e2e/performanceprofile/testdata/render-expected-output/default/cpuFrequency/manual_machineconfig.yaml
```

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -267,6 +268,29 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 		})
 	})
 
+	Context("Using performance profile", func() {
+		It("Should have system services running on the system.slice cgroup", func() {
```
mrniranjan, line 272: Can you add test case id 73107 to this test case.

rbaturov:
> > > /approved
> > > /label backport-risk-assessed
> > > Any reason some of the e2e tests are excluded from the manual backport?
> > > /hold
> > 
> > 
> > Which e2e tests are missing?
> 
> I've noticed the patch difference between 4.16/4.15 in these "testdata" files and wonder if this is intentional:
> 
> ```
> test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-ctrcfg/openshift-bootstrap-master_machineconfig.yaml
> test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-ctrcfg/openshift-bootstrap-worker_machineconfig.yaml
> test/e2e/performanceprofile/testdata/render-expected-output/default/cpuFrequency/manual_machineconfig.yaml
> ```

Yes this was intentional. These testdata files are not part of the 4.15 release.

rbaturov:
> /approve /lgtm
Can we remove the hold label?

PR #1017, opened 2024-04-03
DO NOT MERGE: WIP: test enabled mixcpus feature watches

yanirq:


Tal-or:
But the payload here doesn't contain the problematic API bump from MCO because it was already reverted.

PR #1024, opened 2024-04-08
OCPBUGS-30647: NTO operand (openshift-tuned) fixes

jmencak:
Changes:
  * fix logging in `updateTunedProfile()` and optimize the calls to update node annotations and update Profile.Status
  * clean up `tunedStop()` to return only one value
  * during `TuneD` process shutdown, handle the fact the `TuneD` process might have already exitted
  * the openshift-tuned operand now no longer unnecessarily exits when `TuneD` process exits; when `TuneD` process exits, wait for k8s object changes and only then restart `TuneD`
  * do not use buffered channels
  * the indication that `TuneD` is reloading is now a status bit potentially reportable back to the operator
  * introduce Change type for the `TuneD` event processor to avoid races, where it was previously possible to change `TuneD` configuration during `TuneD` profile reload
  * register the fact `TuneD` finished reloading in case the primary `TuneD` profile does not exist
  * conditional `TuneD` reload when Cloud Provider changes
  * minor logging and comment improvements

### DIFF
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -4,6 +4,7 @@ import (
 	"bufio"   // scanner
 	"bytes"   // bytes.Buffer
 	"context" // context.TODO()
+	"errors"  // errors.Is()
 	"fmt"     // Printf()
 	"math"    // math.Pow()
 	"os"      // os.Exit(), os.Stderr, ...
@@ -34,10 +35,9 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/version"
 )
 
-// Constants
+// Constants used for instantiating Profile status conditions;
+// they will be set to 2^0, 2^1, 2^2, ..., 2^n
 const (
-	// Constants used for instantiating Profile status conditions;
-	// they will be set to 2^0, 2^1, 2^2, ..., 2^n
 	scApplied Bits = 1 << iota
 	scWarn
 	scError
@@ -46,9 +46,10 @@ const (
 	scUnknown
 )
 
+// Constants used for controlling TuneD;
+// they will be set to 2^0, 2^1, 2^2, ..., 2^n
 const (
-	// Constants used for controlling TuneD;
-	// they will be set to 2^0, 2^1, 2^2, ..., 2^n
+	// Should we reload (vs. full restart) TuneD?
 	ctrlReload Bits = 1 << iota
 	// Did the command-line parameters to run the TuneD daemon or tuned-main.conf change?
 	// In other words, is a complete restart of the TuneD daemon needed?
@@ -551,7 +552,7 @@ func (c *Controller) tunedStop() error {
 		// The TuneD daemon rolls back the current profile and should terminate on SIGTERM.
 		klog.V(1).Infof("sending SIGTERM to PID %d", c.tunedCmd.Process.Pid)
 		if err := c.tunedCmd.Process.Signal(syscall.SIGTERM); err != nil {
-			if err == os.ErrProcessDone {
+			if errors.Is(err, os.ErrProcessDone) {
 				// The TuneD process has already finished.
 				return nil
 			}
@@ -569,7 +570,7 @@ func (c *Controller) tunedStop() error {
 		// within tunedGracefulExitWait.
 		klog.V(1).Infof("sending SIGKILL to PID %d", c.tunedCmd.Process.Pid)
 		if err := c.tunedCmd.Process.Signal(syscall.SIGKILL); err != nil {
-			if err == os.ErrProcessDone {
+			if errors.Is(err, os.ErrProcessDone) {
 				// The TuneD process has already finished.
 				return nil
 			}
@@ -607,7 +608,7 @@ func (c *Controller) tunedReload() error {
 	if c.tunedCmd.Process != nil {
 		klog.Infof("sending HUP to PID %d", c.tunedCmd.Process.Pid)
 		if err := c.tunedCmd.Process.Signal(syscall.SIGHUP); err != nil {
-			if err == os.ErrProcessDone {
+			if errors.Is(err, os.ErrProcessDone) {
 				// The TuneD process has already finished.  The following comes to mind:
 				//   * TuneD exitted due to a bug
 				//   * someone or something (systemd) killed TuneD intentionally
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -42,7 +42,18 @@ const (
 	scWarn
 	scError
 	scSysctlOverride
+	scReloading // reloading is true during the TuneD daemon reload.
 	scUnknown
+)
+
+const (
+	// Constants used for controlling TuneD;
+	// they will be set to 2^0, 2^1, 2^2, ..., 2^n
+	ctrlReload Bits = 1 << iota
+	// Did the command-line parameters to run the TuneD daemon or tuned-main.conf change?
+	// In other words, is a complete restart of the TuneD daemon needed?
+	ctrlRestart
+	ctrlDebug // Debugging can be turned on only from the command-line.
 )
 
 // Constants
@@ -74,24 +85,38 @@ const (
 )
 
 // Types
-type Bits uint8
+type Bits uint64
 
 type Daemon struct {
-	// reloading is true during the TuneD daemon reload.
-	reloading bool
-	// reloaded is true immediately after the TuneD daemon finished reloading.
-	// and the node Profile k8s object's Status needs to be set for the operator;
-	// it is set to false on successful Profile update.
-	reloaded bool
-	// debugging flag
-	debug bool
-	// bit/set representaton of Profile status conditions to report back via API.
+	// bit/set representation of TuneD options which are remotely controllable.
+	restart Bits
+	// bit/set representation of Profile status conditions to report back via API.
 	status Bits
 	// stderr log from TuneD daemon to report back via API.
 	stderr string
 	// stopping is true while the controller tries to stop the TuneD daemon.
 	stopping bool
-	// the TuneD profile we wish to be applied.
+	// recommendedProfile is the TuneD profile the operator calculated to be applied.
+	// This variable is used to cache the value which was written to tunedRecommendFile.
+	recommendedProfile string
+}
+
+type Change struct {
+	// Did the node Profile k8s object change?
+	profile bool
+	// Do we need to update Tuned Profile status?
+	profileStatus bool
+	// Did the "rendered" Tuned k8s object change?
+	rendered bool
+
+	// The following keys are set when profile == true.
+	// Was debugging set in Profile k8s object?
+	debug bool
+	// Cloud Provider as detected by the operator.
+	provider string
+	// Should we turn the reapply_sysctl TuneD option on in tuned-main.conf file?
+	reapplySysctl bool
+	// The current recommended profile as calculated by the operator.
 	recommendedProfile string
 }
 
@@ -107,19 +132,7 @@ type Controller struct {
 	listers *ntoclient.Listers
 	clients *ntoclient.Clients
 
-	change struct {
-		// Did the node Profile k8s object change?
-		profile bool
-		// Did the "rendered" Tuned k8s object change?
-		rendered bool
-		// Did tunedBootcmdlineFile change on the filesystem?
-		// It is set to false on successful Profile update.
-		bootcmdline bool
-		// Did the command-line parameters to run the TuneD daemon change?
-		// In other words, is a complete restart of the TuneD daemon needed?
-		daemon bool
-	}
-
+	change Change
 	daemon Daemon
 
 	tunedCmd     *exec.Cmd       // external command (tuned) being prepared or run
@@ -130,9 +143,14 @@ type Controller struct {
 	tunedMainCfg *ini.File       // global TuneD configuration as defined in tuned-main.conf
 }
 
-type wqKey struct {
+type wqKeyKube struct {
 	kind string // object kind
 	name string // object name
+}
+
+type wqKeyTuned struct {
+	kind   string // object kind
+	change Change // object change type
 }
 
 // Get a client from kubelet's kubeconfig to write to the Node object.
@@ -181,10 +199,10 @@ func newController(stopCh <-chan struct{}) (*Controller, error) {
 		kubeclient:  kubeclient,
 		listers:     listers,
 		clients:     clients,
-		tunedExit:   make(chan bool, 1),
+		tunedExit:   make(chan bool),
 		stopCh:      stopCh,
-		changeCh:    make(chan bool, 1),
-		changeChRet: make(chan bool, 1),
+		changeCh:    make(chan bool),
+		changeChRet: make(chan bool),
 	}
 
 	return controller, nil
@@ -200,15 +218,15 @@ func (c *Controller) eventProcessorKube() {
 			return
 		}
 
-		klog.V(2).Infof("got event from workqueue")
+		klog.V(2).Infof("got event from workqueue: %#v", obj)
 		func() {
 			defer c.wqKube.Done(obj)
-			var workqueueKey wqKey
+			var workqueueKey wqKeyKube
 			var ok bool
 
-			if workqueueKey, ok = obj.(wqKey); !ok {
+			if workqueueKey, ok = obj.(wqKeyKube); !ok {
 				c.wqKube.Forget(obj)
-				klog.Errorf("expected wqKey in workqueue but got %#v", obj)
+				klog.Errorf("expected wqKeyKube in workqueue but got %#v", obj)
 				return
 			}
 
@@ -235,7 +253,7 @@ func (c *Controller) eventProcessorKube() {
 	}
 }
 
-func (c *Controller) sync(key wqKey) error {
+func (c *Controller) sync(key wqKeyKube) error {
 	switch {
 	case key.kind == wqKindTuned:
 		if key.name != tunedv1.TunedRenderedResourceName {
@@ -243,22 +261,13 @@ func (c *Controller) sync(key wqKey) error {
 		}
 		klog.V(2).Infof("sync(): Tuned %s", key.name)
 
-		tuned, err := c.listers.TunedResources.Get(key.name)
-		if err != nil {
-			return fmt.Errorf("failed to get Tuned %s: %v", key.name, err)
-		}
-
-		change, err := profilesSync(tuned.Spec.Profile, c.daemon.recommendedProfile)
-		if err != nil {
-			return err
-		}
-		c.change.rendered = change
-		// Notify the event processor that the Tuned k8s object containing TuneD profiles changed.
-		c.wqTuneD.Add(wqKey{kind: wqKindDaemon})
+		// Notify the event processor that the Tuned k8s object containing all TuneD profiles changed.
+		c.wqTuneD.Add(wqKeyTuned{kind: wqKindDaemon, change: Change{rendered: true}})
 
 		return nil
 
 	case key.kind == wqKindProfile:
+		var change Change
 		if key.name != getNodeName() {
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


jmencak:
/cc @yanirq @ffromani  @MarSik 
This is the cleanup PR I talked about during the bug scrub.  Once this merges, I'm no longer aware of any races in the operand code.  Next plan it to remove the `tuned/rendered` resource.  This will come in a separate PR.

ffromani:
thanks, I'll carefully review ASAP

jmencak:
Infra issues
/retest

pkg/tuned/controller.go
```diff
@@ -42,9 +42,20 @@ const (
 	scWarn
 	scError
 	scSysctlOverride
+	scReloading // reloading is true during the TuneD daemon reload.
 	scUnknown
 )
 
+const (
+	// Constants used for controlling TuneD;
+	// they will be set to 2^0, 2^1, 2^2, ..., 2^n
```
ffromani, line 51: this general comment is probably best fit above `const` on line 49 (applies to other consts e.g. like lines 38-47)

pkg/tuned/controller.go
```diff
@@ -42,9 +42,20 @@ const (
...
+	ctrlReload Bits = 1 << iota
```
ffromani, line 52: we seem to miss a docstring specific to `ctrlReload`

pkg/tuned/controller.go
```diff
@@ -74,24 +85,38 @@ const (
...
-	// reloaded is true immediately after the TuneD daemon finished reloading.
-	// and the node Profile k8s object's Status needs to be set for the operator;
-	// it is set to false on successful Profile update.
-	reloaded bool
-	// debugging flag
-	debug bool
-	// bit/set representaton of Profile status conditions to report back via API.
+	// bit/set representation of TuneD options which are remotely controllable.
+	restart Bits
+	// bit/set representation of Profile status conditions to report back via API.
```
ffromani, line 93: why are we moving from booleans to a bitset?

pkg/tuned/controller.go
```diff
@@ -181,10 +199,10 @@ func newController(stopCh <-chan struct{}) (*Controller, error) {
 		kubeclient:  kubeclient,
 		listers:     listers,
 		clients:     clients,
-		tunedExit:   make(chan bool, 1),
+		tunedExit:   make(chan bool),
```
ffromani, line 202: let's record the rationale from moving to buffered to unbuffered

pkg/tuned/controller.go
```diff
@@ -516,27 +535,31 @@ func (c *Controller) tunedRun() {
...
 		// Looks like there has been a termination signal prior to starting tuned.
-		return false, nil
+		return nil
 	}
 	if c.tunedCmd.Process != nil {
 		// The TuneD daemon rolls back the current profile and should terminate on SIGTERM.
 		klog.V(1).Infof("sending SIGTERM to PID %d", c.tunedCmd.Process.Pid)
 		if err := c.tunedCmd.Process.Signal(syscall.SIGTERM); err != nil {
-			return false, fmt.Errorf("failed to signal TuneD process: %v", err)
+			if err == os.ErrProcessDone {
```
ffromani, line 554: I'd use `errors.Is` everywhere

ffromani (commented):
initial review, I'll need more time to process the state machine changes.
So far this seems a very nice set of improvements and cleanups, but more than ever I'd have split each change from the bullet list in its own commit to make it easier to review and to enable git bisect and easier troubleshooting later on.

pkg/tuned/controller.go
jmencak: This is in preparation for the containerized `TuneD`. The way I had the PoC implemented was that we had two bitset files.  One was to control TuneD (remotely from the daemonset), the other one was used to communicate the TuneD status.

pkg/tuned/controller.go
jmencak: FWICT there was never any need to use a buffered channel.  The buffered channel was never intentional and is likely a copy-and-paste mistake.

All I remember in this context is that it the past, we had a bug where TuneD would be agressively restarted by openshift-tuned when a machine would go down for reboot. This would delay reboot by a minute or two before systemd would kill TuneD and openshift-tuned with SIGKILL.

All that matters in my view is not to agressively restart TuneD when it gets SIGTERM or exits for any reason.  This is for two reasons:
  * Blocking machine reboots
  * IPIs

jmencak:
> initial review, I'll need more time to process the state machine changes. So far this seems a very nice set of improvements and cleanups, but more than ever I'd have split each change from the bullet list in its own commit to make it easier to review and to enable git bisect and easier troubleshooting later on.


Thank you for the initial review, Francesco. As for the split to several commits/bisect, OpenShift CI has a [merge_method squash](https://github.com/openshift/release/blob/2f8527b7eb78074a3088b1095e20b0ccc2f6f6f1/core-services/prow/02_config/openshift/cluster-node-tuning-operator/_prowconfig.yaml#L3), so I believe this will not benefit you much unless you rely on github's web interface, right?  If you insist on this, I can try to do the split later on once all the reviewers are happy with the code.

ffromani:
> > initial review, I'll need more time to process the state machine changes. So far this seems a very nice set of improvements and cleanups, but more than ever I'd have split each change from the bullet list in its own commit to make it easier to review and to enable git bisect and easier troubleshooting later on.
> 
> Thank you for the initial review, Francesco. As for the split to several commits/bisect, OpenShift CI has a [merge_method squash](https://github.com/openshift/release/blob/2f8527b7eb78074a3088b1095e20b0ccc2f6f6f1/core-services/prow/02_config/openshift/cluster-node-tuning-operator/_prowconfig.yaml#L3), so I believe this will not benefit you much unless you rely on github's web interface, right? If you insist on this, I can try to do the split later on once all the reviewers are happy with the code.

If we ever split the commits is better to do it sooner to help the review :) Incidentally, I'm also against the merge method squash because it pretty much prevents any hope of (git) bisect in exchange to (AFAICT) only a marginally easier way to backport PRs. The tradeoff is pretty negative, but perhaps there are reasons I'm missing. Anyway I'm not pushing for this split, especially since it seems I'm the odd one here going against the flow.

pkg/tuned/controller.go
```diff
@@ -450,6 +439,35 @@ func providerExtract(provider string) error {
...
+			return len(provider) > 0, providerExtract(provider)
+		}
+		return false, err
+	}
+	defer f.Close()
+
+	var scanner = bufio.NewScanner(f)
+	for scanner.Scan() {
+		providerCurrent = strings.TrimSpace(scanner.Text())
+	}
```
ffromani, line 463: On second thought *IF* we own `openshiftTunedProvider` would it make sense to just use `os.ReadFile` to read its (IIUC?) single-word content?

pkg/tuned/controller.go
```diff
@@ -450,6 +439,35 @@ func providerExtract(provider string) error {
...
+			// The provider file hasn't been created yet.
+			return len(provider) > 0, providerExtract(provider)
+		}
+		return false, err
+	}
+	defer f.Close()
+
+	var scanner = bufio.NewScanner(f)
+	for scanner.Scan() {
+		providerCurrent = strings.TrimSpace(scanner.Text())
```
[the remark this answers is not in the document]
jmencak: There's really only a single provider line. We don't really need the loop. I can get rid of it if you prefer.

pkg/tuned/controller.go
```diff
@@ -546,34 +569,52 @@ func (c *Controller) tunedStop() (bool, error) {
...
-	return true, nil
+	return nil
 }
 
 func (c *Controller) tunedReload() error {
-	c.daemon.reloading = true
+	klog.Infof("tunedReload()")
+
 	c.daemon.status = 0 // clear the set out of which Profile status conditions are created
+	c.daemon.status |= scReloading
```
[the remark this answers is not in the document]
jmencak: This is really a matter of preference.  I did think about doing what you suggest, but then I thought it is better to be **explicit** about clearing the set.  If you insist, we can optimize, but I personally feel this is better.

pkg/tuned/controller.go
jmencak: Yes, will change this, thank you.

pkg/tuned/controller.go
```diff
@@ -450,6 +439,35 @@ func providerExtract(provider string) error {
...
+// Read the Cloud Provider name from openshiftTunedProvider file and
+// extract/write 'provider' only if the file does not exist or it does not
+// match 'provider'.  Returns indication whether the 'provider' changed and
+// an error if any.
+func providerSync(provider string) (bool, error) {
+	var providerCurrent string
+
+	f, err := os.Open(openshiftTunedProvider)
+	if err != nil {
+		if os.IsNotExist(err) {
```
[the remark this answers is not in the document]
jmencak: Not sure I understand "overwriting if exists". Is your suggestion to use `stat` first before trying to open the file?

pkg/tuned/controller.go
ffromani: I'd write as 
```
func providerSync(provider string) (bool, error) {
        providerCurrent, err := os.ReadFile(openshiftTunedProvider)
        if err != nil {
                if os.IsNotExist(err) {
                        return len(provider) > 0, providerExtract(provider)
                }
                return false, err
        }

        if provider == string(providerCurrent) {
                return false, nil            
        }

        return true, providerExtract(provider)
}
```

but besides that, yep, looks nice

pkg/tuned/controller.go
```diff
@@ -733,37 +836,49 @@ func (c *Controller) eventProcessorTuneD() {
...
+				//   * update Profile status
+				//   * extract TuneD configuration
+				//   * reload/restart TuneD
+				// Requeuing based on TuneD reloading is far from ideal, but it is simple/reliable.
+				// When rewriting this, beware of rare deadlocks.
+				c.wqTuneD.AddRateLimited(workqueueKey)
+				return
+			}
+
+			c.change = workqueueKey.change
```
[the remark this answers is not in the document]
jmencak: This is not a question of _how hard_.  IIUC, you'd like to remove the ratelimiting workqueue for the `TuneD` events? If so, it could actually simplify things. However, the idea behind the ratelimiting workqueue is to "deduplicate" the same events, so that you don't need to process all of them, but just the "last" one.  This is the theory.  Now, in practise, whether it this really works and helps, hard to tell.  I've only simulated this in a contrived example. Saying that, not sure we want to make major changes to a working code that we expect to backport. But perhaps we really want to change this to make things simpler.  Happy to hear opinions.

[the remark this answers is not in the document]
ffromani: It is about rate limiting (process no more than N events over T time units) or about mutual exclusion? because the fact we're using a per-controller variable let me think we may have introduced a form of the latter perhaps incidentally. (thinking about https://en.wikipedia.org/wiki/Semaphore_(programming) )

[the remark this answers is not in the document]
ffromani: This is a sketch of what I'm suggesting. It would be really beneficial to document (with a comment) in the code why the rate limiting is needed, to shape further evolutions:
```
index e7d89531..35fc34a9 100644
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -133,13 +133,12 @@ type Controller struct {
 	listers *ntoclient.Listers
 	clients *ntoclient.Clients
 
-	change Change
 	daemon Daemon
 
 	tunedCmd     *exec.Cmd       // external command (tuned) being prepared or run
 	tunedExit    chan bool       // bi-directional channel to signal and register TuneD daemon exit
 	stopCh       <-chan struct{} // receive-only channel to stop the openshift-tuned controller
-	changeCh     chan bool       // bi-directional channel to wake-up the main thread to process accrued changes
+	changeCh     chan Change     // bi-directional channel to wake-up the main thread to process accrued changes
 	changeChRet  chan bool       // bi-directional channel to announce success/failure of change processing
 	tunedMainCfg *ini.File       // global TuneD configuration as defined in tuned-main.conf
 }
@@ -202,7 +201,7 @@ func newController(stopCh <-chan struct{}) (*Controller, error) {
 		clients:     clients,
 		tunedExit:   make(chan bool),
 		stopCh:      stopCh,
-		changeCh:    make(chan bool),
+		changeCh:    make(chan Change),
 		changeChRet: make(chan bool),
 	}
 
@@ -686,10 +685,10 @@ func GetBootcmdline() (string, error) {
 	return responseString, nil
 }
 
-func (c *Controller) changeSyncerProfileStatus() (synced bool) {
+func (c *Controller) changeSyncerProfileStatus(change Change) (synced bool) {
 	klog.V(2).Infof("changeSyncerProfileStatus()")
 
-	if c.change.profileStatus {
+	if change.profileStatus {
 		// One or both of the following happened:
 		// 1) tunedBootcmdlineFile changed on the filesystem.  This is very likely the result of
 		//    applying a TuneD profile by the TuneD daemon.  Make sure the node Profile k8s object
@@ -708,7 +707,7 @@ func (c *Controller) changeSyncerProfileStatus() (synced bool) {
 
 // changeSyncerTuneD synchronizes k8s objects to disk, compares them with
 // current TuneD configuration and signals TuneD process reload or restart.
-func (c *Controller) changeSyncerTuneD() (synced bool, err error) {
+func (c *Controller) changeSyncerTuneD(change Change) (synced bool, err error) {
 	var reload bool
 
 	klog.V(2).Infof("changeSyncerTuneD()")
@@ -720,20 +719,20 @@ func (c *Controller) changeSyncerTuneD() (synced bool, err error) {
 	}
 
 	// Check whether reload of the TuneD daemon is really necessary due to a Profile change.
-	if c.change.profile {
-		changeProvider, err := providerSync(c.change.provider)
+	if change.profile {
+		changeProvider, err := providerSync(change.provider)
 		if err != nil {
 			return false, err
 		}
 		reload = reload || changeProvider
 
-		if c.daemon.recommendedProfile != c.change.recommendedProfile {
-			if err = TunedRecommendFileWrite(c.change.recommendedProfile); err != nil {
+		if c.daemon.recommendedProfile != change.recommendedProfile {
+			if err = TunedRecommendFileWrite(change.recommendedProfile); err != nil {
 				return false, err
 			}
-			klog.Infof("recommended TuneD profile changed from (%s) to (%s)", c.daemon.recommendedProfile, c.change.recommendedProfile)
+			klog.Infof("recommended TuneD profile changed from (%s) to (%s)", c.daemon.recommendedProfile, change.recommendedProfile)
-			c.daemon.recommendedProfile = c.change.recommendedProfile
+			c.daemon.recommendedProfile = change.recommendedProfile
 			reload = true
 		} else {
 			klog.Infof("recommended profile (%s) matches current configuration", c.daemon.recommendedProfile)
@@ -747,10 +746,10 @@ func (c *Controller) changeSyncerTuneD() (synced bool, err error) {
 
 		// Does the current TuneD process have debugging turned on?
 		debug := (c.daemon.restart & ctrlDebug) != 0
-		if debug != c.change.debug {
+		if debug != change.debug {
 			// A complete restart of the TuneD daemon is needed due to a debugging request switched on or off.
 			c.daemon.restart |= ctrlRestart
-			if c.change.debug {
+			if change.debug {
 				c.daemon.restart |= ctrlDebug
 			} else {
 				c.daemon.restart &= ^ctrlDebug
@@ -759,7 +758,7 @@ func (c *Controller) changeSyncerTuneD() (synced bool, err error) {
 
 		// Does the current TuneD process have the reapply_sysctl option turned on?
 		reapplySysctl := c.tunedMainCfg.Section("").Key("reapply_sysctl").MustBool()
-		if reapplySysctl != c.change.reapplySysctl {
+		if reapplySysctl != change.reapplySysctl {
 			if err = iniCfgSetKey(c.tunedMainCfg, "reapply_sysctl", !reapplySysctl); err != nil {
 				return false, err
 			}
@@ -771,7 +770,7 @@ func (c *Controller) changeSyncerTuneD() (synced bool, err error) {
 		}
 	}
 
-	if c.change.rendered {
+	if change.rendered {
 		// The "rendered" Tuned k8s object changed.
 		tuned, err := c.listers.TunedResources.Get(tunedv1.TunedRenderedResourceName)
 		if err != nil {
@@ -790,7 +789,7 @@ func (c *Controller) changeSyncerTuneD() (synced bool, err error) {
 	if reload {
 		c.daemon.restart |= ctrlReload
 	}
-	err = c.changeSyncerRestartOrReloadTuneD()
+	err = changeSyncerRestartOrReloadTuneD()
 
 	return err == nil, err
 }
@@ -817,14 +816,14 @@ func (c *Controller) changeSyncerRestartOrReloadTuneD() error {
 // reloads as needed.  Returns indication whether the change was successfully
 // synced and an error.  Only critical errors are returned, as non-nil errors
 // will cause restart of the main control loop -- the changeWatcher() method.
-func (c *Controller) changeSyncer() (synced bool, err error) {
+func (c *Controller) changeSyncer(change Change) (synced bool, err error) {
 	// Sync k8s Profile status if/when needed.
-	if !c.changeSyncerProfileStatus() {
+	if !c.changeSyncerProfileStatus(change) {
 		return false, nil
 	}
 
 	// Extract TuneD configuration from k8s objects and restart/reload TuneD as needed.
-	return c.changeSyncerTuneD()
+	return c.changeSyncerTuneD(change)
 }
 
 // eventProcessorTuneD is a long-running method that will continually
@@ -860,8 +859,7 @@ func (c *Controller) eventProcessorTuneD() {
 				return
 			}
 
-			c.change = workqueueKey.change
-			c.changeCh <- true
+			c.changeCh <- workqueueKey.change
 			eventProcessed := <-c.changeChRet
 			if !eventProcessed {
 				requeued := c.wqTuneD.NumRequeues(workqueueKey)
@@ -1147,11 +1145,11 @@ func (c *Controller) changeWatcher() (err error) {
 		case err := <-wFs.Errors:
 			return fmt.Errorf("error watching filesystem: %v", err)
 
-		case <-c.changeCh:
+		case ch := <-c.changeCh:
 			var synced bool
 			klog.V(2).Infof("changeCh")
 
-			synced, err := c.changeSyncer()
+			synced, err := c.changeSyncer(ch)
 			if err != nil {
 				return err
 			}
```
the reason why I'm suggesting this change is to go in the direction of reducing the common/implicit state, because doing so very very often leads to easier and simpler code.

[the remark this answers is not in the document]
jmencak: > It is about rate limiting (process no more than N events over T time units) or about mutual exclusion? 

We can probably forget about rate limiting in wqTuneD workqueue.  That's just a convenience that will rarely happen in my view.  I'd say rate limiting only needs to be done for k8s object changes in wqKube workqueue.  What needs to be handled are:
  * requeues of failed attempts at updating profile status for the operator
  * mutual exclusion (not to extract any TuneD config during its reloads/restart and not to reload/restart during ongoing reloads/restarts)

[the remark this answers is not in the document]
jmencak: > This is a sketch of what I'm suggesting. Utterly untested. It would be really beneficial to document (with a comment) in the code why the rate limiting is needed, to shape further evolutions:
> 
> CODE CUT FOR BREVITY
> the reason why I'm suggesting this change is to go in the direction of reducing the common/implicit state, because doing so very very often leads to easier and simpler code.

Oh, I see what you mean now, I was being a bit dumb.  I like the change and thank you for it.  Do you want to co-author or shall I just adapt it?  Testing the code now.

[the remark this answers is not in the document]
ffromani: no need to co-author me for such a simple change. Feel free to take the code and let's move on.

pkg/tuned/controller.go
```diff
@@ -450,6 +439,26 @@ func providerExtract(provider string) error {
...
+	}
+
+	if provider == string(providerCurrent) {
+		return false, nil
+	}
+
+	return true, providerExtract(provider)
+}
+
 func TunedRecommendFileWrite(profileName string) error {
```
ffromani, line 462: I think this function can be simplified quite a lot but this seems to be better done in a followup PR. No changes needed for this PR

ffromani (commented):
/lgtm

at the best of my knowledge I can't see further major changes needed or issues. Unfortunately changes like this can have a tail of fixes, but again nothing I can spot atm.

ffromani:
/hold

I'd welcome more reviews from different ppl

jmencak:
> /hold
> 
> I'd welcome more reviews from different ppl

Thank you.  Any more things I should change, @yanirq , @MarSik so that we unblock Francesco? Thank you!

PR #1025, opened 2024-04-09
OCPBUGS-31905: [release-4.14] [manual] : Add performance real time tuned template

rbaturov:
This is a manual backport for: Add performance real time tuned template (#984)

* Add performance real time tuned template
Kernel parameters that are not supported on RT kernel systems are being applied, causing errors to be logged to the tuned daemon, resulting in a degraded profile state. Therefore, I added the openshift-node-performance-rt profile that would be included if an RT kernel is detected, thereby dropping the unsupported kernel parameters before they are applied.

* Added a test that make sure the tuned profile is not degraded
This could be great to have a generic test that make sure that after performance profile is applied the tuned profile is not degraded. This would prevent future issues like OCPBUGS-23167.
The test will fail only when the tuned profile was found degraded on BM host.
In case of tuned profile found degraded on a VM, only a warning would be reported in the logs.
The reason for that is that the fact CI is using VMs - and some
configurations like certain kernel args can't be applied and therby lead to
the tuned profile being degraded.

* render-sync update
Updated render-sync to include artifacts that are needed for the e2e tests. Committing the rendered items to this commit.

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -1,6 +1,12 @@
 [main]
 summary=Openshift node optimized for deterministic performance at the cost of increased power consumption, focused on low latency network performance. Based on Tuned 2.11 and Cluster node tuning (oc 4.5)
-include=openshift-node,cpu-partitioning
+
+# In case real time kernel is enabled the following include section will be evaluated as:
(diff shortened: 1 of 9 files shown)


jmencak:
/lgtm
/approve
/label backport-risk-assessed

@rbaturov , you need to adjust Jira to fix:
```
release note text must be set and not match the template OR release note type must be set to "Release Note Not Required"
```

PR #1029, opened 2024-04-10
WIP: Replaced tuned hard-included repo with a git submodule

rbaturov:
Until now tuned repo was hard included in the NTO repo. This design could lead to developers mistakenly contribute to assets/tuned. Instead of https://github.com/redhat-performance/tuned, as it is hard to realize it is an external repo. Therefore, we chose to change this design by including the tuned repo as a git submodule, preventing mistakes as I described and better informing tuned is an external repo.


jmencak:
Thank you for the PR, Ronny.  I think this needs a bit more work.


This is what I've done:

````
$ git clone https://github.com/openshift/cluster-node-tuning-operator
$ git fetch origin pull/1029/head:1029
$ git checkout 1029
$ make update-tuned-submodule
(cd assets/tuned/tuned && \
  git pull origin master && \
  cd tuned && git checkout 954bc4624db8fb345e9fb264ee2ac09d1736105a)
From https://github.com/openshift/cluster-node-tuning-operator
 * branch              master     -> FETCH_HEAD
hint: You have divergent branches and need to specify how to reconcile them.
hint: You can do so by running one of the following commands sometime before
hint: your next pull:
hint: 
hint:   git config pull.rebase false  # merge
hint:   git config pull.rebase true   # rebase
hint:   git config pull.ff only       # fast-forward only
hint: 
hint: You can replace "git config" with "git config --global" to set a default
hint: preference for all repositories. You can also pass --rebase, --no-rebase,
hint: or --ff-only on the command line to override the configured default per
hint: invocation.
fatal: Need to specify how to reconcile divergent branches.
make: *** [Makefile:58: update-tuned-submodule] Error 128

````

This also suggests that `TuneD` would be in assets/tuned/tuned/tuned ? Is this intended?

File: Dockerfile [hunk 1]
```diff
@@ -10,7 +10,7 @@ RUN INSTALL_PKGS=" \
       gcc git rpm-build make desktop-file-utils patch dnf-plugins-core \
       " && \
     dnf install --setopt=tsflags=nodocs -y $INSTALL_PKGS && \
-    cd assets/tuned/daemon && \
```
jmencak, line 13: Do we need this change?

rbaturov:
> Thank you for the PR, Ronny. I think this needs a bit more work.
> 
> This is what I've done:
> 
> ```
> $ git clone https://github.com/openshift/cluster-node-tuning-operator
> $ git fetch origin pull/1029/head:1029
> $ git checkout 1029
> $ make update-tuned-submodule
> (cd assets/tuned/tuned && \
>   git pull origin master && \
>   cd tuned && git checkout 954bc4624db8fb345e9fb264ee2ac09d1736105a)
> From https://github.com/openshift/cluster-node-tuning-operator
>  * branch              master     -> FETCH_HEAD
> hint: You have divergent branches and need to specify how to reconcile them.
> hint: You can do so by running one of the following commands sometime before
> hint: your next pull:
> hint: 
> hint:   git config pull.rebase false  # merge
> hint:   git config pull.rebase true   # rebase
> hint:   git config pull.ff only       # fast-forward only
> hint: 
> hint: You can replace "git config" with "git config --global" to set a default
> hint: preference for all repositories. You can also pass --rebase, --no-rebase,
> hint: or --ff-only on the command line to override the configured default per
> hint: invocation.
> fatal: Need to specify how to reconcile divergent branches.
> make: *** [Makefile:58: update-tuned-submodule] Error 128
> ```
> 
> This also suggests that `TuneD` would be in assets/tuned/tuned/tuned ? Is this intended?

Yes. Before, the `assets/tuned/daemon` dir contained the hard-included tuned repo.
After I have changed this design to include a git submodule I removed "daemon" and included the git "tuned" submodule. I could name the submodule "daemon" but it would be misleading as in github it will be displayed as like this:
image

Dockerfile
rbaturov: Yes. the updated dir is `assets/tuned/tuned`. look the the comment above

jmencak: OK, I don't have a big problem with this change, even though I don't think it is needed. What needs to work though is cloning/updating the submodule using the `update-tuned-submodule` target and that doesn't seem to work.

Makefile
```diff
@@ -15,7 +15,7 @@ REV=$(shell git describe --long --tags --match='v*' --always --dirty)
 # Upstream tuned daemon variables
 TUNED_REPO:=https://github.com/redhat-performance/tuned.git
 TUNED_COMMIT:=954bc4624db8fb345e9fb264ee2ac09d1736105a
-TUNED_DIR:=daemon
+TUNED_DIR:=tuned
```
jmencak, line 18: This variable is no longer needed. Let's get rid of it as we're using it only once.

Makefile
```diff
@@ -54,12 +54,14 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
...
-	(cd assets/tuned && \
-	  rm -rf $(TUNED_DIR) && \
-	  git clone -n $(TUNED_REPO) $(TUNED_DIR) && \
-	  cd $(TUNED_DIR) && git checkout $(TUNED_COMMIT) && cd .. && \
-	  rm -rf $(TUNED_DIR)/.git)
+# This target will be run in the Dockerfile to initialize the tuned submodule by cloning it.
+# Moreover, this can be used to update the tuned repo to a specific commit.
+update-tuned-submodule:
+	(git submodule update --init --force && \
+	  cd assets/tuned/$(TUNED_DIR) && \
```
jmencak, line 61: s|assets/tuned/$(TUNED_DIR)|assets/tuned/tuned|
See above.

Makefile
```diff
@@ -54,12 +54,14 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
...
+	  git pull origin master && \
+	  git checkout $(TUNED_COMMIT) &&\
+	  cd -)
```
jmencak, line 64:
```
cd -) 
```
This has no effect, please remove this.

jmencak (changes requested):
Thank you for the changes, looking better now.  Let's address a few nits and also set 
TUNED_COMMIT:=HEAD

Thank you!

Makefile
rbaturov: Right. Also TUNED_REPO. will remove both.

jmencak: yes, also
TUNED_COMMIT:=HEAD
Please

rbaturov:
> Thank you for the changes, looking better now. Let's address a few nits and also set TUNED_COMMIT:=HEAD
> 
> Thank you!

Sorry, missed you asked for that in the previous message.
I have pushed this change now.
This will guarantee tuned repo will always be updated with the latest commit, that's your intention, right?

jmencak:
Thank you for the changes.
/approve
/lgtm

Giving other reviewers a chance to chime in.
/hold

jmencak:
@rbaturov , please add a valid Jira reference, thank you!

jmencak:
@rbaturov feel free to unhold if you don't see any other feedback from reviewers in some time.

Makefile
```diff
@@ -13,9 +13,7 @@ GOFMT_CHECK=$(shell find . -not \( \( -wholename './.*' -o -wholename '*/vendor/
 REV=$(shell git describe --long --tags --match='v*' --always --dirty)
 
 # Upstream tuned daemon variables
-TUNED_REPO:=https://github.com/redhat-performance/tuned.git
-TUNED_COMMIT:=954bc4624db8fb345e9fb264ee2ac09d1736105a
-TUNED_DIR:=daemon
+TUNED_COMMIT:=HEAD
```
bartwensley, line 16: For the upstream NTO build, don't we want to pick a specific tag in the tuned repo to build from?

rbaturov: @jmencak

jmencak: Yes, we do. That's why we keep the `TUNED_COMMIT` variable and it is meant to be used.  It is a lot of maintenance to keep the variable up-to-date in the Makefile.  That's why I've decided to set this to HEAD.  Other thought welcome, but we haven't keep up-to-date with this in the past, which resulted in old `TuneD` in the repo.  I believe that keeping this to HEAD is lesser of the two evils, but I can be wrong.

bartwensley: OK - so you are saying that HEAD will be set by something else when the upstream NTO build is done?

bartwensley: Thanks Jiri - I think it makes sense to use HEAD when building NTO from the master branch. The one tricky part (in the future) might be when someone wants to build upstream NTO from a specific branch (e.g. release-4.xx). In that case, it may be hard to figure out what tuned commit/tag is appropriate and using HEAD will probably not be the right choice. To make it easier, it might be nice to set the TUNED_COMMIT here to an appropriate tag when a new NTO branch is created, but we probably won't remember.

bartwensley (approved):
Looks good

PR #1030, opened 2024-04-11
PSAP-1354: Updated a script for deploying a custom NTO image to a cluster

rbaturov:
Updated a script to deploy a custom NTO image to a cluster
* Using this script developers could easily test their code.
Running it will replace the NTO image in the current running operator deployment with the custom image.

 Updated HACKING.md to build & deploy custom NTO image
* It is important to make sure this file contains correct steps, and does not turn outdated.
Therefore, a new contributor reading HACKING.md file is able to deploy a custom version of NTO.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -2,13 +2,13 @@
 
 # Hacking on the Node Tuning Operator
 
-These instructions have been tested on Fedora 36.
+These instructions have been tested on Fedora 39.
 
 ## Prerequisites
 
 Most changes you'll want to test live against a real cluster. Instructions are TODO.
 
-Go v1.18 is used for development of the NTO.
+Go v1.21 is used for development of the NTO.
 
 # Static Analysis
 
@@ -38,69 +38,14 @@ that require a cluster to properly test your code.
 
 In those situations, you will need to:
 
-1) build the NTO image you're interested in testing
+1) Build the NTO image you're interested in testing
 2) Push the image to an OpenShift cluster
 
-## Build NTO image
+To build the repo's code as an image and push it to the cluster we have a script that 
+can automate this for you: 
 
-To build an image that contains all NTO components run:
+`ORG=quay-user hack/deploy-custom-nto.sh`
 
-```
-# Specify the registry
-REGISTRY=quay.io
-# Specify the registry user
-ORG=username
-# Specify the tag
-TAG=tag
+It is also possible to opt for a specific revision of [TuneD](https://github.com/redhat-performance/tuned) by specifying: `TUNED_COMMIT=commit-hash`
 
-make local-image IMAGE=${REGISTRY}/${ORG}/cluster-node-tuning-operator:${TAG}
-```
-
-After the build is complete, you can push the image to a registry with:
-
-```
-# Specify the registry
-REGISTRY=quay.io
-# Specify the registry user
-ORG=username
-# Specify the tag
-TAG=tag
-
-make local-image-push IMAGE=${REGISTRY}/${ORG}/cluster-node-tuning-operator:${TAG}
-```
-
-## Push the image to an OpenShift cluster
-
-```
-override_tuning_operator ()
-{
-  oc patch clusterversion version --type json -p '[{"op":"add","path":"/spec/overrides","value":[{"kind":"Deployment","group":"apps","name":"cluster-node-tuning-operator","namespace":"openshift-cluster-node-tuning-operator","unmanaged":true}]}]'
-}
-
-custom_tuning_operator ()
-{
-  TUNING_IMAGE=${REGISTRY}/${ORG}/cluster-node-tuning-operator:${TAG}
-  override_tuning_operator
-  oc scale deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator --replicas=0
-  oc patch deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator -p \
-    '{"spec":{"template":{"spec":{"containers":[{"name":"cluster-node-tuning-operator","image":"'${TUNING_IMAGE}'"}]}}}}'
-  oc patch deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator -p \
-    '{"spec":{"template":{"spec":{"containers":[{"name":"cluster-node-tuning-operator","env":[{"name":"CLUSTER_NODE_TUNED_IMAGE","value":"'${TUNING_IMAGE}'"}]}]}}}}'
-  oc describe  deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator | grep --color=auto Image
-  oc describe  deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator | grep --color=auto CLUSTER_NODE_TUNED_IMAGE
-  oc delete ds -n openshift-cluster-node-tuning-operator tuned
-  oc scale deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator --replicas=1
-  sleep 10
-  oc wait deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator --for condition=Available=True --timeout=90s
-  for i in {1..36}; do
-    echo "Waiting for DS to be deployed, try $i"
-    sleep 10
-    if oc get ds -n openshift-cluster-node-tuning-operator tuned 2>/dev/null; then
-      break
-    fi
-  done
-  oc rollout status daemonset -n openshift-cluster-node-tuning-operator tuned --timeout=90s
-  oc get deployment -n openshift-cluster-node-tuning-operator
-  oc get ds -n openshift-cluster-node-tuning-operator
-}
-```
+We recommend running some e2e tests to verify the custom image works as expected.
--- a/hack/deploy-custom-nto.sh
+++ b/hack/deploy-custom-nto.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
 set -e
+WORKDIR=$(dirname "$(realpath "$0")")/..
 
 # Example invocation
 # ~~~~~~~~~~~~~~~~~~
-# ORG=quay-user hack/test-upstream-tuned.sh
+# ORG=quay-user hack/deploy-custom-nto.sh
 #
 # This script assumes a deployed OpenShift cluster and an environment 
 # configured to write the locally built IMAGE to an image repository.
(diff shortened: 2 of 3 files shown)


rbaturov:
cc @jmencak

HACKING.md
```diff
@@ -38,69 +38,37 @@ that require a cluster to properly test your code.
...
-make local-image IMAGE=${REGISTRY}/${ORG}/cluster-node-tuning-operator:${TAG}
 ```
 
+Verify the `IMAGE` variable is being constructed as you expect.
+
+# Build NTO image
+
+To build an image that contains all NTO components run:
+
+`make local-image`
```
ffromani, line 61: this works only if the `IMAGE` environment variable is `export`ed, right?

HACKING.md
```diff
@@ -38,69 +38,37 @@ that require a cluster to properly test your code.
...
 After the build is complete, you can push the image to a registry with:
 
-```
-# Specify the registry
-REGISTRY=quay.io
-# Specify the registry user
-ORG=username
-# Specify the tag
-TAG=tag
+`make local-image-push`
```
ffromani, line 65: ditto

HACKING.md
rbaturov: Yes. The `IMAGE` is set in the makefile.

HACKING.md
```diff
@@ -38,69 +38,37 @@ that require a cluster to properly test your code.
...
 In those situations, you will need to:
 
-1) build the NTO image you're interested in testing
+1) Build the NTO image you're interested in testing
 2) Push the image to an OpenShift cluster
 
-## Build NTO image
-
-To build an image that contains all NTO components run:
+Before continue with the next steps, please change the following environment variables in the **makefile**:
```
jmencak, line 44: No, users shouldn't need to change makefile variables at all.  You can call make with something like:

```
make <target> VARIABLE=xyz
```

HACKING.md
jmencak: Shell variable `export` will not help you.  You need to specifically set make variables on the command line.  We should probably be more specific here and say something like:

```
make local-image IMAGE=$IMAGE
```

This needs to be fixed in the docs.

Makefile
```diff
@@ -161,6 +161,9 @@ local-image:
 local-image-push:
 	$(IMAGE_PUSH_CMD) $(IMAGE_PUSH_EXTRA_OPTS) $(IMAGE)
 
+replace-nto-image:
+	IMAGE=$(IMAGE) hack/replace-nto.sh
```
jmencak, line 165: If possible, let's avoid adding a new Makefile target for this.

1) This will not work as is, you'd have to say something like "make IMAGE=whatever replace-nto-image"
2) I'd say this is overcomplicated and I don't see any value adding a new target for this.  There is plenty of hack/* scripts that don't have a makefile target.

hack/replace-nto.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
jmencak, line 1: There's a lot of copy and paste from `test-upstream-tuned.sh` here.  I wonder if we should refactor `test-upstream-tuned.sh`, make some of the functions into a shell functions file. In any case, the duplication seems bad.

jmencak (commented):
Some food for thought (initial review).

hack/replace-nto.sh
rbaturov: Thanks @jmencak.
Now that I'm thinking of it, could we just get rid of the `test-upstream-tuned.sh` script?  It seems redundant now that we've decided to always track the tuned repo and stay updated (also my newly added script in this case does the same).

jmencak: `test-upstream-tuned.sh` has a clear purpose.  To deploy NTO image with a certain version of TuneD and run e2e tests on it.  FWICS, your script is basically a copy of `test-upstream-tuned.sh` (I'd expect that fact reflected in git history...) adds application of manifests. That's good.  But it drops the e2e test runs.  We can probably live with that and run the tests manually after the custom version is deployed.  So yeah, after you clean the script, we can probably do what you suggest.

hack/replace-nto.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -e
+
+# This script assumes a deployed OpenShift cluster and an environment .
```
jmencak, line 5: Copy and paste without a clear end.

hack/replace-nto.sh
```diff
@@ -0,0 +1,54 @@
...
+cvo_scale() {
+  local replicas=${1:-0}
+  oc scale deploy/cluster-version-operator --replicas=$replicas -n openshift-cluster-version
+}
+
+nto_deploy_custom() {
+  oc project openshift-cluster-node-tuning-operator
+  for file in ./manifests/*; do
+    # We will skip the operator deployment manifest and just patch it instead. 
+    if [[ "$file" != ./manifests/50-operator* && "$file" == *.yaml ]]; then
```
jmencak, line 21: Why do we want to skip that manifest?

hack/replace-nto.sh
```diff
@@ -0,0 +1,54 @@
...
+      oc apply -f "$file"
+    fi
+  done
+oc patch deploy cluster-node-tuning-operator -p '{"spec":{"template":{"spec":{"containers":[{"name":"cluster-node-tuning-operator","image":"'$IMAGE'","env":[{"name":"CLUSTER_NODE_TUNED_IMAGE","value":"'$IMAGE'"}]}]}}}}'
```
jmencak, line 25: Indentation.

hack/replace-nto.sh
rbaturov: Yes, my script is pretty much copy and paste. I'd like to get it unified and have one script for both cases (as we don't need two separate scripts).
We can leave the e2e scripts, I personally thought this should be optional (maybe add documentation to hacking.md how to run them?). If you insist on keeping them, we can do it.

hack/replace-nto.sh
rbaturov: This manifest includes the operator deployment (which is already deployed with OCP). we are patching the image fields and inserting our custom image reference instead of changing and reapplying the yaml (supposed to have the same effect).

hack/replace-nto.sh
jmencak: I do not insist on having two scripts. The original script existed so that we have an easy way of testing new `TuneD` releases without adding extra commands. If you agree, we can e.g. "git mv test-upstream-tuned.sh deploy-custom-nto.sh" and then you can build on it.

hack/replace-nto.sh
jmencak: >  (which is already deployed with OCP).

All the manifests are already deployed with OCP, the operator manifest is no way special and the developer might have changed it to suit his/her own needs.

rbaturov: I agree. I think then we should only skip the `50-operator-ibm-cloud-managed.yaml` then.

hack/deploy-custom-nto.sh
```diff
@@ -31,7 +32,14 @@ cvo_scale() {
 
 nto_deploy_custom() {
   oc project openshift-cluster-node-tuning-operator
-  oc patch deploy cluster-node-tuning-operator -p '{"spec":{"template":{"spec":{"containers":[{"env":[{"name":"CLUSTER_NODE_TUNED_IMAGE","value":"'$IMAGE'"}],"name":"cluster-node-tuning-operator"}]}}}}'
+  cd "${WORKDIR}" || { echo "failed to change dir to ${WORKDIR}"; exit; }
+  for file in ${WORKDIR}/manifests/*; do
+    # We will skip the ibm specific deployment. 
+    if [[ "$file" != ${WORKDIR}/manifests/50-operator-ibm* && "$file" == *.yaml ]]; then
```
jmencak, line 38: Why don't we just list `*.yaml` files instead of the check for a `*.yaml` file here?

hack/deploy-custom-nto.sh
```diff
@@ -31,7 +32,14 @@ cvo_scale() {
 
 nto_deploy_custom() {
   oc project openshift-cluster-node-tuning-operator
-  oc patch deploy cluster-node-tuning-operator -p '{"spec":{"template":{"spec":{"containers":[{"env":[{"name":"CLUSTER_NODE_TUNED_IMAGE","value":"'$IMAGE'"}],"name":"cluster-node-tuning-operator"}]}}}}'
+  cd "${WORKDIR}" || { echo "failed to change dir to ${WORKDIR}"; exit; }
```
jmencak, line 35: Errors should go to `stderr` and we should not exit 0 when there is a failure. Let's exit 1.

jmencak (commented):
I just went through the updated script. Any reasons we still use `nto_run_tuned_tests` there?

hack/deploy-custom-nto.sh
```diff
@@ -31,7 +32,14 @@ cvo_scale() {
 
 nto_deploy_custom() {
   oc project openshift-cluster-node-tuning-operator
-  oc patch deploy cluster-node-tuning-operator -p '{"spec":{"template":{"spec":{"containers":[{"env":[{"name":"CLUSTER_NODE_TUNED_IMAGE","value":"'$IMAGE'"}],"name":"cluster-node-tuning-operator"}]}}}}'
+  cd "${WORKDIR}" || { echo "failed to change dir to ${WORKDIR}" >&2; exit 1; }
```
jmencak, line 35: Changing user's working directory and not changing it back where they were could be a bit annoying for them.  Can we prevent this in a clean way?

jmencak (commented):
Thank you for the changes so far Ronny, much appreciated.  I'd say this nearly looks good to me.  Looking at HACKING.md file, there are many other things I'd like to add/clarify, but I can do this in a follow-up PR myself.

hack/deploy-custom-nto.sh
rbaturov: Adding `cd -` after the for loop?

jmencak: AFAICS, we don't need to change the directory in `nto_deploy_custom()` at all, do we?  It only needs to be slightly updated.  Where we'll likely need to chagne a directory is where you're calling the `make` commands.  For that, you can use `make -C $WORKDIR`.  Then, you shouldn't need to the directory in the script at all.  Please let me know if I've missed something.

hack/deploy-custom-nto.sh
```diff
@@ -57,21 +64,8 @@ wait_for_updated_tuned_pods() {
...
-  # e2e-gcp-pao-updating-profile usually take 4-5h
-  make pao-functests-updating-profile
-}
-
 nto_prepare_image
 cvo_scale 0
 nto_deploy_custom
 wait_for_updated_tuned_pods
-nto_run_tuned_tests
 cvo_scale 1
```
jmencak, line 71: `cvo_scale 1` needs to be removed, otherwise CVO will bring back its own version of NTO.  This was fine after the nto_run_tuned_tests were done, but we cannot leave it here now.

jmencak (changes requested):
Two last changes and we're good to go.

For the second one, could you please remove 
```
find /root/rpms -name \*.rpm -exec basename {} .rpm \; | xargs rpm -e --justdb && \`
```
There is no longer any reason for this and in fact, removing this will help us identify `TuneD` used in the `TuneD` pod.

jmencak:
Thank you for the PR
/approve
/lgtm
To give other reviewers a chance to comment (please unhold if you're not getting any in a few days). This needs to merge soon.
/hold

PR #1031, opened 2024-04-11
NO-JIRA: E2E: use strings.CutPrefix instead of trim function to get containerID

mrniranjan:
earlier when using strings.Trim , if the containerID contains any  characters from "cri-o://" it can remove these characters giving us invalid containerID

Instead of using Trim use the strings.CutPrefix method to fetch the containerID

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/pods/pods.go
+++ b/test/e2e/performanceprofile/functests/utils/pods/pods.go
@@ -205,15 +205,17 @@ func GetContainerIDByName(pod *corev1.Pod, containerName string) (string, error)
 	if err := testclient.Client.Get(context.TODO(), key, updatedPod); err != nil {
 		return "", err
 	}
+	// Find the container by name
 	for _, containerStatus := range updatedPod.Status.ContainerStatuses {
 		if containerStatus.Name == containerName {
-			containerID := containerStatus.ContainerID
-			// Split the container ID at the "://" delimiter, keeping the second part:
-			containerID = strings.SplitN(containerID, "://", 2)[1]
-			return containerID, nil
-
+			containerId, found := strings.CutPrefix(containerStatus.ContainerID, "cri-o://")
+			if !found {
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/utils/pods/pods.go
```diff
@@ -207,7 +207,11 @@ func GetContainerIDByName(pod *corev1.Pod, containerName string) (string, error)
 	}
 	for _, containerStatus := range updatedPod.Status.ContainerStatuses {
 		if containerStatus.Name == containerName {
-			return strings.Trim(containerStatus.ContainerID, "cri-o://"), nil
+			containerID := containerStatus.ContainerID
+			// Split the container ID at the "://" delimiter, keeping the second part:
+			containerID = strings.SplitN(containerID, "://", 2)[1]
+			return containerID, nil
+
```
ffromani, line 214: this works, but it seems a textbook case of https://pkg.go.dev/strings#CutPrefix . Can we use `CutPrefix`?

jmencak: good point
/lgtm cancel

mrniranjan: If the prefix is not found that also means it's junk , as we are  checking object of v1.ContainerStatus where it's of   '<type>://<container_id>'.

ffromani: fair enough! I still think `CutPrefix` can help; but even if it is not, and we need to use `Split` or `SplitN`, we need a bit more robustness check

mrniranjan: updating the patch to use CutPrefix

MarSik (approved):
/lgtm

ffromani (commented):
/lgtm

PR #1032, opened 2024-04-11
CNF-6517: [Part 1]  Hypershift PAO adoption 

Tal-or:
Before we start with the implementation itself we should do some preliminary  work in order to generalize some of the functions and flows so they could fit for the controller when it runs on  hypershift platform.

The main work done here, is extracting bits which are specific for non-hypershift platform and replacing methods with functions.

For example we want to pass the `client` from the caller because on hypershift the client is different than on non-hypershift platform.

NOTE: whether we decide to go with unified controller for both hypershift and none-hypertshift or a separate controller for each, this PR is good improvement in general and will allow us to reuse more code.
This is the main reason for dedicating a separate PR for this code changes.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/resources/resources.go
+++ b/pkg/performanceprofile/controller/performanceprofile/resources/resources.go
@@ -1,13 +1,9 @@
-package controller
+package resources
 
 import (
 	"context"
 	"encoding/json"
 	"reflect"
-
-	apiconfigv1 "github.com/openshift/api/config/v1"
-	mcov1 "github.com/openshift/api/machineconfiguration/v1"
-	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 
 	nodev1 "k8s.io/api/node/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
@@ -16,6 +12,11 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	apiconfigv1 "github.com/openshift/api/config/v1"
+	mcov1 "github.com/openshift/api/machineconfiguration/v1"
+	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 )
 
 func mergeMaps(src map[string]string, dst map[string]string) {
@@ -27,20 +28,20 @@ func mergeMaps(src map[string]string, dst map[string]string) {
 
 // TODO: we should merge all create, get and delete methods
 
-func (r *PerformanceProfileReconciler) getMachineConfig(ctx context.Context, name string) (*mcov1.MachineConfig, error) {
+func GetMachineConfig(ctx context.Context, cli client.Client, name string) (*mcov1.MachineConfig, error) {
 	mc := &mcov1.MachineConfig{}
 	key := types.NamespacedName{
 		Name:      name,
 		Namespace: metav1.NamespaceNone,
 	}
-	if err := r.Get(ctx, key, mc); err != nil {
+	if err := cli.Get(ctx, key, mc); err != nil {
 		return nil, err
 	}
 	return mc, nil
 }
 
-func (r *PerformanceProfileReconciler) getMutatedMachineConfig(ctx context.Context, mc *mcov1.MachineConfig) (*mcov1.MachineConfig, error) {
-	existing, err := r.getMachineConfig(ctx, mc.Name)
+func GetMutatedMachineConfig(ctx context.Context, cli client.Client, mc *mcov1.MachineConfig) (*mcov1.MachineConfig, error) {
+	existing, err := GetMachineConfig(ctx, cli, mc.Name)
 	if errors.IsNotFound(err) {
 		return mc, nil
 	}
@@ -64,23 +65,23 @@ func (r *PerformanceProfileReconciler) getMutatedMachineConfig(ctx context.Conte
 	return mutated, nil
 }
 
-func (r *PerformanceProfileReconciler) getClusterOperator() (*apiconfigv1.ClusterOperator, error) {
+func GetClusterOperator(ctx context.Context, cli client.Client) (*apiconfigv1.ClusterOperator, error) {
 	co := &apiconfigv1.ClusterOperator{}
 	key := types.NamespacedName{
 		Name:      "node-tuning",
 		Namespace: metav1.NamespaceNone,
 	}
-	if err := r.Get(context.TODO(), key, co); err != nil {
+	if err := cli.Get(ctx, key, co); err != nil {
 		return nil, err
 	}
 	return co, nil
 }
 
-func (r *PerformanceProfileReconciler) createOrUpdateMachineConfig(mc *mcov1.MachineConfig) error {
-	_, err := r.getMachineConfig(context.TODO(), mc.Name)
+func CreateOrUpdateMachineConfig(ctx context.Context, cli client.Client, mc *mcov1.MachineConfig) error {
+	_, err := GetMachineConfig(ctx, cli, mc.Name)
 	if errors.IsNotFound(err) {
 		klog.Infof("Create machine-config %q", mc.Name)
-		if err := r.Create(context.TODO(), mc); err != nil {
+		if err := cli.Create(ctx, mc); err != nil {
 			return err
 		}
 		return nil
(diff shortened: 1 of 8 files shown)


pkg/performanceprofile/controller/performanceprofile/resources/resources.go
```diff
@@ -1,21 +1,22 @@
...
-	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
-
 	nodev1 "k8s.io/api/node/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
yanirq, line 15: just a general thought,  can we avoid using controller-runtime in case we would like to eliminate it completely in the future ?

pkg/performanceprofile/controller/performanceprofile/resources/resources.go
```diff
@@ -27,20 +28,20 @@ func mergeMaps(src map[string]string, dst map[string]string) {
 
 // TODO: we should merge all create, get and delete methods
 
-func (r *PerformanceProfileReconciler) getMachineConfig(ctx context.Context, name string) (*mcov1.MachineConfig, error) {
+func GetMachineConfig(ctx context.Context, client client.Client, name string) (*mcov1.MachineConfig, error) {
```
yanirq, line 31: is there a way to use the client without passing it as an argument in every function ? (same for all functions below)

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -57,7 +60,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-const finalizer = "foreground-deletion"
+const (
```
yanirq, line 63: rebase change ?

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -437,36 +437,36 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	ctrRuntime, err := r.getContainerRuntimeName(ctx, instance)
+	profileMCP, err := getMachineConfigPoolByProfile(ctx, r.Client, instance)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("could not determine high-performance runtime class container-runtime for profile %q; %w", instance.Name, err)
-	}
-	klog.Infof("using %q as high-performance runtime class container-runtime for profile %q", ctrRuntime, instance.Name)
```
yanirq, line 453: do we want to drop this info msg ?

yanirq:
overall Looks good.  general questions are in comments.

pkg/performanceprofile/controller/performanceprofile/resources/resources.go
Tal-or: Not in the current architecture.
We're using operator SDK for developing PAO and the controller-runtime client is part of the operator SDK framework.

pkg/performanceprofile/controller/performanceprofile/resources/resources.go
Tal-or: We need to pass the client from outside because these function will be used for both hypershift and non-hypershift clusters. 
Each cluster will be implementing its own logic, hence it should control the type of client it wants to pass.

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: probably, i'll fix that

Tal-or: no, just added the parenthesis

pkg/performanceprofile/controller/performanceprofile/resources/resources.go
ffromani, line 31: It could be worth adding a shallow object to hold a context and a client. When the client code makes repeated calls, it always use the same set of parameters.

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -568,25 +572,25 @@ func (r *PerformanceProfileReconciler) applyComponents(profile *performancev2.Pe
 	}
 
 	// get mutated machine config
-	mcMutated, err := r.getMutatedMachineConfig(context.TODO(), components.MachineConfig)
+	mcMutated, err := resources.GetMutatedMachineConfig(context.TODO(), r.Client, components.MachineConfig)
```
ffromani, line 571: let's create a single context - TODO() is ok and let's reuse alongside the client for all the calls in the function

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -631,20 +635,20 @@ func (r *PerformanceProfileReconciler) applyComponents(profile *performancev2.Pe
 
 func (r *PerformanceProfileReconciler) deleteComponents(profile *performancev2.PerformanceProfile) error {
 	tunedName := components.GetComponentName(profile.Name, components.ProfileNamePerformance)
-	if err := r.deleteTuned(tunedName, components.NamespaceNodeTuningOperator); err != nil {
+	if err := resources.DeleteTuned(context.TODO(), r.Client, tunedName, components.NamespaceNodeTuningOperator); err != nil {
```
ffromani, line 634: let's create a single context - TODO() is ok and let's reuse alongside the client for all the calls in the function

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -296,17 +296,17 @@ func (r *PerformanceProfileReconciler) getInfraPartitioningMode() (pinning apico
 	return infra.Status.CPUPartitioning, nil
 }
 
-func (r *PerformanceProfileReconciler) getContainerRuntimeName(ctx context.Context, profile *performancev2.PerformanceProfile) (mcov1.ContainerRuntimeDefaultRuntime, error) {
+func getContainerRuntimeName(ctx context.Context, client client.Client, profile *performancev2.PerformanceProfile) (mcov1.ContainerRuntimeDefaultRuntime, error) {
```
ffromani: `cli client.Client` ? I'd avoid identifier aliasing unless there's really no other option

pkg/performanceprofile/controller/performanceprofile/status/status.go
```diff
@@ -21,18 +21,18 @@ import (
...
+	ConditionReasonMCPDegraded               = "MCPDegraded"
+	ConditionFailedGettingMCPStatus          = "GettingMCPStatusFailed"
+	ConditionKubeletFailed                   = "KubeletConfig failure"
+	ConditionFailedGettingKubeletStatus      = "GettingKubeletStatusFailed"
+	ConditionReasonTunedDegraded             = "TunedProfileDegraded"
+	ConditionFailedGettingTunedProfileStatus = "GettingTunedStatusFailed"
 )
 
-func (r *PerformanceProfileReconciler) updateStatus(profile *performancev2.PerformanceProfile, conditions []conditionsv1.Condition) error {
+func UpdateStatus(ctx context.Context, client client.Client, profile *performancev2.PerformanceProfile, conditions []conditionsv1.Condition) error {
```
ffromani, line 35: likewise, we may want to add a trivial `struct` to factor `context` and `client` instead of always repeating them

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -484,42 +485,42 @@ func (r *PerformanceProfileReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	}
 
 	// if conditions were not added due to machine config pool status change then set as available
 	if conditions == nil {
-		conditions = getAvailableConditions("")
+		conditions = status.GetAvailableConditions("")
 	}
 
-	if err := r.updateStatus(instance, conditions); err != nil {
+	if err := status.UpdateStatus(ctx, r.Client, instance, conditions); err != nil {
```
ffromani, line 523: nit: `status.Update` is a little nicer, if we can do this rename (not sure there are other implications e.g. 1000 usages)

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -568,25 +572,25 @@ func (r *PerformanceProfileReconciler) applyComponents(profile *performancev2.Pe
 	}
```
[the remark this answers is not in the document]
Tal-or: Thank you for the idea, I agree it's much cleaner but it basically mimic what we had before when the functions were methods of the reconciliation struct.

The problem with the `Manager` or any receiver is that it  should be platform aware, meaning it should know which client to pass (on hypershift we should pass a different client for some of the objects) so we'll have to have something like:
```
if InHypershift() {
// pass different client
}
```
for some of the functions.
I'll play with the implementation a little bit more, let's see what can be done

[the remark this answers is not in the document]
Tal-or: As we discussed offline on hypershift, the client is changing depending on the object,
runtimeclass for example, should be query from the hostedcluster while kubeletconfig from the management.
This is why we want to pass the client from the caller.

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: I'm planning on doing it on a later PR. (part 2 of hypershift adoption)

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: I'm planning on doing it on a later PR. (part 2 of hypershift adoption)

pkg/performanceprofile/controller/performanceprofile/resources/resources.go
Tal-or: https://github.com/openshift/cluster-node-tuning-operator/pull/1032#discussion_r1569036144

pkg/performanceprofile/controller/performanceprofile/status/status.go
Tal-or: Same constrains as here: https://github.com/openshift/cluster-node-tuning-operator/pull/1032#discussion_r1569036144

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: Done

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: Done

pkg/performanceprofile/controller/performanceprofile_controller.go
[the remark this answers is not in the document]
ffromani: fair enough for now

pkg/performanceprofile/controller/performanceprofile_controller.go
ffromani: ok, fair enough

pkg/performanceprofile/controller/performanceprofile_controller.go
ffromani: ack

MarSik (approved):
/lgtm

ffromani (commented):
/lgtm

PR #1037, opened 2024-04-16
CNF-6517: [Part 2] Hypershift PAO adoption

Tal-or:
Since the controller implementation diverged on Hypershift platform, we need to add an abstraction layer
in order to handle a different type of implementation.

In this PR we introduce 2 interfaces:
1. `components.Handler` - which provides an API for handling the different components the controller owns.
2. `status.Writer` which provides an API for reporting the different statuses by the controller.

This PR doesn't add any new functionality but implementing the old reconciliation loop using the new interfaces introduced above.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/handler/handler.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/handler/handler.go
@@ -1,0 +1,211 @@
+package handler
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serros "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	mcov1 "github.com/openshift/api/machineconfiguration/v1"
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/machineconfig"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/manifestset"
+	profileutil "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/resources"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/status"
+)
+
+var _ components.Handler = &handler{}
+
+type handler struct {
+	client.Client
+	scheme *runtime.Scheme
+}
+
+func NewHandler(cli client.Client, scheme *runtime.Scheme) components.Handler {
+	return &handler{Client: cli, scheme: scheme}
+}
+
+func (h *handler) Apply(ctx context.Context, obj client.Object, recorder record.EventRecorder, opts *components.Options) error {
+	profile, ok := obj.(*performancev2.PerformanceProfile)
+	if !ok {
+		return fmt.Errorf("wrong type conversion; want=PerformanceProfile got=%T", obj)
+	}
+
+	if profileutil.IsPaused(profile) {
+		klog.Infof("Ignoring reconcile loop for pause performance profile %s", profile.Name)
+		return nil
+	}
+
+	ctrRuntime, err := h.getContainerRuntimeName(ctx, profile, opts.ProfileMCP)
+	if err != nil {
+		return fmt.Errorf("could not determine high-performance runtime class container-runtime for profile %q; %w", profile.Name, err)
+	}
+	klog.Infof("using %q as high-performance runtime class container-runtime for profile %q", ctrRuntime, profile.Name)
+
+	opts.MachineConfig.DefaultRuntime = ctrRuntime
+	components, err := manifestset.GetNewComponents(profile, opts)
+	if err != nil {
+		return err
+	}
+	for _, componentObj := range components.ToObjects() {
+		if err := controllerutil.SetControllerReference(profile, componentObj, h.scheme); err != nil {
+			return err
+		}
+	}
+
+	// get mutated machine config
+	mcMutated, err := resources.GetMutatedMachineConfig(ctx, h.Client, components.MachineConfig)
+	if err != nil {
+		return err
+	}
+
+	// get mutated kubelet config
+	kcMutated, err := resources.GetMutatedKubeletConfig(ctx, h.Client, components.KubeletConfig)
+	if err != nil {
+		return err
+	}
+
+	// get mutated performance tuned
+	performanceTunedMutated, err := resources.GetMutatedTuned(ctx, h.Client, components.Tuned)
+	if err != nil {
+		return err
+	}
+
+	// get mutated RuntimeClass
+	runtimeClassMutated, err := resources.GetMutatedRuntimeClass(ctx, h.Client, components.RuntimeClass)
+	if err != nil {
+		return err
+	}
+
+	updated := mcMutated != nil ||
+		kcMutated != nil ||
+		performanceTunedMutated != nil ||
+		runtimeClassMutated != nil
+
+	// does not update any resources if it no changes to relevant objects and continue to the status update
+	if !updated {
+		return nil
+	}
+
+	if mcMutated != nil {
+		if err := resources.CreateOrUpdateMachineConfig(ctx, h.Client, mcMutated); err != nil {
+			return err
+		}
+	}
+
+	if performanceTunedMutated != nil {
+		if err := resources.CreateOrUpdateTuned(ctx, h.Client, performanceTunedMutated, profile.Name); err != nil {
+			return err
+		}
+	}
+
+	if kcMutated != nil {
+		if err := resources.CreateOrUpdateKubeletConfig(ctx, h.Client, kcMutated); err != nil {
+			return err
+		}
+	}
+
+	if runtimeClassMutated != nil {
+		if err := resources.CreateOrUpdateRuntimeClass(ctx, h.Client, runtimeClassMutated); err != nil {
+			return err
+		}
+	}
+	recorder.Eventf(profile, corev1.EventTypeNormal, "Creation succeeded", "Succeeded to create all components")
+	return nil
+}
+
+func (h *handler) Delete(ctx context.Context, profileName string) error {
+	tunedName := components.GetComponentName(profileName, components.ProfileNamePerformance)
+	if err := resources.DeleteTuned(ctx, h.Client, tunedName, components.NamespaceNodeTuningOperator); err != nil {
+		return err
+	}
+
+	name := components.GetComponentName(profileName, components.ComponentNamePrefix)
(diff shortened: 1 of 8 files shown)


Tal-or:
/test e2e-gcp-pao
```
Falling back to `oc adm inspect clusteroperators.v1.config.openshift.io` to collect basic cluster information.
  Warning: apps.openshift.io/v1 DeploymentConfig is deprecated in v4.14+, unavailable in v4.10000+
  error running backup collection: inspection completed with the errors occurred while gathering data:
      [skipping gathering secrets/support due to error: secrets "support" not found, skipping gathering endpoints/host-etcd-2 due to error: endpoints "host-etcd-2" not found, skipping gathering sharedconfigmaps.sharedresource.openshift.io due to error: the server doesn't have a resource type "sharedconfigmaps", skipping gathering sharedsecrets.sharedresource.openshift.io due to error: the server doesn't have a resource type "sharedsecrets"]
  error: gather did not start for pod must-gather-d7xjp: unable to pull image: ImagePullBackOff: Back-off pulling image "registry.build03.ci.openshift.org/ci-op-z8lw46nb/stable@sha256:addc9b013a4cbbe1067d652d032048e7b5f0c867174671d51cbf55f765ff35fc"
  [FAILED] in [BeforeSuite] - /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/6_mustgather_testing/test_suite_mustgather_test.go:42 @ 04/18/24 10:12:58.063
[BeforeSuite] [FAILED] [187.436 seconds]
[BeforeSuite] 
/go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/6_mustgather_testing/test_suite_mustgather_test.go:20
  [FAILED] Unexpected error:
      <*exec.ExitError | 0xc00014c060>: 
      exit status 1
      {
          ProcessState: {
              pid: 13528,
              status: 256,
              rusage: {
                  Utime: {Sec: 39, Usec: 823355},
                  Stime: {Sec: 9, Usec: 256492},
                  Maxrss: 193088,
                  Ixrss: 0,
                  Idrss: 0,
                  Isrss: 0,
                  Minflt: 459497,
                  Majflt: 2,
                  Nswap: 0,
                  Inblock: 3808,
                  Oublock: 214320,
                  Msgsnd: 0,
                  Msgrcv: 0,
                  Nsignals: 0,
                  Nvcsw: 209038,
                  Nivcsw: 3664,
              },
          },
          Stderr: nil,
      }
  occurred
  In [BeforeSuite] at: /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/6_mustgather_testing/test_suite_mustgather_test.go:42 @ 04/18/24 10:12:58.063
```

Tal-or:
Depends on #1032

pkg/performanceprofile/controller/performanceprofile/components/handler/handler.go
```diff
@@ -0,0 +1,213 @@
...
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/machineconfig"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/manifestset"
+	profileutil "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/resources"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/status"
+)
+
+var _ components.Handler = &handler{}
```
rbaturov, line 29: Just wondering, why is this needed?

pkg/performanceprofile/controller/performanceprofile/status/status.go
```diff
@@ -32,6 +32,13 @@ const (
 	ConditionFailedGettingTunedProfileStatus = "GettingTunedStatusFailed"
 )
 
+type Writer interface {
+	// Update updates the status reported by the controller
+	Update(ctx context.Context, object client.Object, conditions []conditionsv1.Condition) error
+	// UpdateOwnedConditions updates the conditions of the components owned by the controller
+	UpdateOwnedConditions(ctx context.Context, object client.Object, profileMCP *mcov1.MachineConfigPool) error
```
rbaturov, line 38: In the implementation for `UpdateOwnedConditions` in hypershift, are we counting on calling this function with an empty  `*mcov1.MachineConfigPool` (we said no MCP on hypershift IIRC)?

pkg/performanceprofile/controller/performanceprofile/components/handler/handler.go
ffromani: it's because https://go.dev/doc/faq#guarantee_satisfies_interface

pkg/performanceprofile/controller/performanceprofile/components/handler/handler.go
```diff
@@ -0,0 +1,213 @@
...
+type handler struct {
+	client.Client
+	scheme *runtime.Scheme
+}
+
+func NewHandler(cli client.Client, scheme *runtime.Scheme) components.Handler {
+	return &handler{Client: cli, scheme: scheme}
+}
+
+func (h *handler) Apply(ctx context.Context, obj client.Object, recorder record.EventRecorder, opts *components.Options) (*reconcile.Result, error) {
```
ffromani, line 40: it seems to me that we either return `nil, err` or `reconcile.Result{}, nil`. If so I'd remove the `reconcile.Result` return value which unnecessarily ties the API to the reconcile loop. We can perhaps just return `error` and make logic in the call site. It seems to me just returning `error` gives the same amount of information.

pkg/performanceprofile/controller/performanceprofile/status/status.go
ffromani: this is a good point indeed

pkg/performanceprofile/controller/performanceprofile/status/writer.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (w *writer) Update(ctx context.Context, object client.Object, conditions []conditionsv1.Condition) error {
+	profile, ok := object.(*performancev2.PerformanceProfile)
+	if !ok {
+		return fmt.Errorf("wrong type conversion; want=PerformanceProfile got=%T", object)
+	}
+	return w.update(ctx, profile, conditions)
+}
+
+func (w *writer) UpdateOwnedConditions(ctx context.Context, object client.Object, profileMCP *mcov1.MachineConfigPool) error {
```
ffromani, line 33: how hard it is (perhaps in a later PR) to fetch the correct MCP inside this function instead of requiring as argument? Reason I'm asking is to look for consistent API between hypershift and non-hypershift. Per previous comments, looks like this argument should always be `nil` in hypershift, which suggest the API is not generic enough.

pkg/performanceprofile/controller/performanceprofile/components/handler/handler.go
```diff
@@ -0,0 +1,213 @@
...
+	profile, ok := obj.(*performancev2.PerformanceProfile)
```
ffromani, line 41: I guess in the hypershift implementation we will need a `ConfigMap` object and we will bail out if it is not a `ConfigMap` object, right?

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -683,6 +455,34 @@ func hasFinalizer(profile *performancev2.PerformanceProfile, finalizer string) b
 	return false
 }
 
+func (r *PerformanceProfileReconciler) getAndValidateMCP(ctx context.Context, instance client.Object) (*mcov1.MachineConfigPool, *reconcile.Result, error) {
```
ffromani, line 457: can we push this function (or an amended form of) inside `err = r.StatusWriter.UpdateOwnedConditions(ctx, instance, profileMCP)`, so we can remove the potentially problematic `profileMCP` argument?

ffromani (commented):
When adding interfaces or generalizing APIs, a painful lesson over time is that we almost always benefit from seeing at least two implementation of the same interface, to be reasonnably confident the abstraction is correct and generic enough.

Here we see just one, but somehow the proposed interfaces click right. So I'm positive about moving on, with relatively minor suggestions inside

But still, let me ask: there is any chance to see a sketch of how the hypershift implementation of the interfaces introduced here could look like?

pkg/performanceprofile/controller/performanceprofile/status/status.go
Tal-or: Another option is to get the MCP from within the function

Tal-or:
@ffromani 
> But still, let me ask: there is any chance to see a sketch of how the hypershift implementation of the interfaces introduced here could look like?

I'm sorry if it's going to be too much of hand waving but i'll try to point out the general flow:
1. The object we expect to have is always a `ConfigMap` and not a `PerformanceProfile` as we see here.
2. No MCPs so wherever MCP requested we should pass `nil`
3. For checking the existence of object the operator looks for  for a `ConfigMap`s inside its namespace which encapsulated the owned objects i.e `MachineConfig`, `KubeletConfig`. `RuntimeClass` object is a different case and the operator checks for its existence on the hostedcluster.
4. The apply flows goes similar as described above. the operator creates the objects inside its namespace and encapsulate them in a `ConfigMap`. 

Another sketch can be seen in the old PR here: 
https://github.com/openshift/cluster-node-tuning-operator/pull/484

pkg/performanceprofile/controller/performanceprofile/components/handler/handler.go
Tal-or: Yes. Exactly

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: Sure we do, sounds like a good idea

pkg/performanceprofile/controller/performanceprofile/status/status.go
Tal-or: Done

pkg/performanceprofile/controller/performanceprofile/status/writer.go
Tal-or: Done

pkg/performanceprofile/controller/performanceprofile_controller.go
Tal-or: I removed the `ProfileMCP` from the `UpdateOwnedConditions` as requested. Good idea, thank you

pkg/performanceprofile/controller/performanceprofile/components/handler/handler.go
Tal-or: OK you were right, it doesn't matter, I thought `ctrl.Result{}` and  `reconcile.Result{}` are different, but they are the same

Tal-or: Done

ffromani (commented):
/approve
/lgtm

thanks for the updates

ffromani:
> @ffromani
> 
> > But still, let me ask: there is any chance to see a sketch of how the hypershift implementation of the interfaces introduced here could look like?
> 
> I'm sorry if it's going to be too much of hand waving but i'll try to point out the general flow:
> 
>     1. The object we expect to have is always a `ConfigMap` and not a `PerformanceProfile` as we see here.
> 
>     2. No MCPs so wherever MCP requested we should pass `nil`
> 
>     3. For checking the existence of object the operator looks for  for a `ConfigMap`s inside its namespace which encapsulated the owned objects i.e `MachineConfig`, `KubeletConfig`. `RuntimeClass` object is a different case and the operator checks for its existence on the hostedcluster.
> 
>     4. The apply flows goes similar as described above. the operator creates the objects inside its namespace and encapsulate them in a `ConfigMap`.
> 
> 
> Another sketch can be seen in the old PR here: #484

good enough for now, let's move on

Tal-or:
/retest

infra issue

ffromani:
/retest-required

it seems the hypershift lane is flaky (again)

PR #1039, opened 2024-04-17
NO-JIRA: pao:controller: remove depracated function

Tal-or:
`deleteDeprecatedComponents` function was added in older releases to overcome this issue:
https://github.com/openshift-kni/performance-addon-operators/pull/538

This is no longer relevant and can be removed since it was intented to fix an upgrade issue from an old version.


ffromani:
/approve

I agree this function is both deprecated and can be removed

PR #1041, opened 2024-04-18
CNF-11985: Enhance latency cnf-tests UX

rbaturov:
Currently, after executing the latency test with OSLAT/cyclictest, if latency exceeds the maximum latency threshold, we display the first exceeding measured latency rather than the maximum latency measured. It is more informative to display the maximum latency measured, therefore we made this change.

### DIFF
--- a/test/e2e/performanceprofile/functests/4_latency/latency.go
+++ b/test/e2e/performanceprofile/functests/4_latency/latency.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -463,17 +450,26 @@ func createLatencyTestPod(testPod *corev1.Pod) {
...
+	Expect(len(output)).To(Equal(2))
 
-	return latencies[1]
+	var latencies []int
+	for _, lat := range strings.Split(output[1], " ") {
+		if lat == "" {
+			continue
+		}
+		curr, err := strconv.Atoi(lat)
+		Expect(err).ToNot(HaveOccurred())
```
Tal-or, line 469: In case of asserting on helper function, please add `GinkgoHelper()` at the beginning of the function

Tal-or (commented):
I like this idea.
Did you test it to make sure it works? because AFAIR we do not have a CI for the latency-testing tools

rbaturov:
> I like this idea. Did you test it to make sure it works? because AFAIR we do not have a CI for the latency-testing tools

Yes I have tested it.

MarSik (approved):
/lgtm

PR #1045, opened 2024-04-24
OCPBUGS-26401: TuneD prior to kubelet in one-shot mode

jmencak:
TuneD prior to kubelet in one-shot mode

Motivation.  In case of node reboots, all pods restart again in random order.
Since there's no control on pod restart order, it is possible that tuned
pods will start after the workload pods.  This means the workload pods can
start with partial tuning which can result in degraded performance or even
cause the workload pods crash.

This change is intended mostly for Telco use cases where PerformanceProfiles
are used.  To minimize the risk of affecting other deployments by backporting
this code, TuneD will start prior to kubelet only when PerformanceProfiles
are used.

Changes:
  * Switch to /var/lib/ocp-tuned as the home directory for the operand
    (TuneD) pod on the host.  This directory now holds configuration for the
    TuneD daemon both running prior to kubelet and in the pod itself, notably
    TuneD daemon profiles and the recommended profile.
  * Use the new tuned-main.conf option "profile_dirs" to extract TuneD profiles
    to /var/lib/ocp-tuned/profiles.  This new option also allowed simplifying the
    algorithm for change detection in profilesSync() when extracting TuneD profiles.
  * Move a lot of the logic of creating the environment for TuneD to the operand
    code, so that no external script are needed for starting the operand.
  * Add ocp-tuned-one-shot.service systemd unit when using a PerformanceProfile.
  * Introduce a timeout to running TuneD in one-shot mode so that we don't block
    the kubelet indefinitely.
  * Shorten the constants names from openshift* to ocp*

Resolves: OCPBUGS-26401

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -89,12 +89,13 @@ const (
 )
 
 const (
-	systemdServiceIRQBalance  = "irqbalance.service"
-	systemdServiceKubelet     = "kubelet.service"
-	systemdServiceCrio        = "crio.service"
-	systemdServiceTypeOneshot = "oneshot"
-	systemdTargetMultiUser    = "multi-user.target"
-	systemdTrue               = "true"
+	systemdServiceIRQBalance   = "irqbalance.service"
+	systemdServiceKubelet      = "kubelet.service"
+	systemdServiceCrio         = "crio.service"
+	systemdServiceTunedOneShot = "ocp-tuned-one-shot.service"
+	systemdServiceTypeOneshot  = "oneshot"
+	systemdTargetMultiUser     = "multi-user.target"
+	systemdTrue                = "true"
 )
 
 const (
@@ -334,6 +335,14 @@ func getIgnitionConfig(profile *performancev2.PerformanceProfile, opts *componen
 		Name:     getSystemdService(clearIRQBalanceBannedCPUs),
 	})
 
+	// add ocp-tuned-one-shot service
+	ocpTunedOneShotServiceContent := getTunedOneShotSystemdUnit()
+	ignitionConfig.Systemd.Units = append(ignitionConfig.Systemd.Units, igntypes.Unit{
+		Contents: &ocpTunedOneShotServiceContent,
+		Enabled:  ptr.To(true),
+		Name:     systemdServiceTunedOneShot,
+	})
+
 	if ok, ovsSliceName := MoveOvsIntoOwnSlice(); ok {
 		// Create the OVS slice that will lift the cpu restrictions for better kernel networking performance
 		// This is technically not necessary as systemd is smart enough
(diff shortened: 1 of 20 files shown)


jmencak:
This PR is ready for review.  I still want to do more testing on IBI/IBU, but regular upgrades seem to work.
/cc @yanirq

jmencak:
/hold
to decide with @ffromani if this should merge after his changes.

pkg/tuned/controller.go
```diff
@@ -445,6 +465,75 @@ func providerSync(provider string) (bool, error) {
...
+	klog.Infof("writing %v", ocpTunedImageEnv)
+
+	f, err := os.Create(ocpTunedImageEnv)
+	if err != nil {
+		return fmt.Errorf("failed to create file %q: %v", ocpTunedImageEnv, err)
+	}
+	defer f.Close()
+	if _, err = f.WriteString(fmt.Sprintf("NTO_IMAGE=%s\n", os.Getenv("CLUSTER_NODE_TUNED_IMAGE"))); err != nil {
+		return fmt.Errorf("failed to write file %q: %v", ocpTunedImageEnv, err)
+	}
```
ffromani, line 532: os.Writefile?

pkg/tuned/run.go
```diff
@@ -16,26 +16,31 @@ package tuned
...
 	if debug {
 		args = append(args, "--debug")
 	}
-	return exec.Command("/usr/sbin/tuned", args...)
+	return "/usr/sbin/tuned", args
+}
+
+func TunedCreateCmd(command string, args []string) *exec.Cmd {
+	return exec.Command(command, args...)
 }
```
ffromani, line 38: I'd go as far as remove this function, is now so shallow it's value is debatable

pkg/util/os.go
```diff
@@ -1,16 +1,71 @@
...
+	if _, err := os.Stat(file); os.IsNotExist(err) {
+		file, err := os.Create(file)
 		if err != nil {
 			return err
 		}
+		defer file.Close()
+	}
+
+	return nil
+}
```
ffromani, line 19: this is still racy though. Do we need to take reasonnable steps to be race free or this is "just" defensive programming?

pkg/util/os.go
```diff
@@ -1,16 +1,71 @@
...
+
+// Create a directory 'dir' including its parents.  Returns nil if directory already exists.
+func Mkdir(dir string) error {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return os.MkdirAll(dir, os.ModePerm)
+	}
+
+	return nil
+}
```
ffromani, line 38: does `os.MkdirAll` fail if (part of) the path given as argument already exists? It should not: https://go.dev/play/p/9jO8toFtYgf (hence `util.Mkdir` seems redundant?)

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -544,6 +553,60 @@ func getRPSUnitOptions(rpsMask string) []*unit.UnitOption {
 	}
 }
 
+func getTunedOneShotSystemdUnit() string {
```
ffromani, line 556: I'd prefer have this as loadable (embeddable?) asset rather than a string constant. WDYT? What do other reviewers think?

MarSik: The other units are created programatically even. I do not care much, but an embedded asset it indeed a nice idea.

pkg/util/os.go
MarSik: Correct: "If path is already a directory, MkdirAll does nothing and returns nil. "

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
jmencak: Creating the unit programatically doesn't make sense in this case in my view.  However, making this a loadable asset (if you mean from a new file within a repo) would make sense to me.  Any suggestions as to where to keep the asset?

pkg/util/os.go
jmencak: The aim here is to have a file (even empty) unless it exists.  If it exists, do not overwrite the file. Not sure I follow. What are you suggesting?

pkg/util/os.go
jmencak: Good point, I'll remove the function.

pkg/util/os.go
ffromani: Thing is, a file can be created by a third party in between `Stat` and `Create`. Nothing in this function prevents it. If no other actor is expected to create these files we manage with `util.Create`, and NTO only owns them, then it seems easier and safer to tolerate overwrites?  And preventing it may turn out not trivial. I'm trying to assess why do we need this level of protection.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
ffromani: `assets/performanceprofile/configs` works. In the future we should probably further split the systemd unit templates in their own directory.

pkg/util/os.go
jmencak: I see what you mean now. Thank you. Actually, let me remove the function completely. It was used in my previous `TuneD` prior kubelet PoC which wasn't running in one-shot mode.  I'll also remove the `Copy()` function from the `util`.  Neither of them are used in the current code.  I'm not sure why it wasn't picked up by the linter.

jmencak:
I believe/hope I've addressed all reviewer's concerns.
/hold cancel

pkg/tuned/cmd/operand/cmd.go
```diff
@@ -56,6 +58,7 @@ func NewTunedCommand() *cobra.Command {
 
 func (t *tunedOpts) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&t.inCluster, "in-cluster", true, "In-cluster operand run.")
+	fs.BoolVar(&t.oneShot, "one-shot", false, "Run TuneD in one-shot mode.")
```
ffromani, line 61: non-blocking: should `oneShot` imply (and override) `inCluster`?
IOW, should we document and enforce that if `oneShot=true` then `inCluster=false`? This will allow to simplify our logic below.

pkg/tuned/controller.go
```diff
@@ -384,53 +389,68 @@ func profilesSync(profiles []tunedv1.TunedProfile, recommendedProfile string) (b
...
-		return fmt.Errorf("failed to create cloud provider name file %q: %v", openshiftTunedProvider, err)
+		return fmt.Errorf("failed to create cloud provider name file %q: %v", ocpTunedProvider, err)
 	}
 	defer f.Close()
 	if _, err = f.WriteString(provider); err != nil {
-		return fmt.Errorf("failed to write cloud provider name file %q: %v", openshiftTunedProvider, err)
+		return fmt.Errorf("failed to write cloud provider name file %q: %v", ocpTunedProvider, err)
 	}
 
 	return nil
```
ffromani, line 445: non-blocking:  this seems another case of function which can be almost completely collapsed into a `os.WriteFile` call.

pkg/tuned/controller.go
```diff
@@ -445,9 +465,75 @@ func providerSync(provider string) (bool, error) {
...
+	// Delete the container's home directory.
+	if err := util.Delete(ocpTunedHome); err != nil {
+		return fmt.Errorf("failed to delete: %q: %v", ocpTunedHome, err)
+	}
+
+	if err := util.Symlink(ocpTunedHomeHost, ocpTunedHome); err != nil {
+		return fmt.Errorf("failed to link %q -> %q: %v", ocpTunedHome, ocpTunedHomeHost, err)
+	}
+
+	err := os.Chdir(ocpTunedHome)
```
ffromani, line 489: nit: `return os.Chdir(ocpTunedHome)`?

ffromani (commented):
/lgtm
/hold

holding to let other reviewers chime in, feel free to remove

pkg/tuned/controller.go
jmencak: Yes, I know. I didn't want to change it yesterday because I believe that should be part of a cleanup and not this PR. Adding on my TODO list.

pkg/tuned/cmd/operand/cmd.go
jmencak: I guess, we could do that.  Not something I'm too worried about though.

pkg/tuned/controller.go
jmencak: Absolutely, the current code is ugly. Please feel free to change in your followup PR. If not, adding to my TODO list as a cleanup.

PR #1047, opened 2024-04-26
NO-JIRA: perfprof: e2e: account for node base load

shajmakh:
Currently the test would fail when tested on machines with exactly isolated cpus equal to the requested cpus by the pod. This is because the worker nodes has already other pods running on them and thus they are using already a portion of the isolated cpus of the node. Scheduling a pod requesting all allocatable cpus would fail due to insufficient cpus on node.

Skip the test if the testing environment is not suitable for the test's requirements.


shajmakh:
blocking: https://github.com/openshift/release/pull/47439

MarSik (approved):
/lgtm

PR #1049, opened 2024-04-29
OCPBUGS-32978: fix extra-reboot on upgrade with paused mcp worker

vitus133:
This PR fixes an extra-reboot when upgrading OpenShift with paused
MCP "worker", for example during EUS-to-EUS upgrade. The extra-reboot
occurred because PerformanceProfile controller was reconciling against
rendered MC appearing in the MCP status. While this MC reflects current
MCP conditions, it does not reflect the latest planned state when the
MCP is paused.
Unpausing the MCP lead to one reboot when applying the target MC, and
one additional reboot after the performance profile was reconciled
against it.
This change eliminates this additional reboot because PerformanceProfile
is now reconciling against the latest planned MC before theMCP is
unpaused

/cc @MarSik @yanirq

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile_controller.go
+++ b/pkg/performanceprofile/controller/performanceprofile_controller.go
@@ -19,10 +19,11 @@ package controller
 import (
 	"context"
 	"fmt"
-	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	"os"
 	"reflect"
 	"time"
+
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 
 	apiconfigv1 "github.com/openshift/api/config/v1"
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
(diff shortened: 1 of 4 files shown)


MarSik:
The patch looks good to me, but it might be a good idea to describe the mechanism (why it works) in the commit message or in the code.

MarSik:
And you should put pretty much the same content to the release note.

MarSik (approved):
/lgtm

pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -591,7 +592,7 @@ func (r *PerformanceProfileReconciler) reconcileCgroupsV1(ctx context.Context, i
 	}
 
 	// (TODO) This code can be removed in the future when the cgroupsv2 is supported
-	currentMC, err := r.getCurrentMachineConfigByMCP(ctx, profileMCP)
```
yanirq, line 594: we can remove the getCurrentMachineConfigByMCP function if its not used at all (linter complaints about it anyway)

vitus133: done

PR #1050, opened 2024-04-29
OCPBUGS-33112: E2E: Add automation to check default cpuset from cpu manager state file

mrniranjan:
Add new automation to verify default cpuset from the cpu manager state file pre and post kubelet restart Automates https://issues.redhat.com/browse/OCPBUGS-24366

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -41,6 +41,8 @@ import (
 var workerRTNode *corev1.Node
 var profile *performancev2.PerformanceProfile
 
+const restartCooldownTime = 1 * time.Minute
+
 var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 	var (
 		balanceIsolated          bool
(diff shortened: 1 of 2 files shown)


ffromani:
looks reasonnable and automates the bug per subject, but anything which consumes the state files directly is tech debt and we need to treat is as such.

mrniranjan:
@ffromani i have modified the test case to check with a gu pod which was the original issue.

ffromani:
> @ffromani i have modified the test case to check with a gu pod which was the original issue.

thanks, but the main issue remains. I guess we don't have better approaches handy, though. We need to start doing these checks using supported APIs, like the podresources API

mrniranjan:
> > @ffromani i have modified the test case to check with a gu pod which was the original issue.
> 
> thanks, but the main issue remains. I guess we don't have better approaches handy, though. We need to start doing these checks using supported APIs, like the podresources API

@ffromani okay so if i understand correctly we can use podResources API to fetch defaultCpuSet instead of reading the state file ? Is there any API documentation that i can refer to ?

ffromani (commented):
/approve
/lgtm

migrating to podresources API will need nontrivial effort. For now we prefer to accept the tech debt in order to get the required coverage

PR #1056, opened 2024-05-09
NO-JIRA: E2E: modify test use HaveKey matchter instead of ContainElement

mrniranjan:
Using ContainElement causes test to fail as it checks the map contains an Element with node.Name , however in the case map[string]map[string]int, where the node.Name contains hostname:{dev:N, dev:N}, the match fails.

HaveKey specifically checks if the map has the Key with node.Name


ffromani (commented):
/approve
/lgtm

PR #1058, opened 2024-05-10
Unify OKD and OCP Dockerfiles

jupierce:


jmencak:
Thank you for the PR, Justin. I probably shouldn't be doing reviews Friday evening, so only a few initial thoughts.

* I believe this approach is better than the original PR mentioned above.
* This PR introduces code duplication (potential config drift in the future) between `dockerfile_install_support.sh` and `Dockerfile`.  Can we also adjust `Dockerfile` to use `dockerfile_install_support.sh` ?
* The `dockerfile_install_support.sh` should probably go to the `/hack` directory.

I'll try to do a more thorough review next week. PTO Monday.  Thank you.

.gitignore
```diff
@@ -8,3 +8,5 @@ tmp
 /*.sh
 /TODO
 pkg/manifests/bindata.go
+
+!hack/dockerfile_install_support.sh
```
jmencak, line 12: Is this still needed since we moved the script to `hack/` directory?

Dockerfile
```diff
@@ -4,43 +4,24 @@ COPY . .
...
-    mkdir -p /etc/grub.d/ /boot /var/lib/ocp-tuned && \
-    dnf install --setopt=tsflags=nodocs -y $INSTALL_PKGS && \
-    rpm -V $INSTALL_PKGS && \
-    dnf --setopt=tsflags=nodocs -y install /root/rpms/*.rpm && \
-    rm -rf /etc/tuned/recommend.d && \
-    echo auto > /etc/tuned/profile_mode && \
-    sed -Ei 's|^#?\s*enable_unix_socket\s*=.*$|enable_unix_socket = 1|;s|^#?\s*rollback\s*=.*$|rollback = not_on_exit|;s|^#?\s*profile_dirs\s*=.*$|profile_dirs = /var/lib/ocp-tuned/profiles,/usr/lib/tuned|' \
-      /etc/tuned/tuned-main.conf && \
-    touch /etc/sysctl.conf && \
+RUN touch /etc/sysctl.conf && \
```
jmencak, line 24: Would it make sense to move this to `dockerfile_install_support.sh`?

Dockerfile.rhel9
```diff
@@ -7,25 +7,21 @@ FROM registry.ci.openshift.org/ocp/4.16:base-rhel9
...
-      nmap-ncat procps-ng pciutils" && \
-    mkdir -p /etc/grub.d/ /boot /var/lib/ocp-tuned && \
-    dnf install --setopt=tsflags=nodocs -y $INSTALL_PKGS && \
-    rm -rf /etc/tuned/recommend.d && \
-    echo auto > /etc/tuned/profile_mode && \
-    sed -Ei 's|^#?\s*enable_unix_socket\s*=.*$|enable_unix_socket = 1|;s|^#?\s*rollback\s*=.*$|rollback = not_on_exit|;s|^#?\s*profile_dirs\s*=.*$|profile_dirs = /var/lib/ocp-tuned/profiles,/usr/lib/tuned|' \
-      /etc/tuned/tuned-main.conf && \
-    touch /etc/sysctl.conf && \
+
+RUN touch /etc/sysctl.conf && \
```
jmencak, line 24: As above, would it make sense to move this to `dockerfile_install_support.sh`?

jmencak (approved):
Thank you for the changes. `podman build --squash-all` helps a lot with reducing the resulting images sizes.

Only a couple of nits/questions left from my side.  There is quite a lot of duplication in the if/else clauses in `dockerfile_install_support.sh`, but we/I can clean this up after the fact.  Thank you for the PR.
/approve
Almost looks good to me.

hack/dockerfile_install_support.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+set -euo pipefail
+set -o xtrace
+
+source /etc/os-release
+if [[ "${ID}" == "centos" ]]; then
```
LorbusChris, line 7: Maybe this should be centos OR fedora, in case we bring back OKD/FCOS builds

LorbusChris: actually, nevermind, this wouldn't work, as OKD/FCOS payload images would still be using the centos builds.

Maybe this should instead make use of the `ARG TAGS=ocp`, which would be overriden in the OKD builds with TAGS=scos for OKD/SCOS and TAGS=fcos for OKD/FCOS.

LorbusChris: Anyway, not a blocker.

.gitignore
jupierce: Not needed any longer. Fixed in next push.

Dockerfile.rhel9
jupierce: Moved to support script.

jupierce:
As this is one of the last images we need to get OKD building again and the changes should be a no-op to OCP.

jmencak:
Thank you for the changes, Justin.  I thought we could get rid of the second "RUN" completely in both Dockerfiles, i.e. moving:

```
dnf clean all
rm -rf /var/cache/yum ~/patches /root/rpms
useradd -r -u 499 cluster-node-tuning-operator
```
too.  But I don't think this is a blocker but we can clean this up later on.

So
/lgtm

PR #1060, opened 2024-05-14
NO-JIRA: E2E: Remove explicitly setting Realtime kernel and realtime hint

mrniranjan:
We don't need to explicitly enable realtime kernel and realtime hint, we use what ever the performance profile already has

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/memorymanager.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/memorymanager.go
@@ -109,15 +109,6 @@ var _ = Describe("[rfe_id: 43186][memorymanager] Memorymanager feature", func() 
 				},
 			}
 			profile.Spec.HugePages = requiredHugepages
-
-			if !*profile.Spec.RealTimeKernel.Enabled {
-				profile.Spec.RealTimeKernel = &performancev2.RealTimeKernel{
-					Enabled: pointer.Bool(true),
-				}
-			}
-			profile.Spec.WorkloadHints = &performancev2.WorkloadHints{
-				RealTime: pointer.Bool(true),
-			}
 			profile.Spec.CPU = &performancev2.CPU{
 				Reserved: &reservedSet,
 				Isolated: &isolatedSet,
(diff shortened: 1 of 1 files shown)


ffromani (commented):
is this a code cleanup or are we trying to fix a bug?

In general my own preference is to make the test very explicit about the preconditions and don't depend on the current defaults (unless the test is checking the defaults) because depending on the defaults creates a hidden dependency

mrniranjan:
> is this a code cleanup or are we trying to fix a bug?
> 
> In general my own preference is to make the test very explicit about the preconditions and don't depend on the current defaults (unless the test is checking the defaults) because depending on the defaults creates a hidden dependency

This is mostly a code clean up where we want to rely on the what ever the performance  profile is providing and not set any explicit settings . This is required because on 4.16 with increase in test matrix, where in Down stream we are running tests where RT, Non RT,  So when running tests in Non-RT environment, if the test code explicitly sets Realtime kernel then we are not testing in Non RT environment,  So to be true to the environment, we are removing explicitly setting Realtime kernel and realtime hints and go with what ever the existing performance profile is setting up.

yanirq:
Can we miss test cases in u/s then ? 
What about other occurrences where we explicitly set Realtime kernel ? 
one example from several existing tests: 
https://github.com/openshift/cluster-node-tuning-operator/blob/f37f16f2df5884645a5b97d00ca7dfe9f3881bcc/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go#L144

mrniranjan:
> Can we miss test cases in u/s then ? What about other occurrences where we explicitly set Realtime kernel ? one example from several existing tests:
> 
> https://github.com/openshift/cluster-node-tuning-operator/blob/f37f16f2df5884645a5b97d00ca7dfe9f3881bcc/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go#L144

So we need to patch these too , In U/S  in config.go we are already setting realTimeKernel to true so we can remove setting explicitly here i think.

yanirq:
> > Can we miss test cases in u/s then ? What about other occurrences where we explicitly set Realtime kernel ? one example from several existing tests:
> > https://github.com/openshift/cluster-node-tuning-operator/blob/f37f16f2df5884645a5b97d00ca7dfe9f3881bcc/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go#L144
> 
> So we need to patch these too , In U/S in config.go we are already setting realTimeKernel to true so we can remove setting explicitly here i think.

We might but there are some tests where we set the value to true OR false (sometimes in the same suite) .   We need to make sure we are consistent and also have the test code clear enough for future contributions

Tal-or:
@mrniranjan I think this PR is ready, feel free to remove hold

PR #1061, opened 2024-05-17
OCPBUGS-33682: right-hand-side profile_dirs take precedence

jmencak:
The new TuneD option profile_dirs defines custom TuneD directories.  Fix the order of the directories so that the system TuneD directory (/usr/lib/tuned) is the first one (searched last), and the custom TuneD directory (/var/lib/ocp-tuned/profiles) is the last one (searched first).

Resolves: OCPBUGS-33682


jmencak:
Tested with 
```
kind: Tuned
metadata:
  name: provider-aws
spec:
  profile:
  - data: |
      [main]
      summary=OpenShift profile to test Cloud provider-specific functionality
      [sysctl]
      kernel.pty.max=8192
    name: provider-aws
      
```

Works fine.
/cc @liqcui

liqcui:
/lgtm
Pre-merge testing passed for cluster-provider feature, thank you for quickly resolving this problem.

PR #1064, opened 2024-05-22
OCPBUGS-34720: Add tier0 and tier1 tags to non reboot test cases

mrniranjan:
This PR categorizes different tests under  0_config and 1_performance as tier0, tier1 using ginkgo labels. 

This tagging helps in executing set of test cases filtered using labels defined under utils/labels.go 

Definition of these tiers are found under utils/labels.go


yanirq:
@mrniranjan  can you please add some info in the PR description for future references ?

PR #1067, opened 2024-05-23
NO-JIRA: PPC: Add relatedObjects status field

rbaturov:
Add relatedObjects support to better integrate with must-gather

relatedObjects is a list of objects that are "interesting" or related to this operator.
must-gather uses this field to collect data from these object when data from the operator is collected.

**Note**: this is a cherry-pick from this [abandoned PR](https://github.com/openshift/cluster-node-tuning-operator/pull/819)

### DIFF
--- a/manifests/20-performance-profile.crd.yaml
+++ b/manifests/20-performance-profile.crd.yaml
@@ -761,6 +761,32 @@ spec:
                   - type
                   type: object
                 type: array
+              relatedObjects:
+                description: 'relatedObjects is a list of objects that are "interesting"
+                  or related to this operator.  Common uses are: 1. the detailed resource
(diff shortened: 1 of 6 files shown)


ffromani:
conceptually OK, but let's make sure to have good e2e coverage because we hit issues in the past when doing something similar on NROP.
Most notably, the MG collections _halts_ mid way if we try to collect objects which don't exist in the cluster. This has to be dealt with.

rbaturov:
> conceptually OK, but let's make sure to have good e2e coverage because we hit issues in the past when doing something similar on NROP. Most notably, the MG collections _halts_ mid way if we try to collect objects which don't exist in the cluster. This has to be dealt with.

Meaning we should verify whether we have a test under the `6_mustgather_testing` that will raise an error if what you described is occurring?

PR #1072, opened 2024-06-05
OCPBUGS-14026: release-4.14: check agetty process moved under cpuset cgroup

SargunNarula:


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -304,6 +304,30 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 				Expect(processesFound).To(BeEmpty(), "The node %s has the following processes on the root cgroup: %v", node.Name, processesFound)
 			}
 		})
+
+		It("Check for agetty process system.slice cgroup", func() {
+			for _, node := range workerRTNodes {
+				agettyPidCommand := "pgrep agetty"
+				agettyPidOutput, err := nodes.ExecCommandOnNode([]string{"/bin/bash", "-c", agettyPidCommand}, &node)
+				Expect(err).ToNot(HaveOccurred())
+				agettyPids := strings.Fields(agettyPidOutput)
```
shajmakh, line 313: let's expect  agettyPidOutput not to be empty before checking the pids

SargunNarula: Resolved

shajmakh:
Thanks for this!
/lgtm
/hold 
leaving room for others review

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -304,6 +304,31 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 				Expect(processesFound).To(BeEmpty(), "The node %s has the following processes on the root cgroup: %v", node.Name, processesFound)
 			}
 		})
+
+		It("Check for agetty process system.slice cgroup", func() {
```
shajmakh, line 308: let's add test is that covers this bug

ffromani: test_id? if so I agree

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -304,6 +304,31 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+			for _, node := range workerRTNodes {
+				agettyPidCommand := "pgrep agetty"
+				agettyPidOutput, err := nodes.ExecCommandOnNode([]string{"/bin/bash", "-c", agettyPidCommand}, &node)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(agettyPidOutput).ToNot(BeEmpty())
+				agettyPids := strings.Fields(agettyPidOutput)
+				Expect(len(agettyPids)).To(BeNumerically(">", 0), "No agetty PIDs found on node %s", node.Name)
```
ffromani, line 315: also: ```Expect(agettyPids).ToNot(BeEmpty())```

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -304,6 +304,31 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+				for _, agettyPid := range agettyPids {
+					cgroupCommand := fmt.Sprintf("cat /proc/%s/cgroup", agettyPid)
+					cgroupOutput, err := nodes.ExecCommandOnNode([]string{"/bin/bash", "-c", cgroupCommand}, &node)
```
ffromani, line 318: I'd merge these two lines:
```
cgroupOutput, err := nodes.ExecCommandOnNode([]string{"/bin/bash", "-c", fmt.Sprintf("cat /proc/%s/cgroup", agettyPid)}, &node)
```
(applies also above)

test/e2e/performanceprofile/functests/1_performance/performance.go
yanirq, line 308: Lets phrase this with should  .  e.g. [It] should have agetty system service under the cpuset cgroup controller

ffromani (commented):
LGTM, minor comments inline

test/e2e/performanceprofile/functests/1_performance/performance.go
MarSik: Hmm.. for making the test meaninful yes. But is missing agetty really a bug?

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -304,6 +304,31 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+					Expect(err).ToNot(HaveOccurred())
+					lines := strings.Split(cgroupOutput, "\n")
+					var cpusetLine string
+					for _, line := range lines {
+						if strings.Contains(line, "cpuset") {
+							cpusetLine = line
+							break
+						}
+					}
+					Expect(cpusetLine).ToNot(BeEmpty(), "No cpuset line found in cgroup output for PID %s on node %s", agettyPid, node.Name)
```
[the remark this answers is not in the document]
MarSik: This will break on cgroups v2, because no cpuset hieararchy exists there.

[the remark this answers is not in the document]
MarSik: Why not read /proc/<pid>/cpuset?

MarSik:
Honestly I am not sure the test design is sound. Yes, we move agetty under system.slice, but that is an implementation detail. The important part we should be testing for is that no system process is interfering with the isolated cpus.

The bug was agetty not being moved properly, fine, but next time it might be a totally different process and this test will not catch it.

MarSik:
One other thing. Why is this PR trying to merge into 4.14? We always start with the master branch. Is this a backport?

test/e2e/performanceprofile/functests/1_performance/performance.go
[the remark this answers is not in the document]
ffromani: reading /proc/pid/cpuset is nicer indeed

SargunNarula:
After an offline discussion with @MarSik, we found a test case that identifies all systemd processes and checks if they are running in the system.slice cgroup. We need similar functionality here, as required by this bug: https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/1_performance/performance.go#L305.

This test was introduced to cover all scenarios in this PR: https://github.com/openshift/cluster-node-tuning-operator/pull/992. The PR was back-ported to versions 4.14 and 4.15, so no additional fixes are needed here.

The agetty bug was opened for versions 4.13 and 4.14, but we do not plan to cover this test in either version.

PR #1075, opened 2024-06-06
OCPBUGS-34972: perfprof: bump verbosiness of logs

ffromani:
Increase the verbosiness required to emit log messages to reduce the log amount.
Note this change doesn't address the reason for periodic reconciles, we only review the verbosiness of log messages.


MarSik (approved):
/lgtm

PR #1078, opened 2024-06-09
NO-JIRA:Dockerfile:bump go-builder to 1.22

Tal-or:
In the `go.mod` file, we are using the `toolchain` directive, which is not recognized by golang 1.20.


ffromani:
/lgtm

/cc @jmencak 
for approval

jmencak:
/approve
At some point, we should remove the "Upstream" `Dockerfile`.  `Dockerfile.rhel9` is now badly named and will build with CentOS stream too and also pull upstream `TuneD`.  There is really no need to keep two (three Dockerfile.rhel8) Containerfiles around.  Adding to my TODO (cleanup list) unless someone else wants to pick this up.

PR #1079, opened 2024-06-10
NO-JIRA: remove 4.13->4.14 compatibility code

ffromani:
remove code which was meant to handle the upgrade
from 4.13 to 4.14. We're at least two version past that.


ffromani:
why the bot missed the hold? sigh.

PR #1081, opened 2024-06-10
NO-JIRA: revert: remove 4.13->4.14 compatibility code (#1079)"

ffromani:
This reverts commit aae3135a3dab198ff2503cd98ad37bd854c8d6dd.

Since #1079 was merged too quickly because the bot missed the "hold" command, I'm posting a revert so we can discuss

### DIFF
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -9,8 +9,6 @@ import (
 	"runtime"
 	"time"
 
-	"net/http"
-
 	apiconfigv1 "github.com/openshift/api/config/v1"
 	mcov1 "github.com/openshift/api/machineconfiguration/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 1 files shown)


Tal-or:
/hold
/lgtm

Adding due to https://github.com/openshift/cluster-node-tuning-operator/pull/1079#issuecomment-2158071511

Tal-or:
@ffromani feel free to unhold, I just added the hold to avoid what just happened in the original PR

ffromani:
I still think the original PR has merit, so we should NOT merge this revert, but let's discuss

bartwensley:
@MarSik - I didn't work on the feature, but I believe @eggfoobar did

eggfoobar:
Oh hey! Yeah this should be removed by now, the migration was for 4.13->4.14, this all went GA 4.14 if I remember correctly, so we should be good to remove it. Only issue would be someone jumping from 4.13 to 4.17, which to my knowledge we have no way of supporting that high of a jump correct?

ffromani:
So, I understand we do _not_ want this revert because we are happy with the compat check removed, right?
@eggfoobar @MarSik @yanirq

eggfoobar:
Yeah that's my understanding, unless we ran into an issue I'm not aware of, this code should be past it's usefulness and we don't want this code back in master.

ffromani:
/close

peer feedback and conversations

PR #1083, opened 2024-06-10
CNF-14090: Add vendor and architecture specific tuning options

fontivan:
- This makes use of a new helper function in tuned that was added in tuned 2.24 to get cpu information from lscpu
- Adds separate tuning files for AMD/X86, ARM/AArch64, and Intel/X86
- Platform specific tuning can be added to the appropriate tuning file and will loaded automatically by the openshift-node-performance profile
- Tuning for AMD and ARM should be considered not final at this time

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -140,6 +140,8 @@ initrd_add_dir=
 
 # overrides cpu-partitioning cmdline
 cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} systemd.cpu_affinity=${not_isolated_cores_expanded}
+
+# No default value but will be composed conditionally based on platform
 cmdline_iommu=
 
 {{if .StaticIsolation}}
@@ -153,11 +155,13 @@ cmdline_realtime=+nohz_full=${isolated_cores} nosoftlockup skew_tick=1 rcutree.k
 {{end}}
 
 {{if .HighPowerConsumption}}
-cmdline_power_performance=${high_power_consumption_cstate}
+# No default value but will be composed conditionally based on platform
+cmdline_power_performance=
 {{end}}
 
 {{if and .HighPowerConsumption .RealTimeHint}}
-cmdline_idle_poll=+idle=poll
+# No default value but will be composed conditionally based on platform
+cmdline_idle_poll=
 {{end}}
 
 {{if .DefaultHugepagesSize}}
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -1,12 +1,18 @@
 [main]
 summary=Openshift node optimized for deterministic performance at the cost of increased power consumption, focused on low latency network performance. Based on Tuned 2.11 and Cluster node tuning (oc 4.5)
 
-# In case real time kernel is enabled the following include section will be evaluated as:
-# include=openshift-node,cpu-partitioning,openshift-node-performance-rt-<PerformanceProfile name>
-# Otherwise:
-# include=openshift-node,cpu-partitioning
-include=openshift-node,cpu-partitioning${f:regex_search_ternary:${f:exec:uname:-r}:rt:,openshift-node-performance-rt-{{.PerformanceProfileName}}:}
-
+# The final result of the include depends on cpu vendor, cpu architecture, and whether the real time kernel is enabled
+# The first line will be evaluated based on the CPU vendor and architecture
+# This has three possible results:
+#   include=openshift-node-performance-amd-x86;
+#   include=openshift-node-performance-arm-aarch64;
+#   include=openshift-node-performance-intel-x86;
+# The second line will be evaluated based on whether the real time kernel is enabled
+# This has two possible results:
+#     openshift-node,cpu-partitioning
+#     openshift-node,cpu-partitioning,openshift-node-performance-rt-<PerformanceProfile name>
+include=openshift-node-performance-${f:lscpu_check:Vendor ID\:\s*GenuineIntel:intel:Vendor ID\:\s*AuthenticAMD:amd:Vendor ID\:\s*ARM:arm}-${f:lscpu_check:Architecture\:\s*x86_64:x86:Architecture\:\s*aarch64:aarch64}-{{.PerformanceProfileName}};
+  openshift-node,cpu-partitioning${f:regex_search_ternary:${f:exec:uname:-r}:rt:,openshift-node-performance-rt-{{.PerformanceProfileName}}:}
 
 # Inheritance of base profiles legend:
 # cpu-partitioning -> network-latency -> latency-performance
@@ -24,7 +30,6 @@ isolated_cores={{.IsolatedCpus}}
 {{end}}
 
 not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
-automatic_pstate=${f:intel_recommended_pstate}
 
 {{if .PerPodPowerManagement}}
 [cpu]
@@ -134,7 +139,8 @@ initrd_dst_img=
 initrd_add_dir=
 
 # overrides cpu-partitioning cmdline
-cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} systemd.cpu_affinity=${not_isolated_cores_expanded} intel_iommu=on iommu=pt
+cmdline_cpu_part=+nohz=on rcu_nocbs=${isolated_cores} tuned.non_isolcpus=${not_isolated_cpumask} systemd.cpu_affinity=${not_isolated_cores_expanded}
+cmdline_iommu=
 
 {{if .StaticIsolation}}
 cmdline_isolation=+isolcpus=domain,managed_irq,${isolated_cores}
@@ -143,11 +149,11 @@ cmdline_isolation=+isolcpus=managed_irq,${isolated_cores}
 {{end}}
 
 {{if .RealTimeHint}}
-cmdline_realtime=+nohz_full=${isolated_cores} tsc=reliable nosoftlockup nmi_watchdog=0 mce=off skew_tick=1 rcutree.kthread_prio=11
+cmdline_realtime=+nohz_full=${isolated_cores} nosoftlockup skew_tick=1 rcutree.kthread_prio=11
 {{end}}
 
 {{if .HighPowerConsumption}}
-cmdline_power_performance=+processor.max_cstate=1 intel_idle.max_cstate=0
+cmdline_power_performance=${high_power_consumption_cstate}
 {{end}}
 
 {{if and .HighPowerConsumption .RealTimeHint}}
@@ -158,15 +164,7 @@ cmdline_idle_poll=+idle=poll
 cmdline_hugepages=+ default_hugepagesz={{.DefaultHugepagesSize}} {{end}} {{if .Hugepages}} {{.Hugepages}} {{end}}
 
 {{if .AdditionalArgs}}
-cmdline_additionalArg=+{{.AdditionalArgs}} 
-{{end}}
-
-{{if .PerPodPowerManagement}}
-cmdline_pstate=+intel_pstate=passive
-{{else if .HardwareTuning}}
-cmdline_pstate=+intel_pstate=active
-{{else}}
-cmdline_pstate=+intel_pstate=${automatic_pstate}
+cmdline_additionalArg=+{{.AdditionalArgs}}
 {{end}}
 
 [rtentsk]
--- a/assets/performanceprofile/tuned/openshift-node-performance-amd-x86
+++ b/assets/performanceprofile/tuned/openshift-node-performance-amd-x86
@@ -19,3 +19,7 @@ cmdline_realtime_amd=tsc=reliable nmi_watchdog=0 mce=off
 {{if .HighPowerConsumption}}
 cmdline_power_performance_amd=processor.max_cstate=1
 {{end}}
+
+{{if and .HighPowerConsumption .RealTimeHint}}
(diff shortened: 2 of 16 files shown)


assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -24,7 +24,46 @@ isolated_cores={{.IsolatedCpus}}
 {{end}}
 
 not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
-automatic_pstate=${f:intel_recommended_pstate}
+
+# Intel specific tuning options
+{{if ${f:lscpu_check:"^Vendor ID:\s*(.*)$"} == "GenuineIntel"}}
```
MarSik, line 29: Eventually we will have to deal with sysctls as well. I think separating those to a sub-profile will work better. The main profile will then just include (if we can do that for variables). The current logic here is already becoming hard to read.

fontivan: If I am understanding you correctly, then you think it might be better to look at a more file orientated approach, where we might have separate files that get included when appropriate, such as:

- openshift-node-performance-shared (with all common configuration)
- openshift-node-performance-intel-x86
- openshift-node-performance-amd-x86
- openshift-node-performance-arm-aarch64

And then we could overlay a specific platform profile on top of the shared profile to come up with a final profile?

MarSik: We already have shared, that is the main profile. And we already have -rt for real time changes. But the others might make the whole thing more readable once it is more than just few kernel args.

fontivan: I've updated the draft PR to use separate files as described above with the include tuned directive

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -24,7 +24,22 @@ isolated_cores={{.IsolatedCpus}}
 {{end}}
 
 not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
-automatic_pstate=${f:intel_recommended_pstate}
+
+# Check for x86 tuning
+{{if ${f:lscpu_check:"^Architecture:\s*(.*)$"} == "x86_64"}}
```
MarSik, line 29: I am not sure this is correct, is tuned evaluating the `{{if` or golang in the operator? I suspect the latter. And in that case it will not work.

MarSik: Also the includes need to happen in the [main] section. An example is on line 8

fontivan: I am not sure the order of the evaluation either but I am working on getting a lab up and running to try it and see what actually happens

fontivan: After testing I think you are correct, it seems that the curly braces are invalid characters in the [main] section

I have 2 more ideas for how this may be workable:
1. Include all the platform tuning in the main file and use conditions inside the individual files to no-op non matching platforms
2. Invert the include order so that we actually apply a specific platform profile (intel, amd, aarch64) and each of those import the existing openshift-node-performance profile with all the shared configuration

I am currently leaning more into 2 but I am not sure how the profile selection will work

fontivan: I've updated the performance profiles for idea 2, however this will require additional work on selecting the profile

assets/performanceprofile/tuned/openshift-node-performance-amd-x86
```diff
@@ -1,5 +1,7 @@
 [main]
-Summary=Platform specific tuning for AMD x86
+summary=Platform specific tuning for AMD x86
+
+include=openshift-node-performance
```
MarSik, line 4: Nope, will not work. The operator does not know the architecture of the node and so cannot select the proper active profile this way.

You need to use the conditional include in the same way we do for the RT kernel

fontivan: Though it is a bit messy I did see a way to use only f functions to handle the conditional import like the rt kernel profile uses, and they do seem to be evaluting correctly on my test server

```
2024-06-13 15:41:04,411 DEBUG    tuned.profiles.functions.functions: ${f:lscpu_check:Vendor ID:\s*GenuineIntel:intel:AuthenticAMD:amd:ARM:arm} expands to: 'intel'
2024-06-13 15:41:04,415 DEBUG    tuned.profiles.functions.functions: ${f:lscpu_check:Architecture:\s*x86_64:x86:aarch64:aarch64} expands to: 'x86'
```

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -1,12 +1,16 @@
...
+# include=openshift-node,cpu-partitioning,openshift-node-performance-intel-x86,openshift-node-performance-rt-<PerformanceProfile name>
 # Otherwise:
-# include=openshift-node,cpu-partitioning
-include=openshift-node,cpu-partitioning${f:regex_search_ternary:${f:exec:uname:-r}:rt:,openshift-node-performance-rt-{{.PerformanceProfileName}}:}
-
+# include=openshift-node,cpu-partitioning,okpenshift-node-performance-intel-x86
+# This include will also evaluate for other CPU vendors and architectures, such as:
+# include=openshift-node,cpu-partitioning,openshift-node-performance-amd-x86
+# include=openshift-node,cpu-partitioning,openshift-node-performance-arm-aarch64
+include=openshift-node-performance-${f:lscpu_check:Vendor ID\:\s*GenuineIntel:intel:AuthenticAMD:amd:ARM:arm}-${f:lscpu_check:Architecture\:\s*x86_64:x86:aarch64:aarch64},openshift-node,cpu-partitioning${f:regex_search_ternary:${f:exec:uname:-r}:rt:,openshift-node-performance-rt-{{.PerformanceProfileName}}:}
```
fontivan, line 13: I'm not sure if the platform specific tuning should be first or last in the list

The f function to handle the vendor and architecture do have to be quite long to accomplish this, but it does seem to be working

MarSik: Yeah, it is messy, maybe we could talk to the tuned folks to check if there is a nicer syntax (at least a way to split to multiple lines).

I think the order does not matter much, but we can ask about that as well.

fontivan:
I think this will work pretty well once the new lscpu helper is upstreamed (https://github.com/redhat-performance/tuned/pull/638) but I will need to figure out how to make the tests happy with this change.

It seems like the tests are not evaluating the include statement, so it is not seeing the results it is expecting. I'm seeing 6 test failures locally with the current version of the PR

MarSik:
@fontivan You need to do couple of things. First, deploy the new files to the cluster. Like here https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go#L228

And then run `make render-sync` to refresh the expected output files (and check them of course).

fontivan:
> @fontivan You need to do couple of things. First, deploy the new files to the cluster. Like here https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go#L228
> 
> And then run `make render-sync` to refresh the expected output files (and check them of course).

I'm guessing I'm going to at least need the new helper merged upstreamed so that I can use the `TUNED_COMMIT` option described in HACKING.md to accomplish this ?

MarSik:
That is for building the NTO container and the full cluster tests.

But the initial test you can do is via `make test-e2e-local` and that does not need any cluster or tuned.

MarSik:
@fontivan Please squash at least my fix-up patches with something so this gets ready for merge. I will review the CI.

assets/performanceprofile/tuned/openshift-node-performance-amd-x86
```diff
@@ -0,0 +1,23 @@
+[main]
+summary=Platform specific tuning for AMD x86
+
+[variables]
+
+# AMD specific tuning options
+iommu=+amd_iommu=on iommu=pt
+hpc_cstate=+processor.max_cstate=1
+pppm_pstate=+amd_pstate=active
```
MarSik, line 9: I would move the variable values directly to the block below OR keep just the variables (with nicer names) and move the IF block back to the main profile for readability.

assets/performanceprofile/tuned/openshift-node-performance-intel-x86
```diff
@@ -0,0 +1,23 @@
+[main]
+summary=Platform specific tuning for Intel x86
+
+[variables]
+
+# Intel specific tuning options
+iommu=+intel_iommu=on iommu=pt
```
MarSik, line 7: Same comment as for the AMD profile here

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -24,7 +30,6 @@ isolated_cores={{.IsolatedCpus}}
 {{end}}
 
 not_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}
-automatic_pstate=${f:intel_recommended_pstate}
```
MarSik, line 27: This is where you lost the intel_pstate=disable, it needs to go to the intel specific tuning profile

fontivan:
> @fontivan Please squash at least my fix-up patches with something so this gets ready for merge. I will review the CI.

Will squash into 2 commits (1 for the work, 1 for the render sync) once I have things working correctly

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -143,11 +149,11 @@ cmdline_isolation=+isolcpus=managed_irq,${isolated_cores}
...
-cmdline_realtime=+nohz_full=${isolated_cores} tsc=reliable nosoftlockup nmi_watchdog=0 mce=off skew_tick=1 rcutree.kthread_prio=11
+cmdline_realtime=+nohz_full=${isolated_cores} nosoftlockup skew_tick=1 rcutree.kthread_prio=11
 {{end}}
 
 {{if .HighPowerConsumption}}
-cmdline_power_performance=+processor.max_cstate=1 intel_idle.max_cstate=0
+cmdline_power_performance=${high_power_consumption_cstate}
 {{end}}
 
 {{if and .HighPowerConsumption .RealTimeHint}}
```
bartwensley, line 159: I don't think ARM supports the `idle=poll` argument (see https://github.com/torvalds/linux/blob/42f7652d3eb527d03665b09edac47f85fb600924/Documentation/admin-guide/kernel-parameters.txt#L2001). Maybe the best approach would be just to reject the .HighPowerConsumption hint for ARM servers for now since we have no way support it - is there a way to do that?

fontivan: Good catch, does not look supported on arm so we should only set it for x86.

fontivan: @MarSik I've used composition for this like on the other changes

assets/performanceprofile/tuned/openshift-node-performance-amd-x86
```diff
@@ -0,0 +1,25 @@
...
+[bootloader]
+cmdline_iommu_amd=amd_iommu=on iommu=pt
+
+{{if .PerPodPowerManagement}}
+cmdline_pstate=amd_pstate=active
+{{else if .HardwareTuning}}
+cmdline_pstate=amd_pstate=passive
+{{else}}
+cmdline_pstate=amd_pstate=guided
+{{end}}
```
bartwensley, line 13: Hmm... how did you decide on these? I would have thought we would use active by default and then passive only when .PerPodPowerManagement was enabled.

assets/performanceprofile/tuned/openshift-node-performance-intel-x86
```diff
@@ -0,0 +1,25 @@
...
+summary=Platform specific tuning for Intel x86
+
+[bootloader]
+cmdline_iommu_intel=intel_iommu=on iommu=pt
+
+{{if .PerPodPowerManagement}}
+cmdline_pstate=intel_pstate=active
+{{else if .HardwareTuning}}
+cmdline_pstate=intel_pstate=passive
+{{else}}
```
bartwensley, line 11: I think these are backwards

test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-ctrcfg/openshift-bootstrap-master_tuned.yaml
```diff
@@ -48,16 +55,64 @@ spec:
...
+      cmdline_pstate=amd_pstate=guided
+
+
+
+      cmdline_realtime_amd=tsc=reliable nmi_watchdog=0 mce=off
+
+
+
+
+
```
bartwensley, line 84: Why all the extra blank lines in the rendered output? Is this a bug somewhere?

assets/performanceprofile/tuned/openshift-node-performance-intel-x86
fontivan: I'm not sure how I managed that but it does appear I inverted them from the original at some point, will fix

assets/performanceprofile/tuned/openshift-node-performance-amd-x86
fontivan: I based these on the intel ones but apparently after I mixed them up, so I will update these too

test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-ctrcfg/openshift-bootstrap-master_tuned.yaml
fontivan: I am not sure, I think it is related to the templating using {{}} though hopefully Martin can enlighten us

MarSik:
Ouch.. I missed the submodule update that does not belong here.

MarSik:
Ok, manual cherry pick it is :)

PR #1084, opened 2024-06-11
CNF-7123: Enable and adjust 1_performance suite on Hypershift

rbaturov:
This PR includes all the necessary adaptations for the `1_performance` suite to be compatible with Hypershift. These changes encompass modifications to some of the utility functions used by the suite, as well as updates to the tests within the suite itself. Additionally, this PR integrates this suite to run alongside the `0_config` suite in the Hypershift lane.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,7 +257,7 @@ pao-functests-mixedcpus:
 pao-functests-hypershift:
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "./test/e2e/performanceprofile/functests/0_config" -p "-vv --label-filter="!openshift" -r --fail-fast --flake-attempts=2 --junit-report=report.xml" -m "Running Functional Tests over Hypershift"
+	hack/run-test.sh -t "./test/e2e/performanceprofile/functests/0_config ./test/e2e/performanceprofile/functests/1_performance" -p "-vv --label-filter="!openshift" -r --fail-fast --flake-attempts=2 --timeout=2h --junit-report=report.xml" -m "Running Functional Tests over Hypershift"
 
 .PHONY: cluster-clean-pao
 cluster-clean-pao:
--- a/Makefile
+++ b/Makefile
@@ -257,7 +257,7 @@ pao-functests-mixedcpus:
 pao-functests-hypershift:
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "./test/e2e/performanceprofile/functests/0_config" -p "-vv -r --fail-fast --flake-attempts=2 --junit-report=report.xml" -m "Running Functional Tests over Hypershift"
+	hack/run-test.sh -t "./test/e2e/performanceprofile/functests/0_config" -p "-vv --label-filter="!openshift" -r --fail-fast --flake-attempts=2 --junit-report=report.xml" -m "Running Functional Tests over Hypershift"
 
 .PHONY: cluster-clean-pao
 cluster-clean-pao:
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -17,7 +17,6 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	machineconfigv1 "github.com/openshift/api/machineconfiguration/v1"
 	tunedv1 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/tuned/v1"
 
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
@@ -28,10 +27,10 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/label"
 	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
-	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/mcps"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profilesupdate"
 	e2etuned "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/tuned"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/util"
 	"github.com/openshift/cluster-node-tuning-operator/test/framework"
@@ -45,7 +44,6 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 	var workerRTNodes []corev1.Node
 	var targetNode *corev1.Node
 	var profile, initialProfile *performancev2.PerformanceProfile
-	var performanceMCP string
 	var err error
 
 	BeforeEach(func() {
@@ -60,13 +58,7 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 
 		initialProfile = profile.DeepCopy()
 
-		performanceMCP, err = mcps.GetByProfile(profile)
-		Expect(err).ToNot(HaveOccurred())
-
-		// Verify that worker and performance MCP have updated state equals to true
-		for _, mcpName := range []string{testutils.RoleWorker, performanceMCP} {
-			mcps.WaitForCondition(mcpName, machineconfigv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
-		}
+		profilesupdate.WaitForTuningUpdated(context.TODO(), profile)
 
 		nodeIdx := pickNodeIdx(workerRTNodes)
 		targetNode = &workerRTNodes[nodeIdx]
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -96,7 +96,7 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 
 					condStatus := corev1.ConditionUnknown
 					EventuallyWithOffset(1, context.TODO(), func() bool {
-						tunedProfile, err := e2etuned.GetProfile(context.TODO(), testclient.Client, components.NamespaceNodeTuningOperator, node.Name)
+						tunedProfile, err := e2etuned.GetProfile(context.TODO(), testclient.DataPlaneClient, components.NamespaceNodeTuningOperator, node.Name)
 						Expect(err).ToNot(HaveOccurred(), "failed to get Tuned Profile for node %q", node.Name)
 						for _, cond := range tunedProfile.Status.Conditions {
 							if cond.Type != tunedv1.TunedProfileApplied {
@@ -228,7 +228,7 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 			data, _ := json.Marshal(testpod)
 			testlog.Infof("using testpod:\n%s", string(data))
 
-			err = testclient.Client.Create(context.TODO(), testpod)
+			err = testclient.DataPlaneClient.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
 			defer func() {
 				if testpod != nil {
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/cpuset"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -146,14 +145,8 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 
 			defer func() { // return initial configuration
 				By("reverting the profile into its initial state")
-				spec, err := json.Marshal(initialProfile.Spec)
-				Expect(err).ToNot(HaveOccurred())
-				Expect(testclient.Client.Patch(context.TODO(), profile,
-					client.RawPatch(
-						types.JSONPatchType,
-						[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
-					),
-				)).ToNot(HaveOccurred())
+				profile.Spec = initialProfile.Spec
+				profiles.UpdateWithRetry(profile)
 				// in the initial profile `GloballyDisableIrqLoadBalancing` value should be false,
 				// so we expect the banned cpu set to be empty
 				Eventually(verifyNodes).WithArguments(false).WithPolling(10 * time.Second).WithTimeout(1 * time.Minute).ShouldNot(HaveOccurred())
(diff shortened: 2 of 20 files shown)


pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -65,16 +83,93 @@ func NewControlPlaneClient(c client.Client, ns string) *ControlPlaneClientImpl {
...
+	case *machineconfigv1.KubeletConfigList:
+		objectType = reflect.TypeOf(machineconfigv1.KubeletConfig{})
+	case *machineconfigv1.MachineConfigList:
+		objectType = reflect.TypeOf(machineconfigv1.MachineConfig{})
+	case *tunedv1.TunedList:
+		objectType = reflect.TypeOf(tunedv1.Tuned{})
+	default:
+		return fmt.Errorf("unsupported object list type: %T", listObj)
+	}
+
```
Tal-or, line 105: At this point you already know what type of list you have, so instead of dealing with reflection, why not casting the type and append into its items directly?

Tal-or (commented):
Initial review.
Will have another pass soon.

Tal-or:
/retitle CNF-7123: Enable and adjust 1_performance suite on Hypershift

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -507,7 +507,7 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 		})
 	})
 
-	Context("Create second performance profiles on a cluster", Label(string(label.Tier0)), func() {
+	Context("Create second performance profiles on a cluster", Label(string(label.Tier0), string(label.OpenShift)), func() {
```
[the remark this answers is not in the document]
Tal-or: Edited: We can have more than a single performance profile but it means we need additional nodepool to be associated with our hosted cluster

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -65,16 +83,93 @@ func NewControlPlaneClient(c client.Client, ns string) *ControlPlaneClientImpl {
...
+func (ci *ControlPlaneClientImpl) listFromConfigMaps(ctx context.Context, listObj client.ObjectList, opts ...client.ListOption) error {
+	cmList, err := ci.getCorrespondingConfigMaps(ctx, listObj)
+	if err != nil {
+		return err
 	}
-	err := ci.Client.Get(ctx, key, cm, opts...)
+
+	var objectType reflect.Type
+	switch listObj.(type) {
+	case *performancev2.PerformanceProfileList:
```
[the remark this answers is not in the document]
Tal-or: This functions will be used only for testing and this extra check makes the code more flexible and readable so let's leave it for now

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -119,6 +195,16 @@ func (ci *ControlPlaneClientImpl) createInConfigMap(ctx context.Context, obj cli
 	return ci.Client.Create(ctx, cm, opts...)
 }
 
+func (ci *ControlPlaneClientImpl) updateConfigMap(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+	cm := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: obj.GetName(), Namespace: HostedClustersNamespaceName}}
+	b, err := EncodeManifest(obj, scheme.Scheme)
+	if err != nil {
+		return err
+	}
+	cm.Data = map[string]string{TuningKey: string(b)}
```
Tal-or, line 204: There's a bug here, we should check what kind of object we're creating.
MC and KC are supposed to be created with the `ConfigKey`

rbaturov: I was thinking that we should use Update only for updating performance profile object. I was assuming that non of the other encapsulated objects would need to be updated as part of the E2E tests. But maybe my assumption was wrong

hack/label-worker-cnf.sh
```diff
@@ -6,8 +6,24 @@ set -e
 OC_TOOL="${OC_TOOL:-oc}"
 
 # Label 1 worker node
-echo "[INFO]: Labeling 1 worker node with worker-cnf"
-node=$(${OC_TOOL} get nodes --selector='node-role.kubernetes.io/worker' \
-    --selector='!node-role.kubernetes.io/master' -o name | head -1)
 
-${OC_TOOL} label --overwrite $node node-role.kubernetes.io/worker-cnf=""
+function label_worker_cnf() {
```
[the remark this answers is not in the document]
ffromani: we discussed about making sure hypershift worker nodes have the correct labels after each config (tuned, perfprofile, kubeletconfig...) update because this will cause the reprovisioning of the nodes.
I guess this part is not implemented here yet

test/e2e/performanceprofile/functests/1_performance/performance.go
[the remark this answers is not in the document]
ffromani: So up to one perfprofile per nodepool, right? If so @rbaturov captured in the commit message already, which is good.

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: We cannot assumed that. please change for `Create` as well since I did the same mistake there.

test/e2e/performanceprofile/functests/utils/nodepools/nodepools.go
```diff
@@ -44,3 +45,20 @@ func waitForCondition(ctx context.Context, c client.Client, NpName, namespace st
...
+
+func GetByClusterName(ctx context.Context, c client.Client, hostedClusterName string) (*hypershiftv1beta1.NodePool, error) {
+	npList := &hypershiftv1beta1.NodePoolList{}
+	if err := c.List(ctx, npList); err != nil {
+		return nil, err
+	}
+	var np *hypershiftv1beta1.NodePool
+	for i := 0; i < len(npList.Items); i++ {
+		if npList.Items[i].Spec.ClusterName == hostedClusterName {
+			np = &npList.Items[i]
```
ffromani, line 57: break or return here, likely better to break

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -4,9 +4,12 @@ import (
 	"bytes"
```
ffromani, line 4: this should have been 2 commits, 1 to add List, 1 to add Update. OK for now, let's take into account next time.

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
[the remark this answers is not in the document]
ffromani: Extra check dupes existing code though and create hidden coupling - we will need to fix both places in the same flow unnecessarily though. I won't block if we want to keep this, but should cleaned up later

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
ffromani: We should really avoid reflection unless there's a very strong case for it

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -65,16 +83,93 @@ func NewControlPlaneClient(c client.Client, ns string) *ControlPlaneClientImpl {
...
+		return err
+	}
+	return nil
+}
+
+func (ci *ControlPlaneClientImpl) getCorrespondingConfigMaps(ctx context.Context, obj runtime.Object) (*corev1.ConfigMapList, error) {
+	var labelSelector string
+
+	switch obj.(type) {
+	case *performancev2.PerformanceProfile, *performancev2.PerformanceProfileList:
```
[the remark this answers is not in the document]
ffromani: same comments as above about dupe ode

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -146,14 +145,8 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
...
-				spec, err := json.Marshal(initialProfile.Spec)
-				Expect(err).ToNot(HaveOccurred())
-				Expect(testclient.Client.Patch(context.TODO(), profile,
-					client.RawPatch(
-						types.JSONPatchType,
-						[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
-					),
-				)).ToNot(HaveOccurred())
+				profile.Spec = initialProfile.Spec
+				Expect(testclient.ControlPlaneClient.Update(context.TODO(), profile)).ToNot(HaveOccurred())
```
ffromani, line 141: at glance the logic here seems questionable. We should `Get` in the loop, not just update. Let's address in a followup PR.

test/e2e/performanceprofile/functests/1_performance/netqueues.go
```diff
@@ -86,14 +83,8 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
...
-		spec, err := json.Marshal(initialProfile.Spec)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(testclient.Client.Patch(context.TODO(), profile,
-			client.RawPatch(
-				types.JSONPatchType,
-				[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/spec", "value": %s }]`, spec)),
-			),
-		)).ToNot(HaveOccurred())
+		profile.Spec = initialProfile.Spec
+		testclient.ControlPlaneClient.Update(context.TODO(), profile)
```
ffromani, line 87: likewise, this should be wrapped in a `Eventually` block (or a retry loop in general)

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
```diff
@@ -118,4 +119,19 @@ func WaitForTuningUpdated(ctx context.Context, profile *performancev2.Performanc
...
+		Expect(testclient.DataPlaneClient.List(context.TODO(), nodesList)).ToNot(HaveOccurred())
+		workerRTNode := nodesList.Items[0]
+		workerCNFLabel := fmt.Sprintf("%s/%s", testutils.LabelRole, testutils.RoleWorkerCNF)
+		testlog.Infof("labeling node %s with label %s\n", workerRTNode.Name, workerCNFLabel)
+		labels := workerRTNode.GetLabels()
+		labels[workerCNFLabel] = ""
+		workerRTNode.SetLabels(labels)
+		Expect(testclient.DataPlaneClient.Update(context.TODO(), &workerRTNode))
+		testlog.Infof("worker %s labels are: %v\n", workerRTNode.Name, workerRTNode.Labels)
+	}
```
ffromani, line 136: nope, the intention is good but this is a bit too simplistic.
1. a function which `wait` for something should not mutate the cluster state. Thus
2. let's move the label enforcement in its own function, and let's call it explicitely
3. inside the label updating logic, `get+mutate+update` needs to be wrapped into a `Eventually` or retry loop in general

ffromani:
This PR is unwieldly. Can be split and merged in phases if you are up for it. I'd be willing to assist and merge.

hack/label-worker-cnf.sh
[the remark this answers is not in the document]
Tal-or: Here  the initial labeling is done. inside the code itself we relabel when needed

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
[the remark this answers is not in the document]
Tal-or: The problem is that if we want to use to previous check, we should pass the type and use reflection here

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
Tal-or: > 2. let's move the label enforcement in its own function, and let's call it explicitly

How do we avoid from other developers to forget to call this function?

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: Moved into a separate PR: https://github.com/openshift/cluster-node-tuning-operator/pull/1107

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
Tal-or: Done

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: Fixed here:  https://github.com/openshift/cluster-node-tuning-operator/pull/1107

test/e2e/performanceprofile/functests/utils/nodepools/nodepools.go
Tal-or: Done.

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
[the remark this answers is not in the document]
Tal-or: Fixed in  https://github.com/openshift/cluster-node-tuning-operator/pull/1107

Tal-or:
/hold
depend on #1107

ffromani:
please rebase and remove the hold: #1107 merged

Makefile
```diff
@@ -257,7 +257,7 @@ pao-functests-mixedcpus:
 pao-functests-hypershift:
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "./test/e2e/performanceprofile/functests/0_config" -p "-vv -r --fail-fast --flake-attempts=2 --junit-report=report.xml" -m "Running Functional Tests over Hypershift"
+	hack/run-test.sh -t "./test/e2e/performanceprofile/functests/0_config ./test/e2e/performanceprofile/functests/1_performance" -p "-vv --label-filter="!openshift" -r --fail-fast --flake-attempts=2 --timeout=2h --junit-report=report.xml" -m "Running Functional Tests over Hypershift"
```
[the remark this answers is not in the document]
ffromani: yes, this is the first part of a longer series

ffromani (commented):
/approve
/lgtm

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
```diff
@@ -77,3 +85,82 @@ func ApplyProfile(profile *performancev2.PerformanceProfile) error {
...
+// This is a lower lever function and should not be used directly.
+// Use PostUpdateSync instead
+func WaitForTuningUpdated(ctx context.Context, profile *performancev2.PerformanceProfile) {
+	GinkgoHelper()
+	// In case we are on OCP, we can query the MCP to determine if the update has completed.
+	if !hypershift.IsHypershiftCluster() {
+		performanceMCP, err := mcps.GetByProfile(profile)
+		Expect(err).ToNot(HaveOccurred())
+		testlog.Infof("waiting for MCP being updated")
+		mcps.WaitForCondition(performanceMCP, mcv1.MachineConfigPoolUpdated, corev1.ConditionTrue)
```
yanirq, line 123: previously we were waiting on both performance mcp AND worker mcp:
now I can see its only performance mcp. did we drop that on purpose ?  
Tracking that from the new `PostUpdateSync`

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
```diff
@@ -77,3 +85,82 @@ func ApplyProfile(profile *performancev2.PerformanceProfile) error {
...
+	hostedClusterName, err := hypershift.GetHostedClusterName()
+	Expect(err).ToNot(HaveOccurred())
+	np, err := nodepools.GetByClusterName(ctx, testclient.ControlPlaneClient, hostedClusterName)
+	Expect(err).ToNot(HaveOccurred())
+	testlog.Infof("wait for node pool %q transition into config ready state", client.ObjectKeyFromObject(np).String())
+	err = nodepools.WaitForConfigToBeReady(ctx, testclient.ControlPlaneClient, np.Name, np.Namespace)
+	Expect(err).ToNot(HaveOccurred())
+}
+
+func EnforceLabels() {
```
yanirq, line 137: maybe a comment if this is a public function?  what are we enforcing here ?

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
Tal-or: > previously we were waiting on both performance mcp AND worker mcp:
Could you please add a reference?
This function is new

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
Tal-or: Done. Thanks!

test/e2e/performanceprofile/functests/utils/profilesupdate/profile_update.go
yanirq: see https://github.com/openshift/cluster-node-tuning-operator/pull/1084/files#diff-decd79950ca0682eefd520e1171ff0d6c0af15779a02cc59fc71c4dbe1012bfaL68

Tal-or: I missed this change:
1. This was added by @ffromani back in the time, maybe he remember the reason for waiting on both MCPs https://github.com/openshift/cluster-node-tuning-operator/commit/c5cf0bd33c5afae7695fd86d08c62aef269105d8#diff-decd79950ca0682eefd520e1171ff0d6c0af15779a02cc59fc71c4dbe1012bfa 
Usually we're not waiting for the worker MCP.
2. The tests are passing consistently. 

Anyway this is not an excuse, just stating the facts. I'll try to work something out on Sunday

ffromani: I don't remember the exact reason. I won't rule out it was a testing artifact (= workaround) which wasn't caught in review. I'd not be too worried about waiting for both if tests pass consistently.

Tal-or: Then if it's agreed we can merge and evaluate later if needed

ffromani: yes, let's do this way

ffromani:

this should have merged friday.

PR #1086, opened 2024-06-19
CNF-13193: Update NTO Hugepages size validation

fontivan:
- Include valid sizes for both x86 and aarch64 in the validation check
- Included notes for aarch64 where each kernel page size includes only a single valid hugepage size
- After discusion on Slack I have also added some additional validations:
-- Verify all nodes in the list have the same cpu architecture
-- Verify all nodes in the list have the same cpu capacity
-- Verify that the selected hugepage size is valid for the cpu architecture

### DIFF
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
@@ -249,15 +249,8 @@ func (r *PerformanceProfile) validateAllNodesAreSameCpuArchitecture(nodes corev1
 	var allErrs field.ErrorList
 	// First check if the node list has valid elements
 	if len(nodes.Items) == 0 {
-		allErrs = append(allErrs,
-			field.Invalid(
-				field.NewPath("spec.nodeSelector"),
-				r.Spec.NodeSelector,
-				"Failed to detect any nodes, unable to validate architecture",
-			),
-		)
-
-		// If we failed to detect any nodes we cannot do anything more here
+		// We are unable to validate this if we have no nodes
+		// But no nodes is still a valid profile so skip this validation
 		return allErrs
 	}
 
@@ -304,15 +297,8 @@ func (r *PerformanceProfile) validateAllNodesAreSameCpuCapacity(nodes corev1.Nod
 	var allErrs field.ErrorList
 	// First check if the node list has valid elements
 	if len(nodes.Items) == 0 {
-		allErrs = append(allErrs,
-			field.Invalid(
-				field.NewPath("spec.nodeSelector"),
-				r.Spec.NodeSelector,
-				"Failed to detect any nodes, unable to validate cpu capacity",
-			),
-		)
-
-		// If we failed to detect any nodes we cannot do anything more here
+		// We are unable to validate this if we have no nodes
+		// But no nodes is still a valid profile so skip this validation
 		return allErrs
 	}
 
@@ -362,25 +348,20 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
 		return allErrs
 	}
 
-	// First check if the node list has valid elements
-	if len(nodes.Items) == 0 {
-		allErrs = append(allErrs,
-			field.Invalid(
-				field.NewPath("spec.nodeSelector"),
-				r.Spec.NodeSelector,
-				"Failed to detect any nodes, unable to validate hugepages",
-			),
-		)
+	// We can only partially validate this if we have no nodes
+	// We can check that the value used is legitimate but we cannot check
+	// whether it is supposed to be x86 or aarch64
+	x86 := false
+	aarch64 := false
+	combinedHugepagesSizes := append(x86ValidHugepagesSizes, aarch64ValidHugepagesSizes...)
 
-		// If we failed to detect any nodes we cannot do anything more here
-		return allErrs
+	if len(nodes.Items) > 0 {
+		// `validateHugePages` implicitly relies on `validateAllNodesAreSameCpuArchitecture` to have already been run
+		// Under that assumption we can return any node from the list since they should all be the same architecture
+		// However it is simple and easy to just return the first node
+		x86 = isX86(nodes.Items[0])
+		aarch64 = isAarch64(nodes.Items[0])
 	}
-
-	// `validateHugePages` implicitly relies on `validateAllNodesAreSameCpuArchitecture` to have already been run
-	// Under that assumption we can return any node from the list since they should all be the same architecture
-	// However it is simple and easy to just return the first node
-	x86 := isX86(nodes.Items[0])
-	aarch64 := isAarch64(nodes.Items[0])
 
 	if r.Spec.HugePages.DefaultHugePagesSize != nil {
 		defaultSize := *r.Spec.HugePages.DefaultHugePagesSize
@@ -405,13 +386,13 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
 					fmt.Sprintf("%s %v", errMsg, aarch64ValidHugepagesSizes),
 				),
 			)
-		} else if !x86 && !aarch64 {
+		} else if !x86 && !aarch64 && !slices.Contains(combinedHugepagesSizes, string(defaultSize)) {
 			allErrs = append(
 				allErrs,
 				field.Invalid(
 					field.NewPath(errField),
 					r.Spec.HugePages.DefaultHugePagesSize,
-					"Failed to detect architecture, unable to validate default hugepage size",
+					fmt.Sprintf("%s %v", errMsg, combinedHugepagesSizes),
 				),
 			)
 		}
@@ -438,13 +419,13 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
 					fmt.Sprintf("%s %v", errMsg, aarch64ValidHugepagesSizes),
 				),
 			)
-		} else if !x86 && !aarch64 {
+		} else if !x86 && !aarch64 && !slices.Contains(combinedHugepagesSizes, string(page.Size)) {
 			allErrs = append(
 				allErrs,
 				field.Invalid(
 					field.NewPath(errField),
-					r.Spec.HugePages.Pages,
-					"Failed to detect architecture, unable to validate hugepage size",
+					r.Spec.HugePages.DefaultHugePagesSize,
+					fmt.Sprintf("%s %v", errMsg, combinedHugepagesSizes),
 				),
 			)
 		}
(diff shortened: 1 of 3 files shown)


pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -218,14 +228,14 @@ func (r *PerformanceProfile) validateHugePages() field.ErrorList {
 	// validate that default hugepages size has correct value, currently we support only 2M and 1G(x86_64 architecture)
 	if r.Spec.HugePages.DefaultHugePagesSize != nil {
 		defaultSize := *r.Spec.HugePages.DefaultHugePagesSize
-		if defaultSize != hugepagesSize1G && defaultSize != hugepagesSize2M {
-			allErrs = append(allErrs, field.Invalid(field.NewPath("spec.hugepages.defaultHugepagesSize"), r.Spec.HugePages.DefaultHugePagesSize, fmt.Sprintf("hugepages default size should be equal to %q or %q", hugepagesSize1G, hugepagesSize2M)))
+		if !slices.Contains(validHugepagesSizes, defaultSize) {
```
fontivan, line 231: I think it would be great if we could validate that the page size you have selected is valid for your hardware architecture, but I don't think it would be very easy

Hypothetically NTO can detect the architecture using its existing ghw library, but that would only work on the node where NTO is installed/running

In a distributed model NTO could be running on an x86 hub, but managing aarch64 spokes and I'm not sure if there's a way for us to validate that

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -36,9 +37,18 @@ import (
...
 	hugepagesSize1G = "1G"
 )
 
+var validHugepagesSizes = []string{
+	hugepagesSize2M, // Supported on x86 and aarch64 (with 4k kernel pages)
+	hugepagesSize32M, // Supported on aarch64 (with 16k kernel pages)
+	hugepagesSize512M, // Supported on aarch64 (with 64k kernel pages)
+	hugepagesSize1G, // Supported on x86
+}
+
```
ffromani, line 51: I think we should have per-arch sets, not one giant set

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
ffromani: we do have node arch exposed by core k8s as label:
```
$ oc get nodes MYNODE -o json | jq .metadata.labels
{
  "beta.kubernetes.io/arch": "amd64",
  "beta.kubernetes.io/os": "linux",
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "MYNODE",
  "kubernetes.io/os": "linux",
  "node-role.kubernetes.io/worker": "",
  "node.openshift.io/os_id": "rhcos"
}
```
and I think I can build on that. Actually I'd go a step further and add the requirement that all the node in a machine config pool must have the same arch, which seems a pretty fair requirement to me.

MarSik: I would even add the requirement that CPU capacity is the same. Because otherwise the CPU IDs will not work for some nodes.

MarSik:
I like the direction, we should make sure we are not adding too many repeated API server requests though (too many node listings for example).

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -234,6 +447,28 @@ func (r *PerformanceProfile) validateHugePages() field.ErrorList {
 	return allErrs
 }
 
+func (r *PerformanceProfile) getArchitectureForPerfProfile() (string, error) {
```
bartwensley, line 450: The validation of a performance profile isn't particularly performance sensitive, but we should probably avoid calling the k8s API each time we need to check the architecture. Is there a way to get the architecture once per invocation of the validation and just access the result each time we need to check the architecture?

bartwensley (commented):
Looks good so far

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
fontivan: I've updated this in the latest changes, @MarSik also had the same comment

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -234,6 +447,28 @@ func (r *PerformanceProfile) validateHugePages() field.ErrorList {
...
+	nodes, err := r.getNodesList()
+	if err != nil {
+		return "", err
+	}
+
+	// This funtion implicitly relies on `validateAllNodesAreSameCpuArchitecture` to have already been run
```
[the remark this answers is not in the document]
fontivan: Fixed in my latest changes

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -201,31 +227,218 @@ func (r *PerformanceProfile) validateSelectors() field.ErrorList {
...
+	}
+
+	// Make sure all other nodes have the same value
+	for i := 1; i < len(nodes.Items); i++ {
+		if getCpuCapacityForNode(nodes.Items[i]) != expectedCpuCapacity {
+			allErrs = append(allErrs,
+				field.Invalid(
+					field.NewPath("spec.nodeSelector"),
+					r.Spec.NodeSelector,
+					fmt.Sprintf("Node %s is not the same CPU capacity as expected %s", nodes.Items[i].Status.NodeInfo.MachineID, expectedCpuCapacity),
```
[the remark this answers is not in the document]
fontivan: Fixed in my latest changes

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -201,31 +227,218 @@ func (r *PerformanceProfile) validateSelectors() field.ErrorList {
...
+
+
+	// Make sure all other nodes have the same value
+	for i := 1; i < len(nodes.Items); i++ {
+		if getCpuArchitectureForNode(nodes.Items[i]) != expectedArchitecture {
+			allErrs = append(allErrs,
+				field.Invalid(
+					field.NewPath("spec.nodeSelector"),
+					r.Spec.NodeSelector,
+					fmt.Sprintf("Node %s is not the same architecture as expected %s", nodes.Items[i].Status.NodeInfo.MachineID, expectedArchitecture),
```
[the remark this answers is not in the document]
fontivan: Fixed in my latest changes

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -201,31 +238,175 @@ func (r *PerformanceProfile) validateSelectors() field.ErrorList {
...
 	var allErrs field.ErrorList
 
 	if r.Spec.HugePages == nil {
 		return allErrs
 	}
 
-	// validate that default hugepages size has correct value, currently we support only 2M and 1G(x86_64 architecture)
+    // `validateHugePages` implicitly relies on `validateAllNodesAreSameCpuArchitecture` to have already been run
+    // Under that assumption we can return any node from the list since they should all be the same architecture
+    // However it is simple and easy to just return the first node
```
bartwensley, line 341: Indentation looks off - somehow spaces crept in.

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
bartwensley: @MarSik - can you please elaborate on what you mean about the CPU IDs? Is it really OK for us to force all the nodes in a machine config pool to have the same CPU capacity, even if they are the same CPU architecture? Is this something we should clear with PM before enforcing in code?

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -105,9 +129,22 @@ func (r *PerformanceProfile) validateNodeSelectorDuplication(ppList *Performance
...
+				err.Error(),
+			),
+		)
+	}
+
 	allErrs = append(allErrs, r.validateCPUs()...)
 	allErrs = append(allErrs, r.validateSelectors()...)
-	allErrs = append(allErrs, r.validateHugePages()...)
+	allErrs = append(allErrs, r.validateAllNodesAreSameCpuArchitecture(nodes)...)
+	allErrs = append(allErrs, r.validateAllNodesAreSameCpuCapacity(nodes)...)
```
bartwensley, line 146: I'm concerned that this might be too restrictive - added another comment to discuss.

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -323,30 +371,183 @@ var _ = Describe("PerformanceProfile", func() {
...
+			Expect(nodes.Items).ToNot(BeEmpty())
+		})
+		It("should return an error when nothing is detected", func() {
+			// Get client with no nodes to test this case
+			nodeSpecs := []NodeSpecifications{}
+			validatorClient = GetFakeValidatorClient(nodeSpecs)
+
+			// There should be an empty node list and error present
+			nodes, err := profile.getNodesList()
+			Expect(err).ToNot(BeNil())
```
bartwensley, line 393: General comment - I think it is good to always at least partially match the error string to ensure your unit test is hitting the "right" error path.

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
MarSik: All nodes in each MCP must be the same architecture and the same CPU topology. That is documented and even somewhat enforced by NTO (it would generate two separate kernel args sets for the different nodes and NTO would stop applying that due to conflict).

The validation just catches that with a nicer message.

@bartwensley Do you see a flow where this assumption would not be valid?

bartwensley: OK - thanks Martin - makes sense.

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
fontivan: Updated in latest commit

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
fontivan: Fixed in latest commit

ffromani:
if you se a label name like something.which.looks.like.dns.com/something-else it should pass the validation. Alternatively, try NOT using slashes at all in the labels

ffromani:
/approve

typo in https://github.com/openshift/cluster-node-tuning-operator/pull/1086#issuecomment-2427130800

PR #1087, opened 2024-06-20
NO-JIRA:hypershift: client: fix Get() call implementation

Tal-or:
There was a hidden (wrong) assumption that the `ConfigMap` name is equal to the
internal object's name which encapsulated in the `ConfigMap`.

To make sure we return the correct internal object,
we should list all the `ConfigMap`'s in the HCP namespace,
decode the data and look for the correct object name.

Since the decoding operation is relatively expesive,
we added a label `ConfigMapEncapsulatedKindKey` that would
help with listing only `ConfigMap`'s that contains the correct kind.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/hypershift/components/handler.go
+++ b/pkg/performanceprofile/controller/performanceprofile/hypershift/components/handler.go
@@ -225,8 +225,9 @@ func (h *handler) encapsulateObjInConfigMap(instance *corev1.ConfigMap, object c
 		return nil, fmt.Errorf("annotation %q not found in ConfigMap %q annotations", hypershiftNodePoolLabel, client.ObjectKeyFromObject(instance).String())
 	}
 
-	name := fmt.Sprintf("%s-%s", strings.ToLower(object.GetObjectKind().GroupVersionKind().Kind), instance.Name)
-	cm := configMapMeta(name, profileName, instance.GetNamespace(), nodePoolNamespacedName)
+	kind := object.GetObjectKind().GroupVersionKind().Kind
+	name := fmt.Sprintf("%s-%s", strings.ToLower(kind), instance.Name)
+	cm := configMapMeta(name, profileName, instance.GetNamespace(), nodePoolNamespacedName, kind)
 	err = controllerutil.SetControllerReference(instance, cm, h.scheme)
 	if err != nil {
 		return nil, err
@@ -261,14 +262,15 @@ func createOrUpdateKubeletConfigConfigMap(ctx context.Context, cli client.Client
 
 // configMapMeta return a ConfigMap that can be used to encapsulate
 // cluster scoped objects within the desired Namespace
-func configMapMeta(name, profileName, namespace, npNamespacedName string) *corev1.ConfigMap {
+func configMapMeta(name, profileName, namespace, npNamespacedName, ecapsulatedObjKind string) *corev1.ConfigMap {
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
 			Name:      name,
 			Labels: map[string]string{
-				hypershiftPerformanceProfileNameLabel: profileName,
-				hypershiftNodePoolLabel:               parseNamespacedName(npNamespacedName),
+				hypershiftPerformanceProfileNameLabel:   profileName,
+				hypershiftNodePoolLabel:                 parseNamespacedName(npNamespacedName),
+				hypershift.ConfigMapEncapsulatedKindKey: strings.ToLower(ecapsulatedObjKind),
 			},
 			Annotations: map[string]string{
 				hypershiftNodePoolLabel: npNamespacedName,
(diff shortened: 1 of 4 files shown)


pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -65,36 +75,59 @@ func NewControlPlaneClient(c client.Client, ns string) *ControlPlaneClientImpl {
...
+	}
+	cmList := &corev1.ConfigMapList{}
+	listOpts := &client.ListOptions{}
 	if key.Namespace == metav1.NamespaceNone {
-		key.Namespace = ci.managementClusterNamespaceName
+		listOpts.Namespace = ci.managementClusterNamespaceName
+		listOpts.LabelSelector = labelSelector
 	}
-	err := ci.Client.Get(ctx, key, cm, opts...)
+	err := ci.List(ctx, cmList, listOpts)
```
ffromani, line 98: is the object name subject to change or once we discovered initially the right object name we can safely assume it's going to be stable?
I guess it boils down to the question if we can cache the information we discover here.

pkg/performanceprofile/controller/performanceprofile/hypershift/components/handler.go
```diff
@@ -261,14 +262,15 @@ func createOrUpdateKubeletConfigConfigMap(ctx context.Context, cli client.Client
...
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
 			Name:      name,
 			Labels: map[string]string{
-				hypershiftPerformanceProfileNameLabel: profileName,
-				hypershiftNodePoolLabel:               parseNamespacedName(npNamespacedName),
+				hypershiftPerformanceProfileNameLabel:   profileName,
+				hypershiftNodePoolLabel:                 parseNamespacedName(npNamespacedName),
+				hypershift.ConfigMapEncapsulatedKindKey: strings.ToLower(ecapsulatedObjKind),
```
ffromani, line 273: labels are public. Too public. Can we store this information elsewhere?

ffromani (commented):
I praise the effort here. Let's try to solve the root cause, and let's workaround it only if we really don't have better options.

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: The name in this context is the name of the encapsulated object, so changing the name of the object means, changing the `data` in the ConfigMap, which is something that we can cache and track for changes.

ffromani: ok. I'm trying to explore this direction: there could be a predictable and fixed relationship between the outer ConfigMap name and the inner object name/kind? if not, do we know why?

Tal-or: That's depends - 
There are ConfigMaps which are controller fully by NTO controls and there other which are not.
All the object listed below are of course encapsulated in ConfigMaps:

A. Fully controlled by NTO - 
     1. KubletConfig 
     2. MachineConfig
     3. Tuned

B. Not controlled by NTO - 
    1. PerformanceProfile
    2. ContainerRuntimeConfig
    3. Tuned (which applied directly by the user)

ConfigMap that are fully controlled by NTO (group A) has a predicted names with fixed relationship between the internal object and the ConfigMap.

The others (group B) are being provided by the user and mirrored into the hosted control plane namespace (the namespace where NTO operator's pod resides) and we cannot control the naming convention of those.

ffromani: > B. Not controlled by NTO - 1. PerformanceProfile 2. ContainerRuntimeConfig 3. Tuned (which applied directly by the user)
> 
> ConfigMap that are fully controlled by NTO (group A) has a predicted names with fixed relationship between the internal object and the ConfigMap.
> 
> The others (group B) are being provided by the user and mirrored into the hosted control plane namespace (the namespace where NTO operator's pod resides) and we cannot control the naming convention of those.

OK, so I'm rusty on hypershift but thing is: there is still some option because the mirroring agent (hypershift) does control the object creation into the hosted control plane.
However thinking twice this is not helping much because in order to preserve the original metadata (e.g. name) that controller would likely need to use labels/annotations, which is pretty much the same thing we're doing already :\

pkg/performanceprofile/controller/performanceprofile/hypershift/components/handler.go
ffromani: answering myself: no, and these labels aren't accessible by users anyway. There is not a clearly better alternative yet.

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -33,20 +33,20 @@ type ControlPlaneClientImpl struct {
 }
 
 func (ci *ControlPlaneClientImpl) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
-	if ObjIsEncapsulatedInConfigMap(obj) {
+	if EncapsulatedInConfigMap(obj) {
```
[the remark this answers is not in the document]
ffromani: I guess we don't want to do this? why?

Tal-or:
Superseded by https://github.com/openshift/cluster-node-tuning-operator/pull/1094

PR #1088, opened 2024-06-21
OCPBUGS-35911: E2E: Add test to verify runc process excludes the cpus used by pod.

SargunNarula:
Adding a test to verify that runc does not use CPUs assigned to guaranteed pods.

Original bug link - https://bugzilla.redhat.com/show_bug.cgi?id=1910386

 **What this test verifies:**
 - It checks the configuration flow from kubelet -> cri-o -> runc, ensuring the correct CPU
   configuration is provided to runc during the container creation process.

 What this test does **NOT** verify:
 - It does not monitor the actual CPU usage during the container creation process or track
   the CPUs used by any intermediate forks.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -2,6 +2,7 @@ package __performance
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"regexp"
@@ -31,6 +32,7 @@ import (
 	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/controller"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/runtime"
 	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cluster"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/deployments"
@@ -49,6 +51,34 @@ var profile *performancev2.PerformanceProfile
 
 const restartCooldownTime = 1 * time.Minute
 const cgroupRoot string = "/sys/fs/cgroup"
+
+type CPUVals struct {
+	CPUs string `json:"cpus"`
+}
+
+type CPUResources struct {
+	CPU CPUVals `json:"cpu"`
+}
+
+type LinuxResources struct {
+	Resources CPUResources `json:"resources"`
+}
+
+type Process struct {
+	Args []string `json:"args"`
+}
+
+type Annotations struct {
+	ContainerName string `json:"io.kubernetes.container.name"`
+	PodName       string `json:"io.kubernetes.pod.name"`
+}
+
+type ContainerConfig struct {
+	Process     Process        `json:"process"`
+	Hostname    string         `json:"hostname"`
+	Annotations Annotations    `json:"annotations"`
+	Linux       LinuxResources `json:"linux"`
+}
 
 var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 	var (
@@ -892,7 +922,177 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		})
 	})
 
+	Context("Check container runtimes cpu usage", Label(string(label.OpenShift)), func() {
+		var guaranteedPod, bestEffortPod *corev1.Pod
+		var guaranteedPodCpus, guaranteedInitPodCpus cpuset.CPUSet
+		var bestEffortPodCpus, bestEffortInitPodCpus cpuset.CPUSet
+
+		// What this test verifies:
+		// - It checks the configuration flow from kubelet -> cri-o -> runc, ensuring the correct CPU
+		//   configuration is provided to runc during the container creation process.
+		//
+		// What this test does NOT verify:
+		// - It does not monitor the actual CPU usage during the container creation process or track
+		//   the CPUs used by any intermediate forks.
+		It("[test_id: 74461] Verify that runc excludes the cpus used by guaranteed pod", func() {
+			By("Creating a guaranteed pod")
+			guaranteedPod = makePod(ctx, workerRTNode, true)
+			err := testclient.Client.Create(ctx, guaranteedPod)
+			Expect(err).ToNot(HaveOccurred(), "Failed to create guaranteed pod")
+			defer func() {
+				if guaranteedPod != nil {
+					testlog.Infof("deleting pod %q", guaranteedPod.Name)
+					deleteTestPod(ctx, guaranteedPod)
+				}
+			}()
+
+			// This Test is specific to runc container runtime
+			// Skipping this test for crun.
+			Eventually(func() error {
+				expectedRuntime, err := runtime.GetContainerRuntimeTypeFor(context.TODO(), testclient.ControlPlaneClient, guaranteedPod)
+				if err != nil {
+					testlog.Errorf("Failed to fetch runtime for Guaranteed Pod: %v", err)
+					return err
+				}
+				if expectedRuntime == "crun" {
+					Skip(fmt.Sprintf("Skipping test as the runtime is 'crun', which is not the expected runtime. Found: %s", expectedRuntime))
+				}
+				return nil
+			}).WithTimeout(30*time.Second).WithPolling(2*time.Second).Should(Succeed(), "Expected to successfully determine the container runtime type")
+
+			By("Waiting for guaranteed pod to be ready")
+			_, err = pods.WaitForCondition(ctx, client.ObjectKeyFromObject(guaranteedPod), corev1.PodReady, corev1.ConditionTrue, 5*time.Minute)
+			Expect(err).ToNot(HaveOccurred(), "Guaranteed pod did not become ready in time")
+			Expect(guaranteedPod.Status.QOSClass).To(Equal(corev1.PodQOSGuaranteed), "Guaranteed pod does not have the correct QOSClass")
+			testlog.Infof("Guaranteed pod %s/%s was successfully created", guaranteedPod.Namespace, guaranteedPod.Name)
+
+			By("Creating a best-effort pod")
+			bestEffortPod = makePod(ctx, workerRTNode, false)
+			err = testclient.Client.Create(ctx, bestEffortPod)
+			Expect(err).ToNot(HaveOccurred(), "Failed to create best-effort pod")
+			defer func() {
+				if bestEffortPod != nil {
+					testlog.Infof("deleting pod %q", bestEffortPod.Name)
+					deleteTestPod(ctx, bestEffortPod)
+				}
+			}()
+
+			By("Waiting for best-effort pod to be ready")
+			_, err = pods.WaitForCondition(ctx, client.ObjectKeyFromObject(bestEffortPod), corev1.PodReady, corev1.ConditionTrue, 5*time.Minute)
+			Expect(err).ToNot(HaveOccurred(), "Best-effort pod did not become ready in time")
+			testlog.Infof("BestEffort pod %s/%s was successfully created", bestEffortPod.Namespace, bestEffortPod.Name)
+
+			By("Getting Information for guaranteed POD containers")
+			GuPods := getConfigJsonInfo(guaranteedPod, "test", workerRTNode)
+			for _, pod := range GuPods {
+				if pod.Annotations.ContainerName == "test" {
+					guaranteedPodCpus, err = cpuset.Parse(pod.Linux.Resources.CPU.CPUs)
+				} else if pod.Annotations.ContainerName == "POD" {
+					guaranteedInitPodCpus, err = cpuset.Parse(pod.Linux.Resources.CPU.CPUs)
+				}
+				Expect(err).ToNot(HaveOccurred(), "Failed to parse GU POD cpus")
+			}
+
+			By("Getting Information for BestEffort POD containers")
+			BEPods := getConfigJsonInfo(bestEffortPod, "test", workerRTNode)
+			for _, pod := range BEPods {
+				if pod.Annotations.ContainerName == "test" {
+					bestEffortPodCpus, err = cpuset.Parse(pod.Linux.Resources.CPU.CPUs)
+				} else if pod.Annotations.ContainerName == "POD" {
+					bestEffortInitPodCpus, err = cpuset.Parse(pod.Linux.Resources.CPU.CPUs)
+				}
+				Expect(err).ToNot(HaveOccurred(), "Failed to parse BE POD cpus")
+			}
+
+			By("Validating CPU allocation for Guaranteed and Best-Effort pod containers")
+			isolatedCpus, err := cpuset.Parse(string(*profile.Spec.CPU.Isolated))
+			Expect(err).ToNot(HaveOccurred(), "Failed to parse isolated CPU set from performance profile")
+			reservedCpus, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
+			Expect(err).ToNot(HaveOccurred(), "Failed to parse reserved CPU set from performance profile")
+
+			Expect(guaranteedInitPodCpus.IsSubsetOf(reservedCpus)).
+				To(BeTrue(), "Guaranteed Init pod CPUs (%s) are not strictly within the reserved set (%s)", guaranteedInitPodCpus, reservedCpus)
+			Expect(guaranteedInitPodCpus.IsSubsetOf(isolatedCpus)).
+				To(BeFalse(), "Guaranteed Init pod CPUs (%s) are within the isolated cpu set (%s)", guaranteedInitPodCpus, isolatedCpus)
+			Expect(guaranteedPodCpus.IsSubsetOf(isolatedCpus)).
+				To(BeTrue(), "Guaranteed pod CPUs (%s) are not strictly within the isolated set (%s)", guaranteedPodCpus, isolatedCpus)
+
+			availableForBestEffort := isolatedCpus.Union(reservedCpus).Difference(guaranteedPodCpus)
+			Expect(bestEffortInitPodCpus.IsSubsetOf(reservedCpus)).
+				To(BeTrue(), "Best-Effort Init pod CPUs (%s) include CPUs not allowed (%s)", bestEffortInitPodCpus, availableForBestEffort)
+			Expect(bestEffortPodCpus.IsSubsetOf(availableForBestEffort)).
+				To(BeTrue(), "Best-Effort pod CPUs (%s) include CPUs not allowed (%s)", bestEffortPodCpus, availableForBestEffort)
+		})
+	})
+
 })
+
+func extractConfigInfo(output string) (*ContainerConfig, error) {
+	var config ContainerConfig
+	output = strings.TrimSpace(output)
+	err := json.Unmarshal([]byte(output), &config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal config.json: %v", err)
+	}
+	return &config, nil
+}
+
+func getConfigJsonInfo(pod *corev1.Pod, containerName string, workerRTNode *corev1.Node) []*ContainerConfig {
+	var pods []*ContainerConfig
+	path := "/rootfs/var/lib/containers/storage/overlay-containers/"
+	podName := pod.Name
+	cmd := []string{
+		"/bin/bash", "-c",
+		fmt.Sprintf(
+			`find %s -type f -exec grep -lP '\"io.kubernetes.pod.name\": \"%s\"' {} \; -exec grep -l '\"io.kubernetes.container.name\": \"%s\"' {} \; | sort -u`,
+			path, podName, containerName,
+		),
+	}
+	output, err := nodes.ExecCommand(context.TODO(), workerRTNode, cmd)
+	Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Failed to search for config.json with podName %s and containerName %s", podName, containerName))
+	filePaths := strings.Split(string(output), "\n")
+	for _, filePath := range filePaths {
+		if filePath == "" {
+			continue
+		}
+		cmd = []string{"/bin/bash", "-c", fmt.Sprintf("cat %s", filePath)}
+		output, err = nodes.ExecCommand(context.TODO(), workerRTNode, cmd)
+		Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Failed to read config.json for container : %s", filePath))
+
+		configData := testutils.ToString(output)
+		config, err := extractConfigInfo(configData)
+		if err != nil {
+			testlog.Errorf("Error extracting config info:", err)
+			continue
+		}
+		pods = append(pods, config)
+		testlog.Infof("Pod Name: %s", config.Annotations.PodName)
+		testlog.Infof("Container Name: %s", config.Annotations.ContainerName)
+		testlog.Infof("Hostname: %s", config.Hostname)
+		testlog.Infof("Arguments: %s", config.Process.Args)
+		testlog.Infof("CPUs: %s", config.Linux.Resources.CPU.CPUs)
+	}
+	return pods
+}
+
+func makePod(ctx context.Context, workerRTNode *corev1.Node, guaranteed bool) *corev1.Pod {
+	testPod := pods.GetTestPod()
+	testPod.Namespace = testutils.NamespaceTesting
+	testPod.Spec.NodeName = workerRTNode.Name
+	testPod.Spec.NodeSelector = map[string]string{testutils.LabelHostname: workerRTNode.Name}
+	if guaranteed {
+		testPod.Spec.Containers[0].Resources = corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("2"),
+				corev1.ResourceMemory: resource.MustParse("200Mi"),
+			},
+		}
+	}
+	profile, _ := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+	runtimeClass := components.GetComponentName(profile.Name, components.ComponentNamePrefix)
+	testPod.Spec.RuntimeClassName = &runtimeClass
+	return testPod
+}
 
 func checkForWorkloadPartitioning(ctx context.Context) bool {
 	// Look for the correct Workload Partition annotation in


test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		})
 	})
 
+	Context("Runc cpus isolation", func() {
```
mrniranjan, line 754: This test case is not limited to runc,  since we execute across all container runtimes, you can rename this check container runtimes cpu usage

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+			for _, guaranteedCpu := range guaranteedPodCpus {
+				for _, runcCpu := range runcCpus {
+					if guaranteedCpu == runcCpu {
+						overlapFound = true
+						break
+					}
+				}
+				if overlapFound {
+					break
+				}
```
mrniranjan, line 835: I am confused by the above for loop,  the bug says we need to check if runc process is not using any of the cpus assigned to the pod, what exactly does this for loop achieve ?

SargunNarula: The loop here is to check if there is any common cpu present between the runc cpus and pod cpus after capturing them from config.json. we do not expect any common cpu number to be present.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+				if _, exists := uniqueCombinations[combination]; !exists {
+					uniqueCombinations[combination] = struct{}{}
+					zippedMatches = append(zippedMatches, map[string]string{
+						"hostname": hostnameMatches[i][1],
+						"cpus":     cpusMatches[i][1],
+					})
+				}
+			}
+
+			parseCPUs := func(cpuStr string) []int {
```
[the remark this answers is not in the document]
SargunNarula: Ok, will go by this method. Thanks for clarifying.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: For that you don't need to use the loop, once you get the cpu's used by runtime(runc) ,   you  can use the existing method of getting the pod cpus and compare them using Expect. ?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: we execute the test cases on both runtimes, so make sure that this test works on crun too.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+		AfterEach(func() {
+			cmd := []string{"rm", "-f", "/rootfs/var/roothome/create"}
+			nodes.ExecCommand(ctx, workerRTNode, cmd)
+			deletePod(ctx, guaranteedPod)
+			deletePod(ctx, secondPod)
+		})
+
+		It("[test_id: 74461] Verify that runc excludes the cpus used by guaranteed pod", func() {
+			guaranteedPod = getGuaranteedPod(ctx, workerRTNode)
+			secondPod = getNonGuaranteedPod(ctx, workerRTNode)
```
shajmakh, line 766: can we give it more meaningful name like bestEffortPod or so?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+
+			out := testutils.ToString(output)
+			hostnameRe := regexp.MustCompile(`"hostname":\s+"([^"]+)"`)
+			cpusRe := regexp.MustCompile(`"cpus":\s+"([^"]+)"`)
+
+			hostnameMatches := hostnameRe.FindAllStringSubmatch(out, -1)
+			cpusMatches := cpusRe.FindAllStringSubmatch(out, -1)
+			if len(hostnameMatches) == 0 || len(cpusMatches) == 0 {
+				Fail("Failed to extract hostname or cpus information")
+			}
```
shajmakh, line 791: why not use gomega Expect?

shajmakh (commented):
Thanks for this. I added few initial comments below

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: Can you review the new way, I tried with Intersection()

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		})
 	})
 
+	Context("Runc cpus isolation", func() {
+		var guaranteedPod, secondPod *corev1.Pod
+
+		AfterEach(func() {
+			cmd := []string{"rm", "-f", "/rootfs/var/roothome/create"}
+			nodes.ExecCommand(ctx, workerRTNode, cmd)
```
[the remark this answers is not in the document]
SargunNarula: The test fails at this line with context deadline error, every time Expect error is added. I am manually verified that If expect is removed, the remove task gets completed at the system. I tried using custom context with context.WithDeadline & context.WithTimeout but still the test failed. This happens at the next error handling part as well.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+	testpod2.Spec.RuntimeClassName = &runtimeClass
+
+	err = testclient.Client.Create(ctx, testpod2)
+	Expect(err).ToNot(HaveOccurred())
+	_, err = pods.WaitForCondition(ctx, client.ObjectKeyFromObject(testpod2), corev1.PodReady, corev1.ConditionTrue, 5*time.Minute)
+	Expect(err).ToNot(HaveOccurred())
+	return testpod2
+}
+
+func deletePod(ctx context.Context, pod *corev1.Pod) {
```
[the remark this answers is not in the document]
SargunNarula: Removed this function with latest commit. Thanks

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+
+			var err error
+			containerIDs[0], err = pods.GetContainerIDByName(guaranteedPod, "test")
+			Expect(err).ToNot(HaveOccurred())
+			containerIDs[1], err = pods.GetContainerIDByName(secondPod, "test")
+			Expect(err).ToNot(HaveOccurred())
+			for _, containerID := range containerIDs {
+				path := fmt.Sprintf("/rootfs/var/run/containers/storage/overlay-containers/%s/userdata/config.json", containerID)
+				cmd := []string{"/bin/bash", "-c", fmt.Sprintf("cat %s >> /rootfs/var/roothome/create", path)}
+				nodes.ExecCommand(ctx, workerRTNode, cmd)
```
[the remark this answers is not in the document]
SargunNarula: Added a comment above regarding facing context deadline error, it is applicable here too.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: Improved with latest commit. Thanks

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -751,8 +751,148 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+					break
+				}
+			}
+			Expect(overlapFound).ToNot(BeTrue(), "Overlap of cpus found, not expected behaviour")
+		})
+	})
+
 })
 
+func getGuaranteedPod(ctx context.Context, workerRTNode *corev1.Node) *corev1.Pod {
```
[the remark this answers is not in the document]
SargunNarula: Moved the pod creation steps under the main spec. Thanks

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: I verified the test execution on a crun runtime enabled cluster as well, It passed successfully.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: Updated with better name.

mrniranjan:
looks good to me from my side.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,106 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+	testpod2.Namespace = testutils.NamespaceTesting
+	testpod2.Spec.NodeSelector = map[string]string{testutils.LabelHostname: workerRTNode.Name}
+
+	profile, _ := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+	runtimeClass := components.GetComponentName(profile.Name, components.ComponentNamePrefix)
+	testpod2.Spec.RuntimeClassName = &runtimeClass
+
+	return testpod2, nil
+}
+
```
shajmakh, line 989: if I'm not missing anything, these two can be one function that returns a pod (the second version) and where you call it you can then adjust the pod to make it guaranteed/burstable or whatever QoS

SargunNarula: Revised, to a single method. Thanks

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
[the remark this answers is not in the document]
shajmakh: I wouldn't completely ignore the error though, a warning report or a klog.Error is fine.
also we can make sure that after this command is executed the file was indeed deleted as a workaround for handling the error

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,94 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+		var guaranteedPod, bestEffortPod *corev1.Pod
+
+		AfterEach(func() {
+			cmd := []string{"rm", "-f", "/rootfs/var/roothome/create"}
+			nodes.ExecCommand(ctx, workerRTNode, cmd)
+			deleteTestPod(ctx, guaranteedPod)
+			deleteTestPod(ctx, bestEffortPod)
+		})
+
+		It("[test_id: 74461] Verify that runc excludes the cpus used by guaranteed pod", func() {
```
shajmakh, line 900: let's please add By() and logs where it fits to help view the flow of the test when debugging

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,94 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+
+			containerIDs := make([]string, 2)
+
+			containerIDs[0], err = pods.GetContainerIDByName(guaranteedPod, "test")
+			Expect(err).ToNot(HaveOccurred())
+			containerIDs[1], err = pods.GetContainerIDByName(bestEffortPod, "test")
+			Expect(err).ToNot(HaveOccurred())
+			for _, containerID := range containerIDs {
+				path := fmt.Sprintf("/rootfs/var/run/containers/storage/overlay-containers/%s/userdata/config.json", containerID)
+				cmd := []string{"/bin/bash", "-c", fmt.Sprintf("cat %s >> /rootfs/var/roothome/create", path)}
```
ffromani, line 922: why do we need to copy the data into an intermediate file and can't we read it directly and store in a variable in this test?
if we need an itermediate file, since we go in append mode, let's make sure the file is actually empty before the test starts. If because a bug the file is not created properly, we will have corrupted state very hard to notice.
Another approach (possibly better) is to create a temp file for this test.

I think the simpler solution is to just avoid the intermediate file

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,94 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+			runcCpus, err := cpuset.Parse(zippedMatches[1]["cpus"])
+			Expect(err).ToNot(HaveOccurred())
+			overlapFound := !guaranteedPodCpus.Intersection(runcCpus).IsEmpty()
+			Expect(overlapFound).ToNot(BeTrue(), fmt.Sprintf("Overlap found between guaranteedPod cpus (%s) and runtime Cpus (%s), not expected behaviour", guaranteedPodCpus, runcCpus))
+		})
+	})
+
 })
 
+func getPod(ctx context.Context, workerRTNode *corev1.Node, guaranteed bool) (*corev1.Pod, error) {
```
ffromani, line 960: if this can't fail , let's just omit the error return value.
I'd call it `makePod` rather than `getPod`

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,94 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+						"cpus":     cpusMatches[i][1],
+					})
+				}
+			}
+			guaranteedPodCpus, err := cpuset.Parse(zippedMatches[0]["cpus"])
+			Expect(err).ToNot(HaveOccurred())
+			runcCpus, err := cpuset.Parse(zippedMatches[1]["cpus"])
+			Expect(err).ToNot(HaveOccurred())
+			overlapFound := !guaranteedPodCpus.Intersection(runcCpus).IsEmpty()
+			Expect(overlapFound).ToNot(BeTrue(), fmt.Sprintf("Overlap found between guaranteedPod cpus (%s) and runtime Cpus (%s), not expected behaviour", guaranteedPodCpus, runcCpus))
```
ffromani, line 954: or
```
overlap := guaranteedPodCpus.Intersection(runcCpus).List()
Expect(overlap).To(BeEmpty(), "Overlap found between guaranteedPod cpus (%s) and runtime Cpus (%s), not expected behaviour", guaranteedPodCpus, runcCpus)
```

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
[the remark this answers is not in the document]
ffromani: +1 to what Shereen wrote

ffromani (commented):
the testing logic can be maybe simplified, but no major objections it seems
questions and possible improvements inside

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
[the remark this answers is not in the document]
SargunNarula: I have removed the intermediate file logic, hence this command is not required. On the other places added the error handling.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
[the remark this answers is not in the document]
SargunNarula: With new logic, file handling is not performed. Yet error handling is added at places required.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: Thanks @ffromani , for the simpler solution advice. Revised with latest commit.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: Revised with latest commit.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: This is a much better way, Thanks. Revised.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,104 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+				configData += testutils.ToString(output)
+			}
+
+			By("Extracting hostname and cpus information from config data")
+			hostnameRe := regexp.MustCompile(`"hostname":\s+"([^"]+)"`)
+			cpusRe := regexp.MustCompile(`"cpus":\s+"([^"]+)"`)
+			hostnameMatches := hostnameRe.FindAllStringSubmatch(configData, -1)
+			cpusMatches := cpusRe.FindAllStringSubmatch(configData, -1)
+			Expect(len(hostnameMatches)).ToNot(Equal(0), "Failed to extract hostname information")
+			Expect(len(cpusMatches)).ToNot(Equal(0), "Failed to extract cpus information")
```
shajmakh, line 942: nit: Expect(<slice name>).ToNot(Equal(0)) = Expect(<slice name>).ToNot(BeEmpty())

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,104 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+			Expect(err).ToNot(HaveOccurred(), "Failed to create guaranteed pod")
+
+			By("Waiting for guaranteed pod to be ready")
+			_, err = pods.WaitForCondition(ctx, client.ObjectKeyFromObject(guaranteedPod), corev1.PodReady, corev1.ConditionTrue, 5*time.Minute)
+			Expect(err).ToNot(HaveOccurred(), "Guaranteed pod did not become ready in time")
+			Expect(guaranteedPod.Status.QOSClass).To(Equal(corev1.PodQOSGuaranteed), "Guaranteed pod does not have the correct QOSClass")
+
+			By("Creating a best-effort pod")
+			bestEffortPod = makePod(ctx, workerRTNode, false)
+			err = testclient.Client.Create(ctx, bestEffortPod)
```
shajmakh, line 910: how do we guarantee that the pod will be created with a different name from the GU pod?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,104 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+			By("Creating a guaranteed pod")
+			guaranteedPod = makePod(ctx, workerRTNode, true)
+			err := testclient.Client.Create(ctx, guaranteedPod)
+			Expect(err).ToNot(HaveOccurred(), "Failed to create guaranteed pod")
+
+			By("Waiting for guaranteed pod to be ready")
+			_, err = pods.WaitForCondition(ctx, client.ObjectKeyFromObject(guaranteedPod), corev1.PodReady, corev1.ConditionTrue, 5*time.Minute)
+			Expect(err).ToNot(HaveOccurred(), "Guaranteed pod did not become ready in time")
+			Expect(guaranteedPod.Status.QOSClass).To(Equal(corev1.PodQOSGuaranteed), "Guaranteed pod does not have the correct QOSClass")
+
```
shajmakh, line 907: an info report would be helpful here, something like:
`klog.InfoS("<QoS> pod %s/%s was successfully created", updatedPodObj.Namespace,updatedPodObj.Name)`
same for the best-effort

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -887,8 +887,104 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		})
 	})
 
+	Context("Check container runtimes cpu usage", func() {
+		var guaranteedPod, bestEffortPod *corev1.Pod
+		AfterEach(func() {
+			By("Cleaning up pods")
+			deleteTestPod(ctx, guaranteedPod)
+			deleteTestPod(ctx, bestEffortPod)
+		})
```
shajmakh, line 896: you can avoid after each because the creation of the pods is done in the specific It(), you can convert that to [defer](https://go.dev/tour/flowcontrol/12) after each pods creation (as close as possible to the creation):

```
guaranteedPod = makePod(ctx, workerRTNode, true)
err := testclient.Client.Create(ctx, guaranteedPod)
Expect(err).ToNot(HaveOccurred(), "Failed to create guaranteed pod")
defer deleteTestPod(ctx, guaranteedPod)
```

ffromani (commented):
I'm not sure this tests checks the correct thing. We do check a BE pod has no overlap with CPUs exclusively assigned to a Guaranteed pod, but the problem here is not what happens at runtime, but what happened at pod creation time. Once the pod goes running, `runc` is terminated, and there's no trace of where did it run

SargunNarula:
@ffromani The original issue identified was that when launching a guaranteed pod running a cyclic test, the runc container creation process was observed to be running on isolated CPUs. This process inadvertently utilized the CPUs allocated to the cyclic test.

The resolution involved ensuring that the cpuset.cpus configuration is passed during container creation.

Additionally, since runc follows a two-step creation process, the initialization process (executed as /usr/bin/pod, which is a symlink to /usr/bin/runc) is started within a container. This container is assigned the cpuset.cpus values. This behavior can be confirmed by examining the config.json of the initialization container to verify that the appropriate CPU allocation is applied, reserved CPUs in the case of a guaranteed pod, or all available CPUs in the case of a Best-Effort (BE) pod.

Reference: 
- https://github.com/opencontainers/runc/blob/9b3fe30e19273e88aaa2812e918fdc100d0d49be/libcontainer/README.md?plain=1#L18

- https://github.com/kubernetes/kubernetes/pull/98019

Based on these observations, the current patch may not effectively validate this scenario. I will work on a revised patch to accurately verify the CPUs being utilized.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: Added log report for both Pods.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: Revised with defer, Thanks for this suggestion.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -889,8 +918,154 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		})
 	})
 
+	Context("Check container runtimes cpu usage", func() {
+		var guaranteedPod, bestEffortPod *corev1.Pod
+		var guaranteedPodCpus, guaranteedInitPodCpus cpuset.CPUSet
+		var bestEffortPodCpus, bestEffortInitPodCpus cpuset.CPUSet
```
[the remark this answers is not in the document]
SargunNarula: Init containers here refer to the `runc` container used to initialize and deploy the main pod container.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -889,8 +918,154 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+			_, err = pods.WaitForCondition(ctx, client.ObjectKeyFromObject(guaranteedPod), corev1.PodReady, corev1.ConditionTrue, 5*time.Minute)
+			Expect(err).ToNot(HaveOccurred(), "Guaranteed pod did not become ready in time")
+			Expect(guaranteedPod.Status.QOSClass).To(Equal(corev1.PodQOSGuaranteed), "Guaranteed pod does not have the correct QOSClass")
+			testlog.Infof("Guranteed pod %s/%s was successfully created", guaranteedPod.Namespace, guaranteedPod.Name)
+			defer func() {
+				if guaranteedPod != nil {
+					testlog.Infof("deleting pod %q", guaranteedPod.Name)
+					deleteTestPod(ctx, guaranteedPod)
+				}
+			}()
```
rbaturov, line 942: Is this test expected to run on hypershift as well?
If yes please change `testclient.Client `to `testclient.controlplane` and `testclient.controlplane`.
If not please add the openshift label as we did previously (string(label.OpenShift))

rbaturov: Also, I think you should move the `defer func()` to be done right after the creation (line 931).

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -889,8 +918,154 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+	var config ContainerConfig
+	output = strings.TrimSpace(output)
+	err := json.Unmarshal([]byte(output), &config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal config.json: %v", err)
+	}
+	return &config, nil
+}
+
+func getConfigJsonInfo(pod *corev1.Pod, containerName string, workerRTNode *corev1.Node) []*ContainerConfig {
```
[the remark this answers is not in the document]
ffromani: > One approach to testing is to verify whether `runc`/`crun` isolates itself from the isolated CPUs by following the method defined in the OCI specification, which involves generating a `config.json` and adhering to the specified CPUs based on the applied profile.

Well, does the process which writes `config.json` run only on reserved CPUs? And does `crun` (or `runc`) honor the settings on `config.json` when setting its own affinity? IOW, who decides and who enforces that `runc` (or `crun`) only runs on reserved CPUs?

your test seems fine for workload, but there are open questions for the infra proper

> Alternatively, a tracing tool like `bpftrace` can be used to monitor all `runc` calls and inspect their CPU affinity. However, this method is quite invasive and may be challenging to implement IMO.

I agree, but the problem here is first and foremost testing the right thing.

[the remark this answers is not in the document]
SargunNarula: @ffromani Do you have any other way in mind?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
SargunNarula: Resolved, Thanks

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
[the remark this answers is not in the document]
SargunNarula: To ensure the check is robust, I have added another `Expect` to verify whether it honors the provided CPUs through the profile. You can review the change here: [Line](https://github.com/openshift/cluster-node-tuning-operator/blob/143f8310134136760a5082c520e060780ac1f869/test/e2e/performanceprofile/functests/1_performance/cpu_management.go#L993) The latest commit addresses your concerns.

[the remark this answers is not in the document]
Tal-or: @SargunNarula would it be correct to generalize the problem as whether host's processes are respecting the isolated cpus, when they are allocated exclusively for a Guaranteed pod? 

Another suggestion I had in mind is to examine if we can add a test for that on crun/runc repo.
Maybe the test's context there would be more suitable for this kind of verification.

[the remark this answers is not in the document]
SargunNarula: @Tal-or Yes, the problem statement is correct. However, I cannot definitively state which other host processes we might consider for this purpose.

> Create a short shell script that runs and prints its associated CPUs

The concept of using a shell script was addressed in the original verification process, where we employed a wrapper around the runc process to capture and log its associated CPUs to a separate file. The challenge, however, was that despite replacing the binary with the wrapper, capturing the CPUs remained difficult due to the extremely short-lived nature of the runc process. The original test logic was designed to address that testing approach, as detailed in this [link](https://github.com/SargunNarula/cluster-node-tuning-operator/blob/b4dfe4a03da80f0d9ec1131cbd958edca9514f8a/test/e2e/performanceprofile/functests/1_performance/cpu_isolation.go).

> Add a test for that in the crun/runc repository

For `runc`, it might be feasible to design a test. However, verifying its behavior with CPU ranges defined by the performance profile presents a challenge, where this test could be valuable.

For `crun`, this test would not be applicable due to its distinct workflow. Unlike `runc`, `crun` does not rely on init containers. Instead, its workflow involves the following steps:
- Creating a temporary bind mount of the current executable.
- Remounting it as read-only.
- Opening the remounted file.
- Unmounting and deleting the temporary file.
- Using the file descriptor to execute the program.

This difference in implementation makes this testing approach unsuitable for crun.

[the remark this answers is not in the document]
Tal-or: >the challenge, however, was that despite replacing the binary with the wrapper, capturing the CPUs remained difficult due to the extremely short-lived nature of the runc process. 

If we're creating our own shell script for test purposes we can make it as longer as needed and the script will print the CPUs. The script doesn't have to be related to `runc`/`crun` we only want to make sure it doesn't violate the isolated cpus isolation. 

> as detailed in this [link](https://github.com/SargunNarula/cluster-node-tuning-operator/blob/b4dfe4a03da80f0d9ec1131cbd958edca9514f8a/test/e2e/performanceprofile/functests/1_performance/cpu_isolation.go). 

The link is a go file, not sure if that's what you meant.

>For runc, it might be feasible to design a test. However, verifying its behavior with CPU ranges defined by the performance profile presents a challenge, where this test could be valuable.

>For crun, this test would not be applicable due to its distinct workflow. Unlike runc, crun does not rely on init containers. Instead, its workflow involves the following steps:

>Creating a temporary bind mount of the current executable.
Remounting it as read-only.
Opening the remounted file.
Unmounting and deleting the temporary file.
Using the file descriptor to execute the program.
This difference in implementation makes this testing approach unsuitable for crun.

Ok too complex, I get it.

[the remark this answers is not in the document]
SargunNarula: If we aim to check the CPU usage of the shell script, we could extend its runtime. However, this would deviate from the primary objective, which is to verify whether the `runc` process is violating CPU isolation. 

In the future, this might need to be extended to other processes, but as of now, that is not a requirement. The original bug specifically pertains to the `runc` process utilizing isolated CPUs. [Bug link](https://bugzilla.redhat.com/show_bug.cgi?id=1910386)

The Go file contains a single test for this bug, utilizing the shell script approach. [Link](https://github.com/SargunNarula/cluster-node-tuning-operator/blob/b4dfe4a03da80f0d9ec1131cbd958edca9514f8a/test/e2e/performanceprofile/functests/1_performance/cpu_isolation.go)

MarSik (approved):

PR #1089, opened 2024-06-23
CNF-12680:CNF-13488: Add performance profile status for hypershift support

rbaturov:
This PR follows the implementation of this enhancement:
https://github.com/openshift/enhancements/pull/1619

Up until now, there was no defined resolution over how PerformanceProfile
status conditions with the status of the different components created
from the PerformanceProfile (MachineConfig, KubeletConfig, Tuned) are handled and exposed.
To resolve this, a custom ConfigMap is added and will be used as a middleware object.
It will hold the updated status populated by the performance profile controller
and will be watched by the NodePool controller, which will in turn reflect them under NodePool.status.condition.

- [x]  Implementation
- [ ]  E2E config suite adoption

There are E2E tests that are required to be added. we can do so once we adjust the `performanceprofile_controller_test.go` (or add another controller test for the hypershift implementation)

Some additional gaps, that we might address if possible:
* In this implementation, KubeletConfig status is not collected (on hypershift)


Tal-or:
/retitle CNF-12680: Add performance profile status for hypershift support

Tal-or:
/hold
e2e tests are in progress

Tal-or:
/hold cancel
This PR is reviewable and includes implementation unit-tests and e2e tests

Tal-or:
/retitle: CNF-12680:CNF-13488: Add performance profile status for hypershift support

pkg/performanceprofile/controller/performanceprofile/status/status.go
```diff
@@ -261,3 +266,52 @@ func removeUnMatchedTunedProfiles(nodes []corev1.Node, profiles []tunedv1.Profil
 	}
 	return filteredProfiles
 }
+
+func CalculateUpdatedStatus(prevStatus *performancev2.PerformanceProfileStatus, profileName string, conditions []conditionsv1.Condition) *performancev2.PerformanceProfileStatus {
```
ffromani: nit: the name of the function is a bit clunky: `status.CalculateUpdatedStatus` but I don't have good suggestions, so this is totally not-blocking (and can be addressed last, or even later)

ffromani: minor: returning an object (not a pointer) signals we are making a copy, especially if we deference it anyway in the call sites; alternatively we can make the receiver an object, not a pointer.

pkg/operator/hypershift.go
```diff
@@ -307,3 +307,7 @@ func generatedConfigMapAnnotations(nodePoolName string) map[string]string {
 		hypershiftNodePoolLabel: nodePoolName,
 	}
 }
+
+func GetTunedUniqueName(tunedName, nodePoolName string) string {
```
ffromani, line 311: minor/nit: `Get` Unique Name or `Make` Unique Name?

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+	hypershiftconsts "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/hypershift/consts"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/status"
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
+)
+
+const (
+	ppStatusConfigMapPrefix = "status"
+)
+
+func getStatusConfigMapName(instanceName string) string {
```
ffromani: minor/nit: `get` or `make` status config map name?

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	key := client.ObjectKeyFromObject(cm)
+	if prevStatusFound {
+		klog.InfoS("Updating status", "ConfigMap", key.String())
+		if err := cli.Update(ctx, cm); err != nil {
+			return fmt.Errorf("failed to update status ConfigMap %q: %w", key.String(), err)
+		}
+		klog.InfoS("Status updated ", "ConfigMap", key.String())
+	} else {
```
ffromani, line 55: minor/nit: please remove the `else` block and align to the left

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+		klog.InfoS("Creating status", "ConfigMap", key.String())
+		if err := cli.Create(ctx, cm); err != nil {
+			return fmt.Errorf("failed to create ConfigMap %q: %w", key.String(), err)
+		}
+		klog.InfoS("Status created", "ConfigMap", key.String())
+	}
+	return nil
+}
+
+func ppStatusConfigMap(instance *corev1.ConfigMap, profileName string, scheme *runtime.Scheme) (*corev1.ConfigMap, error) {
```
ffromani, line 65: minor/nit: `makePerformanceProfileStatusConfigMap`?

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	messageString := status.GetTunedProfilesMessage(tunedProfileList.Items)
+	if len(messageString) == 0 {
+		return nil, nil
+	}
+	return status.GetDegradedConditions(status.ConditionReasonTunedDegraded, messageString), nil
+}
+
+func getPreviousStatusFrom(ctx context.Context, cli client.Client, cm *corev1.ConfigMap) (*performancev2.PerformanceProfileStatus, bool, error) {
```
ffromani, line 94: non-blocking question: is this get cached somehow somewhere (client-go perhaps) right?

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -235,3 +266,7 @@ func DecodeManifest(b []byte, scheme *runtime.Scheme, obj runtime.Object) (bool,
 	}
 	return false, err
 }
+
+func GetStatusConfigMapName(instanceName string) string {
```
ffromani, line 336: minor./nit: `get` or `make`?
didn't we have something similar or the same code elsewhere?

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -32,6 +32,30 @@ type ControlPlaneClientImpl struct {
 	// hostedControlPlaneNamespaceName is the namespace name on the management cluster
 	// on which the control-plane objects reside
 	hostedControlPlaneNamespaceName string
+
```
ffromani, line 35: this is well intentioned, but two mode of operations create room for bugs. We need to evaluate the opportunity cost here

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -90,6 +90,14 @@ func (ci *ControlPlaneClientImpl) Update(ctx context.Context, obj client.Object,
 	return ci.Client.Update(ctx, obj, opts...)
 }
 
+func (ci *ControlPlaneClientImpl) Delete(ctx context.Context, obj client.Object, opts ...client.DeleteOption) error {
```
ffromani, line 97: please add in the commit message a line or few about the intended usage for the `Delete` API

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -33,7 +33,8 @@ type ControlPlaneClientImpl struct {
 	// on which the control-plane objects reside
 	hostedControlPlaneNamespaceName string
 
-	opts controlPlaneClientImplOptions
+	yamlSerializer *serializer.Serializer
```
ffromani, line 36: ok, but why? performances?

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
```diff
@@ -3,6 +3,7 @@ package tuned
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
```
ffromani, line 6: I'm not against this approach, but there should be another PR which adds the bindata compilation anyway (#1019 ?). Let's sync about this.

ffromani (commented):
the hottest topic is the testing mode, which is well intentioned but I'm afraid it can backfire later on. Let's spend more time (let me spend more time) assessing the approach. Roughly half of the changes are straightforward fixes or changes which seem fine already. Will need at least another review pass.

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: Spare the need of recreating the serializer on every call

test/e2e/performanceprofile/functests/utils/tuned/tuned.go
Tal-or: Good. I went with the opposite approach but probably in your case it was a must.
I suggest to apply https://github.com/openshift/cluster-node-tuning-operator/pull/1019/commits/426451af8e24552b16c1c339ddbf04b6469f766d on separate PR and the sooner the better.

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
Tal-or: Caching when we're having two clients (i.e. two etcds) is something that deserves more intention and deeper investigation. ATM I do not have all the answers unfortunately, but what I'm knowing for sure is that we're caching only the hosted-cluster objects:
https://github.com/openshift/cluster-node-tuning-operator/blob/master/cmd/cluster-node-tuning-operator/main.go#L137

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: I'm pretty sure I did - from the commit message:
```
The `Delete` call deletes the `ConfigMap` that holds an encoded data of the object that was given for deletion.
```

Tal-or:
/retitle CNF-12680:CNF-13488: Add performance profile status for hypershift support

pkg/performanceprofile/controller/performanceprofile/status/status.go
Tal-or: > minor: returning an object (not a pointer) signals we are making a copy, especially if we deference it anyway in the call sites; alternatively we can make the receiver an object, not a pointer.

We can change it to an actual object. The only reason I see for keeping it a pointer is because it's easier to check whether the status got updated or not, by checking if the function returned a `nil`.
We can add additional argument for signaling that tough

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
Tal-or: similar to: https://github.com/openshift/cluster-node-tuning-operator/blob/570bbfd7e761611f58ad20f3e478aec1f044a814/pkg/performanceprofile/controller/performanceprofile/components/utils.go#L16

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: We do, I made this one: https://github.com/openshift/cluster-node-tuning-operator/pull/1089#discussion_r1685683711 public

pkg/operator/hypershift.go
Tal-or: Done. Thanks

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
Tal-or: Done. Thank you

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
Tal-or: Done

Tal-or:
/hold
Depends on #1116

Tal-or:
> This PR follows the implementation of this enhancement:
https://github.com/openshift/enhancements/pull/1619
>
> Up until now, there was no defined resolution over how PerformanceProfile
status conditions with the status of the different components created
from the PerformanceProfile (MachineConfig, KubeletConfig, Tuned) are handled and exposed.
To resolve this, a custom ConfigMap is added and will be used as a middleware object.
It will hold the updated status populated by the performance profile controller
and will be watched by the NodePool controller, which will in turn reflect them under NodePool.status.condition.
>
> Implementation
 E2E config suite adoption
There are E2E tests that are required to be added. we can do so once we adjust the performanceprofile_controller_test.go (or add another controller test for the hypershift implementation)
>
>Some additional gaps, that we might address if possible:
>
>In this implementation, KubeletConfig status is not collected (on hypershift)

- [x] Implementation
- [x]  E2E config suite adoption 
- [x] performanceprofile_controller_test.go (or add another controller test for the hypershift implementation) 

In this implementation, KubeletConfig status is not collected (on hypershift) - And it probably won't since the status is not reflected by KubeletConfig on Hypershift at all.

Tal-or:
/hold cancel

#1116 merged

pkg/performanceprofile/controller/performanceprofile/hypershift/status/status.go
ffromani: ok, good enough for now

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
ffromani: after some (admittedly not too much) more reflection, I think this is well intentioned but too risky. Test code is meant to exercise the production code as much as possible and this is a nice idea but make us diverge too much. I'm already a bit concerned about the previous comment (which gets `Status` from configmap and transparently merges into the main object), this one feels far too above the threshold.

I think we should accept the performance cost for the time being.

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
ffromani: you indeed described what the `Delete` call would do, but is not obvious which flow needs `Delete` (and how we managed so far?)

ffromani: anyway, not really blocking

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
ffromani: the tradeoff here is between reusing objects, which can possibly have unexpected state, vs recreating every time. The latter seems cheap enough, but the serializer should be stateless, so I'm ok with this change, even though I'm skeptical about the gains here.

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
ffromani: wait, I misunderstood the code. This really ties to the previous commit and basically enables the API backward  compatibility to restore the status in the same perfprof object. Let me rethink.

ffromani (commented):
ok, please:
- remove the testing mode commit
- improve the doc for the commit adding `Delete` (which code needs it? how we managed till now?)

and LGTM

Tal-or:
> ok, please:
> 
> * remove the testing mode commit
> * improve the doc for the commit adding `Delete` (which code needs it? how we managed till now?)
> 
> and LGTM

Done.
See: https://github.com/openshift/cluster-node-tuning-operator/pull/1089/commits/072a8df06e2e661bd385bd6960bca4b70b229a5c

ffromani (commented):
/approve
/lgtm

thanks for the updates and sorry for the churn

PR #1092, opened 2024-06-24
No-JIRA: e2e:run-test.sh: match ginkgo CLI to ginkgo package

Tal-or:
We are usually having a mismatch between the ginkgo CLI and ginkgo package, because we are updating the ginkgo package on the go.mod but forget to do the same on our test's script.
something like:
```
Ginkgo detected a version mismatch between the Ginkgo CLI and the version of Ginkgo imported by your packages:
  Ginkgo CLI Version:
    2.6.1
  Mismatched package versions found:
    2.17.1 used by 0_config, 1_performance, 6_mustgather_testing, 10_performance_ppc
```

let's automate the process to save some work and avoid bugs that might happens due to breaking changes between versions which are not matching

### DIFF
--- a/hack/run-test.sh
+++ b/hack/run-test.sh
@@ -73,7 +73,8 @@ main() {
     which ginkgo
     if [ $? -ne 0 ]; then
         echo "Downloading ginkgo tool"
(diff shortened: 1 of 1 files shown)


ffromani (commented):
/approve
/lgtm

at long last! nice!

PR #1093, opened 2024-06-25
CNF-13394: Changed node inspector to have lazy initialization

rbaturov:
To streamline the deployment of the node inspector, we have implemented a change where the ExecCommand will automatically deploy the node inspector if it is not already running. This approach offers several benefits:

* On-Demand Deployment: The node inspector is deployed only when needed.
* Users no longer need to explicitly deploy the inspector using the 0_config suite, allowing them to run specific specs or suites without verifying the node inspector's deployment status.

Moreover, this PR introduces some improvements including API simplification as we abstracted this process from the user.
* Lowercasing functions that are not intended to be called from outside the package.
* It is obvious that we would always use the Dataplane client, hence removed it from all function signatures.
* Introducing global variables for the namespace and node inspector, as these values remain constant.
* Changed node inspector to use same context
* Added logs and improved error handling
* Switch the cleanup of the node-inspector to happen after each suite, instead of Z_deconfig.


test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -19,14 +20,30 @@ import (
...
 	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/namespaces"
 	testpods "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
 )
 
 const serviceAccountSuffix = "sa"
 const clusterRoleSuffix = "cr"
 const clusterRoleBindingSuffix = "crb"
 
+func init() {
```
Tal-or, line 33: Add a global boolean `initilized` if true, skip the creation

rbaturov: But the `init` function is going to be called only once per process (as each suite runs as a separate process meaning the initialized value will be zeroed every suite we run).
The only benefit I see from your suggestion would be to check if `initialized` is true in the create function prior to execution, but we don't use the create function anywhere else except the `init` function.

let me know if you would still want the change, or if I miss something.

Tal-or: I would still like to have a lazy initialization so the creation will happen only at the first time when we call `node.ExecCommand` and then we'll check the `initialized` value.

The advantage is for example when you're using `ginkgo.focus` for specific test that you would like to run on which you're not calling `node.ExecCommand`. If the specific test reside in a package which import the inspector, it would create the daemonset although it's not needed.

rbaturov: Then what you are suggesting is to not use an init function, and just add an initialized var that will be check upon exec command.

Tal-or: Yes. But I would like to get a second opinion. @ffromani WDYT?

ffromani: I personally dislike usage of `init` function for everything but the most trivial things, like setting few variables. Creating objects in the cluster as side effect of initialization  is likely to backfire in the future, composes poorly and in general is a pretty rich side effect.
I'd actually go as far as having an explicit `Setup` (or `Init` or `New` or whatever) function, but a lazy initialization like @Tal-or suggested seems a good middleground

rbaturov:
@Tal-or  @ffromani Ready for another review iteration

rbaturov:
cc @mrniranjan

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -19,58 +19,91 @@ import (
...
+var nodeInspectorName = testutils.NodeInspectorName
+
+// initialize would be used to lazy initialize the node inspector
+func initialize() error {
+	if initialized {
+		return nil
+	}
+	// Create the test namespace
+	err := testclient.DataPlaneClient.Create(context.TODO(), namespace)
+	if err != nil && !errors.IsAlreadyExists(err) {
```
Tal-or, line 43: We should handle `errors.IsAlreadyExists(err) == true` separately, otherwise we will continue to `create()` and will return with error there

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -19,58 +19,91 @@ import (
...
+		return fmt.Errorf("Failed to create Node Inspector resources: %v", err)
+	}
+
+	return nil
+}
+
+func create() error {
+	serviceAccountName := fmt.Sprintf("%s-%s", nodeInspectorName, serviceAccountSuffix)
+	sa := createServiceAccount(serviceAccountName, namespace.Name)
+	if err := testclient.DataPlaneClient.Create(context.Background(), sa); err != nil && !errors.IsAlreadyExists(err) {
```
Tal-or, line 59: If you're using `context.Background()` pass the same one to all the calls

Tal-or, line 59: `errors.IsAlreadyExists(err)` deserves a separate handling.
Log a warning at least. 
And having the objects already create means the we didn't have a proper cleanup so it might had other sideaffects on the tests

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -19,58 +19,91 @@ import (
...
+	initialized = true
 
 	return nil
 }
 
-func Delete(cli client.Client, namespace, name string) error {
-	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
-	if err := cli.Delete(context.Background(), ns); err != nil && !errors.IsNotFound(err) {
+func Delete() error {
+	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace.Name}}
```
Tal-or, line 88: Use same contexts, and log a warning if not found

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -94,7 +127,7 @@ func getDaemonPodByNode(node *corev1.Node) (*corev1.Pod, error) {
 // ExecCommand executing the command on a daemon pod of the given node
 func ExecCommand(ctx context.Context, node *corev1.Node, command []string) ([]byte, error) {
 	// Ensure the node inspector is running
-	ok, err := isRunning(testclient.DataPlaneClient, testutils.NodeInspectorNamespace, testutils.NodeInspectorName)
+	ok, err := isRunning()
 	if err != nil || !ok {
```
Tal-or, line 131: In case of `ok` and `err == nil` it returns `nil, nil` which means the caller if `ExecCommand` will continue as usual since there was not error reported

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
ffromani: even better would be to add a context to `create` and let the caller set it.

rbaturov: In case of `create()` this will be called once per each testing suite, because the initialized flag will be zeroed.
This follows the four create functions for the components, to verify they are indeed deployed. 
I feel that logging a warning implies unexpected behavior. but this is expected to happen, as the initialized flag is zerored.

Tal-or: You should log a warning only if the components already exists.

If the components are already exist it means the cleanup of the previous run, was not handled as expected.

mrniranjan:
@rbaturov i used this patch and ran some tests and they pass

test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
rbaturov: > If the components are already exist it means the cleanup of the previous run, was not handled as expected.

This is not true in this case.
we don't cleanup in between suites. that means that for every suite, we are going to run this `create` function once and these component are expected to exist

Tal-or (commented):
Now that we have this, do we still need the `Z_deconfig` and the trap?

rbaturov:
> Now that we have this, do we still need the `Z_deconfig` and the trap?
IMO yes. trap provides us the ability to cleanup the node inspector although the user/other has interrupted the tests. this "forces" this cleanup.

Tal-or:
/lgtm
We have many PRs scattered and it became really hard to track. 
Due to lack in time and capacity as long as we don't see any regression on non-hypershift platform, let's merge the code and have a final pass when the e2e tests #1084 will go in

PR #1094, opened 2024-06-26
NO-JIRA:pao:hypershift: fix client implementation

Tal-or:
On hypershift many objects are encapsulated in ConfigMaps.
So if we're given the name of the encapsulated object,
trying to get the ConfigMap with same name is not enough,
because the internal object's name and the ConfigMap name
doesn't have to be equal.

Hence, the client have to `List()` all the ConfigMaps in the
hosted control plane namespace, extracts the data and
compares the names.

In case it was not found, it mimics an IsNotFound error
to be compatible with the error returned from the API server.

The implementation can be improved, by caching some of the information,
but before we come with "fancy-but-not-proved-to-be-efficent" solutions
let's evaluate the simplest solution first and iterate later if needed.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
+++ b/pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
@@ -23,6 +23,7 @@ const (
 	TuningKey                   = "tuning"
 	ConfigKey                   = "config"
 	HostedClustersNamespaceName = "clusters"
+	PerformanceProfileNameLabel = "hypershift.openshift.io/performanceProfileName"
 )
 
 type ControlPlaneClientImpl struct {
--- a/pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
+++ b/pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
@@ -33,20 +33,20 @@ type ControlPlaneClientImpl struct {
 }
 
 func (ci *ControlPlaneClientImpl) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
-	if ObjIsEncapsulatedInConfigMap(obj) {
+	if IsEncapsulatedInConfigMap(obj) {
 		return ci.getFromConfigMap(ctx, key, obj, opts...)
 	}
 	return ci.Client.Get(ctx, key, obj, opts...)
 }
 
 func (ci *ControlPlaneClientImpl) Create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error {
-	if ObjIsEncapsulatedInConfigMap(obj) {
+	if IsEncapsulatedInConfigMap(obj) {
 		return ci.createInConfigMap(ctx, obj, opts...)
 	}
 	return ci.Client.Create(ctx, obj, opts...)
 }
 
-func ObjIsEncapsulatedInConfigMap(obj runtime.Object) bool {
+func IsEncapsulatedInConfigMap(obj runtime.Object) bool {
 	switch obj.(type) {
 	case *performancev2.PerformanceProfile, *performancev2.PerformanceProfileList,
 		*machineconfigv1.KubeletConfig, *machineconfigv1.KubeletConfigList,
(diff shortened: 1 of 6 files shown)


pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -25,6 +25,14 @@ const (
 	HostedClustersNamespaceName = "clusters"
 )
 
+type counter map[string]int
```
Tal-or, line 29: This commit should not go in and it's only meant for the reviewers sake, to provide some rough numbers about  how many time each CRUD operation is being called.

I'll provide some numbers soon.

ffromani:
I'd be willing to merge these 4 commits
```
hypershift: rename ObjIsEncapsulatedInConfigMap -> `IsEncapsulatedInConfigMap`
logs: add log when PerformanceProfile decoded successfully 
hypershift:runtimeclass: pass mutated object 
predicates: reconcile if owned object has changed
```
if you're happy with them in their current form. Feel free to spin off a new PR.
Other than that general direction LGTM, nice work here

Tal-or:
> I'd be willing to merge these 4 commits
> 
> ```
> hypershift: rename ObjIsEncapsulatedInConfigMap -> `IsEncapsulatedInConfigMap`
> logs: add log when PerformanceProfile decoded successfully 
> hypershift:runtimeclass: pass mutated object 
> predicates: reconcile if owned object has changed
> ```
> 
> if you're happy with them in their current form. Feel free to spin off a new PR. Other than that general direction LGTM, nice work here

The problem with splitting the PR is that many of the bugs (and fixes) has been discovered after I've changed the client implementation. I'm not feeling comfortable to enter this changes alone, since the controller might behave differently.

ffromani:
> > I'd be willing to merge these 4 commits
> > ```
> > hypershift: rename ObjIsEncapsulatedInConfigMap -> `IsEncapsulatedInConfigMap`
> > logs: add log when PerformanceProfile decoded successfully 
> > hypershift:runtimeclass: pass mutated object 
> > predicates: reconcile if owned object has changed
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > if you're happy with them in their current form. Feel free to spin off a new PR. Other than that general direction LGTM, nice work here
> 
> The problem with splitting the PR is that many of the bugs (and fixes) has been discovered after I've changed the client implementation. I'm not feeling comfortable to enter this changes alone, since the controller might behave differently.

fair enough

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: Some statistics about client calls:
When some owned object is changing:
```
I0630 10:28:19.469666       1 performanceprofile_controller.go:537] Client counter END of reconciliation:
 FunctionName="Get", Called 2 times
FunctionName="GetFromConfigMap", Called 3 times
```
 
When owned objects are creating:
```
I0630 15:40:13.183077       1 performanceprofile_controller.go:424] Client counter BEGIN of reconciliation:
 FunctionName="Get", Called 1 times
...
I0630 15:40:13.313975       1 performanceprofile_controller.go:425] Client counter END of reconciliation:
 FunctionName="Get", Called 5 times
FunctionName="GetFromConfigMap", Called 3 times
FunctionName="Create", Called 3 times

```

Let's analyze the numbers:
1. When a change being detected (meaning some of the objects has been tampered) 
standard `Get` being called 2 times and `GetFromConfigMap` , which is more costly because it involves with listing `ConfigMap`s a and deserialize   their data, called 3 times.

2. When objects are create (fresh performanceProfilie applied) 
standard `Get` being called 4 times and `GetFromConfigMap` called 3 times and `Create` called 3 times.

3. When nothing is changing, no calls to the API server are being made by the controller as expected

According to the above numbers and considering the tight timelines we have, I think the provided solution for fixing the client implementation in this PR is good enough.

In case of performance degradation/additional bugs/gaps that were not covered, we can re evaluate in the future. 
But I this point in time I believe we should merge quickly in order to unlift the blocker from the testing development.

ffromani: I agree with your assessment given the numbers we have. I'll have another review ASAP.

Tal-or:
Depends on #1096 https://github.com/openshift/cluster-node-tuning-operator/pull/1096
The CI won't pass without it.

ffromani:
xref: https://github.com/openshift/cluster-node-tuning-operator/pull/1094#discussion_r1660206182

PR #1095, opened 2024-06-30
OCPBUGS-36355: Backport fix for OCPBUGS-30647

jmencak:
This change combines PRs 970, 998 and 1024 which fixed OCPBUGS-30647 in 4.16.

Summary of changes:
  * Even though there is currently no namespace collision with TuneD using in/var/lib/tuned, change this path to /var/lib/ocp-tuned
  * Do not inherit --enable-leader-election and --version NTO flags as they are not handled by subcommands anyway (yet)
  * Clean up after #844
  * Remove TuneD timeout code and reload on ERRORs
  * fix logging in updateTunedProfile() and optimize the calls to update node annotations and update Profile.Status
  * clean up tunedStop() to return only one value
  * during TuneD process shutdown, handle the fact the TuneD process might have already exitted
  * the openshift-tuned operand now no longer unnecessarily exits when TuneD process exits; when TuneD process exits, wait for k8s object changes and only then restart TuneD
  * do not use buffered channels
  * the indication that TuneD is reloading is now a status bit potentially reportable back to the operator
  * introduce Change type for the TuneD event processor to avoid races, where it was previously possible to change TuneD configuration during TuneD profile reload
  * register the fact TuneD finished reloading in case the primary TuneD profile does not exist
  * conditional TuneD reload when Cloud Provider changes
  * minor logging and comment improvements

Resolves: OCPBUGS-36355


jmencak:
Infra issues
/retest

MarSik (approved):
/lgtm

PR #1100, opened 2024-07-04
CNF-13530: hypershift: support `containerruntimeconfig`

Tal-or:
NTO supports different runtimes for the high-performance runtime class:
ref: https://github.com/openshift/cluster-node-tuning-operator/pull/588
This commit adds this kind of support when NTO runs on hypershift platform.


pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
```diff
@@ -129,13 +129,17 @@ func EncodeManifest(obj runtime.Object, scheme *runtime.Scheme) ([]byte, error)
 	return buff.Bytes(), err
 }
 
-func DecodeManifest(b []byte, scheme *runtime.Scheme, obj runtime.Object) error {
+// DecodeManifest tries to decode b into obj.
+// if fails to decode b, it returns an error
+// if decoded data returned from b, has different GVK than what stored in obj
+// it returns false
```
ffromani, line 135: is this a legit use case of the code or are we abusing some behavior?

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift_test.go
```diff
@@ -249,3 +251,50 @@ spec:
...
+	if err := machineconfigv1.AddToScheme(scheme.Scheme); err != nil {
+		t.Fatal(err)
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ok, err := DecodeManifest(tc.manifest, scheme.Scheme, tc.into)
+			if err != nil {
+				t.Errorf("failed to decode manifest into %T: %v", tc.into, err)
+			}
```
ffromani, line 294: should we add a test which fails, such as we expect a non-nil err?

ffromani:
/lgtm

assuming we have e2e coverage (#1084 ?)
but please check the inline comments

Tal-or:
>  but please check the inline comments

I modify the Decode function logic a bit and added unit-tests

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift_test.go
Tal-or: Added

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift.go
Tal-or: This abstraction based on the documentation of the underlying `decode` function:
https://github.com/openshift/cluster-node-tuning-operator/blob/32e3df6b94d52dabbd44e82d86bc3e2798b2c616/vendor/k8s.io/apimachinery/pkg/runtime/serializer/json/json.go#L127-#L134

ffromani: nice, TIL

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift_test.go
ffromani: I would have expected a `expectedError` or something similar, but this is not blocking

ffromani:
/lgtm

feel free to unhold anytime

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift_test.go
Tal-or: if `err != nil` it means the decoding itself failed. 
The case we care about is whether the manifest which we decoded matches the provided object.
This check is done in the `!ok && tc.isLoaded`  below

Tal-or:
/hold cancel
Due the lack of time before CF, and due to the fact that the e2e tests  #1084 requires additional work let's merge as much payload as possible and fix later if needed (worse case we can always open as bug after code freeze)

Tal-or:
/hold
Need to make additional last check

pkg/performanceprofile/controller/performanceprofile/hypershift/hypershift_test.go
ffromani: ok, makes sense. Thanks!

ffromani (commented):
/lgtm

PR #1102, opened 2024-07-04
DNM:debug 1_performance suite

Tal-or:


Tal-or:
/hold
DNM!

Tal-or:
Note to self:
For some reason which is not clear, the `testpod` has two `nodeSelectors`:
```
NodeSelector: {
            kubernetes.io/hostname: ip-10-0-129-17.ec2.internal,
            node-role.kubernetes.io/worker-cnf: ""
        },
```
Now, the node (for another reason which I don't understand) does not contains the `worker-cnf` label (although it gets labeled at the beginning of the test).
Checking carefully I see the node that was labeled at the beginning has gone during the test runtime, so the node the pod is scheduled into, does not contain the label anymore.
The coincidence of:
     A. A node which was removed from a cluster.
     B. An addition if undesired `nodeSelector` for the testpod
is the cause for this bug.
All that "left" is to figure out what causes `A` and `B`

rbaturov:
> Note to self: For some reason which is not clear, the `testpod` has two `nodeSelectors`:
> 
> ```
> NodeSelector: {
>             kubernetes.io/hostname: ip-10-0-129-17.ec2.internal,
>             node-role.kubernetes.io/worker-cnf: ""
>         },
> ```
> 
> Now, the node (for another reason which I don't understand) does not contains the `worker-cnf` label (although it gets labeled at the beginning of the test). Checking carefully I see the node that was labeled at the beginning has gone during the test runtime, so the node the pod is scheduled into, does not contain the label anymore. The coincidence of: A. A node which was removed from a cluster. B. An addition if undesired `nodeSelector` for the testpod is the cause for this bug. All that "left" is to figure out what causes `A` and `B`

Seems like the worker-cnf has been appended by the API server. definitely, we didn't append this label to the pod in any phase of the pod construction.

Tal-or:
> > Note to self: For some reason which is not clear, the `testpod` has two `nodeSelectors`:
> > ```
> > NodeSelector: {
> >             kubernetes.io/hostname: ip-10-0-129-17.ec2.internal,
> >             node-role.kubernetes.io/worker-cnf: ""
> >         },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Now, the node (for another reason which I don't understand) does not contains the `worker-cnf` label (although it gets labeled at the beginning of the test). Checking carefully I see the node that was labeled at the beginning has gone during the test runtime, so the node the pod is scheduled into, does not contain the label anymore. The coincidence of: A. A node which was removed from a cluster. B. An addition if undesired `nodeSelector` for the testpod is the cause for this bug. All that "left" is to figure out what causes `A` and `B`
> 
> Seems like the worker-cnf has been appended by the API server. definitely, we didn't append this label to the pod in any phase of the pod construction.

Indeed, but atm we should first make sure that after every time that performance profile is being applied we'll restore the labels for the nodes. IOW set the worker-cnf label.
This will fix the bug and we'll handle with the investigation of the appended node selector to the pod later.

Tal-or:
https://github.com/openshift/cluster-node-tuning-operator/pull/1084 got merged

PR #1104, opened 2024-07-08
NO-JIRA: Fix requested profile doesn't exist

jmencak:
Fix requested profile doesn't exist for upstream builds.  Upstream `TuneD` now ships system profiles in /usr/lib/tuned/profiles vs /usr/lib/tuned.


MarSik (approved):
/lgtm

PR #1105, opened 2024-07-09
NO-JIRA: hypershift:unitests: adjust controller unit-tests

Tal-or:
We want same unit tests to be running for hypershift,
so we convert the tests into table and added one entry for OCP and one for HCP.

As for now most of the tests are not supported/qualified to be running on HCP and
are skipped for now.

But it is better and safer to have a low coverage than nothing.
We'll work to support additional tests in future PRs


ffromani (commented):
/approve

PR #1108, opened 2024-07-10
CNF-13530: controller: predicate for `containerruntimeconfig`

Tal-or:
The predicate addition will trigger the reconciliation loop every time there's a change in the `ConfigMap` that holds the `containerruntimeconfig`.

Signed-off-by: Talor Itzhak [titzhak@redhat.com](mailto:titzhak@redhat.com)

### DIFF
--- a/pkg/operator/hypershift.go
+++ b/pkg/operator/hypershift.go
@@ -33,9 +33,10 @@ const (
 	// TODO remove once HyperShift has switched to using new key.
 	tunedConfigMapConfigKeyDeprecated = "tuned"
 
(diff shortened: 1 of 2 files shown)


Tal-or:
/hold
depends on https://github.com/openshift/hypershift/pull/4150

yanirq:
Can/Should we add a unit test for the controller here ?

Tal-or:
> Can/Should we add a unit test for the controller here ?

We cannot exercise this one through unit-tests since the clients are mock.
we can test it through e2e tests but we need https://github.com/openshift/hypershift/pull/4150 and #1084 first

ffromani (commented):
/lgtm

PR #1111, opened 2024-07-15
NO-JIRA: e2e: performance: skip test on env with insufficient cpus

shajmakh:
Some CIs run the tests on an environment that has exactly 2 isolated
cpus, but that doesn't mean that these are completely free cpus as there
is a node base load (cpu consumption of burstable and besteffort
pods).The problem is that this test will always fail on such
environments while it's irrelevant to run.

Skip the test by adding a condition to verify whether the requested cpus
are greater than what the node can provide.


ffromani (commented):
/approve
/lgtm

ffromani:

test-only change

Tal-or:
/retitle NO-JIRA: e2e: performance: skip test on env with insufficient cpus

PR #1113, opened 2024-07-16
WIP: enable control of the verbosiness of the operand

ffromani:
WIP TBD

### DIFF
--- a/pkg/flagcodec/flagcodec.go
+++ b/pkg/flagcodec/flagcodec.go
@@ -1,0 +1,195 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


jmencak:
Thank you for the PR, Francesco.

I can see a couple of disadvantages with this approach.

1) An extra API call to get a config map.
2) Unability to chage verbosity per operand (this would cause a restart of all NTO operands, not just the ones we are interested in).

While the current way of turning on debugging for operands (TuneD) is not ideal, I believe we probably want something along the lines of what we already have:

```
  recommend:
    operand:
      debug: true
```

and perhaps add something like:
```
  recommend:
    operand:
      debug: true
      verbosity: <number>
```

This might even enable us not to restart the operand/TuneD at all and thus have less impact on the cluster?

ffromani:
shouldn't we need to change the DaemonSet manifest anyway?

jmencak:
> shouldn't we need to change the DaemonSet manifest anyway?

Why?  Wouldn't we just pick up the verbosity from the profile, like we do debugging?  I haven't tried though, so perhaps there are some technical difficulties I'm not seeing.

ffromani:
ok, sounds fair to me. We'll see in a future PR then.

jmencak:
> ok, sounds fair to me. We'll see in a future PR then.

Let me look into that or whoever gets there first...

jmencak:
Alternative approach: https://github.com/openshift/cluster-node-tuning-operator/pull/1114

PR #1114, opened 2024-07-17
PSAP-1471: Add support for setting operand verbosity

jmencak:
This adds support for setting NTO operand (ocp-tuned) klog logging verbosity via configuring Tuned CR.

Example recommend section:

```
      recommend:
      - match:
        - label: profile
        priority: 20
        profile: openshift-profile
        operand:
          verbosity: 2
```

This is on a per Profile/node basis and no operand/`TuneD` restarts are required.

### DIFF
--- a/manifests/20-profile.crd.yaml
+++ b/manifests/20-profile.crd.yaml
@@ -71,6 +71,9 @@ spec:
                   tunedProfile:
                     description: TuneD profile to apply
                     type: string
+                  verbosity:
+                    description: klog logging verbosity
+                    type: integer
                 required:
                 - tunedProfile
                 type: object
--- a/manifests/20-tuned.crd.yaml
+++ b/manifests/20-tuned.crd.yaml
@@ -133,6 +133,9 @@ spec:
                                 for the TuneD daemon: true/false'
                               type: boolean
                           type: object
+                        verbosity:
+                          description: klog logging verbosity
+                          type: integer
                       type: object
                     priority:
                       description: Tuned profile priority. Highest priority is 0.
--- a/pkg/apis/tuned/v1/tuned_types.go
+++ b/pkg/apis/tuned/v1/tuned_types.go
@@ -113,6 +113,10 @@ type OperandConfig struct {
 	// +optional
 	Debug bool `json:"debug,omitempty"`
 
+	// klog logging verbosity
+	// +optional
+	Verbosity int `json:"verbosity,omitempty"`
+
 	// +optional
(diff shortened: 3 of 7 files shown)


Tal-or:
If this is part of the API of Tuned CR, would you consider make the this new field more user friendly? 
I mean setting verbosity as a number doesn't says much.

We can have more readable and informative API which can be translate into `--v` number by the operand, for example
`Default` - `--v=0`
`Normal` - `--v=2`
`Info` - `--v=4`
`Warning` - `--v=5`
`Debug` - `--v=8`

We can discuss about the klog that each level represents

jmencak:
> If this is part of the API of Tuned CR, would you consider make the this new field more user friendly? I mean setting verbosity as a number doesn't says much.
> 
> We can have more readable and informative API which can be translate into `--v` number by the operand, for example `Default` - `--v=0` `Normal` - `--v=2` `Info` - `--v=4` `Warning` - `--v=5` `Debug` - `--v=8`
> 
> We can discuss about the klog level that each field represents

Thank you for looking @Tal-or !  I understand the motivation.  Let's assume we choose better levels than suggested above, such as the ones `library-go` uses:

```
Normal == 2
Debug == 4
Trace == 6
TraceAll == 8
```
Then you still have an inconsistency with what we use for the NTO operator itself.  We start it with `-v=<number>`.  
So if we switch to the "user-friendly" logging levels for the operand, we want to do the same thing for the operator too and also completely review the logging levels in the entire project.  This goes well beyond what this PR set out to do, but sure, we could do this.

Saying all this, I'm actually not a huge fan of having non-numbered logging levels as suggested.  However, I might be in the minority.  In my view the numbering gives you the advantage of having better granularity and flexibility than the 4 "user-friendly" levels, but also (perhaps) lets programmers choose levels which do not make much sense.  So restricting the number of levels (using "user-friendly" levels) might actually help and also clean things up.

Let's see what other reviewers have to say.  If they're silent, I can bring this up in one of our Monday meetings to unblock this PR, because I'm not sure we have an agreement right now.  (cc @ffromani for an opinion/thoughts as he seems to be a logging enthusiast).

MarSik:
I agree with Jirka here. The code uses numbered levels (klog.V(4).Info...) and the common unix tools also use numbered verbosity everywhere. So I think giving pretty names to the log levels would just confuse things.

The only possible exception is Python though, there named levels are more common. So if this gets passed to tuned as well, we will have to think about it. But if not, then lets use numbers.

PR #1115, opened 2024-07-21
CNF-12711: hypershift: enable feature gate support

Tal-or:
In order to support feature gate on Hypershift,
we need to build and provide an `eventRecorder` for the `featureGateAccessor`.
The `eventRecorder` uses the controllerRef of NTO pod when it emits an event:
https://github.com/openshift/cluster-node-tuning-operator/blob/81b7b9d35113848bbcf96758829a4cb54ac5b784/vendor/github.com/openshift/library-go/pkg/operator/events/recorder.go#L199

In order to get the controllerRef, we need to `Get()` NTO pod and extract
the ownerReference from it's metadata:
https://github.com/openshift/cluster-node-tuning-operator/blob/81b7b9d35113848bbcf96758829a4cb54ac5b784/vendor/github.com/openshift/library-go/pkg/operator/events/recorder.go#L76

Theoretically, on Hypershift we need to use the MNG client to `Get()`
NTO pod, hence we should also pass the rest config of the MNG cluster.

But NTO pod doesn't have an access to `Pod` objects under it role
on Hypershift anyway:
https://github.com/openshift/hypershift/blob/c84168e02ddc3f980ee325a790d6e5ff3b4c3f9f/control-plane-operator/controllers/hostedcontrolplane/nto/clusternodetuningoperator.go#L147-#L149

The consequence of granting access for NTO to list all pods on the NS means
that it would be able to list all the control plan pods, since on
Hypershift all the pods are under the same NS.

Considering the price we would pay from security perspective
and comparing to what other controllers are doing, it's preferred to fallback for
using the namespace as the reference name:
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/starter.go#L71

### DIFF
--- a/cmd/cluster-node-tuning-operator/main.go
+++ b/cmd/cluster-node-tuning-operator/main.go
@@ -194,6 +194,11 @@ func operatorRun() {
 			klog.Exitf("unable to create PerformanceProfile v2 webhook: %v", err)
 		}
 	} else {
+		operatorNamespace := config.OperatorNamespace()
+		fg, err := setupFeatureGates(context.TODO(), restConfig, operatorNamespace)
+		if err != nil {
+			klog.Exitf("failed to setup feature gates: %v", err)
+		}
 		// HyperShift configuration
 		restConfig, err := ntoclient.GetInClusterConfig()
 		if err != nil {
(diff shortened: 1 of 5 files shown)


ffromani (commented):
/lgtm

let's have a second review here

ffromani (commented):
/lgtm

PR #1118, opened 2024-07-22
WIP: OCPBUGS-28647: consume deferred updates from performance profile

ffromani:
Consume the deferred updates support from the performance profile. The idea is that if the annotation (the same as NTO) is found on performanceprofile, it is passed throught the generated tuned object.

Initially we don't support hypershift

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
@@ -17,6 +17,7 @@ import (
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
 	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components"
 	profilecomponent "github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/components/profile"
+	ntoutil "github.com/openshift/cluster-node-tuning-operator/pkg/util"
 )
 
 const (
@@ -40,8 +41,17 @@ const (
 	templatePerformanceProfileName          = "PerformanceProfileName"
 )
 
-func new(name string, profiles []tunedv1.TunedProfile, recommends []tunedv1.TunedRecommend) *tunedv1.Tuned {
-	return &tunedv1.Tuned{
+func OptionsFromPerformanceProfile(pp *performancev2.PerformanceProfile) components.TunedOptions {
+	opts := components.TunedOptions{}
+	if pp == nil {
+		return opts
+	}
+	opts.DeferredEnabled = ntoutil.HasDeferredUpdateAnnotation(pp.Annotations)
+	return opts
+}
+
+func newTuned(name string, profiles []tunedv1.TunedProfile, recommends []tunedv1.TunedRecommend, opts *components.TunedOptions) *tunedv1.Tuned {
+	obj := tunedv1.Tuned{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: tunedv1.SchemeGroupVersion.String(),
 			Kind:       "Tuned",
@@ -55,10 +65,14 @@ func new(name string, profiles []tunedv1.TunedProfile, recommends []tunedv1.Tune
 			Recommend: recommends,
 		},
 	}
+	if opts != nil {
+		obj.Annotations = ntoutil.ToggleDeferredUpdateAnnotation(obj.Annotations, opts.DeferredEnabled)
+	}
+	return &obj
 }
 
 // NewNodePerformance returns tuned profile for performance sensitive workflows
-func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tuned, error) {
+func NewNodePerformance(profile *performancev2.PerformanceProfile, opts *components.TunedOptions) (*tunedv1.Tuned, error) {
 	templateArgs := make(map[string]interface{})
 
 	templateArgs[templatePerformanceProfileName] = profile.Name
@@ -244,7 +258,7 @@ func NewNodePerformance(profile *performancev2.PerformanceProfile) (*tunedv1.Tun
 			MachineConfigLabels: profilecomponent.GetMachineConfigLabel(profile),
 		},
 	}
-	return new(name, profiles, recommends), nil
+	return newTuned(name, profiles, recommends, opts), nil
 }
 
 func getProfileData(tunedTemplate string, data interface{}) (string, error) {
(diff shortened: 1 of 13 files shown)


pkg/performanceprofile/controller/performanceprofile_controller.go
```diff
@@ -595,6 +600,17 @@ func (r *PerformanceProfileReconciler) isMixedCPUsEnabled(object client.Object)
 	return profileutil.IsMixedCPUsEnabled(object.(*performancev2.PerformanceProfile))
 }
 
+func (r *PerformanceProfileReconciler) isTunedDeferredEnabled(object client.Object) bool {
+	if ntoconfig.InHyperShift() {
+		return false
```
Tal-or, line 605: I know it's still WIP so I might be judging too early. 
We recently aligned all NTO/PAO features to work on Hypershift let's make sure to keep this spirit, unless there's a really good reason why not.
And even if there's, let's add a comment explaining why we cannot support it on Hypershift.

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
```diff
@@ -40,8 +41,17 @@ const (
 	templatePerformanceProfileName          = "PerformanceProfileName"
 )
 
-func new(name string, profiles []tunedv1.TunedProfile, recommends []tunedv1.TunedRecommend) *tunedv1.Tuned {
-	return &tunedv1.Tuned{
+func OptionsFromPerformanceProfile(pp *performancev2.PerformanceProfile) components.TunedOptions {
```
Tal-or, line 44: This is more a question than a comment, why it's better to return a struct than a pointer here?

ffromani: In this case is debatable indeed. I tend to use return by value (meaning return structs, not pointer to) by default because I like to minimize the mutability side.
Using values make it obvious we work on copies, there's (the promise of: members can be pointers!) no mutability, which simplifies the reasoning.
For small structs the performance loss is negligible, and in general optimization should be data-driven, the code should default to readability.

But again: in this case is mostly habit, returning a pointer wouldn't have raised eyebrows.

pkg/performanceprofile/controller/performanceprofile_controller.go
ffromani: yes, good point. Will fix/clarify by the time the WIP tag is lifted.

yanirq: We need to come up with a mechanism to keep this practice then

Tal-or: @yanirq For new features we're usually adding e2e tests.
Since the same tests are running on both OCP and HCP they likely to be failed on HCP in case no one took care to add support for the new feature on HCP

pkg/performanceprofile/controller/performanceprofile/components/tuned/tuned.go
Tal-or: Ack. Thank you for the explanation

ffromani:
need to clarify if we want to update at all in hypershift. Won't harm, but would it be useful?

yanirq:
> need to clarify if we want to update at all in hypershift. Won't harm, but would it be useful?

Since we intend to downport this PR I am wondering if should actually keep the separation and have a followup PR for hypershift support

ffromani:
to backport the feature we will need code changes anyway (unfortunately). I think the hypershift support is not a concern in this case.

ffromani:
> I am bit worried about this one. PerfProfile does more than just create Tuned. We also post MC and KubeletConfigs that do not support deferred updates. And they need to stay in sync. Moreover, any change to cpu sets or kernel args in PP causes an MCP reboot.
> 
> So I wonder what can be changed in PP that it does not reboot the nodes and goes to Tuned.

each of these can be addressed individually. But it's allowed for a change in the perfprofile to fan out on multiple objects, and each of them can cause a reboot. And by design these objects are reconciled independently.

ffromani:
> > I am bit worried about this one. PerfProfile does more than just create Tuned. We also post MC and KubeletConfigs that do not support deferred updates. And they need to stay in sync. Moreover, any change to cpu sets or kernel args in PP causes an MCP reboot.
> > So I wonder what can be changed in PP that it does not reboot the nodes and goes to Tuned.
> 
> each of these can be addressed individually. But it's allowed for a change in the perfprofile to fan out on multiple objects, and each of them can cause a reboot. And by design these objects are reconciled independently.

uhm the linked bug mentions only tuned though. Fixing all the other reboot cause, assuming we can in the current architecture, is a much larger scope.

ffromani:
> > > I am bit worried about this one. PerfProfile does more than just create Tuned. We also post MC and KubeletConfigs that do not support deferred updates. And they need to stay in sync. Moreover, any change to cpu sets or kernel args in PP causes an MCP reboot.
> > > So I wonder what can be changed in PP that it does not reboot the nodes and goes to Tuned.
> > 
> > 
> > each of these can be addressed individually. But it's allowed for a change in the perfprofile to fan out on multiple objects, and each of them can cause a reboot. And by design these objects are reconciled independently.
> 
> uhm the linked bug mentions only tuned though. Fixing all the other reboot cause, assuming we can in the current architecture, is a much larger scope.

I addressed somehow this part in my last commit, PTAL @MarSik @yanirq

yanirq:
> > > > I am bit worried about this one. PerfProfile does more than just create Tuned. We also post MC and KubeletConfigs that do not support deferred updates. And they need to stay in sync. Moreover, any change to cpu sets or kernel args in PP causes an MCP reboot.
> > > > So I wonder what can be changed in PP that it does not reboot the nodes and goes to Tuned.
> > > 
> > > 
> > > each of these can be addressed individually. But it's allowed for a change in the perfprofile to fan out on multiple objects, and each of them can cause a reboot. And by design these objects are reconciled independently.
> > 
> > 
> > uhm the linked bug mentions only tuned though. Fixing all the other reboot cause, assuming we can in the current architecture, is a much larger scope.
> 
> I addressed somehow this part in my last commit, PTAL @MarSik @yanirq

I'm good with the setup for having the option to have multiple objects.  Lets get to an agreement with the current method presented here (set of objects,using wildcards) so we are consistent in documentation as well (u/s docs as well?)

ffromani:
> We'll need an e2e test for that, unless there's something as part of NTO tests

talking about classic OCP (non hypershift) we do have e2e tests for the tuned functionality and we have controller tests which ensure that the label is propagated correctly. Would this be sufficient coverage?

ffromani:
tuned-level changes are sufficient to close this bug. But we need one more followup, to be filed soon.

PR #1119, opened 2024-07-23
OCPBUGS-28647: deferred updates: cleanups

ffromani:
cleanups suggested late in the #1019 cycle or emerged after the first nightly CI run

### DIFF
--- a/test/e2e/deferred/restart.go
+++ b/test/e2e/deferred/restart.go
@@ -20,7 +20,7 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/util/wait"
 )
 
-var _ = ginkgo.Describe("[deferred][restart] Profile deferred", func() {
+var _ = ginkgo.Describe("[deferred][restart][slow][disruptive][flaky] Profile deferred", ginkgo.Label("deferred", "restart", "slow", "disruptive", "flaky"), func() {
 	ginkgo.Context("when restarting", func() {
 		var (
 			createdTuneds []string
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ $(GOBINDATA_BIN):
 
 test-e2e: $(BINDATA)
 	for d in core basic reboots reboots/sno deferred; do \
-	  KUBERNETES_CONFIG="$(KUBECONFIG)" $(GO) test -v -timeout 40m ./test/e2e/$$d -ginkgo.v -ginkgo.no-color -ginkgo.fail-fast || exit; \
+	  KUBERNETES_CONFIG="$(KUBECONFIG)" $(GO) test -v -timeout 40m ./test/e2e/$$d -ginkgo.v -ginkgo.no-color -ginkgo.fail-fast -ginkgo.label-filter=!flaky || exit; \
 	done
 
 .PHONY: test-e2e-local
--- a/pkg/tuned/cmd/render/render.go
+++ b/pkg/tuned/cmd/render/render.go
@@ -213,7 +213,7 @@ func render(inputDir []string, outputDir string, mcpName string) error {
 	for _, t := range tuneD {
 		tunedProfiles = append(tunedProfiles, t.Spec.Profile...)
 	}
-	_, _, _, _, err = tunedpkg.ProfilesExtract(tunedProfiles, recommendedProfile)
+	_, err = tunedpkg.ProfilesExtract(tunedProfiles, recommendedProfile)
 	if err != nil {
 		klog.Errorf("error extracting tuned profiles : %v", err)
 		return fmt.Errorf("error extracting tuned profiles: %w", err)
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -80,7 +80,7 @@ const (
 	// be generous and give it 10s.
 	tunedGracefulExitWait = time.Second * time.Duration(10)
 	ocpTunedHome          = "/var/lib/ocp-tuned"
-	ocpTunedRunDir        = "/run/" + programName
+	ocpTunedRunDir        = "/run/ocp-tuned"
 	ocpTunedProvider      = ocpTunedHome + "/provider"
 	// With the less aggressive rate limiter, retries will happen at 100ms*2^(retry_n-1):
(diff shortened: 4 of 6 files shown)


jmencak:
Thank you for the cleanup.  This looks good to me, but letting other reviewers comment.  Only one super-nit:
```
index cce3334a..561a60dc 100644
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -403,7 +403,7 @@ type ExtractedProfiles struct {
 
 // ProfilesExtract extracts TuneD daemon profiles to tunedProfilesDirCustom directory.
 // Returns:
-//   - ExtractedProfile with the details of the operation performed
+//   - ExtractedProfiles with the details of the operation performed.
 //   - Error if any or nil.
 func ProfilesExtract(profiles []tunedv1.TunedProfile, recommendedProfile string) (ExtractedProfiles, error) {
 	klog.Infof("profilesExtract(): extracting %d TuneD profiles (recommended=%s)", len(profiles), recommendedProfile)

```

ffromani:
> Thank you for the cleanup. This looks good to me, but letting other reviewers comment. Only one super-nit:
> 
> ```
> diff --git a/pkg/tuned/controller.go b/pkg/tuned/controller.go
> index cce3334a..561a60dc 100644
> --- a/pkg/tuned/controller.go
> +++ b/pkg/tuned/controller.go
> @@ -403,7 +403,7 @@ type ExtractedProfiles struct {
>  
>  // ProfilesExtract extracts TuneD daemon profiles to tunedProfilesDirCustom directory.
>  // Returns:
> -//   - ExtractedProfile with the details of the operation performed
> +//   - ExtractedProfiles with the details of the operation performed.
>  //   - Error if any or nil.
>  func ProfilesExtract(profiles []tunedv1.TunedProfile, recommendedProfile string) (ExtractedProfiles, error) {
>  	klog.Infof("profilesExtract(): extracting %d TuneD profiles (recommended=%s)", len(profiles), recommendedProfile)
> ```

thanks, fixed

ffromani:
/cc @Tal-or 
as he did most of the comments I'm addressing here

test/e2e/deferred/restart.go
```diff
@@ -20,6 +22,23 @@ import (
...
+	}
+	val, err := strconv.ParseBool(envVal)
+	if err != nil {
+		util.Logf("wrong value for %q: %q (err=%v)", enableRestartTestsEnvVar, envVal, err)
+		return false
+	}
+	return val
+}
+
 var _ = ginkgo.Describe("[deferred][restart] Profile deferred", func() {
```
Tal-or, line 42: Maybe we should also add `[disruptive]` ? 
https://github.com/kubernetes/community/blob/0566459c2bdfec36670861be20be26d33f31467e/contributors/devel/sig-testing/e2e-tests.md#kinds-of-tests

test/e2e/deferred/restart.go
```diff
@@ -20,6 +22,23 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/util/wait"
 )
 
+const (
+	enableRestartTestsEnvVar = "DEFERRED_RESTART_TESTS"
+)
+
+func areRestartTestsEnabled() bool {
```
Tal-or, line 29: `Ginkgo` has a built-in Labels mechanism that allow you to skip/unskip tests in a more natural manner. 
Unless the commands the executes the tests sits somewhere else (on different code-base) then we need to actively configure it to skip the label, which is less convenient.

test/e2e/deferred/restart.go
ffromani: indeed, done

test/e2e/deferred/restart.go
ffromani: very good point, let's add labels, and let's fix the matching tags

Tal-or:
@ffromani I don't see, where the actual skip is done?
We should have something like this:
https://github.com/openshift/cluster-node-tuning-operator/blob/master/Makefile#L260C135-L260C162

ffromani:
> @ffromani I don't see where the actual skip is done? We should have something like this: https://github.com/openshift/cluster-node-tuning-operator/blob/master/Makefile#L260C135-L260C162 Am I missing something here?

because is not yet :) I'll still need to do the skipping part consuming the labels

ffromani:
/hold

incomplete PR

ffromani:
/hold cancel

xref: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-node-tuning-operator/openshift-cluster-node-tuning-operator-master.yaml#L119

Tal-or:
/lgtm
Thanks!

ffromani:
/unhold

before the latest update:
```
$ make test-e2e
for d in deferred; do \
  KUBERNETES_CONFIG="/home/fromani/clusters/cnflab/kubeconfig" GOOS=linux GO111MODULE=on GOFLAGS=-mod=vendor go test -v -timeout 40m ./test/e2e/$d -ginkgo.v -ginkgo.no-color -ginkgo.fail-fast -ginkgo.label-filter=\"!flaky\" || exit; \
done
=== RUN   TestNodeTuningOperatorDeferred
Ginkgo detected configuration issues:
Syntax Error Parsing Label Filter
  "!flaky"
  Invalid token '!'.

  Learn more at: http://onsi.github.io/ginkgo/#spec-labels
FAIL	github.com/openshift/cluster-node-tuning-operator/test/e2e/deferred	0.021s
FAIL
make: *** [Makefile:103: test-e2e] Error 1
```

Tal-or:
The label is `Flaky` (begin with capital letter) but the filter is `!flaky` (small letter). 
Are you sure it would match the label?

ffromani:
> The label is `Flaky` (begin with capital letter) but the filter is `!flaky` (small letter). Are you sure it would match the label?

good catch. Partial update. Fixing.

ffromani:
uhm, kube standard is to use `Capitalized` nouns, but the rest of the tests is using `lowercase-dash-separated`

Tal-or:
/lgtm
Thank you for the fixes

ffromani:
thank you for the careful review! low caffeine intake today

jmencak:
> New changes are detected. LGTM label has been removed.

Thank you @ffromani , the last change was requested by me.

/lgtm

PR #1124, opened 2024-07-30
WIP: CNF-1339: ppc: adjust for Hypershift

Tal-or:
WIP


cmd/performance-profile-creator/cmd/root.go
```diff
@@ -578,7 +578,7 @@ type ProfileCreatorArgs struct {
 	EnableHardwareTuning        bool   `json:"enable-hardware-tuning,omitempty"`
 }
 
-func createProfile(profileData ProfileData) error {
+func makePerformanceProfileFrom(profileData *ProfileData) *performancev2.PerformanceProfile {
```
ffromani, line 483: if we don't mutate `profileData`, and it seems to me we don't it seems better to pass a copy rather than a pointer

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -189,25 +167,56 @@ func NewRootCommand() *cobra.Command {
...
-	root.PersistentFlags().StringVar(&pcArgs.MustGatherDirPath, "must-gather-dir-path", "must-gather", "Must gather directory path")
-	root.PersistentFlags().StringVar(&pcArgs.ProfileName, "profile-name", "performance", "Name of the performance profile to be created")
-	root.PersistentFlags().StringVar(&pcArgs.TMPolicy, "topology-manager-policy", kubeletconfig.RestrictedTopologyManagerPolicy, fmt.Sprintf("Kubelet Topology Manager Policy of the performance profile to be created. [Valid values: %s, %s, %s]", kubeletconfig.SingleNumaNodeTopologyManagerPolicy, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy))
-	root.PersistentFlags().StringVar(&pcArgs.Info, "info", infoModeLog, fmt.Sprintf("Show cluster information; requires --must-gather-dir-path, ignore the other arguments. [Valid values: %s]", strings.Join(validInfoModes, ", ")))
-	root.PersistentFlags().BoolVar(pcArgs.PerPodPowerManagement, "per-pod-power-management", false, "Enable Per Pod Power Management")
-	root.PersistentFlags().BoolVar(&pcArgs.EnableHardwareTuning, "enable-hardware-tuning", false, "Enable setting maximum cpu frequencies")
+	initFlags(root.PersistentFlags(), pcArgs)
```
ffromani, line 172: if we want this modularization (I'd be pretty neutral about it) I'd prefer to add a method to `pcArgs` to do the flag bindign there

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -139,6 +139,10 @@ func NewRootCommand() *cobra.Command {
 		Use:   "performance-profile-creator",
 		Short: "A tool that automates creation of Performance Profiles",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			mustGatherDirPath := cmd.Flag("must-gather-dir-path").Value.String()
+			if err := validateMustGatherDirPath(mustGatherDirPath); err != nil {
+				return err
+			}
```
ffromani: well intentioned, but nothing guarantees that the path will be valid after this check: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

for maximum safety we should do the check as late as possible (so there's minimally, ideally none, difference between time to check and time to use)

ffromani (commented):
I like the general direction. We may want to rediscuss some specific commits, but the vast majority of changes here seem good and useful.

cmd/performance-profile-creator/cmd/root.go
Tal-or: Hey, I'm not sure I understand. 
I basically follow the example from:
https://github.com/k8stopologyawareschedwg/deployer/blob/bcc68d884fc94cc004b21de9ac9f36af5c04678d/pkg/commands/root.go#L105

cmd/performance-profile-creator/cmd/root.go
Tal-or: I don't know at what stage of the PR you have reviewed it but if you'll have a look at the latest revision you'll see the path is being used immediately after the validation.

cmd/performance-profile-creator/cmd/root.go
Tal-or: Done. thanks

Tal-or:
/hold
I would like to add another test case

Tal-or:
/hold cancel
Test case has been added

cmd/performance-profile-creator/cmd/info.go
```diff
@@ -0,0 +1,146 @@
...
+const (
+	infoModeJSON = "json"
+	infoModeLog  = "log"
+)
+
+type infoOptions struct {
+	jsonOutput bool
+}
+
+func NewInfoCommand(pcArgs *ProfileCreatorArgs) *cobra.Command {
```
yanirq, line 22: maybe a short comment on the function since its public

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -277,6 +281,20 @@ func getClusterData(mustGatherDirPath string) (ClusterData, error) {
 	return cluster, nil
 }
 
+func validateMustGatherDirPath(path string) error {
```
yanirq: is there a test coverage for these scenarios?

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -128,86 +131,105 @@ func NewRootCommand() *cobra.Command {
...
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			nodes, err := listNodesForPool(pcArgs)
 			if err != nil {
-				return fmt.Errorf("failed to parse the cluster data: %v", err)
+				return err
 			}
-
-			profileCreatorArgsFromFlags, err := getDataFromFlags(cmd)
+			nodesHandlers, err := makeNodesHandlers(pcArgs.MustGatherDirPath, pcArgs.NodePoolName, nodes)
```
yanirq, line 167: where are `nodesHandlers` being used here ?  is this only for validation (catching only the error) ?

cmd/performance-profile-creator/cmd/root.go
Tal-or: yes. we have unit tests that cover the complete flow

cmd/performance-profile-creator/cmd/root.go
Tal-or: Please check the full changes. `nodeHandlers` are used in lines 171 and 177
https://github.com/openshift/cluster-node-tuning-operator/pull/1124/files/9a4aba81c529c6ce074771593be98f40afd2ea6c#diff-baadfd30860582642e86fbeebbe6cb8ca8b7f39266420331fc322cf237e121fcR171 

https://github.com/openshift/cluster-node-tuning-operator/pull/1124/files/9a4aba81c529c6ce074771593be98f40afd2ea6c#diff-baadfd30860582642e86fbeebbe6cb8ca8b7f39266420331fc322cf237e121fcR177

cmd/performance-profile-creator/cmd/info.go
Tal-or: Done. Thanks

test/e2e/performanceprofile/testdata/ppc-expected-profiles/profile1.yaml
```diff
@@ -18,3 +18,5 @@ spec:
     highPowerConsumption: false
     realTime: true
     perPodpowerManagement: false
+  net:
+    userLevelNetworking: false
```
yanirq, line 22: why did we have to add this specifically ?  (same comment for the rest of the files having the same change)

Tal-or: Because we have a slight change in the way of how the flags are being initialized,  we added a default value for some of the variables to make the tests happy.

yanirq:
/approve
looks ok to me, we can add more tests in a separate PR . Giving others a chance to have another review and ack.

cmd/performance-profile-creator/cmd/root.go
ffromani: nevermind, it's minor and we can address later. Just note in general the code I authored is not necessarily a good example :) I try my best, but still.

ffromani (commented):
/lgtm

Very nice work!
there are some minor comments here and there I can make, but nothing too important to further delay this PR, we can re-evaluate later.

cmd/performance-profile-creator/cmd/root.go
Tal-or: no worries, in later commit I moved it to a receiver function

PR #1125, opened 2024-07-30
OCPBUGS-36868: TuneD prior to kubelet in one-shot mode

jmencak:
Manual backport of #1045 due to changes in Dockerfile and deprecated k8s.io/utils/pointer Go module.

Motivation.  In case of node reboots, all pods restart again in random order. Since there's no control on pod restart order, it is possible that tuned pods will start after the workload pods.  This means the workload pods can start with partial tuning which can result in degraded performance or even cause the workload pods crash.

This change is intended mostly for Telco use cases where PerformanceProfiles are used.  To minimize the risk of affecting other deployments by backporting this code, TuneD will start prior to kubelet only when PerformanceProfiles are used.

Changes:
  * Switch to /var/lib/ocp-tuned as the home directory for the operand (TuneD) pod on the host.  This directory now holds configuration for the TuneD daemon both running prior to kubelet and in the pod itself, notably TuneD daemon profiles and the recommended profile.
  * Use the new tuned-main.conf option "profile_dirs" to extract TuneD profiles to /var/lib/ocp-tuned/profiles.  This new option also allowed simplifying the algorithm for change detection in profilesSync() when extracting TuneD profiles.
  * Move a lot of the logic of creating the environment for TuneD to the operand code, so that no external script are needed for starting the operand.
  * Add ocp-tuned-one-shot.service systemd unit when using a PerformanceProfile.
  * Introduce a timeout to running TuneD in one-shot mode so that we don't block the kubelet indefinitely.
  * Shorten the constants names from openshift* to ocp*

Resolves: OCPBUGS-36868

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,19 +26,21 @@ ENV PATH=${APP_ROOT}/bin:${PATH}
 ENV HOME=${APP_ROOT}
 ENV SYSTEMD_IGNORE_CHROOT=1
 WORKDIR ${APP_ROOT}
-COPY --from=tuned   /root/assets ${APP_ROOT}
-COPY --from=tuned   /root/rpmbuild/RPMS/noarch /root/rpms
+COPY --from=tuned /root/assets/bin /usr/local/bin
+COPY --from=tuned /root/rpmbuild/RPMS/noarch /root/rpms
 RUN INSTALL_PKGS=" \
       nmap-ncat procps-ng pciutils \
       " && \
-    mkdir -p /etc/grub.d/ /boot && \
+    mkdir -p /etc/grub.d/ /boot /var/lib/ocp-tuned && \
     dnf install --setopt=tsflags=nodocs -y $INSTALL_PKGS && \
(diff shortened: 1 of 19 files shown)


jmencak:
Manually tested this on `4.15.0-0.nightly-2024-07-30-012242` and:

```
kind: PerformanceProfile
metadata:
  name: performance
spec:
  cpu:
    isolated: "4-7"
    reserved: "0-3"
  globallyDisableIrqLoadBalancing: false
  net:
    userLevelNetworking: true
    devices: []
  nodeSelector:
  numa:
    topologyPolicy: "single-numa-node"
  realTimeKernel:
    enabled: false
```

`ocp-tuned-one-shot.service` started and ran in one-shot mode without issues and terminated prior to kubelet systemd service.

ffromani (commented):
provisional LGTM
will have another pass and then give formal lgtm

ffromani (commented):
/lgtm

large changeset, but nothing outstanding, let's move forward. We need careful testing anyway.

PR #1126, opened 2024-07-30
OCPBUGS-37734: Backport fix for OCPBUGS-36355

jmencak:
This is a backport of #1095 which fixed OCPBUGS-36355 in 4.15.

Summary of changes:
  * Change the operand's home directory from TuneD's artifacts directory /var/lib/tuned to /var/lib/ocp-tuned
  * Do not inherit --enable-leader-election and --version NTO flags as they are not handled by subcommands anyway (yet)
  * Clean up after #844
  * Remove TuneD timeout code and reload on ERRORs
  * Fix logging in updateTunedProfile() and optimize the calls to update node annotations and update Profile.Status
  * Clean up tunedStop() to return only one value
  * During TuneD process shutdown, handle the fact the TuneD process might have already exitted
  * The openshift-tuned operand now no longer unnecessarily exits when TuneD process exits; when TuneD process exits, wait for k8s object changes and only then restart TuneD
  * Do not use buffered channels
  * The indication that TuneD is reloading is now a status bit potentially reportable back to the operator
  * Introduce Change type for the TuneD event processor to avoid races, where it was previously possible to change TuneD configuration during TuneD profile reload
  * Register the fact TuneD finished reloading in case the primary TuneD profile does not exist
  * Conditional TuneD reload when Cloud Provider changes
  * Minor logging and comment improvements

Resolves: OCPBUGS-37734


jmencak:
`e2e-hypershift` looks like an infra issue, `e2e-gcp-pao-updating-profile` is slightly suspicious but could be infra too.
/retest

jmencak:
```
Step e2e-hypershift-create-management-cluster failed after 31m8s. 
```
I don't believe this is due to this PR, let's retest and if this persist, let's debug in the CI.
/retest

jmencak:
With this fix, I also tried to bring the operand code to the 4.15 as possible to make future 4.14 patches simpler.
Manual testing passed.

jmencak:
@ffromani , @swatisehgal , can I have some eyes on this backport?  Many thanks!

ffromani:
/lgtm

I reviewed the changes and they seem OK. The most important part is the verification/testing anyway. The fact CI is passing gives us quite a bit of confidence.

PR #1127, opened 2024-08-05
OCPBUGS-37956: E2E: Add test to verify cpuset.cpus.exclusive is writeable

mrniranjan:
Automates OCPBUGS-34812: cgroupsv2: failed to write on cpuset.cpus.exclusive

To reproduce the bug, we need to create and delete deployment (deploying guaranteed pods with cpu load balancing annotation) in quick succession so that we do not fully wait for the cleanup  causing the pod about to be deleted to still have access to exclusive cpus causing the new pod from to fail because we can't still write cpuset.cpus.exclusive .  As the pre-start hook fails to write to cpuset.cpus.exclusive file in the pods' cgroup. 

This automation PR creates  and deletes deployment in loop to reproduce the issue and checks if the pods fails with Runtime error with message "failed to run pre-start hook for container"

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -9,10 +9,12 @@ import (
 	"strings"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
 	"k8s.io/utils/cpuset"
@@ -31,6 +33,7 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cgroup/controller"
 	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/cluster"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/deployments"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/events"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/images"
@@ -615,6 +618,121 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		)
 
 	})
+	// Automates OCPBUGS-34812: cgroupsv2: failed to write on cpuset.cpus.exclusive
+	Context("Cgroupsv2", func() {
+		It("[test_id:75327] cpus from deleted cgroup can be reassigned to new cgroup", Label(string(label.Tier0)), func() {
+
+			// we need system with more than 10 cpus to execute this test
+			if len(onlineCPUSet.List()) < 10 {
+				Skip("Requires system with  more than 10 cpus")
+			}
+			if !cgroupV2 {
+				Skip("Requires CgroupV2")
+			}
+			// create deployment with 2 replicas and each pod having cpu load balancing disabled
+			// and runtime class. This is required as the cpu id's used by the container are
+			// written to cpuset.cpus.exclusive
+			const DeploymentName = "test-deployment"
+			var numberofReplicas int32 = 2
+			var dp *appsv1.Deployment
+			annotations := map[string]string{
+				"cpu-load-balancing.crio.io": "disable",
+			}
+			p := pods.GetTestPod()
+			p.Spec.NodeSelector = testutils.NodeSelectorLabels
+			p.ObjectMeta = metav1.ObjectMeta{
+				Labels: map[string]string{
+					"app": DeploymentName,
+				},
+			}
+			runtimeClass := components.GetComponentName(profile.Name, components.ComponentNamePrefix)
+			p.Spec.RuntimeClassName = &runtimeClass
+			p.Spec.Containers[0].Image = images.Test()
+			p.Spec.Containers[0].Resources = corev1.ResourceRequirements{
+				Limits: corev1.ResourceList{
+					corev1.ResourceMemory: resource.MustParse("200Mi"),
+					corev1.ResourceCPU:    resource.MustParse("2"),
+				},
+			}
+
+			// Adding a unique label to the deployment
+			deploymentLabels := map[string]string{
+				"app": DeploymentName,
+			}
+			// we delete the deployment either way
+			defer func() {
+				err := testclient.DataPlaneClient.Get(ctx, client.ObjectKey{Name: dp.Name, Namespace: testutils.NamespaceTesting}, dp)
+				if err == nil {
+					// deployment exists, delete it
+					testlog.Infof("Deleting Deployment %v", dp.Name)
+					err := testclient.Client.Delete(ctx, dp)
+					Expect(err).ToNot(HaveOccurred())
+				}
+				testlog.Infof("Deployment %v is deleted %v", dp.Name)
+			}()
+			// we create and delete deployment in loop to create deployments in
+			// quick succession to verify pre-start hook is able to write to
+			// cpuset.cpus.exclusive
+			for i := 0; i < 5; i++ {
+				testlog.Infof("%d Create deployment %s with 2 Guaranteed pods requesting 2 cpus", i, DeploymentName)
+				dp = deployments.Make(DeploymentName, testutils.NamespaceTesting,
+					deployments.WithPodTemplate(p),
+					deployments.WithNodeSelector(testutils.NodeSelectorLabels))
+				dp.Spec.Template.Annotations = annotations
+				dp.Spec.Template.Labels = deploymentLabels // Add labels to the pod template
+				dp.Spec.Replicas = &numberofReplicas
+
+				Expect(testclient.DataPlaneClient.Create(ctx, dp)).ToNot(HaveOccurred())
+				podList := &corev1.PodList{}
+				listOptions := &client.ListOptions{Namespace: testutils.NamespaceTesting, LabelSelector: labels.SelectorFromSet(deploymentLabels)}
+				Eventually(func() bool {
+					if err := testclient.DataPlaneClient.List(context.TODO(), podList, listOptions); err != nil {
+						return false
+					}
+					for _, pod := range podList.Items {
+						Expect(pod.Status.QOSClass).To(Equal(corev1.PodQOSGuaranteed))
+						if isPodReady(&pod) || pod.Status.Phase == corev1.PodRunning {
+							continue
+						}
+						for _, containerStatus := range pod.Status.ContainerStatuses {
+							if containerStatus.State.Waiting == nil {
+								continue
+							}
+							// we want to test container Pending state first so that we do not skip the condition
+							// since we are creating and deleting the deployments, it's possible for this condition
+							// to be skipped if one of the pods is not in pending state
+							if containerStatus.State.Waiting.Reason == "RunContainerError" && strings.Contains(containerStatus.State.Waiting.Message, "failed to run pre-start hook for container") {
+								testlog.Warningf("container %s failed to start with error: %s", pod.Spec.Containers[0].Name, containerStatus.State.Waiting.Message)
+								return false
+							}
+						}
+					}
+					if len(podList.Items) < int(numberofReplicas) {
+						testlog.Warningf("Required number of pods is %d", numberofReplicas)
+						return false
+					}
+					for _, s := range podList.Items[0].Status.ContainerStatuses {
+						if !s.Ready {
+							testlog.Warningf("container status is %q", s.Name)
+							return false
+						}
+					}
+					// if we are here all the pods in the deployment are in ready state
+					return true
+				}, 10*time.Second, 5*time.Second).Should(BeTrue())
+				// delete deployment
+				// since the deployment is called in loop, all the pods are given some time
+				// before we delete them because we want to preserve the current pod state
+				// for some time to allow us to capture the container status messages before
+				// we start deleting the pod
+				testlog.Info("we wait for 5 seconds before deployment is deleted")
+				time.Sleep(5 * time.Second)
+				testlog.Infof("Deleting Deployment %v", dp.Name)
+				err := testclient.Client.Delete(ctx, dp)
+				Expect(err).ToNot(HaveOccurred())
+			}
+		})
+	})
 	Context("Crio Annotations", Label(string(label.Tier0)), func() {
 		var testpod *corev1.Pod
 		var allTestpods map[types.UID]*corev1.Pod
@@ -711,7 +829,6 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 				Expect(cgroupCpuset).To(Equal(podAffinityCpuset), "cpuset.cpus not matching the process affinity")
 			})
 		})
-
 		Describe("Load Balancing Annotation", func() {
 			It("[test_id:32646] cpus used by container should not be load balanced", func() {
 				output, err := getPodCpus(testpod)
@@ -1082,4 +1199,14 @@ func busyCpuImageEnv() string {
 		qeImageRegistry = "quay.io/ocp-edge-qe/"
 	}
 	return fmt.Sprintf("%s%s", qeImageRegistry, busyCpusImage)
+}
+
+// isPodReady checks if the pod is in ready state
+func isPodReady(pod *corev1.Pod) bool {
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
 }


test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -3,6 +3,9 @@ package __performance
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/deployments"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/labels"
```
shajmakh, line 8: nit: let's move the packages to their groups below, k8s & local

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,80 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+				Expect(testclient.DataPlaneClient.Create(ctx, dp)).ToNot(HaveOccurred())
+				podList := &corev1.PodList{}
+				listOptions := &client.ListOptions{Namespace: testutils.NamespaceTesting, LabelSelector: labels.SelectorFromSet(dp.Spec.Selector.MatchLabels)}
+				Eventually(func() bool {
+					if err := testclient.DataPlaneClient.List(context.TODO(), podList, listOptions); err != nil {
+						return false
+					}
+					//readyPods := 0
+					for _, pod := range podList.Items {
+						if pod.Status.Phase != corev1.PodRunning && !isPodReady(&pod) {
```
shajmakh, line 672: can it be ready but not running? or are these equal conditions?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,80 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+											testlog.Infof("container %s failed to start with error: %s", pod.Spec.Containers[0].Name, containerStatus.State.Waiting.Message)
+											return false
+										}
+									}
+								}
+							}
+						}
+					}
+					return true
+				}, 10*time.Second, 5*time.Second).Should(BeTrue())
```
shajmakh, line 686: if I understand correctly this will be checked 3 times with the current interval and timeout, is this what we need? 
wouldn't this be fragile on compact environments like the u/s ci with 4 cpus while 3 only are isolated or is this the stress we desire for verifying the bug?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,80 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+				dp.Spec.Replicas = &numberofReplicas
+				Expect(testclient.DataPlaneClient.Create(ctx, dp)).ToNot(HaveOccurred())
+				podList := &corev1.PodList{}
+				listOptions := &client.ListOptions{Namespace: testutils.NamespaceTesting, LabelSelector: labels.SelectorFromSet(dp.Spec.Selector.MatchLabels)}
+				Eventually(func() bool {
+					if err := testclient.DataPlaneClient.List(context.TODO(), podList, listOptions); err != nil {
+						return false
+					}
+					//readyPods := 0
+					for _, pod := range podList.Items {
```
shajmakh, line 671: if I understand correctly this eventually waits for the deployment to be ready, I think it would be a good idea to add that under a helper here maybe:
https://github.com/openshift/cluster-node-tuning-operator/tree/ea36e6ae28ce38898481d5f4cd8819a6e017c95b/test/e2e/util/wait
 also it is recommended to avoid nested if conditions as much as possible.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,80 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+						if pod.Status.Phase != corev1.PodRunning && !isPodReady(&pod) {
+							for _, containerStatus := range pod.Status.ContainerStatuses {
+								if containerStatus.State.Waiting != nil {
+									if containerStatus.State.Waiting.Reason == "RunContainerError" {
+										if strings.Contains(containerStatus.State.Waiting.Message, "failed to run pre-start hook for container") {
+											testlog.Infof("container %s failed to start with error: %s", pod.Spec.Containers[0].Name, containerStatus.State.Waiting.Message)
+											return false
+										}
+									}
+								}
```
shajmakh, line 681: unless we exactly want to verify the reason for waiting is `RunContainerError`, we can verify the dp is ready by checking the ready vs available vs desired:

`
status:
  availableReplicas: 1
  conditions:
...
  observedGeneration: 1
  readyReplicas: 1
  replicas: 1
  updatedReplicas: 1
[root@registry ~]# 
`

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,80 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+					corev1.ResourceCPU:    resource.MustParse("2"),
+				},
+			}
+
+			// we create and delete deployment in loop to create deployments in
+			// quick succession to verify pre-start hook is able to write to
+			// cpuset.cpus.exclusive
+			for i := 0; i < 5; i++ {
+				// Create Deployment
+				dp = deployments.Make(DeploymentName, testutils.NamespaceTesting,
```
shajmakh, line 658: is it on purpose to create the deployment with same name? since the test doesn't wait for the deployment deletion on purpose I'm not sure if creating a new deployment with same name and configuration is stable

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,80 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+					corev1.ResourceMemory: resource.MustParse("200Mi"),
+					corev1.ResourceCPU:    resource.MustParse("2"),
+				},
+			}
+
+			// we create and delete deployment in loop to create deployments in
+			// quick succession to verify pre-start hook is able to write to
+			// cpuset.cpus.exclusive
+			for i := 0; i < 5; i++ {
+				// Create Deployment
```
shajmakh, line 657: let's add some more reports (By() & klog) where we think it would help us debug the test upon failure

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,80 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+				dp = deployments.Make(DeploymentName, testutils.NamespaceTesting,
+					deployments.WithPodTemplate(p),
+					deployments.WithNodeSelector(testutils.NodeSelectorLabels))
+				dp.Spec.Template.Annotations = annotations
+				dp.Spec.Replicas = &numberofReplicas
+				Expect(testclient.DataPlaneClient.Create(ctx, dp)).ToNot(HaveOccurred())
+				podList := &corev1.PodList{}
```
shajmakh, line 664: do we want to make sure the pods are guaranteed?

shajmakh (commented):
Thanks for this! 
added few comments below

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: we are not interested in the case where the pod is running, hence i am not checking that condition. We are checking where one of the pods is not in a running state

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: In upstream this test is skipped because we need system with more than 10 cpus

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: Yes for the bug we need to create and delete deployment in quick succession.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: In the deployment spec we are making sure that the pod has cpu and memory limits with cpu load balancing annotation. I mean the deployment can create only guaranteed pods.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: Yes the test is only interested in error RunContainerError, and yes we can check dp is ready by checking the number of replicas are equal to desired , here that case will be satisfied in every loop which i don't want to check , the condition of RunContainerError can occur during when the 2nd or 3rd time pods are being created and the pod from the earlier deployment is still being terminated.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
shajmakh: right, but just to validate that in case someone changed the deployment resources. just as a gate, especially because the test verifies normal condition of the deployment being ready

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
shajmakh: ah right I missed that

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -1083,3 +1159,13 @@ func busyCpuImageEnv() string {
...
+
+// isPodReady checks if the pod is in ready state
+func isPodReady(pod *corev1.Pod) bool {
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
```
shajmakh, line 1171: there is a util function to check that https://github.com/openshift/cluster-node-tuning-operator/blob/ea36e6ae28ce38898481d5f4cd8819a6e017c95b/test/e2e/util/wait/pod.go#L7

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: It's bit difficult to avoid here, because there are just[ 5 phases](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) the pod can be  and in our case the pod though in oc status returns RunContainerError, its not a defined phase and this status comes only when the pod is in Pending state and crio is not able to write to cpuset.cpus.exclusive , so we need to check if the pod is in pending state and not all pending states give that error , hence we need to go again deeper to check waiting reason and if that reason is because pre-start hook failed. 

Even if we avoid the nested if's in test case, we have to the nested if's in some other utils function if we move code any where.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: Addressed in latest commit

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: Addressed in latest commit

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
shajmakh: what if the pod failed for different reason? the eventually will return with true and the test will succeed

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -671,12 +671,13 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
-											return false
-										}
+								if containerStatus.State.Waiting == nil {
+									continue
+								}
+								if containerStatus.State.Waiting.Reason == "RunContainerError" {
+									if strings.Contains(containerStatus.State.Waiting.Message, "failed to run pre-start hook for container") {
+										testlog.Infof("container %s failed to start with error: %s", pod.Spec.Containers[0].Name, containerStatus.State.Waiting.Message)
+										return false
 									}
```
shajmakh, line 681: if the code doesn't return false here the test will pass but not because the deployment is ready in all instances but because the test ignored the pod that is in different waiting reason. this will be misleading, I'm not sure this is what we want, is it? 
to have a negative check, maybe increase the requested cpus in such case that the node can't provide resources (not enough resources), and verify if the test pass, if it does then it is false negative

mrniranjan: Added additional checks to verify deployment is indeed successful

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,86 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+					}
+					for _, s := range podList.Items[0].Status.ContainerStatuses {
+						if s.Ready == false {
+							return false
+						}
+					}
+					return true
+				}, 10*time.Second, 5*time.Second).Should(BeTrue())
+				// delete deployment
+				time.Sleep(5 * time.Second)
```
Tal-or, line 694: Why this is needed?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,86 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+				testlog.Infof("Deleting Deployment %v", dp.Name)
+				err := testclient.Client.Delete(ctx, dp)
```
Tal-or, line 696: Since right after the deletion you're about to create the `Deployment` again, I would check that it's actually gone completely, by trying to `Get` it and expect `IsNotFound` error or at least move the cooling time from line 694 to here.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,86 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+							}
+						}
+					}
+					if len(podList.Items) < 1 {
+						return false
+					}
+					for _, s := range podList.Items[0].Status.ContainerStatuses {
+						if s.Ready == false {
+							return false
+						}
```
shajmakh, line 689: these should better be before the for loop `for _, pod := range podList.Items {`

shajmakh: when added then you can remove the for loop as the test knows it is working with one pod only

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,86 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+									if containerStatus.State.Waiting.Reason == "RunContainerError" && strings.Contains(containerStatus.State.Waiting.Message, "failed to run pre-start hook for container") {
+										testlog.Infof("container %s failed to start with error: %s", pod.Spec.Containers[0].Name, containerStatus.State.Waiting.Message)
+										return false
+									}
+								}
+							}
+						}
+					}
+					if len(podList.Items) < 1 {
+						return false
```
shajmakh, line 684: it would be helpful to add warnings logs before returning "false" and when returning true also add a positive klog.info that conditions are met while adding what are the expected conditions

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,86 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+					}
+					for _, s := range podList.Items[0].Status.ContainerStatuses {
```
shajmakh, line 686: when looking into adding multiple condition statements for the same object (in this case, `ContainerStatuses`), I believe it would be better to use these conditions, one after the others, to be able to follow the flow of the test when troubleshooting is needed. If the conditions can be simplified even more it would be evne better

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,80 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		)
 
 	})
+	// Automates OCPBUGS-34812: cgroupsv2: failed to write on cpuset.cpus.exclusive
+	Context("Cgroupsv2", func() {
+		It("[test_id:75327] cpus from deleted cgroup can be reassigned to new cgroup", Label(string(label.Tier0)), func() {
+
+			// we need system with more than 10 cpus to execute this test
+			if len(onlineCPUSet.List()) < 10 {
```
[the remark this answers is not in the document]
mrniranjan: yes its D/S only as it requires more than 4 cpus to execute the test

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: okay we are creating a deployment in quick succession because that's what is required to reproduce the issue.  I updated the PR description but let me put it again. 

this PR automates the bug OCPBUGS-34812 where we need to create  and delete deployment in quick succession so that the old cgroup of the delete pod which still has access to cpus is not released causing the new cgroup creation fail as the pre start hook fails to update cpuset.cpus.exclusive file.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: we want to wait for few seconds before we start the new deployment but at the same time we can't wait for too long.

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: addressed in latest commit

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: addressed in latest commit

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: addressed in latest commit

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
shajmakh: I think it's a good idea that we comment the reason as well and how did we know that 5 seconds is enough, also adding a klog before this to indicate that test is about to go into sleep for few secs

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,116 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+							}
+						}
+					}
+					if len(podList.Items) < int(numberofReplicas) {
+						testlog.Warningf("Required number of pods is %d", numberofReplicas)
+						return false
+					}
+					for _, s := range podList.Items[0].Status.ContainerStatuses {
+						if !s.Ready {
+							testlog.Warningf("pod status is %q", s.Name)
```
shajmakh, line 716: isn't it the container status?

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -615,6 +618,116 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
...
+						}
+						for _, containerStatus := range pod.Status.ContainerStatuses {
+							if containerStatus.State.Waiting == nil {
+								continue
+							}
+							// we want to test container Pending state first so that we do not skip the condition
+							// since we are creating and deleting the deployments, it's possible for this condition
+							// to be skipped if one of the pods is not in pending state
+							if containerStatus.State.Waiting.Reason == "RunContainerError" && strings.Contains(containerStatus.State.Waiting.Message, "failed to run pre-start hook for container") {
+								testlog.Infof("container %s failed to start with error: %s", pod.Spec.Containers[0].Name, containerStatus.State.Waiting.Message)
```
shajmakh, line 705: nit: I'd go with Warningf instead of Infof to be aligned with the rest of the reports in the eventually when it returns false

shajmakh:
Thanks for all the updates!
/lgtm
/hold 
for other reviewers lgtm

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: addressed in the latest commit

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
mrniranjan: addressed in the latest commit

shajmakh:
/lgtm
Thanks for your efforts on this! feel free to unhold as you see fit

PR #1131, opened 2024-08-12
OCPBUGS-36431: Fix generated cpu mask for 512+ cpus

MarSik:
The padding size was hardcoded to support max 64 nibbles which is equivalent to 256 cpus (%064x). Any mask bigger than this that was not itself 32 cpu aligned broke the chunking algorithm.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/utils.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/utils.go
@@ -47,7 +47,7 @@ func CPUListToHexMask(cpulist string) (hexMask string, err error) {
 		x := new(big.Int).Lsh(big.NewInt(1), uint(cpu))
 		currMask.Or(currMask, x)
 	}
-	return fmt.Sprintf("%064x", currMask), nil
+	return fmt.Sprintf("%0x", currMask), nil
 }
 
 // CPUListToMaskList converts a list of cpus into a cpu mask represented
@@ -57,6 +57,14 @@ func CPUListToMaskList(cpulist string) (hexMask string, err error) {
 	if err != nil {
 		return "", nil
 	}
+
+	// Make sure the raw mask can be processed in 8 character chunks
+	padding_needed := len(maskStr) % 8
+	if padding_needed != 0 {
+		padding_needed = 8 - padding_needed
+		maskStr = strings.Repeat("0", padding_needed) + maskStr
+	}
+
 	index := 0
 	for index < (len(maskStr) - 8) {
 		if maskStr[index:index+8] != "00000000" {
--- a/pkg/performanceprofile/controller/performanceprofile/components/utils_test.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/utils_test.go
@@ -19,6 +19,17 @@ var cpuListToMask = []listToMask{
 	{"1,3-7", "000000fa"},
 	{"0-127", "ffffffff,ffffffff,ffffffff,ffffffff"},
 	{"0-255", "ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff"},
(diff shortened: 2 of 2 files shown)


Tal-or:
Seems like a flake
```
[rfe_id:28761][performance] Updating parameters in performance profile Updating of nodeSelector parameter and node labels [It] [test_id:27484]Verifies that node is reverted to plain worker when the extra labels are removed [tier-2]
/go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go:483
  [FAILED] Expected
      <string>: # irqbalance is a daemon process that distributes interrupts across
      # CPUs on SMP systems.  The default is to rebalance once every 10
      # seconds.  This is the environment file that is specified to systemd via the
      # EnvironmentFile key in the service unit file (or via whatever method the init
      # system you're using has).
      
      #
      # IRQBALANCE_ONESHOT
      #    After starting, wait for ten seconds, then look at the interrupt
      #    load and balance it once; after balancing exit and do not change
      #    it again.
      #
      #IRQBALANCE_ONESHOT=
      
      #
      # IRQBALANCE_BANNED_CPUS
      #    64 bit bitmask which allows you to indicate which CPUs should
      #    be skipped when reblancing IRQs.  CPU numbers which have their
      #    corresponding bits set to one in this mask will not have any
      #    IRQs assigned to them on rebalance.
      #
      #IRQBALANCE_BANNED_CPUS=
      
      #
      # IRQBALANCE_BANNED_CPULIST
      #    The CPUs list which allows you to indicate which CPUs should
      #    be skipped when reblancing IRQs. CPU numbers in CPUs list will
      #    not have any IRQs assigned to them on rebalance.
      #
      #      The format of CPUs list is:
      #        <cpu number>,...,<cpu number>
      #      or a range:
      #        <cpu number>-<cpu number>
      #      or a mixture:
      #        <cpu number>,...,<cpu number>-<cpu number>
      #
      #IRQBALANCE_BANNED_CPULIST=
      
      #
      # IRQBALANCE_ARGS
      #    Append any args here to the irqbalance daemon as documented in the man
      #    page.
      #
      #IRQBALANCE_ARGS=
      IRQBALANCE_BANNED_CPUS=0
  not to contain substring
      <string>: 1
  In [It] at: /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go:513 @ 08/14/24 11:16:27.69
```

MarSik:
@Tal-or @mrniranjan The test failure is not a fluke. But it is unrelated to this change.

```
  not to contain substring
      <string>: 1
  In [It] at: /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go:513 @ 08/19/24 09:14:24.633 
```

The character "1" is present in one of the comments in /etc/sysconfig/irqbalance:

```
# irqbalance is a daemon process that distributes interrupts across
# CPUs on SMP systems.  The default is to rebalance once every 10
```

Tal-or:
> @Tal-or @mrniranjan The test failure is not a fluke. But it is unrelated to this change.
> 
> ```
>   not to contain substring
>       <string>: 1
>   In [It] at: /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go:513 @ 08/19/24 09:14:24.633 
> ```
> 
> The character "1" is present in one of the comments in /etc/sysconfig/irqbalance:
> 
> ```
> # irqbalance is a daemon process that distributes interrupts across
> # CPUs on SMP systems.  The default is to rebalance once every 10
> ```

Hmm I'm a bit confused though, why we're not seeing it on other PRs as well?

MarSik:
@Tal-or It was not showing on other PRs, because during the cleanup of mine I moved the padding to the right place. And the test is using the function that till now was returning a wider mask than just plain "1".

PR #1132, opened 2024-08-12
OCPBUGS-34917: Operator: do not use /tmp

jmencak:
Stop writing root certificate bundle of the metrics server for client authentication to /tmp.  There are two reasons for this.  Firstly, it is racy and it could cause the operator to temporarily start without client authentication.  Secondly, it prevents us from making the operator more secure by setting its container securityContext.readOnlyRootFilesystem=true.

Resolves: OCPBUGS-34917

### DIFF
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -15,7 +15,6 @@ import (
 
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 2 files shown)


jmencak:
This look like like infra issues. Also, this change is very unlikely to cause any test failures.
/retest

ffromani (commented):
/lgtm

PR #1136, opened 2024-08-16
PSAP-1428: Read-only filesystem for the operator

jmencak:
This change switches the operator's `securityContext.readOnlyRootFilesystem` to `true`.  This is in line with the OCP's Control Plane required privileges policy,


Tal-or:
/lgtm
Thanks!

PR #1139, opened 2024-08-20
CNF-13014: Allow specifying $TAG separately from the full image path

fontivan:
- Since $TAG by default is using the local git branch name, this can cause problems if special characters (notably '/') are used
- Add section to HACKING.md about the potential error


ffromani:
/lgtm

I'm in favor of overridable values with good defaults

jmencak:
Thank you for the PR Steve.
/lgtm
/approve

PR #1140, opened 2024-08-21
OCPBUGS-39320: E2E: Remove checking of reserved cpus in irqbalance file

mrniranjan:
Remove checking of reserved cpus in the /etc/sysconfig/irqbalance as in the current Performance profile deployment reserved cpus are never added to the irqbalance config file.


MarSik (approved):
/lgtm

The original logic of the test is checking for a condition that can never happen. Moreover the test is comparing reserved cpus with all the content of the irqbalance config files which is not great. There were accidental matches between comments and cpu no.1.

Tal-or:
/retitle NO-JIRA:E2E: Remove checking of reserved cpus in irqbalance file

yanirq:
/retitle OCPBUGS-39320: E2E: Remove checking of reserved cpus in irqbalance file

PR #1141, opened 2024-08-21
CNF-13014:  Add a specific error message for when the TuneD submodule folder is not initialized

fontivan:
- If you did not checkout the git repo using `--recursive` then the tuned submodule folder will be empty
- Without a specific error message it would instead return a vague error that a patch failed to be applied
- Instead we will check explicitly if the folder is empty and return a more helpful error message in these cases

### DIFF
--- a/hack/dockerfile_install_support.sh
+++ b/hack/dockerfile_install_support.sh
@@ -9,6 +9,12 @@ INSTALL_PKGS="nmap-ncat procps-ng pciutils"
 cp -r /root/assets/bin/* /usr/local/bin
 mkdir -p /etc/grub.d/ /boot /run/ocp-tuned
 
+# Verify TuneD submodule is not empty
+if [[ -z $(ls /root/assets/tuned/tuned) ]]; then
+    echo "TuneD submodule is an empty folder. Consider initializing the module: 'git submodule update --init --recursive'"
+    exit 1
+fi
+
 source /etc/os-release
 if [[ "${ID}" == "centos" ]]; then
 


fontivan:
Old error:
```
+ cd /root/assets/tuned/tuned
+ LC_COLLATE=C
+ cat ../patches/030-no-bootloader-errors.diff
+ patch -Np1
can't find file to patch at input line 10
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|NTO currently supports the bootloader plugin only on RHCOS via the integration
|with MCO.  Disable patching of BLS entries and grub configuration files so
|that error messages from the tuned daemon are eliminated and not reported
|further on in k8s objects.
|
|diff --git a/assets/tuned/daemon/tuned/plugins/plugin_bootloader.py b/assets/tuned/daemon/tuned/plugins/plugin_bootloader.py
|index 416df7d6..b0a61905 100644
|--- a/tuned/plugins/plugin_bootloader.py
|+++ b/tuned/plugins/plugin_bootloader.py
--------------------------
File to patch: 
Skip this patch? [y] 
Skipping patch.
3 out of 3 hunks ignored
Error: building at STEP "RUN /bin/bash /tmp/dockerfile_install_support.sh": while running runtime: exit status 1
make: *** [Makefile:175: local-image] Error 1
```

New error:
```
++ ls /root/assets/tuned/tuned
+ [[ -z '' ]]
+ echo 'TuneD submodule is an empty folder. Consider initializing the module: '\''git submodule update --init --recursive'\'''
TuneD submodule is an empty folder. Consider initializing the module: 'git submodule update --init --recursive'
+ exit 1
Error: building at STEP "RUN /bin/bash /tmp/dockerfile_install_support.sh": while running runtime: exit status 1
make: *** [Makefile:175: local-image] Error 1
```

rbaturov:
@fontivan take a look at this fix https://github.com/openshift/cluster-node-tuning-operator/pull/1159.

fontivan:
@rbaturov I like #1159 but I also think we could also do error message addition too since it is a minor error message improvement if you end up in a weird submodule state

ffromani:
if you se a label name like `something.which.looks.like.dns.com/something-else` it should pass the validation. Alternatively, try NOT using slashes at all in the labels

fontivan:
@MarSik @jirimacku @rbaturov do we still want this or is this unnecessary with #1159?

jmencak:
> @MarSik @jmencak @rbaturov do we still want this or is this unnecessary with #1159?

Can you provide steps to show us the issue you're seeing?

fontivan:
I think this is just no longer required with #1159, I don't see the original error any more on a fresh checkout

PR #1142, opened 2024-08-22
OCPBUGS-38795: Fix defer status during recommended profile change

MarSik:
Process recommended profile change even when the profile itself has not changed itself. The internal profile fingerprint tracking was not updated during recommended profile update with no internal changes.

### DIFF
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -1029,6 +1029,7 @@ func (c *Controller) changeSyncerTuneD(change Change) (synced bool, err error) {
 			// Cache the value written to tunedRecommendFile.
 			c.daemon.recommendedProfile = change.recommendedProfile
 			klog.V(1).Infof("recommended TuneD profile updated from %q to %q [inplaceUpdate=%v nodeRestart=%v]", prevRecommended, change.recommendedProfile, inplaceUpdate, change.nodeRestart)
+			changeRecommend = true
 
 			if change.deferredMode == util.DeferUpdate && !inplaceUpdate && c.daemon.recoveredRecommendedProfile == change.recommendedProfile {
 				klog.V(1).Infof("recommended TuneD profile changed; skip TuneD reload [deferred=%v recoveredRecommended=%v]", change.deferredMode, c.daemon.recoveredRecommendedProfile)
--- a/test/e2e/deferred/basic.go
+++ b/test/e2e/deferred/basic.go
@@ -73,6 +73,7 @@ var _ = ginkgo.Describe("Profile deferred", ginkgo.Label("deferred", "profile-st
 				ginkgo.By(fmt.Sprintf("cluster changes rollback: %q", createdTuned))
 				util.ExecAndLogCommand("oc", "delete", "-n", ntoconfig.WatchNamespace(), "-f", createdTuned)
 			}
+			util.ExecAndLogCommand("oc", "label", "node", targetNode.Name, tunedMatchLabelLater+"-")
 		})
 
 		ginkgo.It("should not trigger any actual change", func(ctx context.Context) {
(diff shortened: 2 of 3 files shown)


MarSik:
After the fix (with few extra logging changes):

```
2024-08-22 08:09:02,700 INFO     tuned.daemon.daemon: static tuning from profile 'openshift-node' applied
...
I0822 08:31:04.090721   43551 controller.go:1031] recommended TuneD profile updated from "openshift-node" to "openshift-profile" [inplaceUpdate=false nodeRestart=false]
I0822 08:31:04.090728   43551 controller.go:1039] recommended TuneD profile changed; trigger TuneD reload [deferred=update]
...
I0822 08:31:04.091068   43551 controller.go:478] profilesExtract(): fingerprint of extracted profiles: "46cc8c1e9bcfc1ff5635ea3065a70f3fda4deec61a59a42c7c5586291bae3c3f"
I0822 08:31:04.091108   43551 controller.go:1061] profile sync change:false fp:46cc8c1e9bcfc1ff5635ea3065a70f3fda4deec61a59a42c7c5586291bae3c3f err:<nil>
I0822 08:31:04.091141   43551 controller.go:1068] current unpacked profile fingerprint "57bf1d12635adc2a15c046890cf1d675630b2148fae2cf0231eea83c98fe66b6" -> "46cc8c1e9bcfc1ff5635ea3065a70f3fda4deec61a59a42c7c5586291bae3c3f"
I0822 08:31:04.091152   43551 controller.go:1142] recommended profile change with deferredMode=update: setting reload flag
I0822 08:31:04.091157   43551 controller.go:1175] changeSyncerRestartOrReloadTuneD()
I0822 08:31:04.091160   43551 controller.go:843] tunedReload()
...
2024-08-22 08:31:04,220 INFO     tuned.daemon.daemon: static tuning from profile 'openshift-profile' applied
...
I0822 08:31:04.221247   43551 controller.go:965] changeSyncerPostReloadOrRestart(): current effective profile fingerprint "57bf1d12635adc2a15c046890cf1d675630b2148fae2cf0231eea83c98fe66b6" -> "46cc8c1e9bcfc1ff5635ea3065a70f3fda4deec61a59a42c7c5586291bae3c3f"
...
I0822 08:31:04.235566   43551 controller.go:1369] daemonStatus(): isApplied -> true: fpUnpacked=46cc8c1e9bcfc1ff5635ea3065a70f3fda4deec61a59a42c7c5586291bae3c3f fpEffective=46cc8c1e9bcfc1ff5635ea3065a70f3fda4deec61a59a42c7c5586291bae3c3f
I0822 08:31:04.235585   43551 controller.go:1372] daemonStatus(): change: deferred=true applied=true nodeRestart=false
I0822 08:31:04.235592   43551 controller.go:1386] computed status conditions: []v1.ProfileStatusCondition{v1.ProfileStatusCondition{Type:"Applied", Status:"True", LastTransitionTime:time.Date(2024, time.August, 22, 6, 37, 35, 0, time.Local), Reason:"AsExpected", Message:"TuneD profile applied."}
```

MarSik:

We need this in time for 4.17 and all the tests seem to be passing.

pkg/tuned/controller.go
```diff
@@ -1029,6 +1029,7 @@ func (c *Controller) changeSyncerTuneD(change Change) (synced bool, err error) {
 			// Cache the value written to tunedRecommendFile.
 			c.daemon.recommendedProfile = change.recommendedProfile
 			klog.V(1).Infof("recommended TuneD profile updated from %q to %q [inplaceUpdate=%v nodeRestart=%v]", prevRecommended, change.recommendedProfile, inplaceUpdate, change.nodeRestart)
+			changeRecommend = true
```
jmencak, line 1032: Good catch with the variable not being set anywhere in this function.  However, can we possibly get rid of the variable completely?  Would it work if we replaced it further down just by `change.nodeRestart` ?  I haven't tested this (yet), but looking at the code it should work.  It would also resolve the slight mis-placement of the variable setting below the logging message and its not very intuitive name.

MarSik: I was thinking about that, but there are two separate triggers for this flow:

- !inplaceUpdate -- the recommended profile was changed for a live system
- change.nodeRestart -- first execution after reboot

The !inplaceUpdate is actually the case that was missed and caused the bug.

Both variables are later used for flow branching and the changeRecommend bool is the combination of both. I agree it is not easily visible or understandable, but I was a bit afraid to touch the logic tbh.

jmencak: Yeah you're right.  I was just testing that and my suggestion didn't work.  I agree `!inplaceUpdate` causes that issue.  

> I was a bit afraid to touch the logic tbh.

Understood.  Well, I wonder if the entire flow needs a bit of cleanup / adding comments once @ffromani is back.  In the meantime, I guess we can merge if this fixes the issue.

jmencak:
/approve
/lgtm
I can confirm this fixes https://issues.redhat.com/browse/OCPBUGS-38795 .  (Not fix https://issues.redhat.com/browse/OCPBUGS-38796).
However, I'd appreciate some follow-up cleanup/comments in the affected deferred updates flow.

PR #1144, opened 2024-08-27
OCPBUGS-39005: Add cluster-wide proxy env file

jmencak:
Add cluster-wide proxy environment file /etc/mco/proxy.env to ocp-tuned-one-shot.service file.  This allows podman to pull NTO image in environments which need to use http(s) proxy for out-of-cluster connections.

Resolves: OCPBUGS-39005


yanirq:
/hold 
@jmencak  feel free to remove if no additional reviews/tests are needed

jmencak:
This needs a bit more work.  The `ocp-tuned-one-shot.service` still failed when I tried to this with a reproducer.  I believe the HTTP(S)_PROXY variables need to be exported.  podman in the systemd unit doesn't see the variables.

MarSik (approved):
/lgtm

jmencak:
I was able to successfully test that podman from `ocp-tuned-one-shot.service` was able to pull a custom NTO image via using `HTTPS_PROXY`.  The previous testing incorrectly used the `https://` prefix for the `HTTPS_PROXY` proxy.  Adding @liqcui so that he doesn't fall into the same trap as me when testing.

Lifting hold
/hold cancel

PR #1151, opened 2024-09-02
NO-JIRA: Remove upstream TuneD patch

jmencak:
Remove `030-no-bootloader-errors.diff`.  Any TuneD patch that is not upstream is a bad idea/temporary fix.  Also, none of the patches in NTO repository are shipped officially with OCP and their presence can be misleading to contributors.  The functionality provided by the patch can be worked around by configuring TuneD in the `[bootloader]` section.

Other changes:
  * Simplify switchTunedHome as suggested by a reviewer in an earlier review.


jmencak:
@MarSik / @rbaturov any thoughs on this PR? Thank you.

PR #1153, opened 2024-09-03
[release-4.16] OCPBUGS-39379: E2E: Add test to verify cpuset.cpus.exclusive is writeable

mrniranjan:
Automates OCPBUGS-34812: cgroupsv2: failed to write on  cpuset.cpus.exclusive

To reproduce the bug, we need to create and delete deployment(deploying guaranteed pods with cpu load balancing annotation) in quick succession, so that we do not fully wait for the cleanup causing the pod about to be deleted to still have access to exclusive cpus causing the new pod to fail because cpuset.cpus.exclusive is not yet freed.

As the pre-start hook fails to write to cpuset.cpus.exclusive file in the pods cgroup pod goes to RunContainerError state.

This automation PR creates and deletes deployment in loop to reproduce the issue and checks if the pods fails with Runtime error with message "failed to run pre-start hook for container"

Manual backport of  #1127

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -9,10 +9,12 @@ import (
 	"strings"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -3,6 +3,9 @@ package __performance
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/label"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/labels"
```
shajmakh, line 8: imports can be reordered better

shajmakh:
/lgtm
Thanks!

PR #1149, opened 2024-09-01
[release-4.17] OCPBUGS-38721: tuned: distinguish deferred updates

yanirq:
This is a manual cherry pick of:
https://github.com/openshift/cluster-node-tuning-operator/pull/1129 OCPBUGS-28647: tuned: distinguish deferred updates
https://github.com/openshift/cluster-node-tuning-operator/pull/1142 OCPBUGS-38795: Fix defer status during recommended profile change 
https://github.com/openshift/cluster-node-tuning-operator/pull/1162 NO-JIRA: deferred updates: fix in-place update handling on reboot

PR #1162, opened 2024-09-09
NO-JIRA: deferred updates: fix in-place update handling on reboot

ffromani:
There are conditions on which we should not set the reload flag. This avoid regression in the test
"Profile deferred when applied should trigger changes when applied fist, then deferred when edited, if tuned restart should be kept deferred"

### DIFF
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -357,6 +357,8 @@ func (c *Controller) sync(key wqKeyKube) error {
 		err = util.SetLogLevel(profile.Spec.Config.Verbosity)
 		if err != nil {
 			klog.Errorf("failed to set log level %d: %v", profile.Spec.Config.Verbosity, err)
+		} else {
+			klog.Infof("set log level %d", profile.Spec.Config.Verbosity)
 		}
 		change.reapplySysctl = true
 		if profile.Spec.Config.TuneDConfig.ReapplySysctl != nil {
@@ -1727,7 +1729,7 @@ func RunInCluster(stopCh <-chan struct{}, version string) error {
 	c.daemon.profileFingerprintUnpacked = profileFP
 	// We should *never* recover `c.daemon.recommendedProfile`.
 	// NTO operand supervises the TuneD daemon, so we must trigger
-	// a TuneD restart. Currently, the main trigger for the tuned
+	// a TuneD restart. Currently, the main reason to trigger a tuned
 	// reload to be triggered is when the current recommended profile
 	// is different from the change.recommended profile (see changeSyncerTuneD).
 	// Hence, we MUST NOT recover the recommended profile name
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -1032,8 +1032,15 @@ func (c *Controller) changeSyncerTuneD(change Change) (synced bool, err error) {
 			c.daemon.recommendedProfile = change.recommendedProfile
 			klog.V(1).Infof("recommended TuneD profile updated from %q to %q [inplaceUpdate=%v nodeRestart=%v]", prevRecommended, change.recommendedProfile, inplaceUpdate, change.nodeRestart)
 
-			if change.deferredMode == util.DeferUpdate && !inplaceUpdate && c.daemon.recoveredRecommendedProfile == change.recommendedProfile {
-				klog.V(1).Infof("recommended TuneD profile changed; skip TuneD reload [deferred=%v recoveredRecommended=%v]", change.deferredMode, c.daemon.recoveredRecommendedProfile)
+			// If we get this far, it's either because we detected a node restart or the recommended profile changed.
+			// If it's a node restart, we just process the change. If it's the latter -- TuneD profile switch, i.e. not an in-place update -- we need further
+			// logic to distinguish between the "update" and "always" DeferMode.
+			// For the "update" DeferMode, we only defer profile edits. If we are re-processing a profile which we already applied, then we must
+			// not reload tuned, otherwise we're missing the point of deferred updates.
+			// See the test "Profile deferred when applied should trigger changes when applied first, then deferred when edited, if tuned restart should be kept deferred"
+			// See the commit 3655f22656d4a3aa9f471099305dcd78a9c80320
+			if !inplaceUpdate && change.deferredMode == util.DeferUpdate && c.daemon.recoveredRecommendedProfile == change.recommendedProfile {
+				klog.V(1).Infof("reprocessing profile already in effect; this seems a daemon reload. Skip TuneD reload [deferred=%v recoveredRecommended=%v]", change.deferredMode, c.daemon.recoveredRecommendedProfile)
 				// Reset because we need only once the first time we process the TuneD k8s object. Let's avoid stale data.
 				c.daemon.recoveredRecommendedProfile = ""
 			} else {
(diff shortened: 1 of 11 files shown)


pkg/tuned/controller.go
```diff
@@ -1029,19 +1031,18 @@ func (c *Controller) changeSyncerTuneD(change Change) (synced bool, err error) {
...
 			c.daemon.recommendedProfile = change.recommendedProfile
 			klog.V(1).Infof("recommended TuneD profile updated from %q to %q [inplaceUpdate=%v nodeRestart=%v]", prevRecommended, change.recommendedProfile, inplaceUpdate, change.nodeRestart)
-			changeRecommend = true
 
 			if change.deferredMode == util.DeferUpdate && !inplaceUpdate && c.daemon.recoveredRecommendedProfile == change.recommendedProfile {
 				klog.V(1).Infof("recommended TuneD profile changed; skip TuneD reload [deferred=%v recoveredRecommended=%v]", change.deferredMode, c.daemon.recoveredRecommendedProfile)
 				// Reset because we need only once the first time we process the TuneD k8s object. Let's avoid stale data.
 				c.daemon.recoveredRecommendedProfile = ""
 			} else {
 				klog.V(1).Infof("recommended TuneD profile changed; trigger TuneD reload [deferred=%v]", change.deferredMode)
```
MarSik, line 1040: The log still says tuned should be reloaded. I will confuse people working on this in the future.. I think.

ffromani: Well, the flow will ultimately trigger a tuned reload: see lines around 1065.
But still: how can we improve the logging?

MarSik: Ah, right! So this is a fix that updates the fingerprint in addition to setting the reload flag. The reload part is not changing. Ok then.

MarSik (approved):

pkg/tuned/controller.go
ffromani: we do want to _not_ set the reload flag if very specific conditions are met, like the one captured in the test mentioned in the commit message, which is constantly failing without this fix

ffromani:
/hold cancel

managed to deflake further (on my env) but not completely

pkg/tuned/controller.go
```diff
@@ -1029,19 +1031,18 @@ func (c *Controller) changeSyncerTuneD(change Change) (synced bool, err error) {
...
-				reload = true
+				changeRecommend = true
 			}
 		} else if util.IsImmediateUpdate(change.deferredMode) && (c.daemon.status&scDeferred != 0) {
 			klog.V(1).Infof("detected deferred update changed to immediate after object update")
-			reload = true
+			changeRecommend = true
```
jmencak, line 1045: What does `changeRecommend` mean exactly?  I'm assuming it indicates that the recommended profile changed. Is that the case here too?  If not, I'd say we need a better name or a variable.

pkg/tuned/controller.go
```diff
@@ -1029,19 +1031,18 @@ func (c *Controller) changeSyncerTuneD(change Change) (synced bool, err error) {
 			// Cache the value written to tunedRecommendFile.
 			c.daemon.recommendedProfile = change.recommendedProfile
 			klog.V(1).Infof("recommended TuneD profile updated from %q to %q [inplaceUpdate=%v nodeRestart=%v]", prevRecommended, change.recommendedProfile, inplaceUpdate, change.nodeRestart)
-			changeRecommend = true
 
 			if change.deferredMode == util.DeferUpdate && !inplaceUpdate && c.daemon.recoveredRecommendedProfile == change.recommendedProfile {
```
jmencak, line 1035: Is it possible to improve the comments/logging for this section?   I'm also a bit uneasy with the complex condition without a comment explaining in what state exactly we've currently landed in.  Ideally, the log message below should also explain it.

jmencak (commented):
Thank you for the changes, Francesco.  They look good to me.  I've added some comments which are not exactly fair, because they should've been added them during previous reviews.  Nevertheless, I believe that adding better logging/comments what happens (especially) in the 
```
change.deferredMode == util.DeferUpdate && !inplaceUpdate && c.daemon.recoveredRecommendedProfile == change.recommendedProfile
```
section would help a lot future maintainers.

pkg/tuned/controller.go
ffromani: right, in this case is better to set the master flag `reload` directly, rather than later checking the combination of `changeRecommend` and `changeProfiles`

pkg/tuned/controller.go
ffromani: added improvements, thanks for raising the point

ffromani:
/hold

want to rerun the e2e testsuite

jmencak:
Thank you for the changes/clarification, Francesco!  I've found some typos and things that (IMO) need further clarification.  I'll post a suggested diff later on and we can work from there.
/hold

jmencak:
I tried to clarify the recently added comments and remove the typos.  Hopefully I didn't introduce any false information/typos.  Feel free to adjust as you see fit and happy to discuss here of over Slack:
```
index e6095297..627f4271 100644
--- a/pkg/tuned/controller.go
+++ b/pkg/tuned/controller.go
@@ -1032,11 +1032,12 @@ func (c *Controller) changeSyncerTuneD(change Change) (synced bool, err error) {
 			c.daemon.recommendedProfile = change.recommendedProfile
 			klog.V(1).Infof("recommended TuneD profile updated from %q to %q [inplaceUpdate=%v nodeRestart=%v]", prevRecommended, change.recommendedProfile, inplaceUpdate, change.nodeRestart)
 
-			// if we get this far, it's either because we detected a node restart or the profile changed (not in-place, thus not-edit).
-			// So first: if this is a edit (replacing, not in-place), we do further logic. If it's a node restart, we just go ahead and process it.
-			// Handling then edits. Id we need to defer only edits, and we are re-processing a profile which we already applied, then we must
-			// not reload tuned, otherwise we're missing the point entirely about deferred updates.
-			// See the test "Profile deferred when applied should trigger changes when applied fist, then deferred when edited, if tuned restart should be kept deferred"
+			// If we get this far, it's either because we detected a node restart or the recommended profile changed.
+			// If it's a node restart, we just process the change. If it's the latter -- TuneD profile switch, i.e. not an in-place update -- we need further
+			// logic to distinguish between the "update" and "always" DeferMode.
+			// For the "update" DeferMode, we only defer profile edits. If we are re-processing a profile which we already applied, then we must
+			// not reload tuned, otherwise we're missing the point of deferred updates.
+			// See the test "Profile deferred when applied should trigger changes when applied first, then deferred when edited, if tuned restart should be kept deferred"
 			// See the commit 3655f22656d4a3aa9f471099305dcd78a9c80320
 			if !inplaceUpdate && change.deferredMode == util.DeferUpdate && c.daemon.recoveredRecommendedProfile == change.recommendedProfile {
 				klog.V(1).Infof("reprocessing profile already in effect; this seems a daemon reload. Skip TuneD reload [deferred=%v recoveredRecommended=%v]", change.deferredMode, c.daemon.recoveredRecommendedProfile)

```

jmencak:
/lgtm
the e2e-gcp-pao and e2e-hypershift-pao failures are unrelated
/retest

jmencak:
Hmm, I think `e2e-gcp-pao` test needs looking into.  I'm getting the same failures (`irqbalance.go`) in https://github.com/openshift/cluster-node-tuning-operator/pull/1163 and also when running the test locally.

jmencak:
Overriding 2 tests because of a known change in TuneD. The tests already passed with the previous version of TuneD and need to be adjusted.

/override ci/prow/e2e-gcp-pao
/override ci/prow/e2e-hypershift-pao

yanirq:
/retitle NO-JIRA: deferred updates: fix in-place update handling on reboot

to be added manually to https://github.com/openshift/cluster-node-tuning-operator/pull/1149

PR #1165, opened 2024-09-16
WIP: amend irqbalance tests to check IRQBALANCE_BANNED_CPUS

yanirq:
Use IRQBALANCE_BANNED_CPULIST instead of the deprecated IRQBALANCE_BANNED_CPUS in irqbalance tests to match tuned irqbalance plugin

xref: https://github.com/redhat-performance/tuned/pull/657/commits/7825b1133039f370e0863a0f824e842e237e4123

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -334,7 +334,7 @@ func getIrqBalanceBannedCPUs(ctx context.Context, node *corev1.Node) (cpuset.CPU
 	bannedCPUs := unquote(strings.TrimSpace(items[1]))
 	testlog.Infof("banned CPUs: %q", bannedCPUs)
 
-	banned, err := components.CPUMaskToCPUSet(bannedCPUs)
+	banned, err := cpuset.Parse(bannedCPUs)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jmencak:
If we do this change, will this not break the potential `TuneD` https://github.com/redhat-performance/tuned/pull/657 revert?

yanirq:
> If we do this change, will this not break the potential `TuneD` [redhat-performance/tuned#657](https://github.com/redhat-performance/tuned/pull/657) revert?

it will, currently setting this as WIP.  I might add a more modular option to check for both values if required

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -359,7 +359,7 @@ func findIrqBalanceBannedCPUsVarFromConf(conf string) string {
 		if strings.HasPrefix(line, "#") {
 			continue
 		}
-		if !strings.HasPrefix(line, "IRQBALANCE_BANNED_CPUS") {
+		if !strings.HasPrefix(line, "IRQBALANCE_BANNED_CPULIST") {
```
mrniranjan, line 362: Is this change alone sufficient, because the CPULIST is now contains the list where as the IRQBALANCE_BANNED_CPUS was mask ?,

PR #1172, opened 2024-09-23
NO-JIRA: Drop GOOS=linux from Makefile

jmencak:
Setting the value explicitly causes issues during cross-building and running tests from other OSes, such as darwin.  Drop this explicit setting.


jmencak:
/cc @yanirq
for thoughts whether this would break his development workflow or not.

jmencak:
@fontivan , any thoughts on this? Thank you.

fontivan:
> @fontivan , any thoughts on this? Thank you.

I think its fine if we just drop it, golang should default to whatever you are compiling on anyway I think?

If we really need cross building between darwin and linux then it can be done similar to the work I did for arm/x86 but I doubt it would ever get used

PR #1174, opened 2024-09-24
NO-JIRA: Fix context deadlines in ExecCommandOnPod()

jmencak:
During testing on x86_64 cluster from an arm64 host, I've
observed strange context deadlines when enabling stdin in
ExecCommandOnPod() function.

Disable stdin as it is not needed during non-interactive tests.

Other changes:
  * Drop obsolete test "test_id:32375".
  * Drop unnecessary "/bin/bash -c" in some tests.
    Further cleanup possible, but it could make code
    slightly less readable, so avoiding it for now.


Tal-or:
/lgtm
Thank you for figure this out

PR #1175, opened 2024-09-26
NO-JIRA: Fix failing pao-functests after a MCO change

jmencak:
Recent MCO change https://github.com/openshift/machine-config-operator/pull/4557 enforced custom MCP to have worker label.  This caused an issue with "Create second performance profiles on a cluster" pao-functest. Adjust the tests's MachineConfigPool Spec to use MatchExpressions instead of MatchLabels, so that the new MCO requirement "All custom MachineConfigPools must inherit from the worker pool and must have a machineConfigSelector that matches for a
'machineconfiguration.openshift.io/role: worker' label" is met.


jmencak:
More context: https://redhat-internal.slack.com/archives/CQNBUEVM2/p1727268390501439

ffromani (commented):
/lgtm

PR #1182, opened 2024-10-14
NO-JIRA: CI: unblock 

Tal-or:
This PR should unblock CI in the short term.
We're doing this out of necessary to relief the pressure from the queue 
and we'll keep investigate the root cause in the background.

issues: 
1. https://issues.redhat.com/browse/OCPBUGS-43280 
2. Terminal configuration has changed and now it provides date and color which breaks the tests


jmencak:
```
 [38;5;243m[It] /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/1_performance/netqueues.go:91[0m

  [38;5;9m[FAILED] Unexpected error:
      <*errors.errorString | 0xc000700180>: 
      failed to run command [tuned-adm profile_info openshift-node-performance-performance]: output "Profile name:\nopenshift-node-performance-performance\n\nProfile summary:\nOpenshift node optimized for deterministic performance at the cost of increased power consumption, focused on low latency network performance. Based on Tuned 2.11 and Cluster node tuning (oc 4.5)\n\nProfile description:\n\n"; error "Cannot talk to TuneD daemon via DBus. Is TuneD daemon running?\n"
      {
          s: "failed to run command [tuned-adm profile_info openshift-node-performance-performance]: output \"Profile name:\\nopenshift-node-performance-performance\\n\\nProfile summary:\\nOpenshift node optimized for deterministic performance at the cost of increased power consumption, focused on low latency network performance. Based on Tuned 2.11 and Cluster node tuning (oc 4.5)\\n\\nProfile description:\\n\\n\"; error \"Cannot talk to TuneD daemon via DBus. Is TuneD daemon running?\\n\"",
      }
  occurred[0m
```
`Cannot talk to TuneD daemon via DBus. Is TuneD daemon running?` is expected.  DBus is not supported in the container and never was.

Tal-or:
/retest
Only because I dont have time to take a deeper look, so let's see if it's a flake or real persistent issue

jmencak:
> a flake or real persistent issue

It is a real persistent issue, I can reproduce this locally.  I believe [the check for `err` should be dropped](https://github.com/openshift/cluster-node-tuning-operator/blob/1685a8fc13765c6b0e03bb0b45a92e97fc64dbe5/test/e2e/performanceprofile/functests/1_performance/netqueues.go#L62) and we should only check for exit status 0.  As I stated above, you'll always get this error message from `TuneD`.  No idea why this wasn't an issue with `Tty: true`.  It looks to me this never worked -- the error output wasn't ever sent to `err` and now it finally is?

Tal-or:
> > a flake or real persistent issue
> 
> It is a real persistent issue, I can reproduce this locally. I believe [the check for `err` should be dropped](https://github.com/openshift/cluster-node-tuning-operator/blob/1685a8fc13765c6b0e03bb0b45a92e97fc64dbe5/test/e2e/performanceprofile/functests/1_performance/netqueues.go#L62) and we should only check for exit status 0. As I stated above, you'll always get this error message from `TuneD`. No idea why this wasn't an issue with `Tty: true`. It looks to me this never worked -- the error output wasn't ever sent to `err` and now it finally is?

I don't like the way we're handling issues in the CI lately and surly it will come back at us later on.

I mean we already gave ourselves some slack already didn't we? 

IMO we should at least try understand where this error come from. Otherwise I don't see a point in a test that doesn't check for errors.

Also, there might be another way to check that tuned applied on the node. Maybe we can check Tuned profile instead.

jmencak:
> I don't like the way we're handling issues in the CI lately and surly it will come back at us later on.

Well there is many things "I don't like".  For example the fact this e2e test was ever merged, because it checks for wrong things and doesn't seem to catch real failures.

> IMO we should at least try understand where this error come from.

We already know this.  This error comes from TuneD and (as I stated) was present there since the beginning.  It is not my fault the e2e tests were not written in a way to catch these "issues" in the past.  Putting "issues" in double quotes, because tuned-adm still returns a valid results and exits 0.  That is the key to check. There are many programs that return stuff to stderr which should not be considered failures.  A properly written tool will exit non-zero on failure.  Are the e2e tests checking for this?

What we (I) do not currently understand is why the `stderr` was not passed with `Tty: true` in the past.  

> Otherwise I don't see a point in a test that doesn't check for errors.

Exactly, the tests should check for errors, we agree on this.  But for errors which are real errors and not just warnings thrown on stderr.  There are many standard command-line tools which put help messages on stderr.  Is that an error?  No.

> Also, there might be another way to check that tuned applied on the node. Maybe we can check Tuned profile instead.

Yes, that's a possibility and _perhaps_ a better one.

Tal-or:
> Well there is many things "I don't like". For example the fact this e2e test was ever merged, because it checks for wrong things and doesn't seem to catch real failures.

+1 and we have many others unfortunately. 

> What we (I) do not currently understand is why the stderr was not passed with Tty: true in the past.

It might be related to the recent change we're seeing in the shell in general, and it was masked up until now.

With the recent change the test will check that tuneD posses the right profile name. 
This kind of check is actually meaningful

Tal-or:
/retest 
Seems like infra issue.
At least pao-gcp finally passed

test/e2e/performanceprofile/functests/1_performance/netqueues.go
```diff
@@ -55,11 +55,16 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
 
 		tunedPaoProfile := fmt.Sprintf("openshift-node-performance-%s", performanceProfileName)
 		//Verify the tuned profile is created on the worker-cnf nodes:
-		tunedCmd := []string{"tuned-adm", "profile_info", tunedPaoProfile}
+		// direct the error to /dev/null on purpose because tuneD always shows the following error:
+		// "Cannot talk to TuneD daemon via DBus. Is TuneD daemon running?"
+		// Which causes the test to fail, but it's a false-positive
+		tunedCmd := []string{"/bin/sh", "-c", fmt.Sprintf("tuned-adm profile_info %s 2>/dev/null | grep ^openshift-", tunedPaoProfile)}
```
ffromani, line 61: I get (and I fully agree) with your intent but arguably redirecting stderr is another form of ignoring errors
But still I like more this variant because the test is stricter and seems more targeted

ffromani:
> > I don't like the way we're handling issues in the CI lately and surly it will come back at us later on.
> 
> Well there is many things "I don't like". For example the fact this e2e test was ever merged, because it checks for wrong things and doesn't seem to catch real failures.

I concur. The problem is not really merging bad tests. This happens. Sometimes we don't know better, sometimes we learn as we go, sometimes it's just how things are. The real killer is we rarely, if ever, go back and review our tests to improve them, removing when necessary (hopefully rarely). We do this exercise pretty much only when stuff breaks, which is not idea,

> > IMO we should at least try understand where this error come from.
> 
> We already know this. This error comes from TuneD and (as I stated) was present there since the beginning. It is not my fault the e2e tests were not written in a way to catch these "issues" in the past. Putting "issues" in double quotes, because tuned-adm still returns a valid results and exits 0. That is the key to check. There are many programs that return stuff to stderr which should not be considered failures. A properly written tool will exit non-zero on failure. Are the e2e tests checking for this?
[...]
> _Edit_: Actually, we _likely_ do test for non 0 exit code in ExecCommandOnPod(), but that information is lost due to the the function returns. It is _likely_ combined into stderr, which IMO is just wrong.

Yes, the function should be changed to not interpret stuff sent to stderr as error. Is not necessarily an error.
Better to address this in another PR because it can open a pretty big can of worms.

test/e2e/performanceprofile/functests/utils/pods/pods.go
```diff
@@ -148,7 +148,7 @@ func ExecCommandOnPod(c *kubernetes.Clientset, pod *corev1.Pod, containerName st
 			Stdin:     false,
 			Stdout:    true,
 			Stderr:    true,
-			TTY:       true,
+			TTY:       false,
```
ffromani, line 151: uuuuuuhmmm
from the `client-go` docstrings:
```
	// StreamWithContext initiates the transport of the standard shell streams. It will
	// transport any non-nil stream to a remote system, and return an error if a problem
	// occurs. If tty is set, the stderr stream is not used (raw TTY manages stdout and
	// stderr over the stdout stream).
	// The context controls the entire lifetime of stream execution.
```

this can explain a lot of things

Tal-or: Nice catch. This explains why we didn't face the issue with TuneD so far

Tal-or:
> /hold /lgtm
> 
> This is not perfect, but it seems the best option atm to stop the bleeding, and I do think it's heading in the right direction

Unless there is an actual blocker, I highly suggest to merge this PR  (ofc if all tests are passing) and iterate again later.
The PRs queue is already piling up

ffromani:
> > /hold /lgtm
> > This is not perfect, but it seems the best option atm to stop the bleeding, and I do think it's heading in the right direction
> 
> Unless there is an actual blocker, I highly suggest to merge this PR (ofc if all tests are passing) and iterate again later. The PRs queue is already piling up

I concur, hence the LGTM. Holding to let other ppl chime in.

Tal-or:
@jmencak PTAL and lift the hold if you're happy

jmencak:
Thank you for this PR. Sorry, couldn't review earlier.  Still technically PTO.

Let's improve the `ExecCommandOnPod()` in a follow-up PR.  We need to unblock the CI ASAP.
/approve
/hold cancel

jmencak:
I see the same issue as described by [OCPBUGS-43280](https://github.com/openshift/cluster-node-tuning-operator/pull/1182#top) in 4.17 lane too.

/cherry-pick release-4.17

PR #1184, opened 2024-10-17
OCPBUGS-43832: E2E: Add tests to support NTO over hypershift with multi nodepools.

SargunNarula:
- Added tests for Multiple Performance Profile under different nodepools
- Adjusted nodepool utility to handle multiple nodepool
- Added test suite run in makefile

### DIFF
--- a/test/e2e/performanceprofile/functests/12_hypershift/12_hypershift_suite_test.go
+++ b/test/e2e/performanceprofile/functests/12_hypershift/12_hypershift_suite_test.go
@@ -1,0 +1,48 @@
+package __hypershift_test
+
+import (
+	"context"
+	"log"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/go-logr/stdr"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/namespaces"
+	nodeinspector "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/node_inspector"
+)
+
+var _ = BeforeSuite(func() {
+	Expect(testclient.ClientsEnabled).To(BeTrue())
+	err := testclient.DataPlaneClient.Create(context.TODO(), namespaces.TestingNamespace)
+	if errors.IsAlreadyExists(err) {
+		testlog.Warning("test namespace already exists, that is unexpected")
+		return
+	}
+	Expect(err).ToNot(HaveOccurred())
+})
+
+var _ = AfterSuite(func() {
+	err := testclient.DataPlaneClient.Delete(context.TODO(), namespaces.TestingNamespace)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(namespaces.WaitForDeletion(testutils.NamespaceTesting, 5*time.Minute)).To(Succeed())
+	Expect(nodeinspector.Delete(context.TODO())).To(Succeed())
+})
+
+func TestHypershift(t *testing.T) {
+	ctrllog.SetLogger(stdr.New(log.New(os.Stderr, "", log.LstdFlags|log.Lshortfile)))
+
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "PAO Hypershift tests")
+}
--- a/test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
+++ b/test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
@@ -1,0 +1,205 @@
+package __hypershift
+
+import (
+	"context"
+	"fmt"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	manifestsutil "github.com/openshift/cluster-node-tuning-operator/pkg/util"
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/hypershift"
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodepools"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
+	hypershiftv1beta1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+)
+
+var _ = Describe("Multiple performance profile in hypershift", Label("Hypershift"), Ordered, func() {
+	var profile, secondProfile *performancev2.PerformanceProfile
+	var nodePools []*hypershiftv1beta1.NodePool
+	var np *hypershiftv1beta1.NodePool
+	var workerRTNodes []corev1.Node
+	var err error
+
+	chkKubeletConfig := []string{"cat", "/rootfs/etc/kubernetes/kubelet.conf"}
+	chkKubeletConfigFn := func(ctx context.Context, node *corev1.Node) (string, error) {
+		out, err := nodes.ExecCommand(ctx, node, chkKubeletConfig)
+		if err != nil {
+			return "", err
+		}
+		output := testutils.ToString(out)
+		return output, nil
+	}
+
+	BeforeAll(func() {
+		By("Checking if discovery mode is enabled and performance profile is not found")
+		if discovery.Enabled() && testutils.ProfileNotFound {
+			Skip("Discovery mode enabled, performance profile not found")
+		}
+		nodeLabel := testutils.NodeSelectorLabels
+		testlog.Infof("Node label: %v", nodeLabel)
+		workerRTNodes, err = nodes.GetByLabels(nodeLabel)
+		Expect(err).ToNot(HaveOccurred())
+		profile, err = profiles.GetByNodeLabels(nodeLabel)
+		Expect(err).ToNot(HaveOccurred())
+		hostedClusterName, err := hypershift.GetHostedClusterName()
+		Expect(err).ToNot(HaveOccurred())
+		testlog.Infof("Hosted Cluster: %q", hostedClusterName)
+		nodePools, err = ListNodePools(context.TODO(), testclient.ControlPlaneClient, hostedClusterName)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(nodePools).ToNot(BeEmpty(), "no node pools found")
+		if len(nodePools) < 2 {
+			Skip("Skipping the test as the number of node pools is less than 2")
+		}
+		PrintNodePoolProfiles(nodePools)
+
+		By("Creating a deep copy of the performance profile for the second node pool")
+		secondProfile = profile.DeepCopy()
+		secondProfile.Name = "second-profile"
+		np = nodePools[1]
+		Expect(len(np.Spec.TuningConfig)).To(BeZero(), "Expected no profile attached to nodepool, but found %d profiles", len(np.Spec.TuningConfig))
+		By("Creating the second performance profile in the control plane")
+		Expect(testclient.ControlPlaneClient.Create(context.TODO(), secondProfile)).To(Succeed(), "Failed to create the performance profile")
+
+		By("Attaching the tuning object to the second node pool")
+		Expect(nodepools.AttachTuningObjectToNodePool(context.TODO(), testclient.ControlPlaneClient, secondProfile, nodePools[1])).To(Succeed())
+
+		By("Waiting for the second node pool configuration to start updating")
+		err = nodepools.WaitForUpdatingConfig(context.TODO(), testclient.ControlPlaneClient, np.Name, np.Namespace)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Waiting for the second node pool configuration to be ready")
+		err = nodepools.WaitForConfigToBeReady(context.TODO(), testclient.ControlPlaneClient, np.Name, np.Namespace)
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	Context("Multiple Nodepool", Ordered, func() {
+		isolated := performancev2.CPUSet("1-2")
+		reserved := performancev2.CPUSet("0,3")
+		policy := "best-effort"
+
+		It("[test_id:75077] should verify support for different performance profiles on hosted cluster via multiple node pools", func() {
+			By("Checking if reserved CPUs are correctly set on the worker nodes")
+			for _, node := range workerRTNodes {
+				By(fmt.Sprintf("Checking kubelet configuration for node %s", node.Name))
+
+				result, err := chkKubeletConfigFn(context.TODO(), &node)
+				Expect(err).ToNot(HaveOccurred(), "Failed to fetch kubelet configuration")
+
+				obj, err := manifestsutil.DeserializeObjectFromData([]byte(result), kubeletconfigv1beta1.AddToScheme)
+				Expect(err).ToNot(HaveOccurred(), "Failed to deserialize kubelet configuration")
+
+				kc, ok := obj.(*kubeletconfigv1beta1.KubeletConfiguration)
+				Expect(ok).To(BeTrue(), "Deserialized object is not of type KubeletConfiguration")
+
+				Expect(kc.ReservedSystemCPUs).To(Equal("0"), "ReservedSystemCPUs is not correctly set")
+				Expect(kc.TopologyManagerPolicy).To(Equal("single-numa-node"), "TopologyManagerPolicy is not correctly set")
+			}
+		})
+
+		It("[test_id:75078] should verify that Performance Profile update re-creates only target nodepool nodes", func() {
+			By("Printing node pool profiles before updating the performance profile")
+			PrintNodePoolProfiles(nodePools)
+
+			By("Modifying the second profile CPU and NUMA configurations")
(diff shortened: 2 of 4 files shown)


test/e2e/performanceprofile/functests/12_hypershift/12_hypershift_suite_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/namespaces"
+	nodeinspector "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/node_inspector"
+)
+
+var _ = BeforeSuite(func() {
+	// create test namespace
```
shajmakh, line 29: let's add this check before:
`Expect(testclient.ClientsEnabled).To(BeTrue())`

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/hypershift"
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodepools"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
+	hypershiftv1beta1 "github.com/openshift/hypershift/api/hypershift/v1beta1"
+	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
```
shajmakh, line 23: it's better to have the imports groups in the following shape: 
(
go libs

framework packages like gomega/ginkgo

k8s 3rd party packages

other 3rd party packages (like hypershift)

local packages
)

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+var _ = Describe("Multiple performance profile in hypershift", Ordered, func() {
```
shajmakh, line 26: adding [hypershift] tag and/or `Label("hypershift")" would provide additional more stable way to run this set of tests via focus.

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+	var profile, secondProfile *performancev2.PerformanceProfile
+	var nodePools []*hypershiftv1beta1.NodePool
+	var np *hypershiftv1beta1.NodePool
+	var workerRTNodes []corev1.Node
+	var err error
+
+	nodeLabel := testutils.NodeSelectorLabels
```
shajmakh, line 33: I don't see other use outside BeforeAll, let's move it below close to its use

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+		output := testutils.ToString(out)
+		return output, nil
+	}
+
+	BeforeAll(func() {
+		By("Checking if discovery mode is enabled and performance profile is not found")
+		if discovery.Enabled() && testutils.ProfileNotFound {
+			Skip("Discovery mode enabled, performance profile not found")
+		}
+		workerRTNodes, err = nodes.GetByLabels(nodeLabel)
```
shajmakh, line 49: let's also log the node label we are working with before processing it

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+		Expect(len(nodePools)).To(BeNumerically(">=", 2))
+		PrintNodePoolProfiles(nodePools)
+	})
+
+	Context("Multiple Nodepool", Ordered, func() {
+		isolated := performancev2.CPUSet("1-2")
+		reserved := performancev2.CPUSet("0,3")
+		policy := "best-effort"
+
+		It("should verify support for different performance profiles on hosted cluster via multiple node pools", func() {
```
shajmakh, line 67: if this is part of the test plan let's add a test id when available. not blocking.

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+			By("Creating a deep copy of the performance profile for the second node pool")
+			secondProfile = profile.DeepCopy()
+			secondProfile.Name = "second-profile"
+			np = nodePools[1]
```
shajmakh, line 71: how do you make sure the existed profile isn't configured already on nodePools[1]?

test/e2e/performanceprofile/functests/utils/nodepools/nodepools.go
```diff
@@ -91,14 +93,16 @@ func AttachTuningObject(ctx context.Context, cli client.Client, object client.Ob
...
-		return err
-	}
-	np, err := GetByClusterName(ctx, cli, hostedClusterName)
-	if err != nil {
-		return err
+func DeattachTuningObject(ctx context.Context, cli client.Client, object client.Object, nodePools ...*hypershiftv1beta1.NodePool) error {
+	var np *hypershiftv1beta1.NodePool
+	var err error
+	if len(nodePools) > 0 && nodePools[0] != nil {
+		np = nodePools[0]
```
shajmakh, line 100: same here single nodePool vs multiple

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+	return nodePools, nil
+}
+
+func PrintNodePoolProfiles(nodePools []*hypershiftv1beta1.NodePool) {
+	for _, np := range nodePools {
+		for _, tuningConfig := range np.Spec.TuningConfig {
+			testlog.Infof("NodePool %q is using profile: %q", np.Name, tuningConfig.Name)
+		}
+		if len(np.Spec.TuningConfig) == 0 {
+			testlog.Infof("NodePool %q does not have a tuningConfig profile", np.Name)
```
shajmakh, line 197: when it does I think it'll be handy to know it though

test/e2e/performanceprofile/functests/utils/nodepools/nodepools.go
```diff
@@ -67,14 +67,16 @@ func GetByClusterName(ctx context.Context, c client.Client, hostedClusterName st
...
-		return err
-	}
-	np, err := GetByClusterName(ctx, cli, hostedClusterName)
-	if err != nil {
-		return err
+func AttachTuningObject(ctx context.Context, cli client.Client, object client.Object, nodePools ...*hypershiftv1beta1.NodePool) error {
+	var np *hypershiftv1beta1.NodePool
+	var err error
+	if len(nodePools) > 0 && nodePools[0] != nil {
+		np = nodePools[0]
```
shajmakh, line 74: why do we need multiple nodePools passed if the function will process only the first one? I think we either pass one nodePool object and work with it, or we pass multiple and process everything. The question remains if an object can be attached to multiple nodePools or just a single one, I'm not an expert in this area though

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+			Expect(nodepools.AttachTuningObject(context.TODO(), testclient.ControlPlaneClient, secondProfile, nodePools[1])).To(Succeed())
+
+			By("Waiting for the second node pool configuration to start updating")
+			err = nodepools.WaitForUpdatingConfig(context.TODO(), testclient.ControlPlaneClient, np.Name, np.Namespace)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Waiting for the second node pool configuration to be ready")
+			err = nodepools.WaitForConfigToBeReady(context.TODO(), testclient.ControlPlaneClient, np.Name, np.Namespace)
+			Expect(err).ToNot(HaveOccurred())
+
```
shajmakh, line 85: since in this Context the tests are reflecting on the second profile results, wouldn't it be better to make this configuration part as "BeforeEach/All"?

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,200 @@
...
+				kc, ok := obj.(*kubeletconfigv1beta1.KubeletConfiguration)
+				Expect(ok).To(BeTrue(), "Deserialized object is not of type KubeletConfiguration")
+
+				Expect(kc.ReservedSystemCPUs).To(Equal("0"), "ReservedSystemCPUs is not correctly set")
+				Expect(kc.TopologyManagerPolicy).To(Equal("single-numa-node"), "TopologyManagerPolicy is not correctly set")
+			}
+		})
+
+		It("should verify that Performance Profile update re-creates only target nodepool nodes", func() {
```
shajmakh, line 104: if we ever want to focus on this specific test, would that be possible considering the main configuration of the second profile is done in the former It()?

shajmakh (commented):
Thanks for working on this.

test/e2e/performanceprofile/functests/12_hypershift/12_hypershift_suite_test.go
```diff
@@ -0,0 +1,51 @@
+//go:build !unittests
+// +build !unittests
```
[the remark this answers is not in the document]
SargunNarula: not mandatory, but removed with latest commit. Thanks

test/e2e/performanceprofile/functests/12_hypershift/12_hypershift_suite_test.go
SargunNarula: Added under latest commit, Thanks

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Revised the import order, Thanks

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Revised, Thanks

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Updated with latest commit.

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Agreed, added a log statement

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Revised with test IDs, Thanks

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Ideally, this test suite should be executed after creating a new nodepool, ensuring that no profiles are pre-configured on it. These two test cases are the only ones that deal with the second nodepool. In the AfterAll block, I'm removing the profile, so there shouldn't be any profiles applied left. However, I can add a check before applying the profile to confirm this.

test/e2e/performanceprofile/functests/utils/nodepools/nodepools.go
SargunNarula: The goal here was to avoid affecting references in other test suites. Before this change, all the profile attachment statements for the nodepool didn’t specify any particular node pool, assuming there was only one default node pool in the environment. Now, the function will handle the first nodepool by default, while I explicitly pass the second node pool under this parameter.

SargunNarula:
> Where do we create additional nodePool? I don't see it.

I'm deploying the additional nodepool using kcli, while cluster installation in my test environment. There is an alternative method using hypershift with the command hypershift create nodepool --cluster-name <name>. However, I don't believe it's appropriate to include commands that modify the cluster structure within a test. There is a way to add via applying a manifest - [Link](https://hypershift-docs.netlify.app/how-to/agent/create-heterogeneous-nodepools/#:~:text=%24%7BOCP_RELEASE%7D-,Create%20Heterogeneous%20NodePool,-By%20default%2C%20the)

But i think the arch details are beyond the scope of this test, What do you suggest?

mrniranjan:
> Where do we create additional nodePool? I don't see it.

Creation of nodepool and adding hosts under nodepool is an infrastructure thing, Currently this is not possible in upstream CI,  We need to skip these tests and execute these tests only in downstream

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: The package - hypershiftv1beta1 "github.com/openshift/hypershift/api/hypershift/v1beta1" could not follow this order, gofmt moves it to last place.

SargunNarula: The package - hypershiftv1beta1 "github.com/openshift/hypershift/api/hypershift/v1beta1" could not follow this order, gofmt moves to last place.

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Added a check to ensure no profile is configured already with the nodepool, with latest commit.

test/e2e/performanceprofile/functests/utils/nodepools/nodepools.go
SargunNarula: Resolved with latest commit.

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Info of profile for a nodepool is already getting logged. [Line](https://github.com/openshift/cluster-node-tuning-operator/pull/1184/files#diff-fce62ee3197b8d8d7e8e511b566166d711ca57a406e25bf356adf90f5a3abe74R199)

test/e2e/performanceprofile/functests/utils/nodepools/nodepools.go
SargunNarula: Added a separate method to remove passing of a slice unnecessary. Thanks

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Yes, Indeed. moved the configuration part to before block under latest commit. Thanks

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Moved the profile configuration steps to the Before block to isolate its dependence on the previous It().

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		nodeLabel := testutils.NodeSelectorLabels
+		testlog.Infof("Node label: %v", nodeLabel)
+		workerRTNodes, err = nodes.GetByLabels(nodeLabel)
+		Expect(err).ToNot(HaveOccurred())
+		profile, err = profiles.GetByNodeLabels(nodeLabel)
+		Expect(err).ToNot(HaveOccurred())
+		hostedClusterName, err := hypershift.GetHostedClusterName()
+		Expect(err).ToNot(HaveOccurred())
+		nodePools, err = ListNodePools(context.TODO(), testclient.ControlPlaneClient, hostedClusterName)
```
shajmakh, line 56: let's also add info log for the hostedClusterName

Tal-or:
> > Where do we create additional nodePool? I don't see it.
> 
> Creation of nodepool and adding hosts under nodepool is an infrastructure thing, Currently this is not possible in upstream CI, We need to skip these tests and execute these tests only in downstream

I agree. these kind of tests need to be running on the d/s and skipped u/s

Tal-or:
```
    [test_id:75077] should verify support for different performance profiles on hosted cluster via multiple node pools
    /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/12_hypershift/hypershift.go:87
  > Enter [BeforeAll] Multiple performance profile in hypershift - /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/12_hypershift/hypershift.go:43 @ 11/15/24 08:59:13.066
  STEP: Checking if discovery mode is enabled and performance profile is not found - /go/src/github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/12_hypershift/hypershift.go:44 @ 11/15/24 08:59:13.067
  Nov 15 08:59:13.067: [INFO]: Node label: map[node-role.kubernetes.io/worker:]
  [FAILED] Expected
      <int>: 1
  to be >=
      <int>: 2
```
Probably this should be skipped u/s because we only have a single `nodepool`

test/e2e/performanceprofile/functests/12_hypershift/hypershift.go
SargunNarula: Resolved, Thanks

test/e2e/performanceprofile/functests/12_hypershift/12_hypershift_suite_test.go
```diff
@@ -0,0 +1,48 @@
...
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	testclient "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/client"
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/namespaces"
+	nodeinspector "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/node_inspector"
+)
+
+var _ = BeforeSuite(func() {
+	Expect(testclient.ClientsEnabled).To(BeTrue())
+	err := testclient.Client.Create(context.TODO(), namespaces.TestingNamespace)
```
rbaturov, line 27: Please use the `testclient.dataplane` client here, as we wish to remove the `testclient.clien`t after all suites are adjusted.

test/e2e/performanceprofile/functests/12_hypershift/12_hypershift_suite_test.go
```diff
@@ -0,0 +1,48 @@
...
+	if errors.IsAlreadyExists(err) {
+		testlog.Warning("test namespace already exists, that is unexpected")
+		return
+	}
+	Expect(err).ToNot(HaveOccurred())
+})
+
+var _ = AfterSuite(func() {
+	err := testclient.Client.Delete(context.TODO(), namespaces.TestingNamespace)
```
rbaturov, line 36: Ditto

test/e2e/performanceprofile/functests/12_hypershift/12_hypershift_suite_test.go
SargunNarula: Updated testclient with the latest commit. Thanks!

Tal-or:
Thanks
/lgtm

MarSik (approved):

PR #1190, opened 2024-10-21
NO-JIRA: Default to username and timestamp for custom NTO push

MarSik:
This improves the hack/deploy-custom-nto.sh script to not require the ORG variable all the time. It defaults to username which users commonly use for their quay.io as well.

The second change makes sure the newly built image has an unique tag formed of short git hash and build timestamp. This helps when systems are configured to use cached images. Each tag is unique and so the cluster will always pull the new build.

### DIFF
--- a/hack/deploy-custom-nto.sh
+++ b/hack/deploy-custom-nto.sh
@@ -14,8 +14,8 @@ WORKDIR=$(realpath "${0%/*}/..")
 # overriding the custom NTO image.  If you wish to upgrade your cluster,
 # do not use this script.
 
-ORG=${ORG:-openshift}	# At a minimum, you'll probably want to override this variable.
-TAG=${TAG:-$(git rev-parse --abbrev-ref HEAD)}  # You may need to override this if your git branch contains special characters
+ORG=${ORG:-$USER}	# At a minimum, you'll probably want to override this variable.
+TAG=${TAG:-$(git rev-parse --short HEAD)-$(date +%s)}  # You may need to override this if your git branch contains special characters
 IMAGE=${IMAGE:-quay.io/${ORG}/origin-cluster-node-tuning-operator:$TAG}
 
 nto_prepare_image() {


MarSik:
@ffromani @jmencak WDYT?

hack/deploy-custom-nto.sh
```diff
@@ -14,8 +14,8 @@ WORKDIR=$(realpath "${0%/*}/..")
 # overriding the custom NTO image.  If you wish to upgrade your cluster,
 # do not use this script.
 
-ORG=${ORG:-openshift}	# At a minimum, you'll probably want to override this variable.
-TAG=${TAG:-$(git rev-parse --abbrev-ref HEAD)}  # You may need to override this if your git branch contains special characters
+ORG=${ORG:-$USER}	# At a minimum, you'll probably want to override this variable.
```
ffromani, line 17: this script is not meant to be used in CI, right?

ffromani (commented):
LGTM

if this script is meant to be used by developers (as I recall) and never in CI

hack/deploy-custom-nto.sh
MarSik: Hmm, that is a good question. I do not think we patch CVO in CI this way as it would defeat the purpose, but lets wait for Jirka.

jmencak: The script is not used in the CI.  I understand the motivation for this change.  The only downside I see is less readable tags and extra clutter/mess at image repository.  Happy to adjust my scripts to explicitly set the TAG so that I don't pollute my quay.io account if we really want this. 
/lgtm
/hold

MarSik: Yeah, I treat the tags as throwaways myself and don't care about the temporaries.

If we are agreed @jmencak, can I remove the hold?

jmencak: Sure.  I guess it is user's responsibility then to clean up their image registries then.
/hold cancel

PR #1193, opened 2024-10-22
NO-JIRA: Remove patch attempt from dockerfile script

fontivan:
- There are no more upstream tuned patches since PR #1151, so this is no longer necessary


hack/dockerfile_install_support.sh
```diff
@@ -16,7 +16,6 @@ if [[ "${ID}" == "centos" ]]; then
   BUILD_INSTALL_PKGS="gcc git rpm-build make desktop-file-utils patch dnf-plugins-core"
   dnf install --setopt=tsflags=nodocs -y ${BUILD_INSTALL_PKGS}
   cd /root/assets/tuned/tuned
-  LC_COLLATE=C cat ../patches/*.diff | patch -Np1
```
jmencak, line 19: Would
```
LC_COLLATE=C find ../patches -name \*.diff | patch -Np1
```
work?

fontivan: Negative, still no directory found with find. If you want to keep the  ability though I can wrap it in an if condition instead?

fontivan: @jmencak see latest version of changes for working solution

jmencak: I see, I had a local copy with patches directory still present.  Well, the code would be a lot simpler if you just created the patches (included it in the NTO repo) directory and then just:

```
find ../patches -name \*.diff | LC_COLLATE=C sort | patch -Np1
```

Alternatively, 

```
test -d ../patches && find ../patches -name \*.diff | LC_COLLATE=C sort | patch -Np1
```
No need for a complicated if/then else clause.  But your approach works too.

PR #1194, opened 2024-10-22
CNF-13014: Align dockerfile ocp version

fontivan:
- Since master branch is used for 4.18, we should be using the 4.18 OCP image here like we do in Dockerfile.rhel9 and .ci-operator.yaml


jmencak:
Thank you for the PR!
/approve
/lgtm

PR #1196, opened 2024-10-23
PSAP-1567: Fix linter warnings

jmencak:
https://github.com/openshift/cluster-node-tuning-operator/pull/1195 introduced a k8s bump which deprecated some methods.  Stop using those methods to fix the linter warnings.

PR #1197, opened 2024-10-23
OCPBUGS-43666: Fix kernel arguments ordering on Intel

MarSik:
An upgrade from previous version causes one extra reboot due to differently ordered kernel arguments. This is a side effect of platform specific tuned profile split we merged in https://github.com/openshift/cluster-node-tuning-operator/pull/1083

This fix updates the Intel specific tuned profile to follow the same ordering that was used in the past.

It does so by exploiting a specific tuned behavior of the bootloader plugin. It orders the kernel argument cmdline_suffix keys based on the order of first appearance. Any additional appearance just changes the value, but not the ordering.

The change is only needed for Intel, because we have never supported other platforms before and so upgrade is not an issue.

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -11,8 +11,8 @@ summary=Openshift node optimized for deterministic performance at the cost of in
 # This has two possible results:
 #     openshift-node,cpu-partitioning
 #     openshift-node,cpu-partitioning,openshift-node-performance-rt-<PerformanceProfile name>
-include=openshift-node-performance-${f:lscpu_check:Vendor ID\:\s*GenuineIntel:intel:Vendor ID\:\s*AuthenticAMD:amd:Vendor ID\:\s*ARM:arm}-${f:lscpu_check:Architecture\:\s*x86_64:x86:Architecture\:\s*aarch64:aarch64}-{{.PerformanceProfileName}};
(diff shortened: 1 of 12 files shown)


fontivan:
Raised as bug https://issues.redhat.com/browse/OCPBUGS-43666

fontivan:
/retitle OCPBUGS-43666: Fix kernel arguments ordering onIntel

fontivan:
/retitle OCPBUGS-43666: Fix kernel arguments ordering on Intel

fontivan:
CI failing due to https://github.com/openshift/cluster-node-tuning-operator/pull/1196, will need to be rebased once that is fixed

pkg/performanceprofile/controller/performanceprofile/components/consts.go
```diff
@@ -21,6 +21,8 @@ const (
 	NamespaceNodeTuningOperator = "openshift-cluster-node-tuning-operator"
 	// ProfileNamePerformance defines the performance tuned profile name
 	ProfileNamePerformance = "openshift-node-performance"
+	// ProfileNamePerformance defines the performance tuned profile name for Intel
+	ProfileNamePerformanceIntel = "openshift-node-performance-intel-x86"
```
fontivan, line 25: This is duplicating the existing constant ProfileNameIntelX86 on line 33

PR #1198, opened 2024-10-23
OCPBUGS-43665: Drop amd_iommu=on from amd tuning

fontivan:
- "=on" is not a valid value for amd_iommu
- amd_iommu is enabled by default unless you specify "amd_iommu=off", unlike intel
- See kernel docs for more information (https://docs.kernel.org/admin-guide/kernel-parameters.html)


fontivan:
CI failing due to https://github.com/openshift/cluster-node-tuning-operator/pull/1196, will need to be rebased once that is fixed

MarSik (approved):
/lgtm

PR #1204, opened 2024-11-01
OCPBUGS-44180: Unblock 4.16 CI

jmencak:
This is a minumum change to unblock 4.16 CI based on #1182  to skip OCPBUGS-43280 while the investigation of the real issue continues.


test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -287,6 +288,7 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		})
 		When("kubelet is restart", func() {
 			It("[test_id: 73501] defaultCpuset should not change", func() {
+				testutils.KnownIssueJira("OCPBUGS-43280")
```
yanirq, line 291: change to OCPBUGS-44178  (4.16 bug)

PR #1206, opened 2024-11-05
NO-JIRA: nodeInspector: do not reuse namespace

Tal-or:
Using the global namespace object leads to an error
when trying to recreate the namespace againg.

This is becuase when the first `Create` call return
the object is fullfiled with some metadata from the API server
like `resourceVersion`.

This in turn leads to error when trying to reuse the object for creation
becuase the API doesn't expect the `resourceVersion` to be set.

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
+++ b/test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
@@ -39,8 +39,14 @@ func initialize(ctx context.Context) error {
 	if initialized {
 		return nil
 	}
-	// Create the test namespace
-	err := testclient.DataPlaneClient.Create(ctx, namespace)
+	// NodeInspectorNamespace is the namespace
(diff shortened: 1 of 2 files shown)


test/e2e/performanceprofile/functests/utils/node_inspector/inspector.go
```diff
@@ -31,8 +31,6 @@ const clusterRoleSuffix = "cr"
 const clusterRoleBindingSuffix = "crb"
 
 var initialized bool
-var namespace *corev1.Namespace = namespaces.NodeInspectorNamespace
-var nodeInspectorName = testutils.NodeInspectorName
```
ffromani, line 35: we use this idiom when we want the name to be overridable, instead of always using the hardcoded default. It seems to me in this case we don't need this option, so the change should be fine

ffromani (commented):
/approve

LGTM

PR #1207, opened 2024-11-06
OCPBUGS-44275: Drop sched_migration_cost_ns setting

jmencak:
This is a manual backport of #1203.

When using a PerformanceProfile, a TuneD scheduler was setting sched_migration_cost_ns.  This setting was previously used by throughput-performance and latency-performance profiles.  However, it was dropped from these profiles based on performance evaluation for RHEL-9 as discussed in rhbz#1957829.

Setting sched_migration_cost_ns causes issues when SecureBoot/kernel_lockdown is in effect.  Do not set this tunable when using PerformanceProfiles.

Resolves: OCPBUGS-44275


MarSik (approved):
/lgtm

jmencak:
@liqcui , can we please have the cherry-pick-approved label? Thank you!

PR #1216, opened 2024-11-14
OCPBUGS-44559: Manage TuneD profiles with the same name and different content

jmencak:
It is possible to create Tuned CRs with TuneD profiles of the same name.
This is problematic when the duplicate TuneD profiles have different
contents.  This can cause periodic switches between the duplicate TuneD
profiles and will lead to TuneD restarts.

This change makes the operator control loop stop to avoid switching
between the conflicting TuneD profiles of the same name and warns about
this misconfiguration via:
  * Tuned CR status and prometheus metric (nto_invalid_tuned_exist_info)
  * NTOInvalidTunedExist alert
  * ERROR message issued in the operator logs

Other changes:
  * Switch the ProfileStatusCondition condition type to
    StatusCondition.  This is cosmetic only, all fields are preserved.

Fixes: OCPBUGS-44559.


pkg/operator/profilecalculator.go
```diff
@@ -151,12 +151,13 @@ func (pc *ProfileCalculator) nodeChangeHandler(nodeName string) (bool, error) {
...
-	MCLabels         map[string]string
-	NodePoolName     string
-	Operand          tunedv1.OperandConfig
+	TunedProfileName     string
+	AllProfiles          []tunedv1.TunedProfile
+	Deferred             util.DeferMode
+	MCLabels             map[string]string
+	NodePoolName         string
+	Operand              tunedv1.OperandConfig
+	TuneDProfileConflict []string
```
MarSik, line 160: Is storing this really necessary? The AllProfiles list already has this information and it is enough to just report the duplicate names in the status I'd say?

jmencak:
An example of creating duplicate profiles with conflicting content:
Profile 1
```
kind: Tuned
metadata:
  name: openshift-profile-dup
spec:
  profile:
  - data: |
      [main]
      summary=Custom OpenShift profile
      include=openshift-node
      [sysctl]
      kernel.shmmni=8192
    name: openshift-profile
  recommend:
  - match:
    - label: profile
    priority: 20
    profile: openshift-profile

```

Profile 2
```
kind: Tuned
metadata:
  name: openshift-profile-dup2
spec:
  profile:
  - data: |
      [main]
      summary=Custom OpenShift profile
      include=openshift-node
      [sysctl]
      kernel.shmmni=8191
    name: openshift-profile
  recommend:
  - match:
    - label: profile
    priority: 20
    profile: openshift-profile

```

Output of `oc get co/node-tuning`

```
$ oc get co/node-tuning
NAME          VERSION       AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
node-tuning   4.18.0-ec.3   True        False         True       10m     TuneD profile(s) [openshift-profile] with conflicting content exist

```

pkg/operator/status.go
```diff
@@ -308,6 +308,13 @@ func (c *Controller) computeStatus(tuned *tunedv1.Tuned, conditions []configv1.C
 			availableCondition.Message = fmt.Sprintf("%v/%v Profiles failed to be applied", numDegradedProfiles, len(profileList))
 		}
 
+		if len(c.tunedProfileConflict) > 0 {
```
MarSik, line 311: I know this is cheaper than doing the count over AllProfiles here, but using the direct value instead of a precomputed cache might be easier to maintain.

pkg/operator/profilecalculator.go
jmencak: You'd have to go through the AllProfiles list *again* when calculating the status and check their content.  Same content is not a problem and I don't think we should make co/node-tuning degraded.  The "cache" is an optimization.  Data vs. CPU cycles.  I'm for saving CPU cycles.

pkg/operator/status.go
jmencak: Yes, as I said above, it is cheaper.

jmencak:
@MarSik , can you please take a look if this is the direction we want to go?  I'd like to ask for a pre-merge testing from the QE.

jmencak:
/hold
Until I add some e2e testing.

jmencak:
Infra issues
/retest

jmencak:
@MarSik can you please take a look? Thank you!

liqcui:
@jmencak Pre-merge testing is done, no issue found, thanks!

jmencak:
> @jmencak Pre-merge testing is done, no issue found, thanks!

@MarSik , can we have LGTM if this looks good from your side? Thank you.

PR #1219, opened 2024-11-17
CNF-15505: Remove container runtime selection

rbaturov:
A new patch has been introduced for CRI-O, allowing the container runtime for the high-performance runtime class to be inherited from the system instead of being detected and explicitly set. https://github.com/cri-o/cri-o/issues/8753

As a result, the controller no longer needs to identify the container runtime from the containerRuntimeConfig for setting it in the 99-runtimes.conf snippet.

This PR encompasses all (need to verify) required changes.


assets/performanceprofile/configs/99-runtimes.conf
```diff
@@ -4,11 +4,6 @@ infra_ctr_cpuset = "{{.ReservedCpus}}"
 {{if .SharedCpus}}shared_cpuset = "{{.SharedCpus}}"{{end}}
 {{end}}
 
-# We should copy paste the default runtime because this snippet will override the whole runtimes section
```
ffromani, line 7: Why is this logic unnecessary?

Tal-or: @rbaturov Please add test to verify that.
You can simply list crio config as part of the test and see that everything is in place

ffromani: this should also be documented in the commit message of the related change:
"this logic is unnecessary _because_..."

rbaturov: @ffromani Defentilty will put a more detailed message for this commit (This was just a draft)

test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
```diff
@@ -31,7 +31,7 @@ func GetContainerRuntimeTypeFor(ctx context.Context, c client.Client, pod *corev
 		"/rootfs",
 		"/bin/bash",
 		"-c",
-		fmt.Sprintf("/bin/awk -F '\"'  '/runtime_path.*/ { print $2 }' %s", CRIORuntimeConfigFile),
+		fmt.Sprintf("ps aux | grep '%s' | grep -oP '(?<=-r\\s)[^\\s]+'", pod.Name),
```
Tal-or, line 34: better to provide full path of `ps`

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -1162,56 +1166,67 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
-				if ctrcfg == nil {
-					testlog.Infof("ContainerRuntimeConfig not exist")
+		DescribeTable("verifies container runtime behavior",
+			func(withCTRCfg bool) {
+				var expectedRuntime string
+				if withCTRCfg {
 					ctrcfg = newContainerRuntimeConfig(ContainerRuntimeConfigName, profile, mcp)
 					By(fmt.Sprintf("creating ContainerRuntimeConfig %q", ctrcfg.Name))
-					Expect(testclient.Client.Create(context.TODO(), ctrcfg)).ToNot(HaveOccurred(), "failed to create ctrcfg %#v", ctrcfg)
+					Expect(testclient.ControlPlaneClient.Create(context.TODO(), ctrcfg)).ToNot(HaveOccurred(), "failed to create ctrcfg %#v", ctrcfg)
```
Tal-or, line 1183: It won't work on HyperShift.
You'll have to create the object on the `clusters` Namespace and then attaches it to the nodePool

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -1317,7 +1332,7 @@ func newContainerRuntimeConfig(name string, profile *performancev2.PerformancePr
 				MatchLabels: profilecomponent.GetMachineConfigPoolSelector(profile, profileMCP),
 			},
 			ContainerRuntimeConfig: &machineconfigv1.ContainerRuntimeConfiguration{
-				DefaultRuntime: machineconfigv1.ContainerRuntimeDefaultRuntimeCrun,
+				DefaultRuntime: machineconfigv1.ContainerRuntimeDefaultRuntimeRunc,
```
Tal-or, line 1335: why default changed here to runc?

rbaturov: Since the workflow of this e2e is as follows:
First we test without any ctrcfg in the system
Second, we apply a ctrcfg with a different runtime than what's already in the system (on 4.18 it is crun) as we do want to test a change and not override the runtime with the same runtime.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
rbaturov: Yes you are right. anyway, this test has the openshift label and currently not running on hypershift lanes.
I figured that out later. 
I can modify this to use the `testclient.client` as before if you would like

test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
rbaturov: Will update this in the next iterarion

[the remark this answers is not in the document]
rbaturov: Will do it for the next review iteration.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
Tal-or: Ack. Got confused by the name

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
rbaturov: @SargunNarula I saw your PR. The changes for supporting hypershift belong to a separate (your PR).
Therefore, I have removed my client adjustments here, please ensure you are doing them in your PR.

test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
```diff
@@ -31,7 +31,7 @@ func GetContainerRuntimeTypeFor(ctx context.Context, c client.Client, pod *corev
 		"/rootfs",
 		"/bin/bash",
 		"-c",
-		fmt.Sprintf("/bin/awk -F '\"'  '/runtime_path.*/ { print $2 }' %s", CRIORuntimeConfigFile),
+		fmt.Sprintf("/bin/ps aux | grep '%s' | grep -oP '(?<=-r\\s)[^\\s]+'", pod.Name),
 	}
```
mrniranjan, line 35: Test looks good to me , May be here we can actually verify if we are getting /usr/bin/crun or /usr/bin/runc and not any other value  ?

rbaturov: I think it is more correct to leave this generalized, as we know that the -r conmon argument is container runtime path.
I believe that we should leave the runtime type assertion outside the scope of this function.

ffromani (commented):
my comments are addressed, LGTM pending review of the e2e tests

rbaturov:
> my comments are addressed, LGTM pending review of the e2e tests

@mrniranjan Has reviewed.

PR #1224, opened 2024-11-26
OCPBUGS-45048: Move reboot tests to appropriate make target for hypershift

mrniranjan:
Move all test which reboot nodes to pao-functests-update-only-hypershift

- Use pao-functests-hypershift for tests which do not update profile 
- Use pao-functests-update-only-hypershift for tests which update Performance profile and reboot the nodes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,7 @@ pao-functests-update-only: $(BINDATA)
 pao-functests-update-only-hypershift: $(BINDATA)
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config ./test/e2e/performanceprofile/functests/2_performance_update" -p "-v -r --label-filter="!openshift" --fail-fast --flake-attempts=2 --timeout=5h --junit-report=report.xml" -m "Running Functional Tests"
+	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config ./test/e2e/performanceprofile/functests/2_performance_update ./test/e2e/performanceprofile/functests/7_performance_kubelet_node ./test/e2e/performanceprofile/functests/8_performance_workloadhints ./test/e2e/performanceprofile/functests/12_hypershift" -p "-v -r --label-filter=!(openshift||slow) --fail-fast --flake-attempts=2 --timeout=5h --junit-report=report.xml" -m "Running Functional Tests"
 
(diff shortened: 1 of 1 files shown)


rbaturov:
@mrniranjan 
Thanks for this fix
/lgtm

Makefile
```diff
@@ -223,7 +223,7 @@ pao-functests-update-only: $(BINDATA)
 pao-functests-update-only-hypershift: $(BINDATA)
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
-	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config ./test/e2e/performanceprofile/functests/2_performance_update" -p "-v -r --label-filter="!openshift" --fail-fast --flake-attempts=2 --timeout=5h --junit-report=report.xml" -m "Running Functional Tests"
+	hack/run-test.sh -t "test/e2e/performanceprofile/functests/0_config ./test/e2e/performanceprofile/functests/2_performance_update ./test/e2e/performanceprofile/functests/7_performance_kubelet_node ./test/e2e/performanceprofile/functests/8_performance_workloadhints ./test/e2e/performanceprofile/functests/12_hypershift" -p "-v -r --label-filter=!(openshift||slow) --fail-fast --flake-attempts=2 --timeout=5h --junit-report=report.xml" -m "Running Functional Tests"
```
Tal-or, line 226: why `!(slow)` ? I thought the whole point of this lane is to run tests which requires nodes rollout (equals to reboot on OCP)

mrniranjan: There are some tests in Workload hints that will do multiple reboots for each test case and those test cases will be very slow and also those test cases need Baremetal to run

mrniranjan: Though we can run them on VM most of them which will be slow and few require changes in BIOS to test them

Tal-or: So `slow` is even longer than just a simple test that require a reboot?

mrniranjan: Yes it is slower, In one or two test cases, we do multiple reboots,

Tal-or: Cool.
Then let's update the docs (possible in different PR) because the existing documentation is misleading:
https://github.com/openshift/cluster-node-tuning-operator/blob/3cbae8f848ccc09ff33bd8e2452a51559f8b7294/test/e2e/performanceprofile/functests/utils/label/label.go#L14-L13

Tal-or:
/hold 
non of the tests run

PR #1231, opened 2024-12-02
OCPBUGS-45346: performanceprofile cpuset input validation

rbaturov:
This fix ensures that the input for any CPU set field is valid. If an invalid string that does not represent a valid CPU set is provided, the admission webhook will return an appropriate error, such as:
 `spec.cpu: Internal error: strconv.Atoi: parsing "garbage": invalid syntax`

This replaces the previous behavior where the server would return an error like: 
`Error from server: error when creating "pp.yaml": admission webhook "vwb.performance.openshift.io" denied the request: panic: runtime error: invalid memory address or nil pointer dereference [recovered].`

Moreover, fuzz tests have been added to ensure we exercise all errors when executing the validation function on valid and invalid inputs, ensuring no panic occurs during execution.

### DIFF
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
@@ -265,6 +265,19 @@ var _ = Describe("PerformanceProfile", func() {
 			Expect(errors).NotTo(BeEmpty(), "should have validation error when isolated and shared CPUs have overlap")
 			Expect(errors[0].Error()).To(Or(ContainSubstring("isolated and shared cpus overlap"), ContainSubstring("shared and isolated cpus overlap")))
 		})
+		DescribeTable("should reject invalid input that does not represent CPU sets",
+			func(fieldSetter func(*PerformanceProfile, CPUSet), cpusField string) {
+				garbageInput := CPUSet("garbage")
+				fieldSetter(profile, garbageInput)
+				errors := profile.validateCPUs()
+				Expect(errors).NotTo(BeEmpty(), "should have error when "+cpusField+" is filled with garbage input")
+				Expect(errors[0].Error()).To(Or(ContainSubstring("Internal error: strconv.Atoi: parsing")))
+			},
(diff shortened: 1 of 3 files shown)


Tal-or:
Let's stretch the testing and add some [fuzz](https://go.dev/doc/security/fuzz/) testing

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -153,6 +154,32 @@ func NewPerformanceProfile(name string) *PerformanceProfile {
...
+			"offline":  func(p *PerformanceProfile, input CPUSet) { p.Spec.CPU.Offlined = &input },
+		}
+
+		for fieldName, setField := range cpuFields {
+			t.Run(fieldName, func(t *testing.T) {
+				cpuSet := CPUSet(input)
+				profile := NewPerformanceProfile("test")
+
+				setField(profile, cpuSet)
+				_ = profile.validateCPUs()
```
Tal-or, line 177: Please add a comment explaining (above this line) that we're not checking for error, only for panic

Tal-or (commented):
Usually we're not adding a unit tests for non-public functions like `validateCPUS` so I would consider make it a public function (or at least not a receiver function).

But, considering the fact that we're interested in backporting this PR, I dunno if making this change more invasive will be a good idea.

ffromani:
> Usually we're not adding a unit tests for non-public functions like `validateCPUS` so I would consider make it a public function (or at least not a receiver function).
> 
> But, considering the fact that we're interested in backporting this PR, I dunno if making this change more invasive will be a good idea.

yes, the _general trend_ (not the hard rule) is to not test private functions, but if it's practical, like to better enable backport, we just do it. I agree with your assessment: in this case the practicality wins, keeping the change minimal will better serve backports.

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
rbaturov: Done

Tal-or:
/lgtm
Thank you, great work!

MarSik (approved):

PR #1234, opened 2024-12-03
cmd: PPC: support tolerating heterogeneous core IDs

shajmakh:
**Problem**: In a system with a specific number of cores per socket
the host gives each core a number. So far PPC would generate
performance profile only after verifying that the nodes pointed to by
the specified node pool are all of the same hardware and topology. That
is because the performance profile will apply the same configuration on all
these nodes and it is most important that they have same structure, CPUs
distribution across NUMAs and CPUs availability. This by default
includes having same core IDs for each CPU in the same NUMA cells for
all the compute nodes. for instance:

Worker-0 has NUMA-0 on which the CPUs siblings [2,66] is coupled in
one core numbered 18`.
All other workers should have this info true for them, otherwise the
tool would fail.

**Suggested solution:**
With further investigation, it was learned that core numbering can have
different schemes even with a system from the same vendor. The numbering
the pattern depends on the settings of the hardware, the software and the
firmware (BIOS).While core IDs may vary nodes can still be considered
having same NUMA topology taking into account that core scope is on the
single NUMA. However, core IDs can be important in optimizing the
system's performance and managing isolation of tasks, meaning if the
performance of worker-0 is not comparable to that of worker-1 before
having performance-profile applied on both, having an improved
performance on both after applying PP is not guaranteed.

In this PR, we introduce an experimental flag to be added to the PPC
tool called `tolerate-core-ids-diff` and is `false` by default; if set
to true then the check to ensure same topologies between nodes will
ignore having different core IDs for the same CPUs in the same NUMA
cells.

Disclaimer: We support this option to unblock business matters and is recommended
to use it with caution only if the default value is not providing the
desired results; and only if the user is willing to take on the
responsibility of troubleshooting the settings on their environment which not all
are reflected in the must-gather collection.

Signed-off-by: Shereen Haj <shajmakh@redhat.com>
@shajmakh

### DIFF
--- a/pkg/performanceprofile/profilecreator/profilecreator_test.go
+++ b/pkg/performanceprofile/profilecreator/profilecreator_test.go
@@ -1483,6 +1483,63 @@ var _ = Describe("PerformanceProfileCreator: Test Helper Function ensureSameTopo
 			err := ensureSameTopology(&originTopology, &mutatedTopology)
 			Expect(err).ToNot(HaveOccurred())
 		})
+		It("should fail when core id on same numa is associated with different siblings list", func() {
+			nodes3 := []*topology.Node{
+				{
+					ID: 0,
+					Cores: []*cpu.ProcessorCore{
+						{ID: 0, Index: 0, NumThreads: 2, LogicalProcessors: []int{0, 4}}, // Cores must be sorted by logical processors before passing them to ensureSameTopology
+						{ID: 2, Index: 3, NumThreads: 2, LogicalProcessors: []int{0, 4}},
+						{ID: 2, Index: 2, NumThreads: 2, LogicalProcessors: []int{2, 6}},
+						{ID: 0, Index: 1, NumThreads: 2, LogicalProcessors: []int{2, 6}},
+					},
+				},
+				{
+					ID: 1,
+					Cores: []*cpu.ProcessorCore{
+						{ID: 0, Index: 2, NumThreads: 2, LogicalProcessors: []int{1, 3}},
+						{ID: 0, Index: 2, NumThreads: 2, LogicalProcessors: []int{1, 3}},
+						{ID: 1, Index: 3, NumThreads: 2, LogicalProcessors: []int{5, 7}},
+						{ID: 1, Index: 3, NumThreads: 2, LogicalProcessors: []int{5, 7}},
+					},
+				},
+			}
+
+			t3 := topology.Info{
+				Architecture: topology.ARCHITECTURE_NUMA,
+				Nodes:        nodes3,
+			}
+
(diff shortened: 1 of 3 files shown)


pkg/performanceprofile/profilecreator/profilecreator_test.go
```diff
@@ -1483,6 +1483,63 @@ var _ = Describe("PerformanceProfileCreator: Test Helper Function ensureSameTopo
 			err := ensureSameTopology(&originTopology, &mutatedTopology)
 			Expect(err).ToNot(HaveOccurred())
 		})
+		It("should fail when core id on same numa is associated with different siblings list", func() {
```
ffromani, line 1486: can we please add a comment documenting on which kind of hardware we found this happening?
e.g. server Dell model So-and-So with N Intel CPUs model So-and-So.

ffromani (commented):
Thanks for the extensive investigation!
The first two commit seems to have good value, if you split them out I'm positive we can merge them quickly. The third commit would deserve more consideration though.

jmencak:
I'm probably missing something, but I wonder if this is a complete solution.  Wouldn't it cause NTO [block updating MachineConfigs](https://github.com/openshift/cluster-node-tuning-operator/blob/7d54582ae062232c79a45689678b37caf25025ec/pkg/operator/controller.go#L844) if it saw different kernel parameters for machines within the same MCP?

pkg/performanceprofile/profilecreator/profilecreator_test.go
shajmakh: Thanks for the comment. This was never observed and likely never happens on a healthy system but still we want to have this case covered because we depend on must-gather to get us the data for the PPC to analyze as stated in the commit message. This check can be considered as a backup for tolerating the core ID difference, so we reference to core ID in the NUMA level and verify alignment to same CPU siblings each time especially when HT is enabled.

shajmakh:
> I'm probably missing something, but I wonder if this is a complete solution. Wouldn't it cause NTO [block updating MachineConfigs](https://github.com/openshift/cluster-node-tuning-operator/blob/7d54582ae062232c79a45689678b37caf25025ec/pkg/operator/controller.go#L844) if it saw different kernel parameters for machines within the same MCP?

That is correct and I learned about it just today. Following an offline conversation we agreed we need to relax the core ID check into a warning if found true and leave the decision for the user to make if to use the generated profile or not.

shajmakh:
close in favor of:
https://github.com/openshift/cluster-node-tuning-operator/pull/1235
https://github.com/openshift/cluster-node-tuning-operator/pull/1236

PR #1236, opened 2024-12-03
OCPBUGS-44372: cmd: PPC: support tolerating heterogeneous core IDs

shajmakh:
Problem: In a system with a specific number of cores per socket the host gives each core a number. So far PPC would generate performance profile only after verifying that the nodes pointed to by the specified node pool are all of the same hardware and topology. That is because the performance profile will apply the same configuration on all these nodes and it is most important that they have same structure, CPUs distribution across NUMAs and CPUs availability. This by default includes having same core IDs for each CPU in the same NUMA cells for all the compute nodes. for instance:

Worker-0 has NUMA-0 on which the CPUs siblings [2,66] is coupled in one core numbered 18`.
All other workers should have this info true for them, otherwise the tool would fail.

It was observed (for example on Intel Xeon GoldGold 6438N with 0-127 online CPUs distributed across 2 sockets, 32 cores per socket and 2 threads per core) that core numbering can have different schemes even with a system from the same vendor, which causes the tool to fail to generate a profile.

Suggested solution:
With further investigation, The numbering the pattern depends on the settings of the hardware, the software and the firmware (BIOS).While core IDs may vary nodes can still be considered having same NUMA topology taking into account that core scope is on the single NUMA. However, core IDs can be important in optimizing the system's performance and managing isolation of tasks, meaning if the performance of worker-0 is not comparable to that of worker-1 before having performance-profile applied on both, having an improved performance on both after applying PP is not guaranteed.

In this PR, we loosen the hard requirment of having same core numbering on same NUMA cells on different systems into a warning will be logged out as well as a comment on the generated CSV file. So as long as the NUMA cells have same logical processors' count and IDs and same threads' number, core ID equality is treated as best effort. That is because when scheduling workloads, we care about the logical processors ids and their location on the NUMAs.

Disclaimer: We support this option to unblock business matters and is recommended to use the generated PP with cautiono.

### DIFF
--- a/cmd/performance-profile-creator/cmd/root.go
+++ b/cmd/performance-profile-creator/cmd/root.go
@@ -82,13 +82,6 @@ var (
 	scheme                     = runtime.NewScheme()
 	validTMPolicyValues        = []string{kubeletconfig.SingleNumaNodeTopologyManagerPolicy, kubeletconfig.BestEffortTopologyManagerPolicy, kubeletconfig.RestrictedTopologyManagerPolicy}
 	validPowerConsumptionModes = []string{defaultLatency, lowLatency, ultraLowLatency}
-	hardwareTuningMessage      = `#HardwareTuning is an advanced feature, and only intended to be used if 
-#user is aware of the vendor recommendation on maximum cpu frequency.
-#The structure must follow
-#
-# hardwareTuning:
-#   isolatedCpuFreq: <Maximum frequency for applications running on isolated cpus>
-#   reservedCpuFreq: <Maximum frequency for platform software running on reserved cpus>`
 )
 
 // ProfileData collects and stores all the data needed for profile creation
@@ -135,6 +128,8 @@ func NewRootCommand() *cobra.Command {
 		"must-gather-dir-path",
 	}
 
+	tolerations := profilecreator.TolerationSet{}
+
 	root := &cobra.Command{
 		Use:   "performance-profile-creator",
 		Short: "A tool that automates creation of Performance Profiles",
@@ -168,7 +163,9 @@ func NewRootCommand() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			if err := profilecreator.EnsureNodesHaveTheSameHardware(nodesHandlers); err != nil {
+
+			err = profilecreator.EnsureNodesHaveTheSameHardware(nodesHandlers, tolerations)
+			if err != nil {
 				return fmt.Errorf("targeted nodes differ: %w", err)
 			}
 			// We make sure that the matched Nodes are the same
@@ -178,11 +175,12 @@ func NewRootCommand() *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("failed to make profile data from node handler: %w", err)
 			}
+			tolerations[profilecreator.EnableHardwareTuning] = true
 			profile, err := makePerformanceProfileFrom(*profileData)
 			if err != nil {
 				return err
 			}
-			return writeProfile(profile, profileData.enableHardwareTuning)
+			return writeProfile(profile, tolerations)
 		},
 	}
 	pcArgs.AddFlags(root.PersistentFlags())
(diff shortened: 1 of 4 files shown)


pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -677,76 +677,92 @@ func contains(s []string, str string) bool {
 	return false
 }
 
-// EnsureNodesHaveTheSameHardware returns an error if all the input nodes do not have the same hardware configuration
-func EnsureNodesHaveTheSameHardware(nodeHandlers []*GHWHandler) error {
+// EnsureNodesHaveTheSameHardware returns an error if all the input nodes do not have the same hardware configuration;
+// if core IDs numbering was found different on same NUMA cells of different compute nodes, the function will return true
+// so this be printed out as a warning to the user later
+func EnsureNodesHaveTheSameHardware(nodeHandlers []*GHWHandler) (bool, error) {
```
ffromani, line 683: I'm more and more convinced that every time we use a bool, there's a better approach hidden somewhere. Can we think of a new type, maybe a struct, maybe just a set of strings, which embeds the wrong-but-tolerable (warnings?) we found about a topology, and use it everywhere instead of bools?

ffromani (commented):
not sure what's the CSV hinted in the PR description/commit message.
The general concept is fine to me. The impl can use some possible polishing if we have time for it.

pkg/performanceprofile/profilecreator/profilecreator.go
shajmakh: thanks for raising. I agree, that's a better idea I'd be happy to update

cmd/performance-profile-creator/cmd/root.go
```diff
@@ -523,15 +521,21 @@ func makePerformanceProfileFrom(profileData ProfileData) (runtime.Object, error)
...
 
-	if enableHardwareTuning {
-		if _, err := writer.Write([]byte(hardwareTuningMessage)); err != nil {
+	if val, _ := tolerations[profilecreator.EnableHardwareTuning]; val == true {
+		if _, err := writer.Write([]byte(profilecreator.HardwareTuningMessage)); err != nil {
+			return err
+		}
+	}
+
+	if val, _ := tolerations[profilecreator.DifferentCoreIDs]; val == true {
```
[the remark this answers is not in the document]
shajmakh: I intentionally made that a map allowing other values to be set and not only found/not found. it's true we use so far only true/false but just considered to allow option for other values if needed later on

shajmakh:
> not sure what's the CSV hinted in the PR description/commit message.

it'll be printed in the generated profile, updated that in the latest push

pkg/performanceprofile/profilecreator/helper.go
```diff
@@ -4,6 +4,30 @@ import (
...
+# While core IDs numbering may differ between two systems, it still can be considered 
+# that NUMA and HW topologies are similar; However this depends on the combination 
+# setting of the hardware, software and firmware as that may affect the mapping pattern. 
+# While the performance profile controller depends on the logical processors per NUMA, 
+# having different IDs may affect your system's performance optimization where the cores
+# location matters, thus use the generated profile with caution.`
+)
+
+// TolerationSet records the data to be tolerated or warned about based on the tool handling
+type TolerationSet map[string]interface{}
```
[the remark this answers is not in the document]
shajmakh: thanks! available since 1.18

pkg/performanceprofile/profilecreator/profilecreator.go
```diff
@@ -677,8 +677,10 @@ func contains(s []string, str string) bool {
 	return false
 }
 
-// EnsureNodesHaveTheSameHardware returns an error if all the input nodes do not have the same hardware configuration
-func EnsureNodesHaveTheSameHardware(nodeHandlers []*GHWHandler) error {
+// EnsureNodesHaveTheSameHardware returns an error if all the input nodes do not have the same hardware configuration;
+// if core IDs numbering was found different on same NUMA cells of different compute nodes, the function will return true
+// so this be printed out as a warning to the user later
```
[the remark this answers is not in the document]
shajmakh: right, fixed, thanks!

shajmakh:
/retest
known issue in hypersift

pkg/performanceprofile/profilecreator/helper.go
```diff
@@ -4,6 +4,32 @@ import (
...
+# that NUMA and HW topologies are similar; However this depends on the combination 
+# setting of the hardware, software and firmware as that may affect the mapping pattern. 
+# While the performance profile controller depends on the logical processors per NUMA, 
+# having different IDs may affect your system's performance optimization where the cores
+# location matters, thus use the generated profile with caution.
+`
+)
+
+// TolerationSet records the data to be tolerated or warned about based on the tool handling
+type TolerationSet map[string]any
```
ffromani, line 31: nit: if we always expect `bool`s as value, we should use `bool` instead of `any`.
To put it differently: do we ever expect any value which is not `bool`?

ffromani (commented):
/lgtm
/hold

to let others review (@MarSik )

pkg/performanceprofile/profilecreator/helper.go
shajmakh: we don't, but I kept it as any just in case we need later different kind of values. I don't mind changing it though. will do that now

ffromani:
/retest-required

the changed code is not physically part of any of the failing flows

ffromani:
@shajmakh please rebase on top of current master branch. The failing lane is being interrupted because timeout

ffromani:

we are changing an offline tool and we can't break the product

shajmakh:
/cherry-pick release-4.18 release-4.17 release-4.16 release-4.15 release-4.14 release-4.13 release-4.12

PR #1241, opened 2024-12-05
Update golangci lint

rbaturov:


Makefile
```diff
@@ -55,7 +55,7 @@ PAO_CRD_APIS :=$(addprefix ./$(API_TYPES_DIR)/performanceprofile/,v2 v1 v1alpha1
 PAO_E2E_SUITES := $(shell hack/list-test-bin.sh)
 
 # golangci-lint variables
-GOLANGCI_LINT_VERSION=1.54.2
+GOLANGCI_LINT_VERSION=1.60.1
```
jmencak, line 58: Any reason for not using the latest 1.62.2 release?  Saying that, this won't fix the current issues.

PR #1244, opened 2024-12-09
perf-prof: latency: fix calculating isolated CPUs size

shajmakh:
Calculating the size of the isolated CPUs set from the performance profile was reducing 1 which is incorrect. Remove the reduction line and fix the corresponding setup of the test suite.

### DIFF
--- a/test/e2e/performanceprofile/functests/4_latency/latency.go
+++ b/test/e2e/performanceprofile/functests/4_latency/latency.go
@@ -482,8 +482,7 @@ func isOddCpuNumber(cpusNum int, profile *performancev2.PerformanceProfile) bool
 	if cpusNum == defaultTestCpus {
 		isolatedCpus, err := cpuset.Parse(string(*profile.Spec.CPU.Isolated))
 		Expect(err).ToNot(HaveOccurred(), "failed to parse cpus %q", string(*profile.Spec.CPU.Isolated))
-		isolatedCpusNum := isolatedCpus.Size() - 1
-		return isolatedCpusNum%2 != 0
+		return isolatedCpus.Size()%2 != 0
 	}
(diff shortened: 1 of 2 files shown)


shajmakh:
/hold 
for testing

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -482,8 +482,7 @@ func isOddCpuNumber(cpusNum int, profile *performancev2.PerformanceProfile) bool
 	if cpusNum == defaultTestCpus {
 		isolatedCpus, err := cpuset.Parse(string(*profile.Spec.CPU.Isolated))
 		Expect(err).ToNot(HaveOccurred(), "failed to parse cpus %q", string(*profile.Spec.CPU.Isolated))
-		isolatedCpusNum := isolatedCpus.Size() - 1
-		return isolatedCpusNum%2 != 0
+		return isolatedCpus.Size()%2 != 0
```
Tal-or, line 485: nit:
```
cpusNum := isolatedCpus.Size()
}
return  cpusNum %2 != 0
```

Tal-or (commented):
We should evaluate this change carefully to make sure we're not causing regression. 
I believe there was a reason for this change, unless it's just a pure bug we missed during review process.

shajmakh:
this change is wrong and is inherited from this calculation:
https://github.com/shajmakh/cluster-node-tuning-operator/blob/0f9d7d673c545f49aba096a039a2d58fb13fca63/test/e2e/performanceprofile/functests/4_latency/latency.go#L363
we need first to make this calculation smarter. Filed a card to track this https://issues.redhat.com/browse/OCPBUGS-45979

PR #1248, opened 2024-12-10
[release-4.17] OCPBUGS-45627: Normalize cpu sets when rendering to Tuned profiles

MarSik:
Previously the isolated cpu list was passed verbatim as the user entered it. This could have resulted in very long kernel command line.

Example: 10,9,8,7,6,5,4,3,2,1 (imagine 500+ cpus)

This change consolidates the parsing logic in the Tuned templating and generates simply the following in such case.

Example: 1-10


MarSik:
/retitle [release-4.17] OCPBUGS-45627: Normalize cpu sets when rendering to Tuned profiles

ffromani (commented):
/lgtm

PR #1249, opened 2024-12-10
OCPBUGS-46426: e2e: add irdma to module_blacklist kernel args

rbaturov:
This implements a workaround to prevent CI failures on specific hardware using an Intel E810 network card.
When UserLevelNetworking is set to True, tuned attempts to set the combined channel count equal to the reserved CPUs but fails with the following error:
tuned.utils.commands: Executing 'ethtool -L ens2f0 combined 1' error: netlink error: Device or resource busy
The error occurs because the ice driver: ens2f0: Cannot change channels when RDMA is active.
This issue causes the tuned profile to degrade.
As a temporary solution, by adding 'module_blacklist=irdma' to the kernel Args we will block RDMA, to avoid these errors.

### DIFF
--- a/test/e2e/performanceprofile/functests/0_config/config.go
+++ b/test/e2e/performanceprofile/functests/0_config/config.go
@@ -154,6 +154,18 @@ func testProfile() (*performancev2.PerformanceProfile, error) {
 
 	hugePagesSize := performancev2.HugePageSize("1G")
 
+	// This implements a workaround to prevent CI failures on specific hardware using an Intel E810 network card.
+	// When UserLevelNetworking is set to True, tuned attempts to set the combined channel count equal to the reserved CPUs but fails with the following error:
+	// tuned.utils.commands: Executing 'ethtool -L ens2f0 combined 1' error: netlink error: Device or resource busy
+	// The error occurs because the ice driver: ens2f0: Cannot change channels when RDMA is active.
+	// This issue causes the tuned profile to degrade.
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -154,6 +154,18 @@ func testProfile() (*performancev2.PerformanceProfile, error) {
 
 	hugePagesSize := performancev2.HugePageSize("1G")
 
+	// This implements a workaround to prevent CI failures on specific hardware using an Intel E810 network card.
+	// When UserLevelNetworking is set to True, tuned attempts to set the combined channel count equal to the reserved CPUs but fails with the following error:
+	// tuned.utils.commands: Executing 'ethtool -L ens2f0 combined 1' error: netlink error: Device or resource busy
```
yanirq, line 159: Add:  "Executing **for example**"  if you giving a specific example as you do here

test/e2e/performanceprofile/functests/0_config/config.go
```diff
@@ -154,6 +154,18 @@ func testProfile() (*performancev2.PerformanceProfile, error) {
...
+	// The error occurs because the ice driver: ens2f0: Cannot change channels when RDMA is active.
```
jmencak, line 160: Nit: `s/because the ice driver/because of the Intel Ethernet ice driver/` ?

jmencak:
OK, this is just e2e tests.
/approve
/lgtm
/hold
to give other a chance to comment.  Feel free to unhold if you don't get any feedback.

PR #1255, opened 2024-12-17
NO-JIRA: e2e: common way to fetch poolName

Tal-or:
On OpenShift we're looking for the MCP associated with the
nodes affected by the profile.

On HyperShift we're looking for the nodePool associated the
nodes affected by the profile.

This commit introduce a function to get the pool name
which is essentially MCP/NodePool name depending on
the platform detected at runtime.


ffromani (commented):
/approve
/lgtm

thanks!

ffromani:
it seems we will need a targeted manual backport

PR #1261, opened 2024-12-21
OCPBUGS-48317: PPC_Test_fix: use ROLE_WORKER_CNF environment variable to determine mcp name

mrniranjan:
In the current ppc tests, we are using mcp-name as worker. Instead of using hardcoded value of worker, use the mcp name from ROLE_WORKER_CNF environment variable

### DIFF
--- a/test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
+++ b/test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
@@ -63,6 +63,7 @@ func PPCTestCreateUtil() *PPCTestIntegration {
 
 var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform awareness", Label(string(label.PerformanceProfileCreator)), func() {
 	mustgatherDir := testutils.MustGatherDir
+	mcpName := testutils.RoleWorkerCNF
 	ntoImage := testutils.NTOImage
 	Context("PPC Sanity Tests", Label(string(label.Tier0)), func() {
 		ppcIntgTest := PPCTestCreateUtil()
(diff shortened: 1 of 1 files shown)


swatisehgal (approved):
/lgtm

swatisehgal:
/assign @MarSik @ffromani 
for approval

test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
```diff
@@ -79,7 +80,7 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
 			cmdArgs := []string{
 				fmt.Sprintf("%s:%s:z", mustgatherDir, mustgatherDir),
 				ntoImage,
-				"--mcp-name=worker",
+				fmt.Sprintf("--mcp-name=%s", mcpName),
```
shajmakh, line 83: by default (in case the env variable isn't set) the target mcp would be worker-cnf, see here:
https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/utils/consts.go#L38
does that make any difference in the expected results? is the gathered data have both mcps?

mrniranjan: so when running under SNO environment we want to set ROLE_WORKER_CNF=master and run the tests

shajmakh: the change then solves SNO perfectly; what about MNO , do we always have worker-cnf mcp in the env from which we gathered the must-gather?

mrniranjan: so in our CI for MNO , our CI creates a must-gather before running PPC tests, and by default we have worker-cnf label set.

ffromani:

test-only change, can't break the product

PR #1262, opened 2024-12-22
Add kernel page size field

rbaturov:


rbaturov:
cc @yanirq

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -41,12 +41,14 @@ import (
 )
 
 const (
-	hugepagesSize2M   = "2M"
-	hugepagesSize32M  = "32M"
-	hugepagesSize512M = "512M"
-	hugepagesSize1G   = "1G"
-	amd64             = "amd64"
-	aarch64           = "arm64"
+	kernelPagesSize4K  = "4K"
```
MarSik, line 44: Isn't the proper SI suffix "k" (lowercase)? I am ok with making it size agnostic, but using just the capitalized variant is probably wrong.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -146,8 +148,13 @@ func New(profile *performancev2.PerformanceProfile, opts *components.MachineConf
 		profile.Spec.RealTimeKernel.Enabled != nil &&
 		*profile.Spec.RealTimeKernel.Enabled
 
+	// Real time kernel with 64k-pages for aarch64 not yet supported and rejected in the validation webhook.
 	if enableRTKernel {
 		mc.Spec.KernelType = MCKernelRT
+	} else if profile.Spec.KernelPageSize != nil && *profile.Spec.KernelPageSize == performancev2.KernelPageSize("64K") {
```
MarSik, line 154: Same here about the suffix letter (K vs k). Because see that on line 41 you use 64k-pages as the kernel name (uses lowercase).

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
rbaturov: Yes you are right. I'll change this to a lowered case character.

docs/performanceprofile/performance_profile.md
```diff
@@ -193,6 +195,13 @@ RealTimeKernel defines the set of parameters relevant for the real time kernel.
 
 [Back to TOC](#table-of-contents)
 
+## KernelPageSize
+
+KernelPageSize defines the kernel page size that will be used by the kernel.
+4k is the default value, 64k is only supported on aarch64.
```
swatisehgal, line 201: I think we should add here or in the `PerformanceProfileSpec` section that real time kernel with 64k kernel page size is not supported with on aarch64. WDYT?

rbaturov: I agree

rbaturov:
@MarSik Should we go ahead with this?

swatisehgal (commented):
LGTM from me.

Will defer the final approval to @MarSik.

MarSik (approved):

rbaturov:
cc @bartwensley

bartwensley (changes requested):

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -438,6 +450,67 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
...
+
+	if r.Spec.KernelPageSize == nil {
+		return allErrs
+	}
+
+	// We can only partially validate this if we have no nodes
+	// We can check that the value used is legitimate but we cannot check
+	// whether it is supposed to be x86 or aarch64
+	x86 := false
+	aarch64 := false
```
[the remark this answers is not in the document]
rbaturov: We have the same problem [here](https://github.com/rbaturov/cluster-node-tuning-operator/blob/master/pkg/apis/performanceprofile/v2/performanceprofile_validation.go#L357).
I think we should address both in a later PR

rbaturov:
@bartwensley Ready for another iteration

PR #1269, opened 2024-12-31
PPC: correct EnableHardwareTuning flag value

shajmakh:
By mistake this was overridden to always `true` in this PR: #1236 Correct this by setting the value equal to the passed value from the parsed data.


swatisehgal (commented):
/lgtm

/retest-required

PR #1278, opened 2025-01-09
WIP

ffromani:
WIP TBD


pkg/apis/performanceprofile/v2/performanceprofile_types.go
```diff
@@ -34,6 +34,13 @@ const PerformanceProfileEnablePhysicalRpsAnnotation = "performance.openshift.io/
 // that ignores the removal of all RPS settings when realtime workload hint is explicitly set to false.
 const PerformanceProfileEnableRpsAnnotation = "performance.openshift.io/enable-rps"
 
+// PerformanceProfileAutogenerateEnablementAnnotation is an experimental annotation which let the controller
+// create the enablement file for the tech-preview `prefer-align-cpus-by-uncorecache` cpumanager policy option.
+// If the option is detected, the generated MachineConfig is automatically augmented with the expected enablement
+// file. Still, users wishing to use the feature need to explicitly opt in. To do so, instead of opting in
+// adding the enablement file, they can opt in adding the annotation and setting its value to "true".
+const PerformanceProfileAutogenerateEnablementAnnotation = "performance.openshift.io/autogenerate-enablement"
```
MarSik, line 42: I would prefer if the name was visually close to the cpu manager flag it is enabling. Something like kubeletconfig.enable-llc-feature maybe.

pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset_test.go
```diff
@@ -0,0 +1,95 @@
...
+		It("should not be generated if missing", func() {
+			mfs, err := GetNewComponents(profile, opts)
+			Expect(err).ToNot(HaveOccurred())
+
+			ok, err := findFileInIgnition(mfs, expectedPath)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ok).To(BeFalse(), "expected path %q found in ignition", expectedPath)
+		})
+
+		It("should not be generated if missing control annotation", func() {
```
MarSik, line 38: You should also check that the prefer-align-cpus-by-uncorecache is false in the final KubeletConfig. Otherwise the kubelet dies.

pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go
```diff
@@ -53,7 +56,13 @@ func (ms *ManifestResultSet) ToManifestTable() ManifestTable {
 func GetNewComponents(profile *performancev2.PerformanceProfile, opts *components.Options) (*ManifestResultSet, error) {
 	machineConfigPoolSelector := profilecomponent.GetMachineConfigPoolSelector(profile, opts.ProfileMCP)
 
-	mc, err := machineconfig.New(profile, &opts.MachineConfig)
+	autogen := profileutil.HaveEnablementAutogeneration(profile)
+	llcEnabled := profileutil.IsLLCAlignmentEnabled(profile)
+	klog.V(4).InfoS("components manifests", "autogenerateEnablement", autogen, "LLCEnabled", llcEnabled)
+
+	mcOpts := opts.MachineConfig.Clone()
+	mcOpts.LLCFileEnabled = autogen && llcEnabled
```
MarSik, line 64: What we need is also - when autogen == false force llcEnabled to false.
Then not generating the file is cleaner, but makes no functional difference to kubelet.

pkg/apis/performanceprofile/v2/performanceprofile_types.go
ffromani: fair enough, renamed

pkg/performanceprofile/controller/performanceprofile/components/manifestset/manifestset.go
ffromani: ok, this means we need a different feature then. Is no longer a simplification but rather a master switch. At this point we should evaluate if we should add to our APIs and take full ownership of the flag.

PR #1279, opened 2025-01-09
CNF-17417: Enhanced utility function for fetching container runtime

SargunNarula:
In our tests, we usually create a ```TestPod``` and then configure the ```nodeSelector``` and ```namespace```. Consequently, the ```NodeName``` remains unset, which leads to the runtime utility function failing to retrieve the node. This PR improve the function to retrieve node from pod's ```nodeSelector``` .

Refrence test cases:

- cpu_management [Link](https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/1_performance/cpu_management.go#L286)
- cgroups [Link](https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/7_performance_kubelet_node/cgroups.go#L302)
- workloadhints [Link](https://github.com/openshift/cluster-node-tuning-operator/blob/master/test/e2e/performanceprofile/functests/8_performance_workloadhints/workloadhints.go#L714)

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -939,6 +939,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			guaranteedPod = makePod(ctx, workerRTNode, true)
 			err := testclient.Client.Create(ctx, guaranteedPod)
 			Expect(err).ToNot(HaveOccurred(), "Failed to create guaranteed pod")
+			guaranteedPod, err = pods.WaitForCondition(ctx, client.ObjectKeyFromObject(guaranteedPod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
+			Expect(err).ToNot(HaveOccurred())
 			defer func() {
 				if guaranteedPod != nil {
 					testlog.Infof("deleting pod %q", guaranteedPod.Name)
@@ -1078,7 +1080,6 @@ func getConfigJsonInfo(pod *corev1.Pod, containerName string, workerRTNode *core
 func makePod(ctx context.Context, workerRTNode *corev1.Node, guaranteed bool) *corev1.Pod {
 	testPod := pods.GetTestPod()
 	testPod.Namespace = testutils.NamespaceTesting
-	testPod.Spec.NodeName = workerRTNode.Name
 	testPod.Spec.NodeSelector = map[string]string{testutils.LabelHostname: workerRTNode.Name}
 	if guaranteed {
 		testPod.Spec.Containers[0].Resources = corev1.ResourceRequirements{
--- a/test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
+++ b/test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
@@ -6,7 +6,6 @@ import (
 	"path/filepath"
 
 	corev1 "k8s.io/api/core/v1"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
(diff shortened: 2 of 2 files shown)


test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
```diff
@@ -23,9 +23,17 @@ const (
 // in the node where the given pod is running
 func GetContainerRuntimeTypeFor(ctx context.Context, c client.Client, pod *corev1.Pod) (string, error) {
 	node := &corev1.Node{}
-	if err := c.Get(ctx, client.ObjectKey{Name: pod.Spec.NodeName}, node); err != nil {
-		return "", err
+
+	if pod.Spec.NodeName == "" {
```
yanirq, line 27: what happens if NodeName is not empty ?  would'nt the node object above be empty ?

test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
```diff
@@ -23,9 +23,17 @@ const (
...
+		if err := c.Get(ctx, client.ObjectKey{Name: pod.Spec.NodeSelector[testutils.LabelHostname]}, node); err != nil {
+			return "", err
+		}
+	} else {
```
yanirq, line 31: else if{  ... is also an accepted format

test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
SargunNarula: if NodeName is not empty, we retrieve the node object using it through the client.Get
The above node object is declared to pass with Get methode

shajmakh, line 27: I wouldn't depend on the user using this function correctly so I would go and validate the input before processing it. from what I understand this function is intended to be called with running pods but it can be misused easily by calling it with pod object that was manually created instead of a fetched one. 
If the pod is running then there is no way the pod doesn't have the field `Spec.nodeName` as it must be running on some node. I'd start from here. So fetch the pod freshly from the cluster using its keyobject (name and namespace) and handle error, and then apply the check:

```
if pod.Spec.NodeName == "" {
//meaning the pod is not running so no node is assigned yet thus return with empty string and relevant error
return "", fmt.Errorf(...)
}
// no need for else statement because if you got here it's sure the nodename field in not empty 
// fetch the node by node name and fetch the runtime and other needed code
```

shajmakh (commented):
Thanks for filing this, I see the point of the change, I left a comment let me know if this works. Note that this may need to update the way/place this is used

test/e2e/performanceprofile/functests/utils/cgroup/runtime/runtime.go
SargunNarula: Thanks for the review comment. I'm catching up late. I intended this change for a manually created pod object. Ideally, the function comment suggests that this function is for a running pod. The patch should update the function to handle both scenarios. I will adopt this change in the next commit.

SargunNarula:
**Update:** This fix addresses scenarios where a user or developer omits the `nodeName` setting during pod creation. If a pod is manually created without explicitly setting `nodeName`, the scheduler will need additional time to assign the pod to a node. This delay can lead to test failures and increase test flakiness.  

**Example:**  
If the following [line](https://github.com/openshift/cluster-node-tuning-operator/blob/main/test/e2e/performanceprofile/functests/1_performance/cpu_management.go#L1081) is not explicitly defined, this [test](https://github.com/openshift/cluster-node-tuning-operator/blob/main/test/e2e/performanceprofile/functests/1_performance/cpu_management.go#L952) will fail –
```
  Mar 28 13:57:49.698: [INFO]: found daemon pod node-inspector-jlzv9 for node <node-name>
  STEP: Creating a guaranteed pod @ 03/28/25 13:57:49.854
  Mar 28 13:57:49.891: [ERROR]: Failed to fetch runtime for Guaranteed Pod: resource name may not be empty
  Mar 28 13:57:51.893: [ERROR]: Failed to fetch runtime for Guaranteed Pod: resource name may not be empty
```

test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -939,6 +939,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			guaranteedPod = makePod(ctx, workerRTNode, true)
 			err := testclient.Client.Create(ctx, guaranteedPod)
 			Expect(err).ToNot(HaveOccurred(), "Failed to create guaranteed pod")
+			guaranteedPod, err = pods.WaitForCondition(ctx, client.ObjectKeyFromObject(guaranteedPod), corev1.PodReady, corev1.ConditionTrue, 10*time.Minute)
```
shajmakh, line 942: this commit has 2 different changes, I'd split them because they are not related.
Also let's add commit messages to all commits please

shajmakh (commented):
Thanks for the enhancments. 
Added one comment apart from that code changes looks good.

shajmakh:
Thanks for addressing the comments!
/lgtm
/approve
please add a Jira reference if exists, if not and you don't desire to backport the change, simply add `NO-JIRA: ` prefix to the PR title

SargunNarula:
Affected tests are not broken and pass with the new changes, Results shared via slack.

PR #1280, opened 2025-01-13
OCPBUGS-48319: Add missing Polarion test case id

mrniranjan:
Add polarion test case id to few
tests for reporting results to test case management systems


ffromani:

test-only fix, can't break the product

PR #1282, opened 2025-01-15
OCPBUGS-48422: [release-4.16]  e2e: tuned degraded test fix

rbaturov:
This is a manual backport of #1259.
* Addressed a corner case where nodes based on VMs would always log a warning without verifying if the profile was degraded. This incorrect behavior has been fixed to properly check the profile state before logging warnings.
* Modified isVM func to execute "systemd-detect-virt" command after chroot to rootfs. This will ensures accurate virtualization detection by running the command in the host's context instead of the container.


Tal-or (approved):
Trivial backport

PR #1286, opened 2025-01-20
CNF-16459: Update cnf-tests image

rbaturov:
This PR corrects the cnf-tests image to point to the correct image version of this release.
We plan to create a corresponding PR for this down to 4.14 at least.

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/images/images.go
+++ b/test/e2e/performanceprofile/functests/utils/images/images.go
@@ -13,7 +13,7 @@ func init() {
 	cnfTestsImage = os.Getenv("CNF_TESTS_IMAGE")
 
 	if cnfTestsImage == "" {
-		cnfTestsImage = "cnf-tests:4.14"
+		cnfTestsImage = "cnf-tests:4.19"
 	}
 
 	if registry == "" {


ffromani:
/approve
/lgtm

but we would need some form of automation to make sure images are kept up to date

rbaturov:
> /approve /lgtm
> 
> but we would need some form of automation to make sure images are kept up to date

Yes I'm thinking how this is going to be done

rbaturov:
> /approve /lgtm
> 
> but we would need some form of automation to make sure images are kept up to date

I just realized we were already branched out from 4.18. Will update the PR to point this image to 4.19 instead.

test/e2e/performanceprofile/functests/utils/images/images.go
```diff
@@ -13,7 +13,7 @@ func init() {
 	cnfTestsImage = os.Getenv("CNF_TESTS_IMAGE")
 
 	if cnfTestsImage == "" {
-		cnfTestsImage = "cnf-tests:4.14"
+		cnfTestsImage = "cnf-tests:4.19"
```
bartwensley, line 16: We need a fix that does not hardcode the image version.

bartwensley (changes requested):

test/e2e/performanceprofile/functests/utils/images/images.go
MarSik: That is actually hard to do here, the source has no way to find out the version of a build from a different repo. That is why the env var is here.

ffromani: this is the fallback default, it's a fair assumption to go in lockstop with the NTO branch/version. The existing env var can and should be used for fine tuning, leaving the hardcoded value to sane defaults.

swatisehgal (commented):
/lgtm

Adding label back, not sure why the label was removed.

/retest

bartwensley:
What about this hardcoded (and outdated) reference to a specific commit in the cluster-node-tuning repo (where some of the test code resides):
https://github.com/openshift-kni/cnf-features-deploy/blob/27523944b9eccb991d67dff325d370ddacacd1fb/go.mod#L208

rbaturov:
> What about this hardcoded (and outdated) reference to a specific commit in the cluster-node-tuning repo (where some of the test code resides): https://github.com/openshift-kni/cnf-features-deploy/blob/27523944b9eccb991d67dff325d370ddacacd1fb/go.mod#L208

As far as I understood, we don't run tests from this vendored repo.
All the tests we run are from the git submodules.
We need to investigate which package imports that and see whether we should bump this. I'm on it

PR #1294, opened 2025-02-05
OCPBUGS-49823: Support hugepage size selection for ARM-based clusters

rbaturov:
This PR introduces support for selecting hugepage sizes on ARM-based clusters. It complements the work done in #1086  and ensures that hugepage configurations are correctly applied.

E2E tests are planned to be added in a future update.

### DIFF
--- a/hack/render-sync.sh
+++ b/hack/render-sync.sh
@@ -43,3 +43,4 @@ rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap/no-mcp
 rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap-cluster/extra-mcp bootstrap/extra-mcp
 rendersync --owner-ref none -- base/performance manual-cluster/performance no-ref 
 rendersync --owner-ref none -- base/performance manual-cluster/cpuFrequency default/cpuFrequency 
+rendersync --owner-ref none -- base/performance manual-cluster/arm default/arm
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -429,10 +429,14 @@ func getSystemdContent(options []*unit.UnitOption) (string, error) {
 // GetHugepagesSizeKilobytes retruns hugepages size in kilobytes
 func GetHugepagesSizeKilobytes(hugepagesSize performancev2.HugePageSize) (string, error) {
 	switch hugepagesSize {
-	case "1G":
-		return "1048576", nil
 	case "2M":
 		return "2048", nil
+	case "32M":
+		return "32768", nil
+	case "512M":
+		return "524288", nil
+	case "1G":
+		return "1048576", nil
 	default:
 		return "", fmt.Errorf("can not convert size %q to kilobytes", hugepagesSize)
 	}
--- a/pkg/performanceprofile/controller/performanceprofile_controller_test.go
+++ b/pkg/performanceprofile/controller/performanceprofile_controller_test.go
@@ -503,84 +503,97 @@ var _ = Describe("Controller", func() {
 					Expect(cmdlineRealtimeWithoutCPUBalancing.MatchString(*t.Spec.Profile[0].Data)).To(BeTrue())
 				})
 
-				It("should update MC when Hugepages params change without node added", func() {
-					size := performancev2.HugePageSize("2M")
-					profile.Spec.HugePages = &performancev2.HugePages{
-						DefaultHugePagesSize: &size,
-						Pages: []performancev2.HugePage{
-							{
-								Count: 8,
-								Size:  size,
+				DescribeTable("should update MC when Hugepages params change without node added",
+					func(size performancev2.HugePageSize) {
+						profile.Spec.HugePages = &performancev2.HugePages{
+							DefaultHugePagesSize: &size,
+							Pages: []performancev2.HugePage{
+								{
+									Count: 8,
+									Size:  size,
+								},
 							},
-						},
-					}
+						}
 
-					r := newFakeReconciler(profile, mc, kc, tunedPerformance, profileMCP, infra, clusterOperator)
+						r := newFakeReconciler(profile, mc, kc, tunedPerformance, profileMCP, infra, clusterOperator)
+						Expect(reconcileTimes(r, request, 1)).To(Equal(reconcile.Result{}))
 
-					Expect(reconcileTimes(r, request, 1)).To(Equal(reconcile.Result{}))
+						By("Verifying Tuned profile update")
+						key := types.NamespacedName{
+							Name:      components.GetComponentName(profile.Name, components.ProfileNamePerformance),
+							Namespace: components.NamespaceNodeTuningOperator,
+						}
+						t := &tunedv1.Tuned{}
+						err := r.Get(context.TODO(), key, t)
+						Expect(err).ToNot(HaveOccurred())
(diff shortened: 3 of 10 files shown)


pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -433,6 +433,10 @@ func GetHugepagesSizeKilobytes(hugepagesSize performancev2.HugePageSize) (string
 		return "1048576", nil
 	case "2M":
 		return "2048", nil
+	case "32M":
+		return "32768", nil
+	case "512M":
+		return "524288", nil
```
rshemtov13, line 439: Would be nicer just to put these in order ascending or descending I guess.

test/e2e/performanceprofile/cluster-setup/manual-cluster/arm/performance_profile.yaml
```diff
@@ -0,0 +1,27 @@
...
+  hugepages:
+    defaultHugepagesSize: "32M"
+    pages:
+      - size: "32M"
+        count: 4
+        node: 0
+      - size: "512M"
+        count: 1
+  realTimeKernel:
+    enabled: true
```
rshemtov13, line 19: Do we need to have kernelPageSize: 64k here and turn realTimeKernel to false?

rshemtov13: Ok gotcha understood. Thanks for clarifying.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -429,10 +429,14 @@ func getSystemdContent(options []*unit.UnitOption) (string, error) {
 // GetHugepagesSizeKilobytes retruns hugepages size in kilobytes
 func GetHugepagesSizeKilobytes(hugepagesSize performancev2.HugePageSize) (string, error) {
```
ffromani, line 430: the problem here is that this function doesn't take into account the hardware architecture.
Do we trust at this point in time the flow, so are we guaranteed that if we reach this stage the combination {hugepageSize, machineArchicture} is legal?
If so, why do we need this function at all? we can trivially convert any hugepagesize treated as string to the kilobyte value.

pkg/performanceprofile/controller/performanceprofile_controller_test.go
```diff
@@ -503,84 +503,97 @@ var _ = Describe("Controller", func() {
...
+						Expect(err).ToNot(HaveOccurred())
+						expectedCmdline := fmt.Sprintf(`\s*cmdline_hugepages=\+\s*default_hugepagesz=%s\s+hugepagesz=%s\s+hugepages=8\s*`, size, size)
+						cmdlineHugepages := regexp.MustCompile(expectedCmdline)
+						Expect(cmdlineHugepages.MatchString(*t.Spec.Profile[0].Data)).To(BeTrue())
+					},
+					Entry("Hugepages size 2M", performancev2.HugePageSize("2M")),
+					Entry("Hugepages size 32M", performancev2.HugePageSize("32M")),
+					Entry("Hugepages size 512M", performancev2.HugePageSize("512M")),
+					Entry("Hugepages size 1G", performancev2.HugePageSize("1G")),
+				)
```
swatisehgal, line 537: I suggest adding a test case for an invalid hugepage size to ensure we properly handle incorrect configurations.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
rbaturov: The [validation](https://github.com/openshift/cluster-node-tuning-operator/pull/1086) ensures the combination {hugepageSize, machineArchicture} is legal.
We can implement some sort of converter that will generalize the switch case instead, which will deal with these three cases - hugepages inserted in GB, MB, and KB, and convert accordingly.
WDYT?

swatisehgal: This is a good point, even if the hugepage size is valid, it might not be compatible with a specific machine architecture and we need to handle that.

hack/render-sync.sh
```diff
@@ -42,4 +42,5 @@ rendersync manual-cluster/performance base/performance default
 rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap/no-mcp
 rendersync bootstrap-cluster/performance pinned-cluster/default bootstrap-cluster/extra-mcp bootstrap/extra-mcp
 rendersync --owner-ref none -- base/performance manual-cluster/performance no-ref 
-rendersync --owner-ref none -- base/performance manual-cluster/cpuFrequency default/cpuFrequency 
\ No newline at end of file
+rendersync --owner-ref none -- base/performance manual-cluster/cpuFrequency default/cpuFrequency 
+rendersync --owner-ref none -- base/performance manual-cluster/arm default/arm
```
jmencak, line 46: Nit: Every text file should really end in EOL character in my view (and apparently the tooling's view too). Let's fix this.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
rbaturov: @swatisehgal The validation handles this, it checks that the request size is compatible with the nodes arch referenced by the MCP.

pkg/performanceprofile/controller/performanceprofile_controller_test.go
rbaturov: As I said before, at this point the hugepages size should be valid, assuming we trust the validation is working.

swatisehgal (commented):
/lgtm

Thanks for clarifying that we already have validation in place to ensure that the specified hugepage size is supported on the hardware architecture.

Defer to @ffromani for final approval.

rbaturov:
/hold 
Needs to be verified on ARM cluster - verification in progress.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
ffromani: the best option is to remove this function in a followup PR. We can use https://github.com/docker/go-units which is an indirect dependency anyway

pkg/performanceprofile/controller/performanceprofile_controller_test.go
ffromani: Then we cannot hit the error path in `GetHugepagesSizeKilobytes` and thus
1. we should document that in the function docstring
2. we should remove the `error` return value

ffromani (commented):
let's remove the double validation if we are sure we can't hit the error path. We can replace that shallow function in a later PR.

rbaturov:
> let's remove the double validation if we are sure we can't hit the error path. We can replace that shallow function in a later PR.

Thanks for the review Francesco, I have generalized the `GetHugepagesSizeKilobytes` to handle various inputs using the package you mentioned. 
However, since we are using it and relying on it, we still want to use `error` even though we expect our inputs to be valid because we can't guarantee their logic not to fail.

rbaturov:
> thanks @rbaturov . I think the dependency change is better done in a followup PR. I was not clear enough in my previous statement, sorry about that. I think we should not change deps in this PR, so let's revert that part.

No worries, that's ok

swatisehgal:
/unhold
as per the comment above

PR #1295, opened 2025-02-13
OCPBUGS-49823: machineconfig: add support for various hugepage sizes

rbaturov:
* Generalized GetHugepagesSizeKilobytes function to handle various inputs. This allows us to use more possible hugepages size (e.g 512M on ARM). Pleas note we expect the validation to ensure the node arch + hugepage size combination is valid. Therefore, this function assumes a valid input, so it can safely convert it to bytes.
* This is done by using the `github.com/docker/go-units` which we should vet.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf
 	github.com/coreos/ignition v0.35.0
 	github.com/coreos/ignition/v2 v2.18.0
+	github.com/docker/go-units v0.5.0
 	github.com/go-logr/stdr v1.2.2
 	github.com/google/go-cmp v0.6.0
 	github.com/jaypipes/ghw v0.8.1-0.20210605191321-eb162add542b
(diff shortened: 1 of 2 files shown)


rbaturov:
@ffromani

rbaturov:
@jmencak Can you please review this?
Please note there is a dependency change here.

jmencak:
While I'm not thrilled by another dependency, it seems well established/maintaned. This looks reasonable to me. 
/lgtm
/approve

Giving a chance to others to comment.
/hold

jmencak:
Actually NTO already seems to be using 
```
github.com/docker/go-unit
```
that's why we didn't see it being added.

ffromani:
codewise LGTM
the only concern is the new dep management.
Do we need a round of `go mod tidy`?

rbaturov:
> codewise LGTM the only concern is the new dep management. Do we need a round of `go mod tidy`?

Thanks for noticing that.
@jmencak Please take a look of the last committed changes

ffromani:
LGTM

jmencak:
Still
/lgtm

> Do we need a round of `go mod tidy`?

The dependency was already there.  FWICS, from 4.14.  I regularly do `go mod tidy` when bumping NTO deps.

rbaturov:
> Still /lgtm
> 
> > Do we need a round of `go mod tidy`?
> 
> The dependency was already there. FWICS, from 4.14. I regularly do `go mod tidy` when bumping NTO deps.

We needed to run `go mod tidy` since that package was changed from indirect to direct dependency.

rbaturov:
/lgtm
Just rebased

jmencak:
Still
/lgtm

PR #1296, opened 2025-02-16
CNF-16742: kernelPageSize + hugepages combinations validation

rbaturov:
This PR includes these changes:

* Corrected valid huge page sizes for ARM.
* Improved huge pages validation to depend on the configured kernelPageSize.
* Added unit tests to validate all `kernelPageSize` and `hugepages` combinations, covering both valid and invalid cases.

This improves validation coverage and prevents misconfiguration.

The validation for this work can be found in the Jira ticket.

### DIFF
--- a/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
+++ b/pkg/apis/performanceprofile/v2/performanceprofile_validation.go
@@ -31,6 +31,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	corev1 "k8s.io/api/core/v1"
@@ -43,10 +44,12 @@ import (
 const (
 	kernelPageSize4k  = "4k"
 	kernelPageSize64k = "64k"
+	hugepagesSize64k  = "64k"
 	hugepagesSize2M   = "2M"
 	hugepagesSize32M  = "32M"
 	hugepagesSize512M = "512M"
 	hugepagesSize1G   = "1G"
+	hugepagesSize16G  = "16G"
 	amd64             = "amd64"
 	aarch64           = "arm64"
 )
@@ -56,11 +59,15 @@ var x86ValidHugepagesSizes = []string{
 	hugepagesSize1G,
 }
 
-// Each kernel page size has only a single valid hugepage size on aarch64
-var aarch64ValidHugepagesSizes = []string{
-	hugepagesSize2M,   // With 4k kernel pages
-	hugepagesSize32M,  // With 16k kernel pages
-	hugepagesSize512M, // With 64k kernel pages
+// Each kernel page size has a certain group of valid hugepage sizes on aarch64
+// https://docs.kernel.org/mm/vmemmap_dedup.html
+var aarch64HugePagesByKernelPageSize = map[string][]string{
+	kernelPageSize4k:  {hugepagesSize64k, hugepagesSize2M, hugepagesSize32M, hugepagesSize1G},
+	kernelPageSize64k: {hugepagesSize2M, hugepagesSize512M, hugepagesSize16G},
+}
+
+var x86HugePagesByKernelPageSize = map[string][]string{
+	kernelPageSize4k: {hugepagesSize2M, hugepagesSize1G},
 }
 
 var aarch64ValidKernelPageSizes = []string{
@@ -155,8 +162,8 @@ func (r *PerformanceProfile) ValidateBasicFields() field.ErrorList {
 	allErrs = append(allErrs, r.validateSelectors()...)
 	allErrs = append(allErrs, r.validateAllNodesAreSameCpuArchitecture(nodes)...)
 	allErrs = append(allErrs, r.validateAllNodesAreSameCpuCapacity(nodes)...)
-	allErrs = append(allErrs, r.validateHugePages(nodes)...)
 	allErrs = append(allErrs, r.validateKernelPageSize(nodes)...)
+	allErrs = append(allErrs, r.validateHugePages(nodes)...)
 	allErrs = append(allErrs, r.validateNUMA()...)
 	allErrs = append(allErrs, r.validateNet()...)
 	allErrs = append(allErrs, r.validateWorkloadHints()...)
@@ -367,7 +374,16 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
 	// whether it is supposed to be x86 or aarch64
 	x86 := false
 	aarch64 := false
-	combinedHugepagesSizes := append(x86ValidHugepagesSizes, aarch64ValidHugepagesSizes...)
+
+	// Default to 4k if KernelPageSize is not set
+	// This is required for backward compatibility with performance profiles that don't have a kernelPageSize field.
+	kernelPageSize := kernelPageSize4k
+
+	// Override the default if KernelPageSize is specified
+	if r.Spec.KernelPageSize != nil {
+		kernelPageSize = string(*r.Spec.KernelPageSize)
+	}
+	hugepagesSizes := allHugePageSizes()
 
 	if len(nodes.Items) > 0 {
 		// `validateHugePages` implicitly relies on `validateAllNodesAreSameCpuArchitecture` to have already been run
@@ -381,32 +397,32 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
 		defaultSize := *r.Spec.HugePages.DefaultHugePagesSize
 		errField := "spec.hugepages.defaultHugepagesSize"
 		errMsg := "hugepages default size should be equal to one of"
-
+		docsRef := "https://docs.kernel.org/mm/vmemmap_dedup.html"
 		if x86 && !slices.Contains(x86ValidHugepagesSizes, string(defaultSize)) {
 			allErrs = append(
 				allErrs,
 				field.Invalid(
 					field.NewPath(errField),
 					r.Spec.HugePages.DefaultHugePagesSize,
-					fmt.Sprintf("%s %v", errMsg, x86ValidHugepagesSizes),
+					fmt.Sprintf("%s %v. doc reference=%s", errMsg, x86ValidHugepagesSizes, docsRef),
 				),
 			)
-		} else if aarch64 && !slices.Contains(aarch64ValidHugepagesSizes, string(defaultSize)) {
+		} else if aarch64 && !slices.Contains(aarch64HugePagesByKernelPageSize[kernelPageSize], string(defaultSize)) {
 			allErrs = append(
 				allErrs,
 				field.Invalid(
 					field.NewPath(errField),
 					r.Spec.HugePages.DefaultHugePagesSize,
-					fmt.Sprintf("%s %v", errMsg, aarch64ValidHugepagesSizes),
+					fmt.Sprintf("%s %v. doc reference=%s", errMsg, aarch64HugePagesByKernelPageSize[kernelPageSize], docsRef),
 				),
 			)
-		} else if !x86 && !aarch64 && !slices.Contains(combinedHugepagesSizes, string(defaultSize)) {
+		} else if !x86 && !aarch64 && !hugepagesSizes.Has(string(defaultSize)) {
 			allErrs = append(
 				allErrs,
 				field.Invalid(
 					field.NewPath(errField),
 					r.Spec.HugePages.DefaultHugePagesSize,
-					fmt.Sprintf("%s %v", errMsg, combinedHugepagesSizes),
+					fmt.Sprintf("%s %v. doc reference=%s", errMsg, hugepagesSizes, docsRef),
 				),
 			)
 		}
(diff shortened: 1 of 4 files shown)


rbaturov:
cc @fontivan

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -367,7 +373,8 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
 	// whether it is supposed to be x86 or aarch64
 	x86 := false
 	aarch64 := false
-	combinedHugepagesSizes := append(x86ValidHugepagesSizes, aarch64ValidHugepagesSizes...)
+	kernelPageSize := *r.Spec.KernelPageSize
+	AllHugepageSizesSet := allHugePageSizes()
```
yanirq, line 377: s/AllHugepageSizesSet/allHugepageSizesSet

yanirq: I would actually use a simplified name, perhaps `hugepagesSizes` or `hugepagesSizeSet`

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -647,6 +651,59 @@ var _ = Describe("PerformanceProfile", func() {
...
 			Expect(errors[0].Error()).To(ContainSubstring(("the page size should be equal to one of")))
 		})
+		Context("KernelPageSize with HugePages validation", func() {
+			var kernelToHugePageSizeByArch = map[string]map[string][]string{
+				aarch64: aarch64HugePagesByKernelPageSize,
+				amd64:   x86HugePagesByKernelPageSize,
+			}
+
+			var tableEntries []TableEntry
+			allHugePageSizes := allHugePageSizes()
```
yanirq, line 661: try being consistent with other values above (using set suffix if all should use one for example)

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -684,3 +696,20 @@ func (r *PerformanceProfile) getNodesList() (corev1.NodeList, error) {
 
 	return *nodes, nil
 }
+
+func allHugePageSizes() map[string]struct{} {
```
yanirq, line 700: what is the purpose of this function ?

rbaturov: Create a set of all kinds of hugepage sizes.

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -684,3 +696,20 @@ func (r *PerformanceProfile) getNodesList() (corev1.NodeList, error) {
...
+	hugePageSet := make(map[string]struct{})
+
+	for _, hugePageSizes := range x86HugePagesByKernelPageSize {
+		for _, size := range hugePageSizes {
+			hugePageSet[size] = struct{}{}
```
yanirq, line 705: this is just an init ?  or did you mean to populate with values here?

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
yanirq: init function?

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
rbaturov: No, this is an empty struct. Since in go, there is no set data structure, this is the workaround that can be done.
The values in this map are irrelevant (The keys are the set).

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
rbaturov: a  helper function

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -95,6 +96,7 @@ func GetFakeValidatorClient(nodeSpecs []NodeSpecifications) client.Client {
 // NewPerformanceProfile returns new performance profile object that used for tests
 func NewPerformanceProfile(name string) *PerformanceProfile {
 	size := HugePageSize1G
+	kernelPageSize := KernelPageSize(kernelPageSize4k)
```
swatisehgal, line 99: What happens when `KernelPageSize` is not specified but `hugepages` are and vice versa. Should we add those test cases as well?

rbaturov: Thanks for your review @swatisehgal 

When KernelPageSize is not specified it gets assigned by the [default 4k value](https://github.com/openshift/cluster-node-tuning-operator/pull/1262/files#diff-606d4678e0de6f457b4585b3799194ab2ad477d9eb01960b0aeb3f717e788675R586).
Therefore it is covered.
For the vice versa scenario, i.e kernelPageSize is specified without hugepages, we have test cases for this:
https://github.com/openshift/cluster-node-tuning-operator/pull/1262/files#diff-14ca8bee9fe60c76c3ecd29064edd3572e53e9b1f6f9ad46b368d021d57d08aa

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -43,10 +43,12 @@ import (
 const (
 	kernelPageSize4k  = "4k"
 	kernelPageSize64k = "64k"
+	hugepagesSize64K  = "64K"
```
rbaturov, line 46: @MarSik In this doc all the references are with capital K.
https://docs.kernel.org/mm/vmemmap_dedup.html 

I wonder if we should stay consistent and define something unified for both kernelPageSize and hugepages

fontivan: That is probably a good idea but I feel like that will opening a large can of worms. Every existing profile would need to be co-erced or updated manually to a capital K?

rbaturov: @fontivan Hugepages with the new 64K is a new option we didn't had before. this is correct also for the kernelPageSize option. Therefore, now we do have the opportunity to define a unified notation for both fields.

yanirq: what is the once source of truth here ?  cc @MarSik

rbaturov:
Ready for another review iteration

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -684,3 +704,20 @@ func (r *PerformanceProfile) getNodesList() (corev1.NodeList, error) {
 
 	return *nodes, nil
 }
+
+func allHugePageSizes() map[string]struct{} {
```
Tal-or, line 708: Use [set](https://pkg.go.dev/github.com/golang-collections/collections/set)

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -647,6 +651,59 @@ var _ = Describe("PerformanceProfile", func() {
...
+							Node:  ptr.To(int32(0)),
+							Size:  HugePageSize(hugePageSize),
+						},
+					}
+					KernelPageSize := KernelPageSize(kernelPageSize)
+					profile.Spec.KernelPageSize = &KernelPageSize
+
+					errors := profile.validateHugePages(nodes)
+					if expectError {
+						Expect(errors).NotTo(BeEmpty(), "should have validation error")
```
Tal-or, line 697: `ToNot` is more readable

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -887,6 +944,9 @@ var _ = Describe("PerformanceProfile", func() {
 			incorrectDefaultSize := HugePageSize("!#@")
 			profile.Spec.HugePages.DefaultHugePagesSize = &incorrectDefaultSize
 
+			incorrectKernelPageSize := KernelPageSize("!#@")
+			profile.Spec.KernelPageSize = &incorrectKernelPageSize
```
Tal-or, line 948: `profile.Spec.KernelPageSize = ptr.To(KernelPageSize("!#@"))`

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -647,6 +651,59 @@ var _ = Describe("PerformanceProfile", func() {
...
+			}
+			DescribeTable("KernelPageSize with HugePages validation",
+				func(arch string, kernelPageSize string, hugePageSize string, expectError bool) {
+					nodeSpecs := []NodeSpecifications{
+						{architecture: arch, cpuCapacity: 1000, name: "node"},
+					}
+					validatorClient = GetFakeValidatorClient(nodeSpecs)
+
+					nodes, err := profile.getNodesList()
+					Expect(err).To(BeNil())
```
Tal-or, line 682: `Expect(err).ToNot(HaveOccured())`

pkg/apis/performanceprofile/v2/performanceprofile_validation_test.go
```diff
@@ -647,6 +651,59 @@ var _ = Describe("PerformanceProfile", func() {
...
+					*profile.Spec.HugePages.DefaultHugePagesSize = HugePageSize(hugePageSize)
+					profile.Spec.HugePages.Pages = []HugePage{
+						{
+							Count: 1,
+							Node:  ptr.To(int32(0)),
+							Size:  HugePageSize(hugePageSize),
+						},
+					}
+					KernelPageSize := KernelPageSize(kernelPageSize)
+					profile.Spec.KernelPageSize = &KernelPageSize
```
Tal-or, line 693: `profile.Spec.KernelPageSize = ptr.To(KernelPageSize(kernelPageSize))`

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -381,67 +396,72 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
...
 				),
 			)
+		} else if !x86 && !aarch64 {
+			if _, exists := hugepagesSizes[string(defaultSize)]; !exists {
+				allErrs = append(
+					allErrs,
+					field.Invalid(
+						field.NewPath(errField),
+						r.Spec.HugePages.DefaultHugePagesSize,
+						fmt.Sprintf("%s %v. %s", errMsg, hugepagesSizes, docsRef),
```
Tal-or, line 425: ditto

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -381,67 +396,72 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
...
 		errMsg := "the page size should be equal to one of"
+		docsRef := "for more info: https://docs.kernel.org/mm/vmemmap_dedup.html"
 		if x86 && !slices.Contains(x86ValidHugepagesSizes, string(page.Size)) {
 			allErrs = append(
 				allErrs,
 				field.Invalid(
 					field.NewPath(errField),
 					r.Spec.HugePages.Pages,
-					fmt.Sprintf("%s %v", errMsg, x86ValidHugepagesSizes),
+					fmt.Sprintf("%s %v. %s", errMsg, x86ValidHugepagesSizes, docsRef),
```
Tal-or, line 442: ditto

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -381,67 +396,72 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
...
-				),
-			)
-		} else if !x86 && !aarch64 && !slices.Contains(combinedHugepagesSizes, string(page.Size)) {
-			allErrs = append(
-				allErrs,
-				field.Invalid(
-					field.NewPath(errField),
-					r.Spec.HugePages.DefaultHugePagesSize,
-					fmt.Sprintf("%s %v", errMsg, combinedHugepagesSizes),
+					fmt.Sprintf("%s %v. %s", errMsg, aarch64HugePagesByKernelPageSize[(kernelPageSize)], docsRef),
```
Tal-or, line 451: ditto

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -381,67 +396,72 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
...
 				),
 			)
+		} else if !x86 && !aarch64 {
+			if _, exists := hugepagesSizes[string(page.Size)]; !exists {
+				allErrs = append(
+					allErrs,
+					field.Invalid(
+						field.NewPath(errField),
+						r.Spec.HugePages.Pages,
+						fmt.Sprintf("%s %v. %s", errMsg, hugepagesSizes, docsRef),
```
Tal-or, line 461: ditto

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -381,67 +396,72 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
...
-				),
-			)
-		} else if !x86 && !aarch64 && !slices.Contains(combinedHugepagesSizes, string(defaultSize)) {
-			allErrs = append(
-				allErrs,
-				field.Invalid(
-					field.NewPath(errField),
-					r.Spec.HugePages.DefaultHugePagesSize,
-					fmt.Sprintf("%s %v", errMsg, combinedHugepagesSizes),
+					fmt.Sprintf("%s %v. %s", errMsg, aarch64HugePagesByKernelPageSize[kernelPageSize], docsRef),
```
Tal-or, line 415: ditto

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
```diff
@@ -381,67 +396,72 @@ func (r *PerformanceProfile) validateHugePages(nodes corev1.NodeList) field.Erro
...
-
+		docsRef := "for more info: https://docs.kernel.org/mm/vmemmap_dedup.html"
 		if x86 && !slices.Contains(x86ValidHugepagesSizes, string(defaultSize)) {
 			allErrs = append(
 				allErrs,
 				field.Invalid(
 					field.NewPath(errField),
 					r.Spec.HugePages.DefaultHugePagesSize,
-					fmt.Sprintf("%s %v", errMsg, x86ValidHugepagesSizes),
+					fmt.Sprintf("%s %v. %s", errMsg, x86ValidHugepagesSizes, docsRef),
```
Tal-or, line 406: Maybe more clear:
`fmt.Sprintf("%s %v. doc reference=%s", errMsg, hugepagesSizes, docsRef),`

Tal-or (commented):
Looking good 
added couple of comments

pkg/apis/performanceprofile/v2/performanceprofile_validation.go
rbaturov: The package at https://github.com/golang-collections/collections hasn't been updated in over 12 years, so adding it as a dependency might not be the best choice. I tried importing the set package, but it doesn't even support iteration, which we need. I submitted a PR to add this feature, but I'm not confident it will be reviewed anytime soon. Given this, I think it's better to stick with the current approach—while it may be a bit less readable, it avoids adding any new dependencies that would need to be vetted.

Tal-or: My bad, this is the right package: k8s.io/apimachinery/pkg/util/sets

rbaturov: Updated

Tal-or (approved):
/lgtm

PR #1301, opened 2025-02-26
CNF-16638: Reduce code duplication of RPS enabled checks

MarSik:
Centralize the RPS enabled logic into just one function and add a way to force RPS off.

This is prep-work for disabling RPS globally for all use cases.

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
@@ -94,6 +94,30 @@ var _ = Describe("Machine Config", func() {
 		})
 	})
 
+	Context("machine config creation with disabled RPS", func() {
+		It("should create machine config with no trace of RPS", func() {
+			profile := testutils.NewPerformanceProfile("test")
+			profile.Annotations = map[string]string{}
+			profile.Annotations[performancev2.PerformanceProfileEnableRpsAnnotation] = "false"
+
+			mc, err := New(profile, &components.MachineConfigOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			result := igntypes.Config{}
+
+			err = json.Unmarshal(mc.Spec.Config.Raw, &result)
+			Expect(err).ToNot(HaveOccurred())
+
+			for _, f := range result.Storage.Files {
+				Expect(f.Node.Path).To(Not(ContainSubstring("rps")), "rps configuration %s should not be present", f.Node.Path)
+			}
+
+			for _, f := range result.Systemd.Units {
+				Expect(f.Name).To(Not(ContainSubstring("rps")), "rps systemd unit %s should not be present", f.Name)
+			}
+		})
+	})
+
 	Context("with hugepages with specified NUMA node and offlinedCPUs", func() {
 		var manifest string
 
--- a/pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go
@@ -71,15 +71,19 @@ func IsPhysicalRpsEnabled(profile *performancev2.PerformanceProfile) bool {
 
 // IsRpsEnabled checks if all RPS should be applied
 func IsRpsEnabled(profile *performancev2.PerformanceProfile) bool {
-	if profile.Annotations == nil {
-		return false
-	}
-	isRpsEnabled, ok := profile.Annotations[performancev2.PerformanceProfileEnableRpsAnnotation]
-	if ok && isRpsEnabled == "true" {
-		return true
+	if profile.Annotations != nil {
+		// First check overrides
+		isRpsEnabled, ok := profile.Annotations[performancev2.PerformanceProfileEnableRpsAnnotation]
+		if ok && isRpsEnabled == "true" {
+			return true
+		} else if ok && isRpsEnabled == "false" {
+			return false
+		}
 	}
 
-	return false
+	// The default behavior enables RPS for real time workloads
+	return profile.Spec.WorkloadHints == nil ||
+		profile.Spec.WorkloadHints.RealTime == nil || *profile.Spec.WorkloadHints.RealTime
 }
 
 func IsMixedCPUsEnabled(profile *performancev2.PerformanceProfile) bool {
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -375,8 +375,7 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 			}
 		})
 		It("[test_id: 59572] Check RPS Mask is applied to at least one single rx queue on all veth interface", func() {
-			if profile.Spec.WorkloadHints != nil && profile.Spec.WorkloadHints.RealTime != nil &&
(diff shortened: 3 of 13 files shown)


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -415,8 +414,7 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 			}
 		})
 		It("[test_id:55012] Should have the correct RPS configuration", func() {
-			if profile.Spec.WorkloadHints != nil && profile.Spec.WorkloadHints.RealTime != nil &&
-				!*profile.Spec.WorkloadHints.RealTime && !profileutil.IsRpsEnabled(profile) {
+			if !profileutil.IsRpsEnabled(profile) {
```
mrniranjan, line 417: @MarSik don't we want to check the conditions of whether workloadhints is enabled or if RealTime workloadhints is enabled ?

test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -375,8 +375,7 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 			}
 		})
 		It("[test_id: 59572] Check RPS Mask is applied to at least one single rx queue on all veth interface", func() {
-			if profile.Spec.WorkloadHints != nil && profile.Spec.WorkloadHints.RealTime != nil &&
-				!*profile.Spec.WorkloadHints.RealTime && !profileutil.IsRpsEnabled(profile) {
+			if !profileutil.IsRpsEnabled(profile) {
```
mrniranjan, line 378: isRpsEnabled checks if Profile is annotations related to RPS,   By default in our E2E testing we don't add any RPS Annotations

test/e2e/performanceprofile/functests/1_performance/performance.go
mrniranjan, line 417: Test wise looks good to me

mrniranjan:
@MarSik we don't have any E2E test which checks the RPS annotation , should we add a test ?

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
```diff
@@ -94,6 +94,23 @@ var _ = Describe("Machine Config", func() {
...
+			profile.Annotations[performancev2.PerformanceProfileEnableRpsAnnotation] = "false"
+
+			mc, err := New(profile, &components.MachineConfigOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			y, err := yaml.Marshal(mc)
+			Expect(err).ToNot(HaveOccurred())
+
+			manifest := string(y)
+			Expect(manifest).To(Not(ContainSubstring("rps")))
```
Tal-or, line 110: Usually the content is base64 coded.
I think we should decode the content first and than perform this check.

pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go
```diff
@@ -71,15 +71,19 @@ func IsPhysicalRpsEnabled(profile *performancev2.PerformanceProfile) bool {
 
 // IsRpsEnabled checks if all RPS should be applied
 func IsRpsEnabled(profile *performancev2.PerformanceProfile) bool {
-	if profile.Annotations == nil {
-		return false
-	}
-	isRpsEnabled, ok := profile.Annotations[performancev2.PerformanceProfileEnableRpsAnnotation]
-	if ok && isRpsEnabled == "true" {
-		return true
+	if profile.Annotations != nil {
```
Tal-or, line 74: The convention is usually align the happy path to left.

yanirq:
looks good to me in general , once the small nits are addressed we can merge the PR

pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go
MarSik: Hmm, but here it is a protection against dereferencing nil. Are you suggesting I move the final return up and place it under profile.Annotations == nil?

Tal-or: yea and then the code can be simplified 
```
if profile.Annotations == nil {
// The default behavior enables RPS for real time workloads
		profile.Spec.WorkloadHints == nil ||
		profile.Spec.WorkloadHints.RealTime == nil || *profile.Spec.WorkloadHints.RealTime
	}

isRpsEnabled, ok := profile.Annotations[performancev2.PerformanceProfileEnableRpsAnnotation]
return ok && isRpsEnabled == "true"
```

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
MarSik: Well.. true it could happen. I do not think it is likely (especially in the filename). This allows a check without parsing the text back to yaml. A crude one, yes, but it works and catches both the systemd unit and the sysctl file.

pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go
MarSik: But that changes the logic for isRpsEnabled == "somethingstupid". I can move the ifs around, that is no bother, but I would like to keep just the two values active "true" and "false" and default to realtime workload hint value otherwise.

MarSik: isRpsEnabled != "false" is less readable I think

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
Tal-or: >This allows a check without parsing the text back to yaml.

Which is not so bad especially since we're talking about tests where we prefer correctness over performance.

pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go
Tal-or: > But that changes the logic for isRpsEnabled == "somethingstupid". I can move the ifs around, that is no bother, but I would like to keep just the two values active "true" and "false" and default to realtime workload hint value otherwise.

If you prefer the workload hint as the last resort then let's keep the original code

MarSik: Yeah, the hint is the default when no override is present.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig_test.go
```diff
@@ -94,6 +94,30 @@ var _ = Describe("Machine Config", func() {
...
+			profile := testutils.NewPerformanceProfile("test")
+			profile.Annotations = map[string]string{}
+			profile.Annotations[performancev2.PerformanceProfileEnableRpsAnnotation] = "false"
+
+			mc, err := New(profile, &components.MachineConfigOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			result := igntypes.Config{}
+
+			err = json.Unmarshal(mc.Spec.Config.Raw, &result)
```
MarSik, line 108: @Tal-or Is this better?

Tal-or (approved):
Thank you for the changes. Looking good

PR #1303, opened 2025-03-05
OCPBUGS-52352: Treat all aarch64 platforms as equals in tuned

MarSik:
We are commonly testing on two different ARM based platforms that use different cpu architecture designators ARM and APM.

The per-platform tuning profile selector did not expect APM and failed to include the proper profile.

Since we currently do not distinguish between the ARN vendors in any way, the patch changes the detection to be based on Architecture instead of on Vendor ID.


MarSik:
@bartwensley PTAL

bartwensley (commented):
/lgtm

nit: Spelling mistake in commit text (ARN)

PR #1304, opened 2025-03-06
PSAP-1608: Obsolete pod label-based matching

jmencak:
The Pod label matching functionality has been mentioned as a candidate for deprecation since OCP 4.4.  Reasons for this change:
  * the Pod label matching functionality does not scale well in larger clusters
  * ease merging the two NTO control loops into one

Report Tuned CRs which use Pod label matching functionality as invalid and trigger an alert (NTOPodLabelsUsed) for this to give cluster administrators time to update their configuration.

Implements: [PSAP-1608](https://issues.redhat.com//browse/PSAP-1608)

### DIFF
--- a/manifests/30-monitoring.yaml
+++ b/manifests/30-monitoring.yaml
@@ -133,6 +133,17 @@ spec:
       for: 30m
       labels:
         severity: warning
+    - alert: NTOPodLabelsUsed
+      annotations:
+        description: >-
+          The Node Tuning Operator is using deprecated functionality.
+          Using pod label matching has been discouraged since OCP 4.4 and this functionality will be removed in future versions.
+          Please revise and adjust your configuration (Tuned custom resources).
+        summary: The Node Tuning Operator is using deprecated functionality.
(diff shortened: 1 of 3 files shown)


jmencak:
@MarSik let's get this into 4.19. Feature Freeze approaching soon 18th April.  Thank you.
/cc @yanirq

yanirq:
just to understand - we can still specify on the tuned CR pod matching but the user will get an alert its deprecated and in a later version remove this completely ?

jmencak:
> just to understand - we can still specify on the tuned CR pod matching but the user will get an alert its deprecated and in a later version remove this completely ?

Yes, that's the intent. It will work similarly to [duplicate TuneD profiles with different content](https://github.com/openshift/cluster-node-tuning-operator/pull/1216).  The deprecation was mentioned as a candidate for removal in the docs since 4.4, but customers are not paying attention to this.  I've seen a recent case where pod labels were being used.  Once they see an alert, hopefully they should change their configuration.  Complete removal in the future will help us in the future to simplify the control loop.

PR #1305, opened 2025-03-06
NO-JIRA: Add observedGeneration to Profile status

jmencak:
`profiles.tuned.openshift.io` are managed both by the NTO operator and the operands.  The operator manages the `.Spec` part, while the operand reports the status of the TuneD profile application via the `.Status` part.

As it is impossible to tell whether the current `.Status` reflects potential `.Spec` changes without `.Status.ObservedGeneration`, introduce the new field.


MarSik (approved):
/lgtm

PR #1306, opened 2025-03-06
OCPBUGS-52484: Treat all aarch64 platforms as equals in tuned [4.18]

yanirq:
Manual backport of https://github.com/openshift/cluster-node-tuning-operator/pull/1303/

We are commonly testing on two different ARM based platforms that use different cpu architecture designators ARM and APM.

The per-platform tuning profile selector did not expect APM and failed to include the proper profile.

Since we currently do not distinguish between the ARM vendors in any way, the patch changes the detection to be based on Architecture instead of on Vendor ID.


MarSik:
/retitle OCPBUGS-52352: Treat all aarch64 platforms as equals in tuned

MarSik:
/retitle OCPBUGS-52484: Treat all aarch64 platforms as equals in tuned [4.18]

Tal-or (approved):

PR #1307, opened 2025-03-09
CNF-16013: e2e: arm: Add arm kernelPageSize e2e testing suite

rshemtov13:
This PR introduces the following changes:

* A new 14_arm directory, which contains the e2e test suite for validating support of the 64k kernelPageSize under PerformanceProfile for the aarch64 family of CPUs.

* A new test suite to validate, at the node level, the kernelPageSize values of 4k and 64k on aarch64. This is done by applying changes to the spec field and inspecting the node to verify that the correct values are reflected on the associated node.

The tests in the suite have been tested in an ARM cluster

### DIFF
--- a/test/e2e/performanceprofile/functests/14_arm/kernelpagesize.go
+++ b/test/e2e/performanceprofile/functests/14_arm/kernelpagesize.go
@@ -1,0 +1,151 @@
+package __arm
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/infrastructure"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/label"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/nodes"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/poolname"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profilesupdate"
+)
+
+const (
+	kernelPageSize4k       = "4k"
+	kernelPageSize64k      = "64k"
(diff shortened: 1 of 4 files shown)


test/e2e/performanceprofile/functests/14_arm/test_suite_arm_test.go
```diff
@@ -0,0 +1,28 @@
...
+	ctrllog.SetLogger(stdr.New(log.New(os.Stderr, "", log.LstdFlags|log.Lshortfile)))
+
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Kernel page size suite")
+}
+
+var _ = ReportAfterSuite("kernel page size suite", func(r Report) {
+	if qe_reporters.Polarion.Run {
+		reporters.ReportViaDeprecatedReporter(&qe_reporters.Polarion, r)
+	}
```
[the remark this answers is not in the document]
mrniranjan: Yes, this just adds the relevant fields, the job only pushes the results

test/e2e/performanceprofile/functests/14_arm/kernelpagesize.go
```diff
@@ -0,0 +1,139 @@
...
+		Expect(err).ToNot(HaveOccurred())
+		if !isArm {
+			Skip("This test requires an aarch64 arm CPU")
+		}
+
+		By("Make sure that the performance profile starts the test suite with 4k kernelPageSize and real time kernel disabled")
+		currentKernelPageSize, err := getKernelPageSizeFromNode(ctx, workerRTNode)
+		Expect(err).ToNot(HaveOccurred())
+
+		if *perfProfile.Spec.RealTimeKernel.Enabled || currentKernelPageSize != kernelPageSize4k {
```
Tal-or, line 68: This is a pointer dereference here, so I would check that it's not `nil` first

Tal-or (commented):
Looking good, left a small comment.

Do we also need to run this test u/s? probably not because we cannot control (AFAIK) on the machine we're getting from CI

Anyway if we do, we should add this suite to one of the makefile targets

test/e2e/performanceprofile/functests/14_arm/kernelpagesize.go
mrniranjan: Thanks Talor,  Agree with Makefile targets . Currently tests need to be run manually

Tal-or:
/lgtm

Thank you

test/e2e/performanceprofile/functests/14_arm/kernelpagesize.go
```diff
@@ -0,0 +1,150 @@
...
+					Expect(kernelPageSize).To(Equal(expectedKernelPageSize))
+				},
+				Entry("[test_id:80459] should accept 64k kernelPageSize", kernelPageSize64k, kernelPageSizeBytes64k),
+				Entry("[test_id:80461] should accept 4k kernelPageSize", kernelPageSize4k, kernelPageSizeBytes4k),
+			)
+		})
+	})
+})
+
+func getKernelPageSizeFromNode(ctx context.Context, node corev1.Node) (string, error) {
```
rbaturov, line 141: Please add `GinkgoHelper()` as the first line in this function.
Search for this in the [Logging Output](https://onsi.github.io/ginkgo/) section for the explanation why.

rshemtov13: Sure np, will add. Thanks!

yanirq:

test only change. not affecting core payload

PR #1313, opened 2025-03-28
OCPBUGS-54628: Add retry method if we hit jira rate limit

mrniranjan:
Jira implemented rate limit due to which tests
which should be skipped due to known issues do not get skipped.

This PR implements retry with exponential backoff
to retry the request if we hit http status code 429(too many requests)

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/jira/status.go
+++ b/test/e2e/performanceprofile/functests/utils/jira/status.go
@@ -9,6 +9,8 @@ import (
 	"net/url"
 	"strconv"
 	"time"
+
+	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
 )
 
 const (
@@ -55,12 +57,15 @@ func RetrieveJiraStatus(key string) (*JiraIssueStatusResponse, error) {
 		case http.StatusTooManyRequests:
 			if retryAfter := resp.Header.Get("Retry-After"); retryAfter != "" {
 				retryTime, err := convertRetryAfterTime(retryAfter)
+				testlog.Infof("Using Retry-After header to retry after %s seconds", retryTime.String())
 				if err != nil {
 					return nil, err
 				}
 				time.Sleep(retryTime)
 			} else {
+				testlog.Info("No Retry-After header found using exponential backoff method to retry")
 				delay := retryWithBackOff(retryCount)
+				testlog.Infof("Wait for %s seconds before retry to fetch bug status", delay.String())
 				time.Sleep(delay)
 			}
 		default:
--- a/test/e2e/performanceprofile/functests/utils/jira/status.go
+++ b/test/e2e/performanceprofile/functests/utils/jira/status.go
@@ -2,10 +2,12 @@ package jira
 
 import (
 	"encoding/json"
+	"fmt"
 	"github.com/pkg/errors"
 	"math/rand"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


MarSik:
I wonder if this is the right solution. It could potentially make stuff worse. Isn't there a header that would tell us when the rate limiting is going to end? I know GitHub API uses one like that, not sure about Jira.

mrniranjan:
> I wonder if this is the right solution. It could potentially make stuff worse. Isn't there a header that would tell us when the rate limiting is going to end? I know GitHub API uses one like that, not sure about Jira.

Thanks martin, will check and update

mrniranjan:
> I wonder if this is the right solution. It could potentially make stuff worse. Isn't there a header that would tell us when the rate limiting is going to end? I know GitHub API uses one like that, not sure about Jira.

Addressed in the latest commit

test/e2e/performanceprofile/functests/utils/jira/status.go
```diff
@@ -2,12 +2,17 @@ package jira
...
 	"github.com/pkg/errors"
+	"math/rand"
 	"net/http"
 	"net/url"
+	"strconv"
+	"time"
 )
 
 const JIRA_BASE_URL = "https://issues.redhat.com"
+const MAX_RETRIES = 3
```
rshemtov13, line 15: We can add this to a const block with JIRA_BASE_URL

test/e2e/performanceprofile/functests/utils/jira/status.go
```diff
@@ -26,28 +31,72 @@ type JiraIssueStatusResponse = struct {
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create request for Jira status of %s", key)
+		}
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create client to get jira status of %s", key)
+		}
+		defer resp.Body.Close()
+		switch resp.StatusCode {
+		case 200:
```
rshemtov13, line 47: Instead of using 200 here we can use http.StatusOk

test/e2e/performanceprofile/functests/utils/jira/status.go
```diff
@@ -26,28 +31,72 @@ type JiraIssueStatusResponse = struct {
...
+			decoder := json.NewDecoder(resp.Body)
+			err = decoder.Decode(&jiraResponse)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to decode jira status of %s", key)
+			}
+		case 429:
```
rshemtov13, line 53: Same here we can use http.StatusTooManyRequests

rshemtov13 (commented):
Overall looks good to me just these nits

test/e2e/performanceprofile/functests/utils/jira/status.go
mrniranjan: fixed in latest commit

test/e2e/performanceprofile/functests/utils/jira/status.go
mrniranjan: fixed in latest commit

test/e2e/performanceprofile/functests/utils/jira/status.go
```diff
@@ -26,28 +33,72 @@ type JiraIssueStatusResponse = struct {
...
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to decode jira status of %s", key)
+			}
+		case http.StatusTooManyRequests:
+			if retryAfter := resp.Header.Get("Retry-After"); retryAfter != "" {
+				retryTime, err := convertRetryAfterTime(retryAfter)
+				if err != nil {
+					return nil, err
+				}
+				time.Sleep(retryTime)
```
MarSik, line 61: Sequential sleep, it might make the job look like it is stuck I think we should log this so whoever is watching the logs will know why it is waiting.

test/e2e/performanceprofile/functests/utils/jira/status.go
```diff
@@ -26,28 +33,72 @@ type JiraIssueStatusResponse = struct {
...
+			decoder := json.NewDecoder(resp.Body)
+			err = decoder.Decode(&jiraResponse)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to decode jira status of %s", key)
+			}
+		case http.StatusTooManyRequests:
+			if retryAfter := resp.Header.Get("Retry-After"); retryAfter != "" {
+				retryTime, err := convertRetryAfterTime(retryAfter)
+				if err != nil {
+					return nil, err
```
MarSik, line 59: Wrong Retry-After header could just log the error and fallback to exponential backoff. Unless we want to know the parsing broke right away.

test/e2e/performanceprofile/functests/utils/jira/status.go
mrniranjan: fixed in latest commit

test/e2e/performanceprofile/functests/utils/jira/status.go
mrniranjan: fixed in latest commit

mrniranjan:
test ci/prow/e2e-gcp-pao

MarSik (approved):

PR #1316, opened 2025-04-03
PSAP-1655: Make machineConfigLabels-related misconfiguration more visible

jmencak:
NTO Tuned CRs support machine config labels (mcLabels) based matching. This involves finding a MachineConfigPool with machineConfigSelector matching mcLabels and setting a TuneD profile on all nodes that are assigned the found MachineConfigPool.  We do not support configurations, where more than 1 MachineConfigPool matches the mcLabels.  While this configuration is not supported, users often unwittingly configure their clusters in this way.  In this case, NTO issues an error message in the operator logs.

This is often not sufficient, so we make this misconfiguration more visible by making ClusterOperator/node-tuning object Degraded.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -163,6 +163,20 @@ func (c *Controller) numProfilesWithBootcmdlineConflict(profileList []*tunedv1.P
 	return numConflict
 }
 
+// numMCLabelsAcrossMCP returns the total number
+// of Profiles in the internal operator's cache (mcLabelsAcrossMCP)
+// tracked as using machineConfigLabels that match across multiple MCPs.
+func (c *Controller) numMCLabelsAcrossMCP(profileList []*tunedv1.Profile) int {
+	n := 0
+	for _, profile := range profileList {
+		if c.mcLabelsAcrossMCP[profile.Name] {
+			n++
+		}
+	}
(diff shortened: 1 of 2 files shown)


jmencak:
/remove-lifecycle stale
I still believe there is value in this PR.  Any reviews? Thanks.

jmencak:
@MarSik , support engineers are asking for this functionality.  Can we have a review?  Thank you!

pkg/operator/status.go
```diff
@@ -318,6 +332,14 @@ func (c *Controller) computeStatus(tuned *tunedv1.Tuned, conditions []configv1.C
 			degradedCondition.Message = fmt.Sprintf("%v/%v Profiles with bootcmdline conflict", numConflict, len(profileList))
 		}
 
+		numMCLabelsAcrossMCP := c.numMCLabelsAcrossMCP(profileList)
```
MarSik, line 335: Will this report a degraded condition for all profiles when any one reports a conflict?

jmencak: Thank you for looking! The degraded condition will be reported only for the ClusterOperator/node-tuning object.

MarSik (commented):
/lgtm

PR #1318, opened 2025-04-07
PSAP-1767: Bump to k8s 1.32.3

jmencak:
This change bumps k8s to 1.32.3 and also bumps other direct dependencies.


jmencak:
Hi @ffromani , @yanirq .  Sorry for the direct ping.  I'd like to allow sufficient time for adjustments to this PR based on reviewer comments and merge this before the freeze 18th April. Would you have a few minutes to take a look at the changes? Thank you!

go.mod
```diff
@@ -1,57 +1,58 @@
 module github.com/openshift/cluster-node-tuning-operator
 
-go 1.22.0
+go 1.23.0
 
-toolchain go1.22.4
+toolchain go1.23.4
```
Tal-or, line 5: Is go bump is mandatory as part of the k8s bump?

go.mod
```diff
@@ -115,124 +113,131 @@ require (
...
+	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.32.3
+	k8s.io/code-generator => k8s.io/code-generator v0.32.3
+	k8s.io/component-base => k8s.io/component-base v0.32.3
+	k8s.io/component-helpers => k8s.io/component-helpers v0.32.3
+	k8s.io/controller-manager => k8s.io/controller-manager v0.32.3
+	k8s.io/cri-api => k8s.io/cri-api v0.32.3
+	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.32.3
+	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.32.3
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.32.3
+	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.32.3
```
Tal-or, line 227: Did we avoid updating kube-openapi on purpose? (line below)

go.mod
jmencak: Hey Talor thank you for looking!  Good question.  So [you have added](https://github.com/openshift/cluster-node-tuning-operator/commit/4d6cb12e3) the `toolchain` option in NTO 4.17+.  Is it actually needed at all?When removed completely,

```
go mod tidy
go mod vendor
```
doesn't add it back.

When you leave it and do the same thing, it seems to be populated by your current toolchain (in my case old go 1.23.4).  In my view it make no sense to keep the old 1.22.4 as we now require go 1.23 with this PR.  The question is, whether we need the toolchain line at all.

go.mod
jmencak: That's a great eye for detail. Thank you for caching this!  Bumped.

go.mod
Tal-or: now i see that I comment on the wrong line, I was referring to the go bump in general.

if go 1.23 is align with k8s 1.32 then that's super and we're good.

Regarding the toolchain, it was added when I ran:
```
go mod tidy
go mod vendor
```
If now running the same command, removes the toolchain then I'm completely fine to go with `go mod` recommendations

Tal-or:
/lgtm

Thanks

Tal-or (approved):

jmencak:
> /lgtm
> 
> Thanks

Thanks Talor, much appreciated. However, you raised a good point here.  I wonder if we should reconsider actually dropping the `toolchain` option alltogether.  Any opinions, @Tal-or , @yanirq ?

/hold

Tal-or:
> > /lgtm
> > Thanks
> 
> Thanks Talor, much appreciated. However, you raised a good point here. I wonder if we should reconsider actually dropping the `toolchain` option alltogether. Any opinions, @Tal-or , @yanirq ?
> 
> /hold

If 
```
go mod tidy
```
remove that then we can.
 Anyway I added LGTM because it's not critical for this PR, we can reevaluate later if needed.

jmencak:
> > > /lgtm
> > > Thanks
> > 
> > 
> > Thanks Talor, much appreciated. However, you raised a good point here. I wonder if we should reconsider actually dropping the `toolchain` option alltogether. Any opinions, @Tal-or , @yanirq ?
> > /hold
> 
> If
> 
> ```
> go mod tidy
> ```
> 
> remove that then we can. Anyway I added LGTM because it's not critical for this PR, we can reevaluate later if needed.

OK thanks. let's do it later then, perhaps during the next bump.  After studying what it actually does, I'm leaning towards removing it unless you have a specific need for it as you added it.

/unhold

jmencak:
Infra issues
/retest

PR #1321, opened 2025-04-14
OCPBUGS-55063: E2E: Use cpuset.CPUSet for reserved, isolated and offline  fields

mrniranjan:
When adding or removing cpus to reserved, isolated and offline use cpuset.CPUSet struct


ffromani (commented):
/approve
/lgtm

PR #1325, opened 2025-04-30
NO-JIRA: Set minimum TLS 1.3 for metrics and webhook servers

jmencak:
Set the minimum default version to TLS 1.3.  PQ algorithms will only be supported in TLS 1.3+.  Hybrid key agreements for TLS 1.3 X25519MLKEM768 is automatically supported by default in go 1.24.

See: [OCPSTRAT-1858](https://issues.redhat.com//browse/OCPSTRAT-1858)


jmencak:
Infra issues
/retest

PR #1326, opened 2025-05-05
OCPBUGS-56416: E2E: use  4 cpus for  reserved  from available cpus

mrniranjan:
Earlier we were using only cpu as reserved cpu
and the rest for isolated, this causes issues as
one reserved cpu for control plane nodes is not enough and can cause issues.

This PR fixes this issues by using 4 reserved cpus and the rest for isolated

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -1047,9 +1047,13 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				}
 				sort.Ints(coreids)
 				Expect(len(coreids)).ToNot(Equal(0))
-				middleCoreIds := coreids[len(coreids)/2]
-				coresiblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, middleCoreIds)
-				reserved = reserved.Union(coresiblings)
+				middleIndex := len(coreids) / 2
+				// we need at least 4 cpus for reserved cpus
+				middleCoreIds := coreids[middleIndex-2 : middleIndex+2]
+				for _, cores := range middleCoreIds {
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -1047,9 +1047,13 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				}
 				sort.Ints(coreids)
 				Expect(len(coreids)).ToNot(Equal(0))
-				middleCoreIds := coreids[len(coreids)/2]
-				coresiblings := nodes.GetAndRemoveCpuSiblingsFromMap(numaCoreSiblings, middleCoreIds)
-				reserved = reserved.Union(coresiblings)
+				middleIndex := len(coreids) / 2
+				// we need at least 4 cpus for reserved cpus
```
ffromani, line 1051: how do we know we can afford 4 reserved cores and then there will be enough isolated cores?

mrniranjan: We have a condition above to check if we have more than 8 cpus to run the tests:
https://github.com/openshift/cluster-node-tuning-operator/blob/d2c881fb8e2a08516885455f8024c4bcaacf1d5a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go#L1033

ffromani: ok then

ffromani (commented):
/approve
/lgtm

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
MarSik: So the tests from here are skipped in all prow CI runs? Afaik we only have 4 cpus available in Prow, right?

ffromani:
/hold

Let's address questions from @MarSik

ffromani:
/hold cancel

it seems the question has been answered

PR #1328, opened 2025-05-12
latency-tests: support FIPS compliancy

shajmakh:
TBD


ffromani:
the intent fully makes sense, but isn't CGO_ENABLED=1 the default? if so, why do we need to set it explicitely? can be to make it very obvious, but if so let's comment accordingly

Issue #1331, opened 2025-05-30
Not possible to use cluster-node-tuning-operator as a dependency because of its go.mod containing non existing references


miguelsorianod:
Hi,

In our repository we are currently referencing github.com/openshift/cluster-node-tuning-operator as a dependency in our go.mod. Specifically, we have the following commit being specified:
```
github.com/openshift/cluster-node-tuning-operator v0.0.0-20220614214129-2c76314fb3cc
```

We have detected that parts of the Go tooling fail when the github.com/openshift/cluster-node-tuning-operator dependency is specified in our go.mod, and the reason is due to github.com/openshift/cluster-node-tuning-operator having in its go.mod dependencies pointing to git tag references that do not exist. *This occurs even when the go.mod of the github.com/openshift/cluster-node-tuning-operator dependency itself has a replace to an existing git tag*

This can be reproduced by doing the following:
* Create your own git repository
* Initialize go.mod and specify github.com/openshift/cluster-node-tuning-operator as a dependency. This can be done pointing to the same commit that I mentioned above, or even pointing to the commit of the current `main` branch
* Run the `go list -m -json all` command. Observe how it fails with an error indicating that there's an unknown revision of one of the dependencies specified. In case it helps for something, we are using Go 1.24 in our project.

In the case of the dependency we have specified, when we run the `go list -m -json all` command, we get the following error:

```
msoriano@localhost:~/go/src/... (master)(ocm:?)$ go list -m -json all 
go: k8s.io/kubernetes@v0.23.3: invalid version: unknown revision v0.23.3
```

The reason is because the tag v0.23.3 of the k8s.io/kubernetes does not exist. You will notice though that in github.com/openshift/cluster-node-tuning-operator v0.0.0-20220614214129-2c76314fb3cc in its go.mod that's specified, but also a `replace` that points to a correctly existing git tag reference:
```
	k8s.io/kubernetes => k8s.io/kubernetes v1.23.3
```
However, this does still produce the error I mentioned when using some parts of the Go tooling. This is particularly relevant because some IDEs like goland make use of parts of the Go tooling that end up hitting the error.

This prevents us and developers from using the dependency in our project. This impacts anyone that might reference this library as a dependency.

We have also tried to update our dependency to the current commit in the `main` branch, but the same error is reproduced, now pointing to other non existent dependencies git tags. When trying to update the dependency to the `main` branch the following can be observed:

```
[...]

PR #1333, opened 2025-06-02
WIP: Remove dependency on k8s.io/kubernetes

MarSik:
... at the cost of a slightly less sensitive test.

This makes a KubeletConfig test less sensitive, because it does not compare the default values for eviction thresholds anymore. Those thresholds are only part of kubernetes code, however that dependency pulls in a lot of other stuff we do not want.


miguelsorianod:
> Those thresholds are only part of kubernetes code, however that dependency pulls in a lot of other stuff we do not want.

To add a bit more detail, it is not only about that. It is that importing `k8s.io/kuberntees` as a dependency is unsupported as per their own documentation (https://github.com/kubernetes/kubernetes?tab=readme-ov-file#to-start-using-k8s), and importing it causes different kinds of dependencies related issues that have an impact in the Go tooling as reported in https://github.com/openshift/cluster-node-tuning-operator/issues/1331

PR #1335, opened 2025-06-04
OCPBUGS-52853: Do not add iommu.passthrough to ARM by default

MarSik:


yanirq:
/close
in favor of https://github.com/openshift/cluster-node-tuning-operator/pull/1354

PR #1340, opened 2025-06-18
PSAP-1827: Add startup_udev_settle_wait option option to tunedConfig

jmencak:
The new `startup_udev_settle_wait` TuneD global option has recently been added to `/etc/tuned/tuned-main.conf`.  This option should help in situations where there are a lot of udev events during TuneD startup and TuneD is too slow to handle them.

Users may want/need to adjust this period or completely disable this functionality.

Other changes:
  - fix the logic when `reapply_sysctl` is undefined in `/etc/tuned/tuned-main.conf`

Also see: OCPBUGS-56442, OCPBUGS-57470

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,7 +220,9 @@ pao-functests-updating-profile: cluster-label-worker-cnf pao-functests-update-on
 pao-functests-update-only: $(BINDATA)
 	@echo "Cluster Version"
 	hack/show-cluster-version.sh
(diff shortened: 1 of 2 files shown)


jmencak:
/cc @yanirq 
as we talked about this briefly in the past.

jmencak:
The errors from e2e-hypershift-pao could be genuine, this needs looking into.

jmencak:
@yanirq , after a lot of debugging, I figured out why the `e2e-hypershift-pao` tests are failing.  The tuned API used by openshift/hypershift needs to be synchronized with the tuned API in this repository together with CRDs.  Please let me know if want this PR.  I'd take additional steps then.

PR #1341, opened 2025-06-19
[release-4.18] OCPBUGS-57947: CNF-18142: Add kernel page size field [1/3]

rbaturov:
This PR is a backport of #1262 and introduces the following changes:
* KernelPageSize field has been added to performance profile API.
This field defines the kernel page size: for x86/amd64, the only valid value is 4k, while for aarch64, the valid values are 4k, 64k.
The default value is 4k if none is specified.
* Validation has been added
* Modified [kernel Type](https://github.com/openshift/machine-config-operator/blob/f0d60f54b8318a84c4d6e6e9e54319b02bdc4dc9/pkg/controller/common/constants.go#L26-L32) selection process, adding the new kernel type `64k-pages` alongside `default` and `realtime` 
* Documented this

### DIFF
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -39,6 +39,8 @@ const (
 	MCKernelRT = "realtime"
 	// MCKernelDefault is the value of the kernel setting in MachineConfig for the default kernel
 	MCKernelDefault = "default"
+	// MCKernel64kPages is the value of the kernel setting in MachineConfig for 64k-pages kernel on aarch64
+	MCKernel64kPages = "64k-pages"
 	// HighPerformanceRuntime contains the name of the high-performance runtime
 	HighPerformanceRuntime = "high-performance"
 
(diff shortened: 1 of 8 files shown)


yanirq:
cc @MarSik @Tal-or

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -148,8 +150,13 @@ func New(profile *performancev2.PerformanceProfile, opts *components.MachineConf
 		profile.Spec.RealTimeKernel.Enabled != nil &&
 		*profile.Spec.RealTimeKernel.Enabled
 
+	// Real time kernel with 64k-pages for aarch64 not yet supported and rejected in the validation webhook.
 	if enableRTKernel {
 		mc.Spec.KernelType = MCKernelRT
+	} else if profile.Spec.KernelPageSize != nil && *profile.Spec.KernelPageSize == performancev2.KernelPageSize("64k") {
+		// During validation, we ensure that nodes are based on aarch64 when the administrator specifies 64k for this field.
```
MarSik, line 157: Do we call the validator inside the operator again or do we rely on the webhook only? Because the webhook is optional.

rbaturov: We are not calling the validator here. (I'm not aware that we are calling it anywhere except the validation webhook).
We do rely on it.
However, if there was wrong value there (neither 4k or 64k),and the validation webhook is turned off, it will choose 4k.

MarSik: I am not 100% happy about this due to discoverability, but fallback to a valid value is probably good enough for a backport.

MarSik (approved):

yanirq:
/retitle [release-4.18] OCPBUGS-57947: CNF-18142: Add kernel page size field [1/3]

PR #1346, opened 2025-06-24
OCPBUGS-55399: cluster-node-tuning-operator Pod of the hosted cluster is not listening on 60000 port which makes the target to be down

georgelipceanu:
**What this PR does / why we need it**:
This PR changes the NTO to run NTO related metrics on port 60000 by changing the port protocol from HTTPS to HTTP, which allows port 60000 to run in HyperShift without the need for additional permissions to RBAC resources in the kube-system namespace to the HyperShift operator.
**Which issue(s) this PR fixes** 
Fixes #[OCPBUGS-55399](https://issues.redhat.com/browse/OCPBUGS-55399)
Currnetly Blocking [HyperShift PR](https://github.com/openshift/hypershift/pull/6313)


pkg/metrics/server.go
```diff
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	ntoconfig "github.com/openshift/cluster-node-tuning-operator/pkg/config"
```
ffromani: misplaced import

pkg/metrics/server.go
```diff
@@ -47,7 +48,11 @@ func init() {
 // it does so only if the CA bundle changed from the current CA bundle on record.
 func DumpCA(caBundle string) {
 	if caBundle != server.caBundle {
-		server.caBundleCh <- caBundle
+		select {
+		case server.caBundleCh <- caBundle:
+		default:
+			klog.Infof("Metrics server CA channel not ready, skipping CA bundle update")
+		}
```
ffromani, line 56: this is welcome and seems correct, but why is it needed in the context of this PR to change the metrics server?

pkg/metrics/server.go
```diff
@@ -126,6 +131,28 @@ func (Server) Start(ctx context.Context) error {
 // and restarted with the current files.  Every non-nil return from this function is fatal
 // and will restart the whole operator.
 func RunServer(port int, ctx context.Context) error {
+	if ntoconfig.InHyperShift() {
+		// HYPERSHIFT-SPECIFIC PATH: simple HTTP server.
```
ffromani: this is pretty obviously hypershift-specific, the very line before is `if ntoconfig.InHyperShift() ` so this comment should be amended to bring value or deleted

ffromani (commented):
thanks for the contribution. I would like to explore options before to resort to just disable tls and expose the port to everyone. I'm concerned about information leaking. How can we prevent it?

pkg/metrics/server.go
georgelipceanu: Removed that comment there, appreciate the feedback 🙌

pkg/metrics/server.go
georgelipceanu: I'm just after retesting this without the change and it seems to work without it. When trying out fixes for this bug, I was of the belief that the NTO was blocked on this line as it would be trying to receive a response that would never come due to the switch to HTTP but it seems that it works just fine without it. Either way, I'm willing to either keep this block in this PR or open a new one with just this little piece of code. Nice catch 🙌

pkg/metrics/server.go
georgelipceanu: Might be a dumb question but where can I look to find how imports should be arranged in this repo? I know that in the `.golangci.yaml` in HyperShift has a specific section on how imports are linted, as follows:
```
linters-settings:
  gci:
    sections:
      - standard
      - dot
      - prefix(github.com/openshift/hypershift)
      - prefix(github.com/openshift)
      - prefix(github.com/aws)
      - prefix(github.com/Azure)
      - prefix(k8s.io)
      - prefix(sigs.k8s.io)
      - default
    custom-order: true
```
Would this repo also be the same, or perhaps something similar?

pkg/metrics/server.go
georgelipceanu: > I would like to explore options before to resort to just disable tls and expose the port to everyone. I'm concerned about information leaking. How can we prevent it?

Regarding the switch to HTTP, from what I've found and experimented with, I found that the NTO needed RBAC permissions to get the `ConfigMap` from the hosted cluster's `kube-system` namespace while the Control-Plane-Operator in the hosted cluster, responsible for deploying the operator, did not have permission to create RBAC resources in the hosted cluster's kube-system namespace.

I personally wasn't able to find a way to keep HTTPS with the current configuration but I would be willing to take a deeper dive into it HTTP won't be secure enough.

pkg/metrics/server.go
ffromani: you didn't find because there is not, it was a convention enforced in review. You mentioned also a very good way for us to automate this check, we will implement. Thanks!

georgelipceanu: Made the import changes to align with the linting rules I commented.

pkg/metrics/server.go
jmencak: Thank you for the PR, IIRC, we had two issues with the metrics server reported to us by the Security team.
* not using https://
* [authentication was required to use https:// endpoint](https://issues.redhat.com/browse/OCPBUGS-4647)

I'd be very reluctant to merge this without meeting both requirements.

georgelipceanu: Makes sense, in that case I'll keep looking into keeping HTTPS with the metrics server for HyperShift. However if the new `DumpCA()` code looks good then I would be willing to repurpose this PR for this code only. Is this something that the team would be interested in?

PR #1349, opened 2025-06-26
autosize control plane support for performance profile creator

ffromani:
Implement reserved cpu (aka infra+control plane) sizing using a the linear programming optimization (gonum/optimize).

The core idea is to model the constraints and let the optimization package compute the desired target.

These changes where AI-Assisted (hence the AA tag), then largely amended by a human (hence the HI tag - Human Intervention).

The initial penalty cost structure was suggested by google Gemini 2.5 flash, and then amended by human intervention.

### DIFF
--- a/pkg/performanceprofile/profilecreator/autosize/autosize.go
+++ b/pkg/performanceprofile/profilecreator/autosize/autosize.go
@@ -31,6 +31,12 @@ import (
 // We want to maximize x_w, or, equivalently, minimize x_c
 
 const (
+	// x86 limit. 256 hardware entries, of those 32 reserved. 256-32 = 224.
+	// see: https://en.wikipedia.org/wiki/Interrupt_request
+	maxIRQsPerPhysicalCore int = 224
+)
+
+const (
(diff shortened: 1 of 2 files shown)


MarSik:
This is an interesting approach, I wonder if we can express constraints like allocate the whole (multiples) L3 CCDs for reserved or integrate the NIC queue count (no more than 16/32) and interrupt counts (224 per cpu).

pkg/performanceprofile/profilecreator/autosize/autosize.go
```diff
@@ -0,0 +1,233 @@
...
+	"math"
+
+	"gonum.org/v1/gonum/optimize"
+
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/profilecreator"
+)
+
+// Assumptions:
+// 1. All the machines in the node pool have identical HW specs and need identical sizing.
+// 2. We cannot distinguyish betwee infra/OS CPU requirements and control plane CPU requirement.
```
ffromani: typo: distinguish

PR #1350, opened 2025-06-26
script: add IRQBALANCE_BANNED_CPULIST variable

Tal-or:
A new version of irqbalance start using IRQBALANCE_BANNED_CPULIST to mark the banned CPUs
that are not suppose to participate in the irq balancing.

We're setting the new variable with a default value `-` which
means that all CPUs are participating in the irq balancing.


assets/performanceprofile/configs/99-runtimes.conf
```diff
@@ -12,3 +12,6 @@ infra_ctr_cpuset = "{{.ReservedCpus}}"
 [crio.runtime.runtimes.high-performance]
 inherit_default_runtime = true
 allowed_annotations = ["cpu-load-balancing.crio.io", "cpu-quota.crio.io", "irq-load-balancing.crio.io", "cpu-c-states.crio.io", "cpu-freq-governor.crio.io"{{ if .CrioSharedCPUsAnnotation }}{{ printf ", %q" .CrioSharedCPUsAnnotation}}{{end}}]
+
+[crio.runtime]
+irqbalance_config_restore_file="disable"
```
jmencak: Missing `\n`

assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
```diff
@@ -11,15 +11,16 @@ NONE=0
 
 [ ! -f "${IRQBALANCE_CONF}" ] && exit 0
 
-${SED} -i '/^\s*IRQBALANCE_BANNED_CPUS\b/d' "${IRQBALANCE_CONF}" || exit 0
+${SED} -i '/^\s*IRQBALANCE_BANNED_CPULIST\b/d' "${IRQBALANCE_CONF}" || exit 0
 # CPU numbers which have their corresponding bits set to one in this mask
 # will not have any irq's assigned to them on rebalance.
 # so zero means all cpus are participating in load balancing.
```
jmencak: This comment needs to be adjusted to reflect the change to `IRQBALANCE_BANNED_CPULIST`.  For example "so zero means all cpus are participating in load balancing" is no longer true.

assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
```diff
@@ -11,15 +11,16 @@ NONE=0
...
+# '-' is a special value that means all cpus are participating in load balancing.
```
jmencak: The value `-` is an edge case of _number1-number2_ where both _number1_ and _number2_ are not defined.  While `-` works, I wonder if it would make sense to try and document this in irqbalance upstream.  By doing that, we'd get an official confirmation this is the intended behaviour and that it will not change in the future.

assets/performanceprofile/configs/99-runtimes.conf
Tal-or: Fixed. Thanks

assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
Tal-or: Fixed. Thanks

Tal-or:
/tests won't pass until we'll have https://github.com/cri-o/cri-o/pull/8931 in

PR #1351, opened 2025-06-29
NO-JIRA:chore:e2e: replace Getenv with LookupEnv

Tal-or:
It's usually preferred to use `LookupEnv` because it allows to distinguish between empty value and unset value.

The reason why this sudden change is only because
now we have AI tools that make these refactoring
tasks easier and faster.


Tal-or:
> Thank you for the change. /lgtm /approve
> 
> Out of curiosity, which AI tool did you use for this?

Curser with claude-4-sonnet as the LLM

jmencak:
:= vs = , nice one. 
/lgtm

PR #1353, opened 2025-07-01
NO-JIRA: Add support for running binaries from NVIDIA driver directory

jmencak:
This PR adds support for running binaries from NVIDIA driver directory mounted by NVIDIA GPU operator to `/run/nvidia/driver/usr/bin`.  This is in support of the new TuneD functionality requested by [RHEL-88914](https://issues.redhat.com//browse/RHEL-88914) and supports both one-shot and daemon mode TuneD invocations.

### DIFF
--- a/pkg/util/os.go
+++ b/pkg/util/os.go
@@ -1,7 +1,13 @@
 package util
 
 import (
+	"fmt"
 	"os"
+	"strings"
+)
+
+const (
+	pathSeparator = string(os.PathListSeparator)
 )
 
 // Delete a file if it exists.  Returns nil if 'file' does not exist.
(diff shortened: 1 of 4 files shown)


ffromani (commented):
/lgtm

pkg/util/os.go
```diff
@@ -25,3 +27,22 @@ func Symlink(target, linkName string) error {
 
 	return nil
 }
+
+// AddToPath adds 'path' to PATH environment variable unless it exists.
+// Returns nil unless os.Setenv() call fails.
+func AddToPath(path string) error {
```
ffromani, line 37: I'd add a few unit tests for this function

jmencak: Added as suggested. I'm personally not into unit tests too much, but in this case it made me think about special cases such as `PATH=`, `PATH=/somepath1:` or `PATH=/somepath1::/somepath2`, so thank you for this.

PR #1356, opened 2025-07-03
Remove dependency on k8s.io/kubernetes

ffromani:
Alternative take to #1333 

We can accept a less precise test, because now the test not compare the default values for eviction thresholds anymore. Those thresholds are only part of kubernetes code, however that dependency pulls in a lot of other stuff we do not want.

Or, we can pseudo-vendor just the tiny bit we care about.


ffromani:
/cc @jmencak 

any concerns or objections? getting rid of the `k8s.io/kubernetes` dep would be hugely beneficial, the only real open point is how.

ffromani:
> I like this as it doesn't hard-code the 4 values in `kubeletconfig_test.go`. I presume that `hack/update-k8s-eviction-defaults.sh` should be run (and adjusted the current k8s version) every time we bump k8s.

This, plus we may need to followup code reorganization changes done in k8s upstream. This seems relatively unlikely, but still possible.

Unfortunately, the defaults are not exported by officially supported libraries, so we have to play some catch up one way or the other, and this seems a possible middle ground.

> If so, I wonder whether we should perhaps document that fact in the script. An alternative would be adding this to the Makefile in some way, i.e. target to run after vendor dependencies bump so that the people doing the bump do not forget to do this.

I can add comments and that's no bother. We don't seem to have automation/helpers to bump vendored deps (and this does not surprise me, is it job pretty well handled by native goo tooling)

hack/update-k8s-eviction-defaults.sh
Tal-or, line 1: Looks familiar ;)

pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig_test.go
```diff
@@ -169,10 +170,15 @@ var _ = Describe("Kubelet Config", func() {
 
 			manifest := string(y)
 
-			memoryAvaialable := "memory.available: " + eviction.DefaultEvictionHard[evictionHardMemoryAvailable]
-			nodefsAvailable := "nodefs.available: " + eviction.DefaultEvictionHard[evictionHardNodefsAvaialble]
-			imagefsAvailable := "imagefs.available: " + eviction.DefaultEvictionHard[evictionHardImagefsAvailable]
-			nodefsInodesFree := "nodefs.inodesFree: " + eviction.DefaultEvictionHard[evictionHardNodefsInodesFree]
+			// When this test fails, compare whether the defaults in kubernetes documentation
```
Tal-or, line 173: I would add a sentence about the existance of the `update-k8s-eviction-defaults.sh` script that might help when this test fails.

Tal-or (commented):
Overall LGTM
A cheap price to pay for getting rid of more than half a million lines of code dependencies.

pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig_test.go
ffromani: good idea, adding

hack/update-k8s-eviction-defaults.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# *********************************************************************************
+# *** possibly update and run this script each time we update the k8s libraries ***
+# *********************************************************************************
+
+# this script pseudo-vendors kubernetes source code in order to let our
```
jmencak, line 7: nit: s/this/This/

hack/update-k8s-eviction-defaults.sh
```diff
@@ -0,0 +1,32 @@
...
+# code access the API defaults for eviction values.
+# We currently use these values just and only in unit tests
+# we cannot just vendor `k8s.io/kuberntes` because using that as library
+# is strongly discouraged and not supported by k8s.
+# Defaults are part of the API (TODO(fromani): citation needed) so they are very
+# unlikely to change.
+# Should the code of 8s be updated and the defaults reorganized (e.g.
+# k8s moves to arch-independent defaults) this script needs to be updated
```
jmencak, line 15: Nit: s/updated/updated./

jmencak (approved):
Thank you for the changes.
Found only non-blocking nits.
/lgtm

ffromani:
thanks @jmencak , I will fix the nits if I need to resubmit

PR #1359, opened 2025-07-07
performance profile: internal cleanup

ffromani:
internal cleanups, code movement and minimal refactoring in preparation for the autosizing feature (https://github.com/openshift/cluster-node-tuning-operator/pull/1349)

No intended changes in behavior.

The switch to the `Alert` mechanism is mostly meant to remove our remaining dependency from `logrus`.


jmencak:
Thank you for the PR. Any vendored dependency removal is welcome.  The changes look fine to me, even at the expense of issuing an extra "\n" in `dumpClusterInfo()`.

/lgtm
from my perspective, not adding hold for other reviewers to comment -- using the (in)valid Jira reference for that.

ffromani:
> Thank you for the PR. Any vendored dependency removal is welcome. The changes look fine to me, even at the expense of issuing an extra "\n" in `dumpClusterInfo()`.
> 
> /lgtm from my perspective, not adding hold for other reviewers to comment -- using the (in)valid Jira reference for that.

Thanks Jiri! Be aware that my https://github.com/openshift/cluster-node-tuning-operator/pull/1349 wants to add back another dependency though. But the balance should still be positive.

ffromani:
/hold cancel

I'll own and fix any regression. I asked previously and the output text format is not part of the API/stability guarantee, so we should be good

PR #1369, opened 2025-07-28
WIP: CNF-16642: Dynamic memory enforcement

MarSik:
This series adds a systemd service to the tuned system. This service runs in the cpu manager-like reconcile loop fashion, monitors all pod cgroups' memory and sets the limits to kubepods-burstable and kubepods-besteffort to protect the system from runaway pods without limits of their own.

kubepods-burstable is set to = node allocatable memory - sum(all guaranteed containers memory limit)
kubepods-besteffort is set to = kubepods-burstable limit - kubepods-bustable current consumption

Note: The boilerplate code was generated by the claude-4-sonnet AI model as interpreted by the Cursor IDE. The python service is my own.

### DIFF
--- a/assets/performanceprofile/scripts/dynamic-memory-enforcement.py
+++ b/assets/performanceprofile/scripts/dynamic-memory-enforcement.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+# Dynamic Memory Enforcement Service
+# This service is used to protect the cluster from memory exhaustion.
+# It is used to set the memory.max value for the kubepods-burstable.slice and kubepods-besteffort.slice.
+
+from datetime import datetime
+import os
+import os.path
+
+ROOT="/sys/fs/cgroup/kubepods.slice"
+ROOT_MAX=os.path.join(ROOT, "memory.max")
+
+BESTEFFORT_MAX=os.path.join(ROOT, "kubepods-besteffort.slice", "memory.max")
+
+BURSTABLE_CURRENT=os.path.join(ROOT, "kubepods-burstable.slice", "memory.current")
+BURSTABLE_MAX=os.path.join(ROOT, "kubepods-burstable.slice", "memory.max")
+
+def guaranteed_max():
+    sum = 0
+    
+    for d in os.listdir(ROOT):
+        if not d.startswith("kubepods-pod"):
+            continue
+        try:
+            sum += int(open(os.path.join(ROOT, d, "memory.max")).read().strip())
+        except IOError as e:
+            continue
+
+    return sum
+
+if __name__ == "__main__":
+    root_max = int(open(ROOT_MAX).read().strip())
+    burstable_current = int(open(BURSTABLE_CURRENT).read().strip())
+    guaranteed_current = guaranteed_max()
+
+    burstable_max = root_max - guaranteed_current
+    besteffort_max = burstable_max - burstable_current
+
+    now = datetime.now().isoformat(' ', timespec='microseconds')
+    print(f"\n{now}\n")
+    print(f"root: {root_max}\nburstable: {burstable_current}\nguaranteed: {guaranteed_current}\n\nburstable max: {burstable_max}\nbest-effort max: {besteffort_max}\n")
+    
+    with open(BURSTABLE_MAX, "w") as f:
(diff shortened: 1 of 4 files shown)


assets/performanceprofile/scripts/dynamic-memory-enforcement.py
```diff
@@ -0,0 +1,48 @@
...
+# Dynamic Memory Enforcement Service
+# This service is used to protect the cluster from memory exhaustion.
+# It is used to set the memory.max value for the kubepods-burstable.slice and kubepods-besteffort.slice.
+
+from datetime import datetime
+import os
+import os.path
+
+ROOT="/sys/fs/cgroup/kubepods.slice"
+ROOT_MAX=os.path.join(ROOT, "memory.max")
```
Tal-or, line 11: The base line or the entrypoint from which we are performing all the calculation is `/sys/fs/cgroup/kubepods.slice/memory.max` 

Does this value has already the memory system reserved subtract from?

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -315,8 +315,9 @@ func getIgnitionConfig(profile *performancev2.PerformanceProfile, opts *componen
 		addContent(ignitionConfig, content, dst, &mode)
 	}
 
-	// Add dynamic memory enforcement service only if annotation is enabled
-	if profilecomponent.IsEnforceReservedMemoryEnabled(profile) {
+	// Add dynamic memory enforcement service only if annotation is enabled and workload partitioning is enabled
+	clusterHasWorkloadPartitioning := opts.PinningMode != nil && *opts.PinningMode == apiconfigv1.CPUPartitioningAllNodes
+	if profilecomponent.IsEnforceReservedMemoryEnabled(profile) && clusterHasWorkloadPartitioning {
```
Tal-or, line 321: I think we should also check that MemoryManager configure to `Static` which is suppose to be obvious if the topologyManager configured to `single-numa-node` or `restricted` but we're not checking that in this flow

MarSik: Why? This has nothing to do with NUMA or logical memory tracking. The code enforces the runtime values only.

assets/performanceprofile/scripts/dynamic-memory-enforcement.py
MarSik: msivak@msivak-thinkpadx1carbongen12 ~> oc describe node testnode | grep -i 'memory:'
  memory:                                  130895396Ki # capacity
  memory:                                  129768996Ki # allocatable

sh-5.1# cat /sys/fs/cgroup/kubepods.slice/memory.max 
132988309504
msivak@msivak-thinkpadx1carbongen12 ~> expr 132988309504 / 1024
129871396

I would say: Yes, it already does.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
Tal-or: But the whole point was to avoid consuming the reserved memory used for the house keeping tasks, right?

Then if we won't have topologyManager configured to `single-numa-node`/`restricted` we won't have the memory manager enabled which required configuring reserved memory for the system.

Tal-or: On second thought, in the worse case the script will work but will do nothing useful

assets/performanceprofile/scripts/dynamic-memory-enforcement.py
```diff
@@ -0,0 +1,48 @@
...
+            continue
+
+    return sum
+
+if __name__ == "__main__":
+    root_max = int(open(ROOT_MAX).read().strip())
+    burstable_current = int(open(BURSTABLE_CURRENT).read().strip())
+    guaranteed_current = guaranteed_max()
+
+    burstable_max = root_max - guaranteed_current
```
MarSik, line 36: I need to double check that memory reserved for workload partitioning is subtracted here as well.

PR #1371, opened 2025-08-07
Address false reports of bootcmdline conflicts

jmencak:
This significantly reduces false reports of bootcmdline conflicts in operator logs and ClusterOperator/node-tuning object.

This is not a bullet-proof solution.  A bullet-proof solution would involve adding a new annotation such as
`tuned.openshift.io/lastProfileObservedGen` to the Node tied to the Profile and comparing it to the Profile's generation prior to updating a MachineConfig.  However, we want to avoid putting extra load on the API server as much as possible and this simpler solution already significantly reduces false reports of bootcmdline conflicts.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -731,6 +731,7 @@ func (c *Controller) syncProfile(tuned *tunedv1.Tuned, nodeName string) error {
 	profile.Spec.Config.ProviderName = providerName
 	profile.Spec.Profile = computed.AllProfiles
 	profile.Status.Conditions = tunedpkg.InitializeStatusConditions()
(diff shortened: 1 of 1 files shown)


jmencak:
/cc @liqcui 
as he initially reported this.

jmencak:
ci/prow/e2e-gcp-pao and ci/prow/e2e-hypershift-pao will keep failing until https://github.com/openshift/cluster-node-tuning-operator/pull/1374 merges.

liqcui:
verified  in a small size cluster, did several testing, no issue, the issue reproduced in the same small size worker nodes using 4.20 nightly version. the new pr did fix the issue

PR #1373, opened 2025-08-12
OCPBUGS-60461: E2E: llc test fix:  Get updated Performance profile before modifying

mrniranjan:
Get latest performance Profile object before modifying. 
In latest downstream tests we saw that test case related to Disabling the annotation "prefer-align-cpus-by-uncorecache"  did not propagate. This PR fixes by fetching latest performance profile object before modifying the profile.

### DIFF
--- a/test/e2e/performanceprofile/functests/13_llc/llc.go
+++ b/test/e2e/performanceprofile/functests/13_llc/llc.go
@@ -221,6 +221,9 @@ var _ = Describe("[rfe_id:77446] LLC-aware cpu pinning", Label(string(label.Open
 		When("align-cpus-by-uncorecache annotations is removed", func() {
 			It("[test_id:77723] should disable align-cpus-by-uncorecache cpumanager policy option", func() {
 				ctx := context.Background()
+				// Get latest profile
+				perfProfile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/13_llc/llc.go
```diff
@@ -234,12 +237,23 @@ var _ = Describe("[rfe_id:77446] LLC-aware cpu pinning", Label(string(label.Open
...
-				for _, node := range workerRTNodes {
-					kubeletconfig, err := nodes.GetKubeletConfig(ctx, &node)
-					Expect(err).ToNot(HaveOccurred())
-					Expect(kubeletconfig.CPUManagerPolicyOptions).ToNot(HaveKey("prefer-align-cpus-by-uncorecache"))
-				}
+				// Use Eventually in case of any delays in updating kubelet
+				Eventually(func() bool {
+					for _, node := range workerRTNodes {
+						kubeletconfig, err := nodes.GetKubeletConfig(ctx, &node)
+						if err != nil {
```
Tal-or: I think that here we should abort (`Expect(err...`. 
If we're failing the get kubelet from the API server, it means wer'e in a pretty bad shape

mrniranjan: fixed in latest commit

MarSik (approved):
/lgtm

PR #1377, opened 2025-08-13
OCPBUGS-60258: Return 401 Unauthorized without client cert

jmencak:
This PR fixes the behaviour of the metrics server to return "401 Unauthorized" when no or an invalid (e.g. expired) client certificate is passed to the metrics server.

Resolves: OCPBUGS-60258

### DIFF
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -66,8 +66,11 @@ func buildServer(port int, caBundle string) *http.Server {
 
 	tlsConfig := &tls.Config{}
 	caCertPool := x509.NewCertPool()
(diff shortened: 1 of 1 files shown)


jmencak:
What I've tested:
* metrics are still visible in the OpenShift console
* requests to the metrics server with a valid certificate return valid metrics as expected.
* requests to the metrics server without a valid certificate no longer return TLS errors. They return regular "401 Unauthorized" responses.
* requests to the metrics server with an invalid certificate no longer return TLS errors. They return regular "401 Unauthorized" responses.

PR #1378, opened 2025-08-13
OCPBUGS-60573: E2E: RPS: Avoid multiple profile updates 

mrniranjan:
Update Performance profile once and update the mcp once instead of Adding annotation and modify reserved and isolated cpus

Also Update profile only when Profile has enable-rps annotation


Tal-or (approved):
/lgtm

MarSik (approved):
/lgtm

PR #1384, opened 2025-08-27
OCPBUGS-41219: Only enable timer.migration on RT kernel

MarSik:
High resolution kernel timers are broken on non-RT kernel when timer migration is enabled. This causes high latency     results when running cyclictest on non-RT kernel like some users do.


assets/performanceprofile/tuned/openshift-node-performance-rt
```diff
@@ -6,3 +6,11 @@ summary=Real time profile to override unsupported settings
 #The openshift-node-performance profile inherits these kernel parameters from the network-latency profile. 
 #Therefore, if the real time kernel is detected they will be dropped, meaning won't be applied.
 drop=kernel.numa_balancing,net.core.busy_read,net.core.busy_poll
+
+{{if .RealTimeHint}}
+# cpu-partitioning and RealTimeHint for RHEL disable it (= 0)
+# OCP is too dynamic when partitioning and needs to evacuate
+# scheduled timers when starting a guaranteed workload (= 1)
+# But only when kernel-rt i in use or HRTIMERs break
```
bartwensley: typo

MarSik: Indeed. Fixed.

PR #1386, opened 2025-08-29
Fixed oslat & cyclictest failure due to missing max latency value

SargunNarula:
The oslat & cyclictest tests were getting skipped in the 4_latency test execution due to missing max_latency parameter.

### DIFF
--- a/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
+++ b/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
@@ -284,14 +284,14 @@ func getNegativeTests(toolToTest string) []latencyTest {
 		testSet = append(testSet, latencyTest{testRuntime: "2", oslatMaxLatency: fmt.Sprint(math.MaxInt32 + 1), outputMsgs: []string{invalidNumberOslatMaxLatency, fail}, toolToTest: toolToTest})
 		testSet = append(testSet, latencyTest{testRuntime: "2", oslatMaxLatency: "-3", outputMsgs: []string{invalidNumberOslatMaxLatency, fail}, toolToTest: toolToTest})
 		// Covers the scenario of BZ 2094046
-		testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "2", outputMsgs: []string{unexpectedError, fail}, toolToTest: toolToTest})
+		testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "2", oslatMaxLatency: untunedLatencyThreshold, outputMsgs: []string{unexpectedError, fail}, toolToTest: toolToTest})
 	}
 	if toolToTest == cyclictest {
 		testSet = append(testSet, latencyTest{testRuntime: "2", cyclictestMaxLatency: "&", outputMsgs: []string{incorrectCyclictestMaxLatency, fail}, toolToTest: toolToTest})
 		testSet = append(testSet, latencyTest{testRuntime: "2", cyclictestMaxLatency: fmt.Sprint(math.MaxInt32 + 1), outputMsgs: []string{invalidNumberCyclictestMaxLatency, fail}, toolToTest: toolToTest})
 		testSet = append(testSet, latencyTest{testRuntime: "2", cyclictestMaxLatency: "-3", outputMsgs: []string{invalidNumberCyclictestMaxLatency, fail}, toolToTest: toolToTest})
 		// Covers the scenario of BZ 2094046
-		testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "2", outputMsgs: []string{unexpectedError, fail}, toolToTest: toolToTest})
+		testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "2", cyclictestMaxLatency: untunedLatencyThreshold, outputMsgs: []string{unexpectedError, fail}, toolToTest: toolToTest})
 	}
 	if toolToTest == hwlatdetect {
 		testSet = append(testSet, latencyTest{testRuntime: "2", hwlatdetectMaxLatency: "&", outputMsgs: []string{incorrectHwlatdetectMaxLatency, fail}, toolToTest: toolToTest})


test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -219,6 +235,48 @@ func clearEnv() {
...
+	}
+
+	// Fallback: first online CPU
+	if cpuID == -1 {
+		set, err := nodes.GetOnlineCPUsSet(ctx, workerNode)
+		if err != nil || set.Size() == 0 {
+			return false, fmt.Errorf("get online CPUs: %w", err)
+		}
+		cpuID = set.List()[0]
+	}
```
shajmakh: by design of PP, it is an incorrect configuration not to have isolated CPUs, so this consition is unreachable

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -219,6 +235,48 @@ func clearEnv() {
 	os.Unsetenv(latencyTestCpus)
 }
 
+// checkHyperthreading determines if the BZ 2094046 test should run based on HT status.
+// Returns true if the test should run (HT enabled), false if it should be skipped (HT disabled).
+func checkHyperthreading() (bool, error) {
```
shajmakh: this sounds a good function that fits more under utils packages

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -107,6 +113,16 @@ var _ = DescribeTable("Test latency measurement tools tests", func(testGroup []l
...
+			// Special handling for BZ 2094046 test case - check if it should be skipped based on Hyper Threading status
+			if test.testCpus == "2" {
+				htEnabled, htErr := checkHyperthreading()
+				if htErr != nil {
+					testlog.Warningf("Failed to check hyperthreading status, running test anyway: %v", htErr)
+				} else if !htEnabled {
+					testlog.Infof("Skipping BZ 2094046 %s test case - hyperthreading is disabled, test would pass unexpectedly", test.toolToTest)
+					continue
+				}
+			}
```
shajmakh: I don't recall this applies to hwlatdetect. However, I would move this filter to an earlier phase where the  specific test of the bug is added to the list of tests (https://github.com/SargunNarula/cluster-node-tuning-operator/blob/ab260875109b2688935d2511ee153289fb5a911e/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go#L352 & https://github.com/SargunNarula/cluster-node-tuning-operator/blob/ab260875109b2688935d2511ee153289fb5a911e/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go#L345) , if the condition is met, then we avoid adding the test object to the set.

shajmakh (commented):
Thanks for catching and addressing this. 
I left a few comments. 
/approve

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
SargunNarula: Thanks for catching, Removed the un-reachable fallback in the latest commit.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
SargunNarula: I think this specific checkHyperthreading function fixes a particular node and CPU to check hyperthreading on, which overlaps with the actual check in [IsHyperThreadingEnabled()](https://github.com/openshift/cluster-node-tuning-operator/pull/1386/files#diff-2b8f0be1f17f99ddfbb871fc0cc460e1e21570e01b80ff508e3c73756ce81116R279). It might be better to keep this logic within the test code itself.

shajmakh (commented):
the updates lgtm. regarding the commit (and PR title), I think it should highlight that this fix is derived from the fact that there might be different HT configurations: with HT enabled and without, rather than having the max latency missing or not.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
SargunNarula: Moved the filtering to testset phase.

shajmakh:
LGTM
let's please confirm the tests pass on both configurations (HT enabled and disabled) before merging this

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -219,6 +225,33 @@ func clearEnv() {
...
+	profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+	if err != nil {
+		return false, fmt.Errorf("get performance profile: %w", err)
+	}
+
+	workerNodes, err := nodes.GetByLabels(testutils.NodeSelectorLabels)
+	if err != nil {
+		return false, fmt.Errorf("get worker nodes: %w", err)
+	}
+	workerNode := &workerNodes[0]
```
ffromani: why do we need to pick a random node which matches the labels? can't we just pick the node by name?

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -219,6 +225,33 @@ func clearEnv() {
...
+
+	ctx := context.Background()
+	set, err := cpuset.Parse(string(*profile.Spec.CPU.Isolated))
+	if err != nil || set.Size() == 0 {
+		return false, fmt.Errorf("failed to parse isolated CPUs from profile")
+	}
+	cpuID := set.List()[0]
+
+	isHTEnabled := nodes.IsHyperthreadingEnabled(ctx, cpuID, workerNode)
```
ffromani: I'd check the node settings (possibly `/proc/cmdline`) or actually any random CPU. To put it differently, why the first isolated CPU is significant and why is it better than, say, cpu#0 ?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -274,6 +274,13 @@ func GetSMTLevel(ctx context.Context, cpuID int, node *corev1.Node) int {
 	return cpus.Size()
 }
 
+// IsHyperthreadingEnabled checks if hyperthreading (SMT) is enabled on the node
+// Returns true if SMT level > 1, false otherwise
+func IsHyperthreadingEnabled(ctx context.Context, cpuID int, node *corev1.Node) bool {
+	smtLevel := GetSMTLevel(ctx, cpuID, node)
+	return smtLevel > 1
```
ffromani: I'd just inline `GetSMTLevel` in the one and only calling site

ffromani (commented):
conceptually OK, but questions about the implementation

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
SargunNarula: By specifying index 0, we fix the node among those that have the appropriate labels. To ensures that if a performance profile has applied any kernel argument, such as nosmt, we can verify it through an actual runtime check.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
SargunNarula: There is no particular significance in choosing the first isolated CPU. An ID was simply required to perform the check, so selected one from the isolated set. Do you suggest checking any random cpu ?

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
SargunNarula: Resolved, with latest commit.

SargunNarula:
> 
LGTM
let's please confirm the tests pass on both configurations (HT enabled and disabled) before merging this


@shajmakh I can now confirm the tests pass on both HT enabled and disabled environments. More specifically pass on HT enabled and gets skipped on HT-disabled ones.

shajmakh:
/lgtm
Thanks for the updates, I'll leave room for oter reviewers if they still have comments
/hold

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
ffromani: ok, but I still don't follow why we need to use `the node selector labels vs picking a specific node and checking that node

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -219,6 +225,37 @@ func clearEnv() {
...
+		return false, fmt.Errorf("get worker nodes: %w", err)
+	}
+	workerNode := &workerNodes[0]
+
+	ctx := context.Background()
+	set, err := cpuset.Parse(string(*profile.Spec.CPU.Isolated))
+	if err != nil || set.Size() == 0 {
+		return false, fmt.Errorf("failed to parse isolated CPUs from profile")
+	}
+	cpuID := set.List()[0]
```
ffromani: I still don't get why this code is better than just checking cpuID 0 (which is much simpler) or the kernel command line arguments (`/proc/cmdline`)

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
SargunNarula: Majorly because using NodeSelectorLabels guarantees we're checking HT status on a node that has the matching profile configuration.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
SargunNarula: Resolved this by introducing SMTActive function.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -278,6 +279,26 @@ func GetSMTLevel(ctx context.Context, cpuID int, node *corev1.Node) (int, error)
...
+	s := strings.TrimSpace(testutils.ToString(out))
+	switch s {
+	case "1":
+		return true, nil
+	case "0":
+		return false, nil
+	default:
+		return false, fmt.Errorf("unexpected %s value: %q", sysDevicesSMTActive, s)
+	}
+}
```
ffromani, line 300: for the future: use `strconv.ParseBool`

PR #1392, opened 2025-09-19
OCPBUGS-61986: E2E: check kernel.timer_migration is not enabled when RealTimeKernel is false

mrniranjan:
Modify test case to check kernel.timer_migration is not enabled when rt kernel is false

This change was done as part of PR #1384

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -348,6 +348,8 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 
 	Context("Tuned kernel parameters", Label(string(label.Tier0)), func() {
 		It("[test_id:28466][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should contain configuration injected through openshift-node-performance profile", func() {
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -348,20 +348,25 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+			profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
+			Expect(err).ToNot(HaveOccurred(), "Unable to fetch latest profile")
 			sysctlMap := map[string]string{
 				"kernel.hung_task_timeout_secs": "600",
 				"kernel.nmi_watchdog":           "0",
 				"kernel.sched_rt_runtime_us":    "-1",
 				"vm.stat_interval":              "10",
 				"kernel.timer_migration":        "1",
 			}
+			if !*profile.Spec.RealTimeKernel.Enabled {
```
MarSik, line 360: This checks the perf profile config, should we check the actual kernel installed? In theory both are the same, but is it possible some race will cause a mismatch? Or maybe the PR that made sure we wait for the perf profile update makes sure we are always in sync.

MarSik (approved):
/lgtm

PR #1399, opened 2025-09-26
OCPBUGS-62605: e2e: refactor GetSMTLevel to remove Gomega assertions

SargunNarula:
The GetSMTLevel util function previously used gomega assertions which could cause panics when called outside of Ginkgo It blocks.

This PR addresses:
- Modified the GetSMTLevel signature to return (int, error)
- Replaced ExpectWithOffset assertions with standard error handling
- Updated all function refrences to handle the error return

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -109,7 +109,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		onlineCPUSet, err = nodes.GetOnlineCPUsSet(ctx, workerRTNode)
 		Expect(err).ToNot(HaveOccurred())
 		cpuID := onlineCPUSet.UnsortedList()[0]
-		smtLevel = nodes.GetSMTLevel(ctx, cpuID, workerRTNode)
+		smtLevel, err = nodes.GetSMTLevel(ctx, cpuID, workerRTNode)
+		Expect(err).ToNot(HaveOccurred(), "Unable to fetch SMT level on node %s, Error: %v", workerRTNode.Name, err)
 		getter, err = cgroup.BuildGetter(ctx, testclient.DataPlaneClient, testclient.K8sClient)
 		Expect(err).ToNot(HaveOccurred())
 		cgroupV2, err = cgroup.IsVersion2(ctx, testclient.DataPlaneClient)
@@ -251,7 +252,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 
 		DescribeTable("Verify CPU usage by stress PODs", func(ctx context.Context, guaranteed bool) {
 			cpuID := onlineCPUSet.UnsortedList()[0]
-			smtLevel := nodes.GetSMTLevel(ctx, cpuID, workerRTNode)
+			smtLevel, err := nodes.GetSMTLevel(ctx, cpuID, workerRTNode)
+			Expect(err).ToNot(HaveOccurred(), "Unable to fetch SMT level on node %s, Error: %v", workerRTNode.Name, err)
 			if smtLevel < 2 {
 				Skip(fmt.Sprintf("designated worker node %q has SMT level %d - minimum required 2", workerRTNode.Name, smtLevel))
 			}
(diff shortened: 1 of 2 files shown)


test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -109,7 +109,8 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 		onlineCPUSet, err = nodes.GetOnlineCPUsSet(ctx, workerRTNode)
 		Expect(err).ToNot(HaveOccurred())
 		cpuID := onlineCPUSet.UnsortedList()[0]
-		smtLevel = nodes.GetSMTLevel(ctx, cpuID, workerRTNode)
+		smtLevel, err = nodes.GetSMTLevel(ctx, cpuID, workerRTNode)
+		Expect(err).ToNot(HaveOccurred())
```
mrniranjan: Here, we can error with message so that it's clear 
`Expect(err).ToNot(HaveOccurred(), "Unable to fetch  SMT level on node %s, Error: %v", workerRTNode.Name, err)`

mrniranjan: same for others also.

SargunNarula: Thanks, enhanced at all places.

test/e2e/performanceprofile/functests/utils/nodes/nodes.go
```diff
@@ -262,16 +262,20 @@ func GetOnlineCPUsSet(ctx context.Context, node *corev1.Node) (cpuset.CPUSet, er
 
 // GetSMTLevel returns the SMT level on the node using the given cpuID as target
 // Use a random cpuID from the return value of GetOnlineCPUsSet if not sure
-func GetSMTLevel(ctx context.Context, cpuID int, node *corev1.Node) int {
+func GetSMTLevel(ctx context.Context, cpuID int, node *corev1.Node) (int, error) {
```
Tal-or, line 265: GinkgoHelper() in the first line will solve exactly that without having to change the function API
https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-handles-failure

ffromani: @Tal-or is very correct. In addition, that would allow us to remove the fragile `WithOffset` expectations (mostly my fault here).

That said, changing the function signature and return an explicit error LGTM and I slightly prefer that approach for library code.

SargunNarula: @Tal-or If i understand correctly, if we used GinkgoHelper(), it would remove the panic and point the failure to Test itself, which originally is not the motive of the API change.

Considering a library function, it should work when called outside a test spec block, where it would panic. Please let me know if I’ve misunderstood anything here.

Tal-or: What is considered to be panic? a go assertion? because it's expected to abort a test with an assertion in Ginkgo.
All GinkgoHelper() is doing is point to the caller of the function (in the stack trace), it's basically equal to call `Fail` with offset.

> Considering a library function, it should work when called outside a test spec block

this function scope is only for the NTO e2e test, it should not be imported into other projects and not even in the none-test code of this specific project, so assertion here is completely fine.

Bottom line, I don't mind whether this function return an errors or asserts, my initial comment was mainly for the sake of discussion and broadening the knowledge and expertise as part of the reviewing process.

ffromani: > this function scope is only for the NTO e2e test, it should not be imported into other projects and not even in the none-test code of this specific project, so assertion here is completely fine.

I fully agree with this sentence

ffromani: I guess the remaining open point boils down to: where is this function planned to be used?

SargunNarula: @ffromani Here is a use- case where this function was tend to be used - [Link](https://github.com/openshift/cluster-node-tuning-operator/pull/1386/files#diff-c5ab91410e41b148d87d486cec9773349c60c605ac9dbf5810643b1325243588R320). Ignore the checkHyperthreding() for now as it is just a wrapper over getSMTLevel and would be removed once it itself gets fixed.

ffromani (commented):
/approve
/lgtm

PR #1400, opened 2025-09-28
NO-JIRA:e2e:hugepages: changing tests to use hugepages-allocator tool

Tal-or:
We want to use rhel9 as the image for the test pod which does not have libhugetlbfs.so.
Instead we're using a small in-house tool allocate the hugepages
and test functionality.

### DIFF
--- a/test/e2e/performanceprofile/functests/utils/images/images.go
+++ b/test/e2e/performanceprofile/functests/utils/images/images.go
@@ -17,7 +17,7 @@ func init() {
 
 	cnfTestsImage, ok = os.LookupEnv("CNF_TESTS_IMAGE")
 	if !ok {
-		cnfTestsImage = "cnf-tests:4.14"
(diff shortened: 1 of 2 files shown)


test/e2e/performanceprofile/functests/utils/images/images.go
```diff
@@ -17,7 +17,7 @@ func init() {
 
 	cnfTestsImage, ok = os.LookupEnv("CNF_TESTS_IMAGE")
 	if !ok {
-		cnfTestsImage = "cnf-tests:4.14"
+		cnfTestsImage = "cnf-tests:4.19"
```
yanirq, line 20: why not 4.20 ?

Tal-or: No special reason, we usually are not keeping track of update this value according to latest GA release.
I went with 4.19 because I remembered it already contains the hugepage-allocator tool

PR #1403, opened 2025-10-02
OCPBUGS-60218: PPC: ghw: filter out namespaces dir

shajmakh:
The previous code fails to run PPC when there is a namespace that has a
suffix `nodes`. ensure filtering the namespace directory to allow PPC to
work when such namespaces exist.

### DIFF
--- a/pkg/performanceprofile/profilecreator/ghwhandler.go
+++ b/pkg/performanceprofile/profilecreator/ghwhandler.go
@@ -34,7 +34,7 @@ import (
 func NewGHWHandler(mustGatherDirPath string, node *v1.Node) (*GHWHandler, error) {
 	nodeName := node.GetName()
 	nodePathSuffix := path.Join(nodes)
-	nodepath, err := getMustGatherFullPathsWithFilter(mustGatherDirPath, nodePathSuffix, clusterScopedResources)
+	nodepath, err := getMustGatherFullPathsWithFilter(mustGatherDirPath, nodePathSuffix, clusterScopedResources, namespaces)
 	if err != nil {
 		return nil, fmt.Errorf("can't obtain the node path %s: %v", nodeName, err)
 	}
--- a/pkg/performanceprofile/profilecreator/mustgather.go
+++ b/pkg/performanceprofile/profilecreator/mustgather.go
@@ -42,21 +42,26 @@ const (
 	yamlSuffix = ".yaml"
 	// nodes defines the subpath, relative to top-level must-gather directory, on which we find node-specific data
 	nodes = "nodes"
+	// namespaces defines the subpath, relative to top-level must-gather directory, on which we find openshift namespace-specific data
+	namespaces = "namespaces"
 	// sysInfoFileName defines the name of the file where ghw snapshot is stored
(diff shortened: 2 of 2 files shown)


pkg/performanceprofile/profilecreator/mustgather.go
```diff
@@ -42,21 +42,26 @@ const (
...
+	// namespaces defines the subpath, relative to top-level must-gather directory, on which we find openshift namespace-specific data
+	namespaces = "namespaces"
 	// sysInfoFileName defines the name of the file where ghw snapshot is stored
 	sysInfoFileName = "sysinfo.tgz"
 	// configOCPInfra defines the sub path relative to ClusterScopedResources, on which OCP infrastructure object is
 	configOCPInfra = "config.openshift.io/infrastructures"
 )
 
-func getMustGatherFullPathsWithFilter(mustGatherPath string, suffix string, filter string) (string, error) {
+func getMustGatherFullPathsWithFilter(mustGatherPath string, suffix string, filter ...string) (string, error) {
```
ffromani: should be `filters`

pkg/performanceprofile/profilecreator/ghwhandler.go
```diff
@@ -34,7 +34,7 @@ import (
 func NewGHWHandler(mustGatherDirPath string, node *v1.Node) (*GHWHandler, error) {
 	nodeName := node.GetName()
 	nodePathSuffix := path.Join(nodes)
-	nodepath, err := getMustGatherFullPathsWithFilter(mustGatherDirPath, nodePathSuffix, clusterScopedResources)
+	nodepath, err := getMustGatherFullPathsWithFilter(mustGatherDirPath, nodePathSuffix, clusterScopedResources, namespaces)
```
ffromani, line 37: so the core idea is to filter out `clusterScopedResources` and  `namespaces` right?

shajmakh: correct. namespaces is affected by the user configuration so we should filter it out anyway

pkg/performanceprofile/profilecreator/mustgather.go
shajmakh: right. Thanks, fixed

pkg/performanceprofile/profilecreator/ghwhandler.go
ffromani: makes sense but need to review again why it causes issues (and why now)

ffromani (commented):
IIRC PPC doesn't care about the content in `namespaces` at all, and if so this change should be good. I want to verify that before approving. With this assumption, codewise LGTM

pkg/performanceprofile/profilecreator/ghwhandler.go
shajmakh: sure, regarding "why now" it has always been the case but it is less common to have namespaces with "nodes" suffix. The bug has additional info.

ffromani (commented):
/approve

LGTM once the unnecessary file is removed

PR #1397, opened 2025-09-24
OCPBUGS-62153: E2E: skip SMT disabled test when L3 cache spans entire NUMA node

mrniranjan:
Skipping LLC tests on virtual machines as these
dont have uncore cache similar to Baremetal systems

PR #1406, opened 2025-10-07
OCPBUGS-62153: E2E: Fix for checking L3 Cache spans entire numa node

mrniranjan:
The earlier pr #1397 used to compare if the cpus
returned by ccx are same as numa node 0, which
may not be same, even though the size matches. This cause the test to not skip on vm.

### DIFF
--- a/test/e2e/performanceprofile/functests/13_llc/llc.go
+++ b/test/e2e/performanceprofile/functests/13_llc/llc.go
@@ -786,11 +786,14 @@ var _ = Describe("[rfe_id:77446] LLC-aware cpu pinning", Label(string(label.Open
 				ccx, err := getCCX(0)
 				Expect(err).ToNot(HaveOccurred())
 				L3CacheGroupSize = ccx.Size()
-				// Get actual CPU ID's from Numa Node 0
-				numaNode0Cpus := cpuset.New(numaInfo[0]...)
-
-				// Compare if L3 cache group CPUs match NUMA Node 0 CPUs
-				if ccx.Equals(numaNode0Cpus) {
+				// Compare length of numa node 0 cpus with ccx size
+				// if they are same then L3 Cache spans the whole of numa node
+				// we cannot compare the L3 cache cpus are same as numa node 0 cpus as
+				// as the cpu topology as they many not be same . For example
+				// ccx cpus may be 0-11, and numa node0 cpus could be 0,2,4,6,8,10,12,14,16,18,22
+				// if we were to compare the equality, the equality fails and tests will not be
+				// skipped atleast on Virtual machines
+				if len(numaInfo[0]) == L3CacheGroupSize {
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/13_llc/llc.go
```diff
@@ -786,11 +786,14 @@ var _ = Describe("[rfe_id:77446] LLC-aware cpu pinning", Label(string(label.Open
...
-				// Compare if L3 cache group CPUs match NUMA Node 0 CPUs
-				if ccx.Equals(numaNode0Cpus) {
+				// Compare length of numa node 0 cpus with ccx size
+				// if they are same then L3 Cache spans the whole of numa node
+				// we cannot compare the L3 cache cpus are same as numa node 0 cpus as
+				// as the cpu topology as they many not be same . For example
+				// ccx cpus may be 0-11, and numa node0 cpus could be 0,2,4,6,8,10,12,14,16,18,22
+				// if we were to compare the equality, the equality fails and tests will not be
+				// skipped atleast on Virtual machines
+				if len(numaInfo[0]) == L3CacheGroupSize {
```
ffromani, line 796: comparing the size is too coarse grained and makes for unnecessary skips. What creates this difference between CCX cpus and NUMA node cpus?

mrniranjan: we don't know why .  On libvirt when we create vm's we specify the  cpu id's of numa nodes in the vm definition , the cache hierarchy that libvirt tries to emulate probably doesn't take in to account the cpu id's. Either way if the total cpus sharing the cache and number of cpus in that node match, shouldn't that be enough to say that there is no specific subset of cpus sharing L3 cache ?

ffromani: This seems to work by coincidence; or, alternatively, I can't see a reliable way to skip which takes in account this odd case. Rather than guess, let's label VM nodes and skip based on labels. At least it's very explicit and consistent and easy to understand.

mrniranjan: @ffromani I am closing this PR, I have modified the cpu topology on VM's to use the same toplogy as it lists in shared_cpu_list

  ```
NUMA node(s):              2
  NUMA node0 CPU(s):         0-11
  NUMA node1 CPU(s):         12-23

```

```
sh-5.1# cat /sys/devices/system/cpu/cpu0/cache/index3/shared_cpu_list
0-11
sh-5.1# cat /sys/devices/system/cpu/cpu1/cache/index3/shared_cpu_list
0-11
sh-5.1# cat /sys/devices/system/cpu/cpu12/cache/index3/shared_cpu_list
12-23
sh-5.1# cat /sys/devices/system/cpu/cpu13/cache/index3/shared_cpu_list
12-23
```
so the current code should work.

PR #1442, opened 2025-12-03
OCPBUGS-66337: E2E:PPC:Use default worker mcp instead worker-cnf

mrniranjan:
PPC tests relied on using worker-cnf mcp
but this assumes that mustgather provided to ppc
tests contains worker-cnf mcp, which may not be the case.

So modifying to use default worker mcp instead worker-cnf

### DIFF
--- a/test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
+++ b/test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
@@ -63,7 +63,7 @@ func PPCTestCreateUtil() *PPCTestIntegration {
 
 var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform awareness", Label(string(label.PerformanceProfileCreator)), func() {
 	mustgatherDir := testutils.MustGatherDir
-	mcpName := testutils.RoleWorkerCNF
+	mcpName := "worker"
 	ntoImage := testutils.NTOImage
 	Context("PPC Sanity Tests", Label(string(label.Tier0)), func() {
 		ppcIntgTest := PPCTestCreateUtil()
--- a/test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
+++ b/test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
@@ -63,7 +63,7 @@ func PPCTestCreateUtil() *PPCTestIntegration {
 
 var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform awareness", Label(string(label.PerformanceProfileCreator)), func() {
 	mustgatherDir := testutils.MustGatherDir
-	mcpName := "worker"
+	mcpName := testutils.RoleWorker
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
```diff
@@ -63,7 +63,7 @@ func PPCTestCreateUtil() *PPCTestIntegration {
 
 var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform awareness", Label(string(label.PerformanceProfileCreator)), func() {
 	mustgatherDir := testutils.MustGatherDir
-	mcpName := testutils.RoleWorkerCNF
+	mcpName := "worker"
```
Tal-or: The `testutils.RoleWorkerCNF` is configurable
https://github.com/openshift/cluster-node-tuning-operator/blob/c544e974701734b8d4c89937b5100535f870a743/test/e2e/performanceprofile/functests/utils/consts.go#L37

wouldn't it be better to simply set the environment variable correctly?

mrniranjan: that environment variable if i change will have an effect on whole tests suite apart from ppc, we were using testutils.RoleWorkerCNF, but if  i use ROLE_WORKER_CNF=worker, it will have affect on other tests because , we apply performance profile only on systems which worker-cnf label

mrniranjan: in the newer ci's which will run on prow ci, the mustgather is collected even before we label the systems and we want to use the same mustgather for ppc tests, so it's safe to use "worker" label , since it's default  and we don't have to make changes to ROLE_WORKER_CNF environment variable

Tal-or (approved):
/lgtm

test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
rshemtov13: Maybe create in the testutilss another role worker called ```RoleWorker``` just to keep it consistent.

mrniranjan: we could , the reason i didn't go through that because , this requires a new environment variable to be added to existing CI infrastructure . Considering that this will run  in prow CI where the number of environment variables are already a lot , i didn't go through that

Tal-or: we have https://github.com/openshift/cluster-node-tuning-operator/blob/c544e974701734b8d4c89937b5100535f870a743/test/e2e/performanceprofile/functests/utils/consts.go#L94

mrniranjan: Updated the code to use testutils.RoleWorker

Tal-or (approved):
/lgtm

PR #1464, opened 2026-01-30
Added GINKGO_LABEL_FILTER env variable to run-test.sh

SargunNarula:
To support removing of tests from upstream lanes based on filters,
added this env variable support to the run-test.sh file 

Signed-off-by: Sargun Narula [snarula@redhat.com](mailto:snarula@redhat.com)

### DIFF
--- a/hack/run-test.sh
+++ b/hack/run-test.sh
@@ -84,10 +84,15 @@ main() {
         NO_COLOR="--no-color"
     fi
 
+    LABEL_FILTER_FLAG=""
+    if [ -n "$GINKGO_LABEL_FILTER" ]; then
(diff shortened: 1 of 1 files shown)


hack/run-test.sh
```diff
@@ -84,10 +84,15 @@ main() {
         NO_COLOR="--no-color"
     fi
 
+    LABEL_FILTER_FLAG=""
```
Tal-or, line 87: Can't we just use the `EXTRA_PARAMS` flag?
If we do, no code change are needed

Tal-or: @SargunNarula if you're happy with suggestion above lets close the PR

SargunNarula: Yes, The purpose of this PR was resolved quite early. Closing in.

PR #1470, opened 2026-02-11
OCPBUGS-77377: E2E: Add test case to check Infrastructure pods affinity

mrniranjan:
Add test case to check affinity of Infrastructure pods. we use tuned as an example to check if the infrastructure pods have the affinity of reserved plus available isolated cpus

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
+++ b/test/e2e/performanceprofile/functests/1_performance/cpu_management.go
@@ -194,6 +194,33 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			Expect(reservedCPUSet.IsSubsetOf(maskSet)).To(Equal(true), fmt.Sprintf("The init process (pid 1) should have cpu affinity: %s", reservedCPU))
 		})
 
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/cpu_management.go
```diff
@@ -194,6 +194,27 @@ var _ = Describe("[rfe_id:27363][performance] CPU Management", Ordered, func() {
 			Expect(reservedCPUSet.IsSubsetOf(maskSet)).To(Equal(true), fmt.Sprintf("The init process (pid 1) should have cpu affinity: %s", reservedCPU))
 		})
 
+		It("[test_id:87722][crit:high][level:acceptance] verify Infrastructure pods have cpu affinity of reserved plus isolated cpuset", func() {
```
MarSik, line 197: This is only valid when workload partitioning is not enabled. And it is only valid when strict cpu isolation is false in cpu manager.

mrniranjan: I have added a condition to check affinity when workload partitioning is enabled.

MarSik (approved):
/lgtm

PR #1486, opened 2026-03-25
NO-JIRA: E2E: move dra resource test to reboot tests directory

mrniranjan:
DRA resource tests needed to be moved to 2_performance_update directory as these test modify the profile which reboots the node.

since 1_performance directory is primarily used to have non reboot tests, hence moving to 2_performance_update


Tal-or (approved):
Good call
/approve
/lgtm

MarSik (approved):
/lgtm

PR #1493, opened 2026-04-14
OCPBUGS-84938: e2e: Add irqbalance StartLimitBurst >= 100 config test

oblau:
This pr Is test automation for OCPBUGS-45112

The bug:
The underlying cause was the default configuration:
`StartLimitIntervalUSec=10s ` `StartLimitBurst=5` meaning over 5 restarts in a 10 sec window would cause `ActiveState=failed` 
Each pod create/delete with `irq-load-balancing.crio.io = "disable"` annotation causes irqbalance to restart.
In addition restarts coalesce meaning this is timing dependent thus not every deployment with more than 5 pods actually causes over 5 restarts.

The fix:
drop-in sets `StartLimitBurst=100`.
Checking this irqbalance systemd property is enough - no functional test is required 

## Summary by CodeRabbit

* **Tests**
  * Renamed a performance test suite for clearer identification of IRQBalance checks, improving test readability.
  * Added a new Tier0 end-to-end test that verifies the IRQBalance service startup limit burst is exactly 100, increasing coverage for system startup configuration and catching regressions early.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -34,6 +35,7 @@ import (
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/pods"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profilesupdate"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/systemd"
 	e2etuned "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/tuned"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/util"
 	"github.com/openshift/cluster-node-tuning-operator/test/framework"
@@ -49,7 +51,7 @@ var (
 	cs = framework.NewClientSet()
 )
 
-var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
+var _ = Describe("[performance] IRQBalance", Ordered, func() {
 	var workerRTNodes []corev1.Node
 	var targetNode *corev1.Node
 	var profile, initialProfile *performancev2.PerformanceProfile
@@ -380,6 +382,19 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 			Entry("[test_id:86343] Verify housekeeping CPUs are sibling paired and applied to IRQ affinity", context.TODO(), 1),
 			Entry("[test_id:86345] Verify first cpu core defines housekeeping CPUs in multi-container pod", context.TODO(), 2),
 		)
+	})
+
+	// Automates OCPBUGS-45112 - Config test
+	// Default StartLimitBurst=5 was too low - each pod created/deleted would trigger a restart of irqbalance
+	// Systemd coalesces concurrent requests, making the reproduction by deployment impractical.
+	It("[test_id:88711] should have irqbalance StartLimitBurst >= 100", Label(string(label.Tier0)), func() {
+		startLimitBurst, err := systemd.ShowPropertyValue(context.TODO(), "irqbalance.service", "StartLimitBurst", targetNode)
+		Expect(err).ToNot(HaveOccurred())
+		startLimitBurst = strings.TrimSpace(startLimitBurst)
+		testlog.Infof("irqbalance.service StartLimitBurst=%s on node %s", startLimitBurst, targetNode.Name)
+		startLimitBurstInt, err := strconv.Atoi(startLimitBurst)
+		Expect(err).ToNot(HaveOccurred(), "unexpected StartLimitBurst value %q", startLimitBurst)
+		Expect(startLimitBurstInt).To(BeNumerically(">=", 100), "irqbalance StartLimitBurst must be >=100 (OCPBUGS-45112)")
 	})
 
 })
--- a/test/e2e/performanceprofile/functests/utils/systemd/systemd.go
+++ b/test/e2e/performanceprofile/functests/utils/systemd/systemd.go
@@ -19,3 +19,9 @@ func ShowProperty(ctx context.Context, unitfile string, property string, node *c
 	out, err := nodes.ExecCommand(ctx, node, cmd)
 	return string(out), err
 }
+
+func ShowPropertyValue(ctx context.Context, unitfile string, property string, node *corev1.Node) (string, error) {
+	cmd := []string{"/bin/bash", "-c", fmt.Sprintf("chroot /rootfs systemctl show -p %s %s --value --no-pager", property, unitfile)}
+	out, err := nodes.ExecCommand(ctx, node, cmd)
+	return string(out), err
+}


test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -382,6 +385,52 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 		)
 	})
 
+	It("[test_id:88711] should not crash irqbalance when deploying ten pods with irq-load-balancing disabled", Label(string(label.Tier0)), func() {
```
mrniranjan, line 389: You can add a comment above this line about the bug number 
// automates OCPBUGS-45112

oblau: Thanks, added the comment

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -382,6 +385,53 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
...
+			"test-deployment",
+			testutils.NamespaceTesting,
+			deployments.WithPodTemplate(testpod),
+			deployments.WithReplicas(int32(repCount)),
+		)
+
+		Expect(testclient.DataPlaneClient.Create(context.TODO(), dp)).To(Succeed())
+		testlog.Infof("Created deployment %s with %d replicas (irq-load-balancing disabled)", dp.Name, repCount)
+		defer func() {
+			Expect(testclient.DataPlaneClient.Get(context.TODO(), client.ObjectKeyFromObject(dp), dp)).To(Succeed())
```
shajmakh: I doubt this line to fail but if it does we likely don't want to fail the test. I see the intention that this block expects the deployment to be there and then deletes it, but it it is not found then no need to delete it.  implementing something like this https://github.com/oblau/cluster-node-tuning-operator/blob/373a65819314a40f259e2f536984e88c78431b7e/test/e2e/performanceprofile/functests/utils/pods/pods.go#L99 would be helpful for deployment deletion management

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -382,6 +385,53 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
 		)
 	})
 
+	// automates OCPBUGS-45112
+	It("[test_id:88711] should not crash irqbalance when deploying ten pods with irq-load-balancing disabled", Label(string(label.Tier0)), func() {
+		By("Creating workload Deployment (irq-load-balancing disabled, 10 replicas)")
+		const workloadCPUsPerPod = 2
```
shajmakh, line 391: this won't run on u/s ci, will the d/s lanes have at least 24 cpus - including node baseload and reserved- ? 
asking because we need to have this running somewhere periodically

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -382,6 +385,53 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
...
+		By("Verifying irqbalance.service is still active after workload is running")
+		activeState, err := systemd.ShowProperty(context.TODO(), "irqbalance.service", "ActiveState", targetNode)
+		Expect(err).ToNot(HaveOccurred())
+		testlog.Infof("irqbalance.service %s on node %s", strings.TrimSpace(activeState), targetNode.Name)
+		Expect(strings.TrimSpace(activeState)).To(Equal("ActiveState=active"), "irqbalance must stay active while workload runs")
+
+		By("Verifying irqbalance has StartLimitBurst=100 from CRI-O drop-in")
+		startLimitBurst, err := systemd.ShowProperty(context.TODO(), "irqbalance.service", "StartLimitBurst", targetNode)
+		Expect(err).ToNot(HaveOccurred())
+		testlog.Infof("irqbalance.service %s on node %s", strings.TrimSpace(startLimitBurst), targetNode.Name)
```
shajmakh: maybe extract  `strings.TrimSpace(startLimitBurst)` into a var

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
oblau: The d/s lanes have 24 cpus - 20 of which are isolated and 4 reserved.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
oblau: Done.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
oblau: I implemented the same pod deletion mechanism for deployments and used it. 
Hope that's what you meant

test/e2e/performanceprofile/functests/utils/deployments/deployments.go
```diff
@@ -116,3 +121,33 @@ func WaitForDesiredDeploymentStatus(ctx context.Context, deployment *appsv1.Depl
 		return false, nil
 	})
 }
+
+// WaitForDeletion waits until the deployment will be removed from the cluster
+func WaitForDeletion(ctx context.Context, cli client.Client, deployment *appsv1.Deployment, timeout time.Duration) error {
```
shajmakh, line 126: I'd keep this unexported. Usually, the right usage of "wait for deletion" should be post-validation that the deployment is there to be deleted.

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
shajmakh: uhm then how come 10 pods each with 2 cpus and all are of gu QoS managed to run with other non-gu running pods? are we sure that all of the pods went running?

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -382,6 +385,53 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
...
 	})
 
+	// automates OCPBUGS-45112
+	It("[test_id:88711] should not crash irqbalance when deploying ten pods with irq-load-balancing disabled", Label(string(label.Tier0)), func() {
+		By("Creating workload Deployment (irq-load-balancing disabled, 10 replicas)")
+		const workloadCPUsPerPod = 2
+		const repCount = 10
+		if isolatedCPUSet.Size() < repCount*workloadCPUsPerPod {
+			Skip(fmt.Sprintf("need %d isolated cpus, have %d", repCount*workloadCPUsPerPod, isolatedCPUSet.Size()))
+		}
```
shajmakh, line 395: This is not enough. The node has other pods running on it and they consume resources, please use https://github.com/openshift/cluster-node-tuning-operator/blob/3d98f7ee7fe0651d182b8a780956512eb407a963/test/e2e/performanceprofile/functests/utils/baseload/baseload.go#L21 to ensure the test continues with enough amount of free CPUs

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -382,6 +385,52 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		Expect(testclient.DataPlaneClient.Get(context.TODO(), client.ObjectKeyFromObject(dp), dp)).To(Succeed())
+		testlog.Infof("Deployment %s reached %d/%d available replicas", dp.Name, dp.Status.AvailableReplicas, *dp.Spec.Replicas)
+
+		By("Verifying irqbalance.service is still active after workload is running")
+		activeState, err := systemd.ShowProperty(context.TODO(), "irqbalance.service", "ActiveState", targetNode)
+		Expect(err).ToNot(HaveOccurred())
+		testlog.Infof("irqbalance.service %s on node %s", strings.TrimSpace(activeState), targetNode.Name)
+		Expect(strings.TrimSpace(activeState)).To(Equal("ActiveState=active"), "irqbalance must stay active while workload runs")
+
```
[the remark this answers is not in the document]
oblau: no long relevant

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -382,6 +385,53 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
...
+			deployments.WithPodTemplate(testpod),
+			deployments.WithReplicas(int32(repCount)),
+		)
+
+		Expect(testclient.DataPlaneClient.Create(context.TODO(), dp)).To(Succeed())
+		testlog.Infof("Created deployment %s with %d replicas (irq-load-balancing disabled)", dp.Name, repCount)
+		defer func() {
+			Expect(testclient.DataPlaneClient.Get(context.TODO(), client.ObjectKeyFromObject(dp), dp)).To(Succeed())
+			Expect(testclient.DataPlaneClient.Delete(context.TODO(), dp)).To(Succeed())
+		}()
```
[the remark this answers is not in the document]
oblau: no longer relevant

test/e2e/performanceprofile/functests/utils/deployments/deployments.go
oblau: Removed it completely - didn't need it with the new approach

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
oblau: Now running a config test instead - so this is outdated

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
oblau: Now running a config test instead - so this is outdated

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
oblau: Now running a config test instead - so this is outdated

test/e2e/performanceprofile/functests/1_performance/irqbalance.go
```diff
@@ -382,6 +383,17 @@ var _ = Describe("[performance] Checking IRQBalance settings", Ordered, func() {
...
 
+	// Automates OCPBUGS-45112 - Config test
+	// Default StartLimitBurst=5 was too low - each pod created/deleted would trigger a restart of irqbalance
+	// Systemd coalesces concurrent requests, making the reproduction by deployment impractical.
+	It("[test_id:88711] should have irqbalance StartLimitBurst = 100", Label(string(label.Tier0)), func() {
+		startLimitBurst, err := systemd.ShowProperty(context.TODO(), "irqbalance.service", "StartLimitBurst", targetNode)
+		Expect(err).ToNot(HaveOccurred())
+		startLimitBurst = strings.TrimSpace(startLimitBurst)
+		testlog.Infof("irqbalance.service %s on node %s", startLimitBurst, targetNode.Name)
+		Expect(strings.TrimSpace(startLimitBurst)).To(Equal("StartLimitBurst=100"), "irqbalance must have StartLimitBurst=100 (OCPBUGS-45112)")
```
MarSik: It would be better to check it is 100 or more here. In case we change it in the future. That is why I did not want to depend on the config, but I agree the full e2e test is more fragile and tests systemd behavior that should be covered by systemd itself.

oblau: Added util to see property's value using  --value flag. 
Now checking if the value is >= 100. 
My worry is - would we not prefer this to fail if the number increases in the future? 
If it increases its likely because 100 wasn't enough - so the test would not flag a system with "old" value of 100.

For example if the old test was for >= 5 we wouldn't know if the old value or the new were there. 

Maybe i'm overthinking or missing something. would like to hear your thoughts.

MarSik: Well, that is an interesting point. Failing the test to make sure the test stays aligned with the code. However, we do not own the code in this case and so a cri-o change would suddently fail our tests. I am not sure what is better.

MarSik (approved):
/lgtm
/hold change the PR title please

PR #1517, opened 2026-05-15
OCPBUGS-80952: perf: latency: compute memory resources dynamically

shajmakh:
When CPUs are very high the pod's fixed memory resources may become too
low to run the latency checks. Add an environment variable to allow more
flexibility while preserving the old behavior for backward
compatibility.
The new behavior goes like this:
1. If no env var is set, scales up the memory amount per CPU floored at
defaultTestMemory.
2. else, if it was set to a an explicit memory value and it's a valid
   quantity then use that in the latency pod, otherwise throw an error.

`32Mi` was picked as the per-CPU memory factor based on input from
consumers of the application; If happened that the memory is still not
enough, the user has the flexibility to override the total memory with
an explicit value.

## Summary by CodeRabbit

* **New Features**
  * Added configurable memory for latency tests via an environment variable, including a dynamic mode that scales memory based on CPU count.
  * Invalid memory values are now validated and will surface as configuration errors.

* **Documentation**
  * Updated docs to describe the new memory configuration and dynamic scaling behavior.

### DIFF
--- a/test/e2e/performanceprofile/functests/4_latency/latency.go
+++ b/test/e2e/performanceprofile/functests/4_latency/latency.go
@@ -36,13 +36,23 @@ import (
 )
 
 const (
-	oslatTestName        = "oslat"
-	cyclictestTestName   = "cyclictest"
-	hwlatdetectTestName  = "hwlatdetect"
-	defaultTestDelay     = 0
-	defaultTestRuntime   = "300"
-	defaultMaxLatency    = -1
-	defaultTestCpus      = -1
+	//tools names
+	oslatTestName       = "oslat"
+	cyclictestTestName  = "cyclictest"
+	hwlatdetectTestName = "hwlatdetect"
+
+	//default values
+	defaultTestDelay   = 0
+	defaultTestRuntime = "300"
+	defaultMaxLatency  = -1
+	defaultTestCpus    = -1
+	defaultTestMemory  = "1Gi"
+
+	//dynamic memory mode values
+	// 32Mi per requested CPU should be reasonable for the test
+	perCpuMemoryFactor = 32
+	memoryFactorFormat = "Mi"
+
 	minCpuAmountForOslat = 2
 )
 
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -278,6 +281,28 @@ func getLatencyTestCpus() (int, error) {
 	return defaultTestCpus, nil
 }
 
+func getLatencyTestMemory(cpus int) (string, error) {
+	if val, ok := os.LookupEnv("LATENCY_TEST_MEMORY"); ok {
+		if val == dynamicMemory {
+			if cpus == defaultTestCpus {
+				return defaultTestMemory, nil
+			}
```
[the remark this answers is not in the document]
shajmakh: correct but I wouldn't want to depend on the CPU filter being applied before computing the memory. I'd keep it as an extra safety net.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -278,6 +281,29 @@ func getLatencyTestCpus() (int, error) {
...
 }
 
+func getLatencyTestMemory(cpus int) (string, error) {
+	if val, ok := os.LookupEnv("LATENCY_TEST_MEMORY"); ok {
+		if val == dynamicMemory {
+			// Defensive check: fall back to default memory if the cpus was not normalized
+			if cpus == defaultTestCpus {
+				return defaultTestMemory, nil
+			}
+			// 32Mi per requested CPU should be reasonable for the test; if needed more, it can
```
MarSik: The PR description says 16, it might be a good idea to align it.

shajmakh: right, I updated the commit message only. Thanks for catching.

MarSik (approved):
/lgtm
/hold You might want to update the PR description too.

PR #1525, opened 2026-05-25
OCPBUGS-83863: Strip debug symbols from Go binaries

sdodson:
## Summary
- Add `-s -w` to ldflags in `GO_BUILD_RECIPE` to strip DWARF debug info and symbol tables from the main operator binary
- Fix `build-gather-sysinfo` and `build-performance-profile-creator` targets which intended to strip but had a make-vs-shell variable expansion bug (`$(LDFLAGS)` is expanded by make before the shell runs, so the shell variable set on the previous line was never used)

## Impact
The cluster-node-tuning-operator image (617 MB) contains three unstripped Go binaries totaling 195 MB:
- `cluster-node-tuning-operator` (96.8 MB)
- `gather-sysinfo` (63.8 MB)
- `performance-profile-creator` (34.6 MB)

Stripping typically reduces Go binary size by 20-30%, reducing container image pull time during node scale-up.

## Test plan
- [ ] Verify image builds successfully
- [ ] Verify binaries are stripped (`file /usr/bin/cluster-node-tuning-operator` should show "stripped")
- [ ] Verify operator functions correctly

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to reduce binary sizes and streamline compilation across multiple build targets.
  * Applied the same build optimizations to additional build and test targets to ensure consistent output and faster, smaller artifacts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ TESTDATA_OTE_DIR=test/extended/testdata
 TESTDATA_OTE=$(shell find $(TESTDATA_OTE_DIR) -type f \( -name '*.yaml' -o -name '*.yml' \) 2>/dev/null)
 ASSETS=$(shell find assets -name \*.yaml)
 GO=GOARCH=$(GOARCH) GO111MODULE=on GOFLAGS=-mod=vendor go
-GO_BUILD_RECIPE=$(GO) build -o $(OUT_DIR)/$(PACKAGE_BIN) -ldflags '-X $(PACKAGE)/version.Version=$(REV)' $(PACKAGE_MAIN)
+GO_BUILD_RECIPE=$(GO) build -o $(OUT_DIR)/$(PACKAGE_BIN) -ldflags '-s -w -X $(PACKAGE)/version.Version=$(REV)' $(PACKAGE_MAIN)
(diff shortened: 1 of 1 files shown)


jmencak:
Thank you for the PR, Scott.  Do we want to do the same thing for `cluster-node-tuning-operator-test-ext`?  It will grow large soon as we add more tests.

sdodson:
Good call, done — added `-ldflags "-s -w"` to the `cluster-node-tuning-operator-test-ext` build target in the latest commit.

sdodson:
The gzipped binary is about half as large once stripped.

```
~ podman run -it --entrypoint /bin/sh `oc adm release info 4.22.0 --image-for cluster-node-tuning-operator`
sh-5.1# dnf install -y binutils
sh-5.1# ls -la cluster-node-tuning-operator*
-rwxr-xr-x. 1 root root 101535656 May 19 16:47 cluster-node-tuning-operator
-rwxr-xr-x. 1 root root  11113396 May 19 16:47 cluster-node-tuning-operator-test-ext.gz
sh-5.1# gzip -d cluster-node-tuning-operator-test-ext.gz 
sh-5.1# strip cluster-node-tuning-operator*
sh-5.1# gzip cluster-node-tuning-operator-test-ext 
sh-5.1# ls -la cluster-node-tuning-operator*
-rwxr-xr-x. 1 root root 72069048 Jun  9 17:48 cluster-node-tuning-operator
-rwxr-xr-x. 1 root root  5684280 Jun  9 17:48 cluster-node-tuning-operator-test-ext.gz
```

PR #1526, opened 2026-05-28
OCPBUGS-87132: E2E: Remove ovs-vswitchd from platform services

mrniranjan:
ovs-vswitchd will always use all the online cpus
and will not be restricted to reserved cpus even
on schedulabel control plane nodes with workload partition enabled

## Summary by CodeRabbit

* **Tests**
  * Updated control-plane scheduling test to refine platform services CPU affinity validation by excluding one service from the test scope while maintaining reserved CPU parsing validation.


MarSik (approved):
/lgtm

PR #1528, opened 2026-05-28
OCPBUGS-86803: e2e: Remove unnecessary len(numa) < 2 skip gates

oblau:
Allow tests to run on systems with single numa
These tests never needed 2 numa nodes 

Changes:

llc.go:
- BeforeAll (SMT Enabled) — removed len(numaInfo) < 2 skip. Affects: 77725, 77726, 81668, 77728, 77729, 77730, 81670, 81671, 87072, 87073, 87074
- BeforeAll (SMT Disabled) — removed len(numaInfo) < 2 skip. Affects: 81672, 81673

updating_profile.go:
- 50966 — removed NUMA check
- 50968 — removed NUMA check
- 50970 — added comment why numa check is required here

## Summary by CodeRabbit

* **Tests**
  * Removed NUMA-count gating from LLC-aware CPU-pinning tests so they run on more hardware configurations.
  * Relaxed NUMA-based skip logic in offline-CPU update tests so they proceed more often, increasing test coverage across platforms.
  * Preserved cores-per-socket and empty-topology checks where needed to avoid unreliable runs.
  * Added clarifying test notes about offlining CPUs across NUMA nodes for maintainability.

### DIFF
--- a/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
+++ b/test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
@@ -777,9 +777,6 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 		It("[test_id:50966]verify offlined parameter accepts multiple ranges of cpuid's", func() {
 			var reserved, isolated, offlined cpuset.CPUSet
 			numaTopology := copyNumaCoreSiblings(numaCoreSiblings)
-			if len(numaCoreSiblings) < 2 {
-				Skip(fmt.Sprintf("This test need 2 NUMA nodes, available only %d", len(numaCoreSiblings)))
-			}
 			if len(numaCoreSiblings[0]) < 20 {
 				Skip("This test needs systems with at least 20 cores per socket")
 			}
@@ -850,9 +847,6 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 		It("[test_id:50968]verify cpus mentioned in reserved or isolated cannot be offline", func() {
 			var reserved, isolated, offlined cpuset.CPUSet
 			numaTopology := copyNumaCoreSiblings(numaCoreSiblings)
-			if len(numaCoreSiblings) < 2 {
-				Skip(fmt.Sprintf("This test need 2 NUMA nodes, available only %d", len(numaCoreSiblings)))
-			}
 			// Get reserved core siblings from 0, 1
 			for reservedCores := 0; reservedCores < 2; reservedCores++ {
 				// Get the cpu siblings from the selected core and delete the siblings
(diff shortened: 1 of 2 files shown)


mrniranjan:
> Tests used "skip if <2 NUMA nodes" as a proxy for VM detection.

This is not a proxy for VM detection but rather detection of Baremetal

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -777,11 +777,10 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
...
 			var reserved, isolated, offlined cpuset.CPUSet
 			numaTopology := copyNumaCoreSiblings(numaCoreSiblings)
-			if len(numaCoreSiblings) < 2 {
-				Skip(fmt.Sprintf("This test need 2 NUMA nodes, available only %d", len(numaCoreSiblings)))
-			}
-			if len(numaCoreSiblings[0]) < 20 {
-				Skip("This test needs systems with at least 20 cores per socket")
+			onlineCPUs, err := nodes.GetOnlineCPUsSet(context.TODO(), &workerRTNodes[0])
+			Expect(err).ToNot(HaveOccurred())
+			if onlineCPUs.Size() < 24 {
```
Tal-or: Those numbers are seems completely random for reviewers, it would be better to add a comment and/or constant that explain the reason for this specific number.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -915,6 +916,13 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 		It("[test_id:50970]Offline CPUID's from multiple numa nodes", func() {
 			var reserved, isolated, offlined cpuset.CPUSet
 			numaTopology := copyNumaCoreSiblings(numaCoreSiblings)
+			onlineCPUs, err := nodes.GetOnlineCPUsSet(context.TODO(), &workerRTNodes[0])
+			Expect(err).ToNot(HaveOccurred())
+			if onlineCPUs.Size() < 60 {
```
Tal-or: ditto

Tal-or (commented):
I'm fine with the change in general if it has a benefit,
but please always explain in the PR+commit the rational for a change. 

Right now, it might be clear why we went with a specific change, but as time goes by we eventually forget why we did what we did.
Hence, it become crucial to understand and document the reasons behind the decisions we're making now, for the sake of our future selves :)

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
oblau: Reverted the cpu changes.
Now the scope of the pr is clear.
Thanks for the constructive criticism

Tal-or (approved):
/lgtm
Thanks!

PR #1534, opened 2026-06-03
OCPBUGS-86541: Enable timer migrations for all use-cases

MarSik:
This has an unfortunate side-effect that HRTIMERs break and cyclictest results will be affected, but there is no way around that. Not enabling this breaks cpu pinned workload isolation for example when any TCP timeout and keep alive timers are present.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized kernel timer-migration setting across tuned performance profiles so timer behavior is consistent between regular and real-time configurations.

* **Tests**
  * Updated performance tests and test data to expect the new, consistent timer-migration behavior.


bartwensley (approved):
Thanks Martin - looks good to me.

PR #1545, opened 2026-06-12
NO-JIRA: E2E: Remove checking of container runtime since crun is default

mrniranjan:
With crun being default and runc being removed
we don't need to check the default container runtime. 
Also the method used to get container runtime is error prone and doesn't work on ARM 

## Summary by CodeRabbit

* **Chores**
  * Removed an obsolete test utility that detected the container runtime.

* **Refactor**
  * Simplified cgroup v1/v2 test logic to remove runtime detection; controller helpers now derive configuration solely from cgroup mount paths.

* **Tests**
  * Updated a mixed-CPU test to reference the CRI-O runtime configuration directly via a local config path.


SargunNarula:
Thanks for checking up on this. Looks good to me.

MarSik (approved):
/lgtm

PR #1546, opened 2026-06-14
CNF-23471: Dedicate cpus for dpdk vswitch

Tal-or:
Add `spec.cpu.ovsDpdk` field to the PerformanceProfile v2 API, enabling users to designate CPUs for exclusive use by OVS-DPDK.

The OVS-DPDK CPUs receives full kernel-level scheduling isolation, are excluded from kubelet scheduling and IRQ balancing, and remain protected from interference across pod lifecycle events.

The controller validates that the cluster meets the prerequisites for the isolation and reports a Degraded status if they are not satisfied.
In addition `"performance.openshift.io/cpu-load-balancing-ovs-dpdk"` annotation was added to disable kernel load balancing on the `ovsDpdk` CPUs. This PR makes it optional since an extra step is needed from the OVS process side to make this feature fully functional, so we don't want to decouple this extra isolation capability from the general set of parameters we configure for the ovs dpdk set.

Includes unit and e2e test coverage.

This work inspired by: https://github.com/mcoquelin/cluster-node-tuning-operator/commits/release-4.20-ovs-dpdk/ 

## Summary by CodeRabbit

* **New Features**
  * Added `spec.cpu.ovsDpdk` to reserve a fully isolated CPU set for OVS-DPDK, with prerequisite validation during reconciliation.
  * Added an annotation to disable OVS-DPDK CPU load balancing.
  * Generated OVS-DPDK CPU pinning (OVS slice/cgroup + `ovs-dpdk-cpus-configure`) and updated kubelet reserved CPUs and tuned CPU isolation.

* **Bug Fixes**
  * Improved IRQ-balance banned CPU and `default_smp_affinity` masking so OVS-DPDK CPUs are reliably excluded, including consistent restore behavior.

* **Documentation**
  * Updated performance profile documentation for the new CPU fields and related options.

* **Tests**
  * Expanded unit, Ginkgo, and E2E coverage for OVS-DPDK validation, degradation reporting, and CPU/affinity behavior.


assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
```diff
@@ -9,17 +9,51 @@ IRQBALANCE_CONF="${1:-/etc/sysconfig/irqbalance}"
...
+	# clear banned bits from each 32-bit group: result = smp & ~banned
+	result=""
+	for (( i=0; i<n; i++ )); do
+		ban="${padded:$(( i * 8 )):8}"
+		val=$(printf "%08x" $(( 0x${smp[$i]} & ~0x${ban} )))
+		result+="${result:+,}${val}"
+	done
+	echo "Setting default_smp_affinity to ${result} (removing dedicated CPUs ${BANNED_CPUS} from default_smp_affinity mask)"
+	echo "${result}" > "${SMP_AFFINITY}"
 fi
```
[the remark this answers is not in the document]
Tal-or: Configuring dedicated CPU field with CPUs which aren't exist will cause kubelet to fail and the node won't go up anyway. 
This is also true for the other CPU fields.

amorenoz:
> Setting disableOvsDynamicPinning prevents OVN-Kubernetes from dynamically adjusting OVS thread CPU affinity on the target nodes.

@Tal-or Disabling OVS Pinning will have a negative impact if we enable workload partitioning. For our use-case we are trying to make ovn-kubernetes skip pmd threads instead: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6500.

Are there any additional reasons you want to expose this to the API?

Tal-or:
> > Setting disableOvsDynamicPinning prevents OVN-Kubernetes from dynamically adjusting OVS thread CPU affinity on the target nodes.
> 
> @Tal-or Disabling OVS Pinning will have a negative impact if we enable workload partitioning. For our use-case we are trying to make ovn-kubernetes skip pmd threads instead: [ovn-kubernetes/ovn-kubernetes#6500](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6500).
> 
> Are there any additional reasons you want to expose this to the API?

@amorenoz we're disabling the dynamic pinning so they won't interfere with the OVS DPDK.
I see that that's what you're trying to do on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6500.

But setting `disableOvsDynamicPinning: true` will give you exactly that - it will remove the enabler file which in turn shut the ovs dynamic pinning and then the `ovsvswitchd` and `ovsdb` services will be bound to reserved CPUs, and won't spill over to the other cores.

mcoquelin:
Hi @Tal-or,

I am currently testing this PR with OVS-DPDK.
I am not clear once the 'dedicated' CPU set is configured in the performance profile, how ovs-vswitchd can use it.
Given ovs-vswitchd uses ovs.slice and not dedicatedcpus.slice, how can I pin only the PMD threads to dedicated CPUs?

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
```diff
@@ -521,6 +584,17 @@ func getIRQBalanceBannedCPUsOptions() []*unit.UnitOption {
 		// WantedBy
 		unit.NewUnitOption(systemdSectionInstall, systemdWantedBy, systemdTargetMultiUser),
 	}
+	if dedicatedCPUs != "" {
+		dedicatedMask, err := components.CPUListToHexMask(dedicatedCPUs)
+		if err != nil {
+			klog.Errorf("failed to convert dedicated CPUs %q to hex mask: %v", dedicatedCPUs, err)
```
jmencak: Can this ever happen?  If so (e.g. no webhook to check user-supplied values, or insufficient check for user error input), is the error in the logs sufficient to inform the user about the failure?  I.e. the fact IRQ banning is not applied properly as user expected (`DEDICATED_CPUS` unset)?  There will probably be no error status condition set in PerformanceProfile, right?  Also, isn't the bad value (0)  going to get propagated to `default_smp_affinity`?

Edit: empty dedicated CPUs will end up with Degraded PerformanceProfile, so I guess that's what we want.  Alternatively, we may decide not to allow empty dedicated CPUs.

pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go
```diff
@@ -130,3 +132,32 @@ func IsDRAManaged(profile *performancev2.PerformanceProfile) bool {
 	}
 	return parsed
 }
+
+// ValidateDedicatedCPUsPrerequisites checks that if dedicated CPUs are set,
+// either Workload Partitioning or strict-cpu-reservation is enabled.
+func ValidateDedicatedCPUsPrerequisites(profile *performancev2.PerformanceProfile, opts *components.Options, kc *mcov1.KubeletConfig) error {
+	if profile.Spec.CPU == nil || profile.Spec.CPU.Dedicated == nil {
```
jmencak: If `*profile.Spec.CPU.Dedicated == ""`, does that mean dedicated CPUs are set or should we treat it here as we're treating `profile.Spec.CPU.Dedicated == nil` ?

assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
```diff
@@ -9,17 +9,51 @@ IRQBALANCE_CONF="${1:-/etc/sysconfig/irqbalance}"
...
+	# pad BANNED_CPUS with leading zeros to match the same number of hex chars
+	padded="${BANNED_CPUS}"
+	while [ ${#padded} -lt $(( n * 8 )) ]; do
+		padded="0${padded}"
+	done
+	# clear banned bits from each 32-bit group: result = smp & ~banned
+	result=""
+	for (( i=0; i<n; i++ )); do
+		ban="${padded:$(( i * 8 )):8}"
+		val=$(printf "%08x" $(( 0x${smp[$i]} & ~0x${ban} )))
```
jmencak, line 54: It would be nice to have some test that ensures this really does what it is supposed to do.  For example on systems with more than 64 CPUs.  This can be potentially a source of nasty silent bugs if implemented incorrectly.

test/e2e/performanceprofile/functests/utils/pods/pods.go
```diff
@@ -340,3 +341,43 @@ func CheckPODSchedulingFailed(c client.Client, pod *corev1.Pod) (bool, error) {
 	}
 	return false, nil
 }
+
+// DumpPodStateOnFailure logs detailed diagnostics for a pod that failed to
```
jmencak: `s/DumpPodStateOnFailure/DumpStateOnFailure/`

jmencak (commented):
I was looking at this from the end-user perspective.  ~~I wanted to test this, but I didn't get past~~ issues starting kubelet when configuring:

```
  kubeletConfig: 
    cpuManagerPolicy: static
    cpuManagerPolicyOptions:
      strict-cpu-reservation: "true"
```

I left some comments WRT validating user input.  ~~I also wonder if we're handling overlapping of Dedicated CPUs with other CPU types correctly.~~

pkg/performanceprofile/controller/performanceprofile/components/profile/profile.go
Tal-or: we should treat `Dedicated` as we're treating to the other optional CPU parameters, such `offlined` and `shared` I'll revise the behavior. Thanks

assets/performanceprofile/scripts/clear-irqbalance-banned-cpus.sh
Tal-or: Agreed, will do.

pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
Tal-or: it should be checked during the API validation. I'll revise this code, thanks.

Tal-or:
@coderabbitai review

Tal-or:
@coderabbitai reedit release note summary for PR description

Tal-or:
@coderabbitai review

assets/performanceprofile/scripts/ovs-dpdk-cpus-configure.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [ ! -d "$VSWITCHD_CGROUP" ]; then
+	echo "ERROR: ovs-vswitchd.service cgroup does not exist at $VSWITCHD_CGROUP" >&2
+	exit 2
+fi
+
+# The ovs-vswitchd.service drop-in sets Delegate=cpuset cpu pids, so systemd
+# enables those controllers in ovs.slice/cgroup.subtree_control automatically.
+# We only need to enable them inside the delegated subtree.
+echo "+cpuset +cpu +pids" > "$VSWITCHD_CGROUP/cgroup.subtree_control"
```
MarSik, line 30: Why not use the same method (drop-in w/ Delegate) for the inner cgroup?

MarSik:
I like it. Except the procedural cgroup setup, but I guess we lack the necessary APIs in systemd slice configs, right?

assets/performanceprofile/scripts/ovs-dpdk-cpus-configure.sh
Tal-or: Because we're not creating the inner cgroup (`ovsDpdk.slice`) using a unit file, as we do for the ovs-vswitchd.service.

we're creating the `ovsDpdk.slice` in the script simply by calling `mkdir`

Tal-or: and we cannot create the `ovsDpdk.slice` using a unit file because systemd won't allow nesting slice under a service (ovs-vswitchd.service)

MarSik: Ack.

MarSik (approved):
/lgtm

Tal-or:
Must go in before EOW

PR #1547, opened 2026-06-15
CNF-24293: Let cri-o manage the GOMAXPROC for burstable pods

MarSik:
This enables a recently merged feature of cri-o that sets GOMAXPROC for containers based on the reserved resources or to the configured minimum whatever is higher.

The effect is that golang applications will no longer default GOMAXPROC to half the available cpus and will no longer overload all the cpus when no limits are configured.

This is related to https://github.com/cri-o/cri-o/pull/9860

## Summary by CodeRabbit

* **New Features**
  * Added support for injecting a new CRI-O runtime configuration parameter to optimize CPU allocation when mixed CPUs are enabled on the system.

* **Tests**
  * Added comprehensive unit and end-to-end tests to verify the new CRI-O runtime configuration parameter is correctly applied and rendered in machine configurations.


assets/performanceprofile/configs/99-runtimes.conf
```diff
@@ -2,6 +2,7 @@
 [crio.runtime]
 infra_ctr_cpuset = "{{.ReservedCpus}}"
 {{if .SharedCpus}}shared_cpuset = "{{.SharedCpus}}"{{end}}
+min_injected_gomaxprocs = {{.MinInjectedGOMAXPROCS}}
```
yanirq, line 5: just to make sure , this goes under the main runtime hook section (and not under highPerformanceHooks section) ?

MarSik: See here https://github.com/harche/cri-o/blob/1b111f660d863a72b32f01d978d41f663f3a2815/docs/crio.conf.5.md#crioruntime-table

PR #1557, opened 2026-07-02
OCPBUGS-97809: e2e: fix broken checks and rework netqueue tests to avoid ARM ethtool blackout flakes

oblau:
This is a rework of the netqueues test suite

Investigation started from ARM CI hitting unexpected timeouts, but it
uncovered that most of these tests could never actually fail in the first
place, regardless of whether the feature worked:

- 4 of the 5 tuned-config assertions used `strings.ContainsAny` instead of
  `strings.Contains` -- ContainsAny matches on individual characters, not
  substrings, so these checks passed for almost any tuned output.

- Every test body was wrapped in a guard that was only ever true because
  BeforeEach/AfterEach forced a specific profile state around it; if the
  cluster's real starting state differed, the guard evaluated false and
  the entire test body was skipped -- reported as a PASS.

- Convergence-check failures called `Skip(...)` instead of failing,
  turning "the feature is broken" into "test skipped, no signal."

On top of that, on ARM `ethtool -L` on the br-ex uplink NIC causes ~30s of
node unreachability, and every test independently re-discovered NICs via
exec during/after profile updates, so it kept landing exec calls inside
that blackout window. The resulting timeout looked identical whether the
feature was actually broken or NIC discovery was just slow -- there was
no way to tell them apart. Between that and the bugs above masking real
signal, this was enough to justify a full rework of the test structure
rather than a handful of small patches.

Changes:

- Fixed the `strings.ContainsAny` -> `strings.Contains` bug in all 4
  affected tuned-config assertions.
- Removed the always-true/always-false profile-state guards -- test
  bodies now always execute and assert for real.
- Removed test_id:40542 as a duplicate of test_id:40308 (identical once
  the guard above was gone).
- Replaced per-test `BeforeEach`/`AfterEach` profile setup/teardown with a
  single `BeforeAll`/`AfterAll` pass -- cuts profile updates (and their
  ethtool blackouts) per suite run from 10-16 down to 5-6.
- Discover all multi-queue-capable NICs once in `BeforeAll` into a
  baseline map instead of every test (and every poll tick) rediscovering
  them; each test now builds its own scoped view of that baseline before
  triggering a profile update, so exec calls needing node access happen
  before the blackout window, not during/after it. Skips the suite
  upfront if no multi-queue-capable NICs are found.
- Convergence check now requires every NIC in the target set to reach the
  reserved CPU count, instead of returning success on the first matching
  NIC anywhere -- meaningfully stricter for the "all devices" and
  wildcard/vendor-ID tests.
- Unified Ginkgo step labels/logging across all 5 tests for consistency.

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end validation for network queue behavior on multi-queue NICs.
  * Added a more reliable NIC discovery, baseline setup, and polling strategy to confirm queue resizing using the reserved CPU baseline.
  * Expanded coverage for interface name, wildcard, negated selection, and vendor/device ID matching.
  * Validate tuned configuration by waiting for the expected udev regex updates and confirming convergence via polling ethtool combined channel counts.
  * Restores the original performance profile after test runs to keep environments consistent.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/netqueues.go
+++ b/test/e2e/performanceprofile/functests/1_performance/netqueues.go
@@ -36,6 +36,7 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
 	var profile, initialProfile *performancev2.PerformanceProfile
 	var tunedConfPath, performanceProfileName string
 	var reservedCPUCount int
+	var baselineMultiQueueNICs map[string]map[nodes.NodeInterface]int
 
 	BeforeAll(func() {
 		if discovery.Enabled() && testutils.ProfileNotFound {
@@ -79,6 +80,12 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
 
 		tunedConfPath = filepath.Join(tunedprofilesDirectory, tunedPaoProfile, "tuned.conf")
 
+		By("Discovering multi-queue capable NICs before any profile changes")
+		baselineMultiQueueNICs = discoverMultiQueueNICs(context.TODO(), workerRTNodes)
+		if len(baselineMultiQueueNICs) == 0 {
+			Skip("No multi-queue capable NICs found on worker nodes")
+		}
+
 		By("Ensuring a baseline of UserLevelNetworking=true, no device filter")
 		desiredNet := &performancev2.Net{UserLevelNetworking: ptr.To(true)}
 		if !equality.Semantic.DeepEqual(profile.Spec.Net, desiredNet) {
@@ -100,30 +107,26 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
 
 	Context("Updating performance profile for netqueues", func() {
 		It("[test_id:40308][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Network device queues Should be set to the profile's reserved CPUs count", func() {
-			nodesDevices := make(map[string]map[string]int)
 			By("To all non virtual network devices when no devices are specified under profile.Spec.Net.Devices")
-			err := checkDeviceSetWithReservedCPU(context.TODO(), workerRTNodes, nodesDevices, reservedCPUCount)
+			err := waitForNICsToMatchReservedCPU(context.TODO(), workerRTNodes, baselineMultiQueueNICs, reservedCPUCount)
 			if err != nil {
 				Skip("Skipping Test: Unable to set Network queue size to reserved cpu count")
 			}
 		})
 
 		It("[test_id:40543] Add interfaceName and verify the interface netqueues are equal to reserved cpus count.", func() {
-			nodesDevices := make(map[string]map[string]int)
-			deviceSupport, err := checkDeviceSupport(context.TODO(), workerRTNodes, nodesDevices)
-			Expect(err).ToNot(HaveOccurred())
-			if !deviceSupport {
-				Skip("Skipping Test: There are no supported Network Devices")
-			}
-			nodeName, device := getRandomNodeDevice(nodesDevices)
+			nodeName, device := getRandomNodeDevice(baselineMultiQueueNICs)
+
+			var err error
 			profile, err = profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
 			Expect(err).ToNot(HaveOccurred())
+
 			By("Enable UserLevelNetworking and add Devices in Profile")
 			profile.Spec.Net = &performancev2.Net{
 				UserLevelNetworking: ptr.To(true),
 				Devices: []performancev2.Device{
 					{
-						InterfaceName: &device,
+						InterfaceName: &device.Name,
 					},
 				},
 			}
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/netqueues.go
```diff
@@ -68,332 +83,360 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
...
+			profile.Spec.Net.Devices = []performancev2.Device{
+				{InterfaceName: &device.Name},
+			}
+
+			By("Updating the performance profile")
+			profiles.UpdateWithRetry(profile)
+
+			By("Verify the tuned profile is created on the worker-cnf nodes")
+			tunedCmd := []string{"bash", "-c",
+				fmt.Sprintf("grep devices_udev_regex %s", tunedConfPath)}
```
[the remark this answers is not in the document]
oblau: This is part of the original code - kept as is

test/e2e/performanceprofile/functests/1_performance/netqueues.go
```diff
@@ -68,332 +83,360 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
...
-					},
+
+			node, err := nodes.GetByName(nodeName)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Building matched NIC map for wildcard pattern")
+			matchedNICs := make(map[string]map[nodes.NodeInterface]int)
+			for nodeKey, nics := range baselineMultiQueueNICs {
+				for nic, channels := range nics {
+					if matched, _ := filepath.Match(devicePattern, nic.Name); matched {
```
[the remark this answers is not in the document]
oblau: Adding assertion to fail on err

test/e2e/performanceprofile/functests/1_performance/netqueues.go
```diff
@@ -68,332 +83,360 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
...
+					if nic.Name == device.Name {
+						continue
+					}
+					if expectedNICs[nodeKey] == nil {
+						expectedNICs[nodeKey] = make(map[nodes.NodeInterface]int)
+					}
+					expectedNICs[nodeKey][nic] = channels
 				}
-				profiles.UpdateWithRetry(profile)
 			}
```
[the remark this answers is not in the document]
oblau: Valid gap - code correct as is but this will increase coverage to other nodes if such exist

test/e2e/performanceprofile/functests/1_performance/netqueues.go
```diff
@@ -53,8 +64,12 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
 
 		performanceProfileName = profile.Name
 
+		reservedCPUs, err := cpuset.Parse(string(*profile.Spec.CPU.Reserved))
+		Expect(err).ToNot(HaveOccurred())
+		reservedCPUCount = reservedCPUs.Size()
+
 		tunedPaoProfile := fmt.Sprintf("openshift-node-performance-%s", performanceProfileName)
-		//Verify the tuned profile is created on the worker-cnf nodes:
+		By("Verify the tuned profile is created on the worker-cnf nodes")
```
yanirq, line 72: not sure this needs to be converted to a BY clause

oblau: The Reason i switched to By is that in ginkgo there is no section for the beforeAll logs - they show up under the under the firsts tests section. i find that this helps backtrack from logs to code. 
Given that, do you still think i should revert?

yanirq: ok we can keep it as is

test/e2e/performanceprofile/functests/1_performance/netqueues.go
```diff
@@ -68,332 +83,362 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
...
-		profile, err := profiles.GetByNodeLabels(testutils.NodeSelectorLabels)
-		Expect(err).ToNot(HaveOccurred())
-		if profile.Spec.Net == nil {
-			By("Enable UserLevelNetworking in Profile")
-			profile.Spec.Net = &performancev2.Net{
-				UserLevelNetworking: ptr.To(true),
-			}
-			By("Updating the performance profile")
+
+		By("Ensuring a baseline of UserLevelNetworking=true, no device filter")
```
yanirq, line 93: what is the device filter here?

oblau: "Device filter" = `profile.Spec.Net.Devices` which limits which NICs get netqueue tuning from `UserLevelNetworking` . Baseline profile should have UserLevelNetworking with no device filters.

test/e2e/performanceprofile/functests/1_performance/netqueues.go
```diff
@@ -68,332 +83,362 @@ var _ = Describe("[ref_id: 40307][pao]Resizing Network Queues", Ordered, Label(s
...
+			profile.Spec.Net = desiredNet
 			profiles.UpdateWithRetry(profile)
 		}
 	})
 
-	AfterEach(func() {
-		By("Reverting the Profile")
-		profile.Spec = initialProfile.Spec
-		profiles.UpdateWithRetry(profile)
+	AfterAll(func() {
```
yanirq, line 102: Notice that you changed the test to after ALL instead of each .  Can you use the same profile in every test?  i assume you override values but the ones which are not overriden will have a value (or stay empty)

oblau: Yes, intentional.

We share one profile for the whole suite. Each test that needs a device filter fully replaces `profile.Spec.Net.Devices` (it does not merge into whatever the previous test left behind). `UserLevelNetworking` stays `true` from the `BeforeAll` baseline.

The old `AfterEach` looked like isolation, but it was not. It reverted with no wait, then the next test updated immediately, so leftover `devices_udev_regex` from the previous test could still be around. It also meant a lot more profile updates (and ARM blackouts) for little gain.

So isolation now works like this: overwrite `Devices`, then wait until this test's expected state shows up. That wait greps `devices_udev_regex` and uses `strings.Contains` on that test's own pattern (iface / wildcard / `!iface` / vendor+device), then checks the matching NICs converged. Tests do not assume a clean profile at the start; they assume their update eventually wins.

`AfterAll` is only suite cleanup. It restores the original profile once at the end so we leave the cluster as we found it. It is not what keeps tests from interfering with each other.

PR #1566, opened 2026-07-19
OCPBUGS-98060: BUG-FIX Latency Test

tavital:
# OCPBUGS-98060 BUG-FIX Latency Test

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring latency test setup overhead via `LATENCY_TEST_SETUP_DELAY` (default 150s).
  * Latency e2e suite now automatically derives per-tool timeouts and applies maximum-latency thresholds per test via environment overrides.
* **Bug Fixes**
  * Improved validation and handling of invalid latency test settings, including clearer failure/timeout behavior.
  * More robust selection of suitable worker resources and conditional skipping when requirements aren’t met.
* **Documentation**
  * Updated latency test documentation to describe `LATENCY_TEST_SETUP_DELAY` and how it combines with other wait settings.

### DIFF
--- a/docs/performanceprofile/performance_controller.md
+++ b/docs/performanceprofile/performance_controller.md
@@ -117,6 +117,7 @@ You can run the container with different ENV variables, but the bare minimum is 
 
 - `LATENCY_TEST_DELAY` indicates an (optional) delay in seconds to be used between the container is created and the tests actually start. Default is zero (start immediately).
 - `LATENCY_TEST_RUNTIME` the amount of time in seconds that the latency test should run.
+- `LATENCY_TEST_TIMEOUT_BUFFER` overhead seconds (not per-tool runtime) for in-pod CPU setup/init and for the pod to reach Succeeded after the tool finishes.
 - `LATENCY_TEST_IMAGE` the image that used under the latency test.
 - `LATECNY_TEST_CPUS` the amount of CPUs the pod which run the latency test should request
 - `OSLAT_MAXIMUM_LATENCY` the expected maximum latency for all buckets in us in the oslat test.
--- a/test/e2e/performanceprofile/functests/4_latency/latency.go
+++ b/test/e2e/performanceprofile/functests/4_latency/latency.go
@@ -42,12 +42,12 @@ const (
 	hwlatdetectTestName = "hwlatdetect"
 
 	//default values
-	defaultTestDelay   = 0
-	defaultTestRuntime = "300"
-	defaultMaxLatency  = -1
-	defaultTestCpus    = -1
-	defaultTestMemory  = "1Gi"
-
+	defaultTestDelay         = 0
+	defaultTestRuntime       = "300"
+	defaultMaxLatency        = -1
+	defaultTestCpus          = -1
+	defaultTestMemory        = "1Gi"
+	defaultTestTimeoutBuffer = 150
 	//dynamic memory mode values
 	// 32Mi per requested CPU should be reasonable for the test
 	perCpuMemoryFactor = 32
@@ -57,15 +57,17 @@ const (
 )
 
 var (
-	latencyTestDelay   = defaultTestDelay
-	latencyTestRuntime = defaultTestRuntime
-	maximumLatency     = defaultMaxLatency
-	latencyTestCpus    = defaultTestCpus
-	latencyTestMemory  = defaultTestMemory
+	latencyTestDelay         = defaultTestDelay
+	latencyTestRuntime       = defaultTestRuntime
+	latencyTestTimeoutBuffer = defaultTestTimeoutBuffer
+	maximumLatency           = defaultMaxLatency
+	latencyTestCpus          = defaultTestCpus
+	latencyTestMemory        = defaultTestMemory
 )
 
 // LATENCY_TEST_DELAY delay the run of the binary, can be useful to give time to the CPU manager reconcile loop
 // to update the default CPU pool
+// LATENCY_TEST_TIMEOUT_BUFFER: extra seconds for pod CPU setup and Succeeded wait (not per-tool runtime)
 // LATENCY_TEST_RUNTIME: the amount of time in seconds that the latency test should run
 // LATENCY_TEST_CPUS: the amount of CPUs the pod which run the latency test should request
 // LATENCY_TEST_MEMORY: the amount of memory the pod which run the latency test should request
@@ -77,6 +79,9 @@ var _ = Describe("[performance] Latency Test", Ordered, func() {
 
 	BeforeEach(func() {
 		latencyTestDelay, err = getLatencyTestDelay()
+		Expect(err).ToNot(HaveOccurred())
+
+		latencyTestTimeoutBuffer, err = getLatencyTestTimeoutBuffer()
 		Expect(err).ToNot(HaveOccurred())
 
 		latencyTestCpus, err = getLatencyTestCpus()
@@ -273,6 +278,23 @@ func getLatencyTestDelay() (int, error) {
 		return val, nil
 	}
 	return defaultTestDelay, nil
+}
+
+func getLatencyTestTimeoutBuffer() (int, error) {
+	if latencyTestTimeoutBufferEnv, ok := os.LookupEnv("LATENCY_TEST_TIMEOUT_BUFFER"); ok {
+		val, err := strconv.Atoi(latencyTestTimeoutBufferEnv)
+		if err != nil {
+			return val, fmt.Errorf("the environment variable LATENCY_TEST_TIMEOUT_BUFFER has incorrect value %q, it must be a non-negative integer with maximum value of %d: %w", latencyTestTimeoutBufferEnv, math.MaxInt32, err)
+		}
+		if val < 0 || val > math.MaxInt32 {
+			return val, fmt.Errorf("the environment variable LATENCY_TEST_TIMEOUT_BUFFER has an invalid number %q, it must be a non-negative integer with maximum value of %d", latencyTestTimeoutBufferEnv, math.MaxInt32)
+		}
+		if val < defaultTestTimeoutBuffer {
+			testlog.Warningf("LATENCY_TEST_TIMEOUT_BUFFER=%d is below %d; for safe execution set it to %d or higher, as a lower value may cause timeouts", val, defaultTestTimeoutBuffer, defaultTestTimeoutBuffer)
+		}
+		return val, nil
+	}
+	return defaultTestTimeoutBuffer, nil
 }
 
 func getLatencyTestCpus() (int, error) {
@@ -489,13 +511,14 @@ func createLatencyTestPod(testPod *corev1.Pod) {
 		Expect(isEqual(RequestsCpusQuantity, latencyTestCpus)).To(BeTrue(), fmt.Sprintf("actual requests of cpus number used for the latency pod is not as set in LATENCY_TEST_CPUS, actual number is: %s", RequestsCpusQuantity))
 	}
 
-	By("Waiting another two minutes to give enough time for the cluster to move the pod to Succeeded phase")
-	podTimeout := time.Duration(timeout + latencyTestDelay + 120)
+	podTimeout := time.Duration(timeout + latencyTestDelay + latencyTestTimeoutBuffer)
+	By(fmt.Sprintf("Waiting up to %d seconds (runtime=%d, delay=%d, timeoutBuffer=%d) for the pod to reach Succeeded phase", int(podTimeout), timeout, latencyTestDelay, latencyTestTimeoutBuffer))
+
 	testPod, err = pods.WaitForPhase(context.TODO(), client.ObjectKeyFromObject(testPod), corev1.PodSucceeded, podTimeout*time.Second)
 	if err != nil {
 		logEventsForPod(testPod)
 	}
-	Expect(err).ToNot(HaveOccurred(), "pod %q did not reach %q phase; error: %v", podKey, corev1.PodSucceeded, err)
+	Expect(err).ToNot(HaveOccurred(), "pod %q did not reach %q phase; error: %v. Please Increase LATENCY_TEST_TIMEOUT_BUFFER to allow the cluster to move the pod to Succeeded phase and run again.", podKey, corev1.PodSucceeded, err)
 }
 
 func extractLatencyValues(exp string, pod *corev1.Pod) []int {
--- a/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
+++ b/test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
@@ -26,13 +26,14 @@ const (
 	cyclictest  = "cyclictest"
 	hwlatdetect = "hwlatdetect"
 	//Environment variables names
-	latencyTestDelay     = "LATENCY_TEST_DELAY"
-	latencyTestRuntime   = "LATENCY_TEST_RUNTIME"
-	maximumLatency       = "MAXIMUM_LATENCY"
-	oslatMaxLatency      = "OSLAT_MAXIMUM_LATENCY"
-	hwlatdetecMaxLatency = "HWLATDETECT_MAXIMUM_LATENCY"
-	cyclictestMaxLatency = "CYCLICTEST_MAXIMUM_LATENCY"
-	latencyTestCpus      = "LATENCY_TEST_CPUS"
+	latencyTestDelay         = "LATENCY_TEST_DELAY"
+	latencyTestRuntime       = "LATENCY_TEST_RUNTIME"
+	latencyTestTimeoutBuffer = "LATENCY_TEST_TIMEOUT_BUFFER"
+	maximumLatency           = "MAXIMUM_LATENCY"
+	oslatMaxLatency          = "OSLAT_MAXIMUM_LATENCY"
+	hwlatdetecMaxLatency     = "HWLATDETECT_MAXIMUM_LATENCY"
+	cyclictestMaxLatency     = "CYCLICTEST_MAXIMUM_LATENCY"
+	latencyTestCpus          = "LATENCY_TEST_CPUS"
 	//invalid values error messages
 	unexpectedError = "Unexpected error"
 	//incorrect values error messages
@@ -47,6 +48,8 @@ const (
 	invalidCpuNumber                   = incorrectMsgPart1 + latencyTestCpus + invalidNumber + mustBePositiveInt
 	incorrectDelay                     = incorrectMsgPart1 + latencyTestDelay + incorrectMsgPart2 + mustBeNonNegativeInt
 	invalidNumberDelay                 = incorrectMsgPart1 + latencyTestDelay + invalidNumber + mustBeNonNegativeInt
+	incorrectTimeoutBuffer             = incorrectMsgPart1 + latencyTestTimeoutBuffer + incorrectMsgPart2 + mustBeNonNegativeInt
+	invalidNumberTimeoutBuffer         = incorrectMsgPart1 + latencyTestTimeoutBuffer + invalidNumber + mustBeNonNegativeInt
 	incorrectMaxLatency                = incorrectMsgPart1 + maximumLatency + incorrectMsgPart2 + mustBeNonNegativeInt
 	invalidNumberMaxLatency            = incorrectMsgPart1 + maximumLatency + invalidNumber + mustBeNonNegativeInt
 	incorrectOslatMaxLatency           = incorrectMsgPart1 + "\"" + oslatMaxLatency + "\"" + incorrectMsgPart2 + mustBeNonNegativeInt
@@ -85,6 +88,7 @@ const (
 type latencyTest struct {
 	testDelay             string
 	testRuntime           string
+	testTimeoutBuffer     string
 	testMaxLatency        string
 	oslatMaxLatency       string
 	cyclictestMaxLatency  string
@@ -181,6 +185,9 @@ func setEnvAndGetDescription(tst latencyTest) string {
 	if tst.testDelay != "" {
 		setEnvWriteDescription(latencyTestDelay, tst.testDelay, sb, &nonDefaultValues)
 	}
+	if tst.testTimeoutBuffer != "" {
+		setEnvWriteDescription(latencyTestTimeoutBuffer, tst.testTimeoutBuffer, sb, &nonDefaultValues)
+	}
 	if tst.testRuntime != "" {
 		setEnvWriteDescription(latencyTestRuntime, tst.testRuntime, sb, &nonDefaultValues)
 	}
@@ -214,6 +221,7 @@ func setEnvWriteDescription(envVar string, val string, sb *bytes.Buffer, flag *b
 
 func clearEnv() {
 	os.Unsetenv(latencyTestDelay)
+	os.Unsetenv(latencyTestTimeoutBuffer)
 	os.Unsetenv(latencyTestRuntime)
 	os.Unsetenv(maximumLatency)
 	os.Unsetenv(oslatMaxLatency)
@@ -240,6 +248,7 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 	testSet = append(testSet, latencyTest{testDelay: "1", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 	testSet = append(testSet, latencyTest{testDelay: "60", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 	testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "5", testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{skip, skipOddCpuNumber}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
+	testSet = append(testSet, latencyTest{testTimeoutBuffer: "155", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 
 	if toolToTest != hwlatdetect {
 		testSet = append(testSet, latencyTest{testRuntime: "1", outputMsgs: []string{skip, skipMaxLatency}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
@@ -290,6 +299,11 @@ func getNegativeTests(toolToTest string) []latencyTest {
 	testSet = append(testSet, latencyTest{testRuntime: "2", testMaxLatency: "1", testCpus: fmt.Sprint(math.MaxInt32 + 1), outputMsgs: []string{invalidCpuNumber, fail}, toolToTest: toolToTest})
 	testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "-1", outputMsgs: []string{invalidCpuNumber, fail}, toolToTest: toolToTest})
 	testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "0", outputMsgs: []string{invalidCpuNumber, fail}, toolToTest: toolToTest})
+
+	// LATENCY_TEST_TIMEOUT_BUFFER must be a valid integer, reject non-numeric, and negative values.
+	testSet = append(testSet, latencyTest{testTimeoutBuffer: "J", outputMsgs: []string{incorrectTimeoutBuffer, fail}, toolToTest: toolToTest})
+	testSet = append(testSet, latencyTest{testTimeoutBuffer: fmt.Sprint(math.MaxInt32 + 1), outputMsgs: []string{invalidNumberTimeoutBuffer, fail}, toolToTest: toolToTest})
+	testSet = append(testSet, latencyTest{testTimeoutBuffer: "-5", outputMsgs: []string{invalidNumberTimeoutBuffer, fail}, toolToTest: toolToTest})
 
 	if toolToTest == oslat {
 		testSet = append(testSet, latencyTest{testRuntime: "2", oslatMaxLatency: "&", outputMsgs: []string{incorrectOslatMaxLatency, fail}, toolToTest: toolToTest})


test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -57,32 +57,36 @@ const (
...
+// LATENCY_TEST_DELAY_TIMEOUT delay the run of the binary, can be useful to give time to the CPU manager reconcile loop
 // to update the default CPU pool
-// LATENCY_TEST_RUNTIME: the amount of time in seconds that the latency test should run
+// LATENCY_TEST_RUN_TIMEOUT: the amount of time in seconds that the latency test should run
 // LATENCY_TEST_CPUS: the amount of CPUs the pod which run the latency test should request
 // LATENCY_TEST_MEMORY: the amount of memory the pod which run the latency test should request
+// LATENCY_TEST_IMAGE_PULL_TIMEOUT: seconds to wait for the image to be ready (Pulled); default 120
+// LATENCY_TEST_POD_START_TIMEOUT: seconds to wait for schedule + container start → Running (after image is ready); default 120
+// LATENCY_TEST_STARTUP_TIMEOUT: seconds after Running (excluding LATENCY_TEST_DELAY_TIMEOUT) to wait for the tool to start (default 600)
+// LATENCY_TEST_EXIT_TIMEOUT: seconds beyond LATENCY_TEST_RUN_TIMEOUT to wait for Succeeded (default 120)
```
Tal-or: This bunch of new ENV variables will complicated the API we're exposing to the users.
We're striving to simplify, i.e minimizing the learning curve for the users for them to run the tests properly.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -57,32 +57,36 @@ const (
...
-	latencyTestDelay   = defaultTestDelay
-	latencyTestRuntime = defaultTestRuntime
+	latencyTestDelayTimeout = defaultTestDelay
+	latencyTestRunTimeout   = defaultTestRuntime
 	maximumLatency     = defaultMaxLatency
 	latencyTestCpus    = defaultTestCpus
 	latencyTestMemory  = defaultTestMemory
 )
 
-// LATENCY_TEST_DELAY delay the run of the binary, can be useful to give time to the CPU manager reconcile loop
```
Tal-or, line 67: we may expand the `LATENCY_TEST_DELAY` to include the time it should take for the test to pull the image.

Tal-or: also since this ENV behaves as a "stable" API we can't simply remove it.
this will require changing in our official docs/scripts and also users already count on this variable name, so unless there's a really really good reason to change the name, that's a no go.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -57,32 +57,36 @@ const (
...
+// LATENCY_TEST_DELAY_TIMEOUT delay the run of the binary, can be useful to give time to the CPU manager reconcile loop
 // to update the default CPU pool
-// LATENCY_TEST_RUNTIME: the amount of time in seconds that the latency test should run
```
Tal-or, line 69: ditto

Tal-or (commented):
Nice work.
I left comments inside, but as a general rule we cannot remove existing API, and exposing few more fields will make the UX (which is already not great) more complex.

Maybe it worth consider to run the pod in a high level object such as deployment/cronjob? maybe those object will provide richer status about the pod progress that we can expose to the user? I'm genuinely  asking, I never tried this approach before.

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: > also since this ENV behaves as a "stable" API we can't simply remove it. this will require changing in our official docs/scripts and also users already count on this variable name, so unless there's a really really good reason to change the name, that's a no go.


ENV var changed back to original - LATENCY_TEST_DELAY

ffromani:
Thanks for the PR. In addition to checking CI issues, coderabbit review, please make sure to write meaningful, albeit possibly short, commit messages describing the changes

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: > we may expand the `LATENCY_TEST_DELAY` to include the time it should take for the test to pull the image.

`LATENCY_TEST_DELAY` makes the runner sleep for the given time so the kubelet CPU manager can finish assigning/pinning exclusive CPUs to the latency pod before oslat / cyclictest / hwlatdetect start measuring.

I don’t think mixing `LATENCY_TEST_DELAY` with `LATENCY_TEST_IMAGE_PULL_TIMEOUT` is a good idea as they serve different roles.

What do you think?

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: I have a solution for that I do understand that those four new timeout ENVs is too much surface for customers.
I have an idea:

I’ll keep the four separate waits (that’s what fixes the shared-120s bug), but drop the four public ENVs:

`LATENCY_TEST_IMAGE_PULL_TIMEOUT`
`LATENCY_TEST_POD_START_TIMEOUT`
`LATENCY_TEST_STARTUP_TIMEOUT`
`LATENCY_TEST_EXIT_TIMEOUT`

Instead I can hardcode default per-phase bases in code (used when no buffer is set):
   | Phase | Default base |
   |-------|----------------|
   | Image pull (wait 1) | **120s** |
   | Pod start → Running (wait 2) | **120s** |
   | Startup / init / CPU detect (wait 3) | **300s** |
   | Exit → Succeeded (wait 4) | **120s** |

Add a single optional env: **`LATENCY_TEST_TIMEOUT_BUFFER`** (default **0**).
Each phase budget is: `default_base + LATENCY_TEST_TIMEOUT_BUFFER`.


Wait budgets become:
wait 1:  120 + `LATENCY_TEST_TIMEOUT_BUFFER`
wait 2:  120 + `LATENCY_TEST_TIMEOUT_BUFFER`
wait 3:  300 + `LATENCY_TEST_TIMEOUT_BUFFER` + `LATENCY_TEST_DELAY`
wait 4:  `LATENCY_TEST_RUNTIME` + 120 + `LATENCY_TEST_TIMEOUT_BUFFER`

Customers still use `LATENCY_TEST_RUNTIME` and `LATENCY_TEST_DELAY` as today, 
and `LATENCY_TEST_TIMEOUT_BUFFER` will be used when the cluster needs extra slack.

What do you think?

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -460,16 +465,66 @@ func logEventsForPod(testPod *corev1.Pod) {
...
+		}
+	}
+	for _, ev := range evList.Items {
+		if ev.Reason == "Pulled" && !*loggedPulled {
+			*pulledMsg = ev.Message
+			*loggedPulled = true
+			return
+		}
+	}
+}
```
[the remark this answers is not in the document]
tavital: Noted.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -69,6 +69,10 @@ var (
 // LATENCY_TEST_RUNTIME: the amount of time in seconds that the latency test should run
 // LATENCY_TEST_CPUS: the amount of CPUs the pod which run the latency test should request
 // LATENCY_TEST_MEMORY: the amount of memory the pod which run the latency test should request
+// LATENCY_TEST_IMAGE_PULL_TIMEOUT: seconds to wait for the image to be ready (Pulled); default 120
+// LATENCY_TEST_POD_START_TIMEOUT: seconds to wait for schedule + container start → Running (after image is ready); default 120
+// LATENCY_TEST_STARTUP_TIMEOUT: seconds after Running (excluding LATENCY_TEST_DELAY) to wait for the tool to start (default 600)
+// LATENCY_TEST_EXIT_TIMEOUT: seconds beyond LATENCY_TEST_RUNTIME to wait for Succeeded (default 120)
```
Tal-or: I don't think dividing the timeout into sections is something that user should care about. 
From user perspective the test should run from start to finish without hitting timeouts due to slow infra or different hardware layout (many CPUs).

Either we consolidate all of them under a single ENV like `LATENCY_TEST_SETUP_DELAY` or we consolidate that under the existing `LATENCY_TEST_DELAY` extending the comment about its usage.

Tal-or (commented):
Please squash the commits together when resubmit.

Also, maybe it worth consider to run the pod in a high level object such as deployment/cronjob? maybe those objects will provide richer status about the pod progress that we can expose to the user? that could be an option instead of filtering specific events for the pod.

Tal-or (commented):
/assign @Tal-or

tavital:
> Also, maybe it worth consider to run the pod in a high level object such as deployment/cronjob? maybe those objects will provide richer status about the pod progress that we can expose to the user? that could be an option instead of filtering specific events for the pod.

The latency check has these stages:

1. Image download.
2. Pod startup and CPU detect.
3. Wait inside the runner before starting the tool.
4. Tool runtime.
5. Finish and exit the pod in a successful state.

I looked for a high-level object that would give richer status for that progress. There isn’t one. Pull, start, and runtime detail still come from the pod.

I also checked a few higher-level options instead of a bare pod. They don’t add real value for our test logic:

Deployment is meant to keep a workload available and always running. We create a pod, run a test, and exit, so it doesn’t help.

CronJob is meant for scheduled or recurring runs. Customers typically run latency tests once to validate hardware, not on a schedule, so it doesn’t help either.

I also checked Job which is a natural fit for one-shot work, but for a single completion, job complete is basically the same as the pod reaching succeeded. 
Pull, start, and runtime progress still come from the child pod, so a job only adds creating the job and then looking up 
that child pod to watch its status and logs, without richer status or simpler code.

I agree that a useful improvement for users is a clear progress indicator. It can sit on top of filtering specific pod events, which remains the best way to track pull, start, and completion progress, as we do now. A higher-level object is not a better alternative for that.

Tal-or:
> > Also, maybe it worth consider to run the pod in a high level object such as deployment/cronjob? maybe those objects will provide richer status about the pod progress that we can expose to the user? that could be an option instead of filtering specific events for the pod.
> 
> The latency check has these stages:
> 
> 1. Image download.
> 2. Pod startup and CPU detect.
> 3. Wait inside the runner before starting the tool.
> 4. Tool runtime.
> 5. Finish and exit the pod in a successful state.
> 
> I looked for a high-level object that would give richer status for that progress. There isn’t one. Pull, start, and runtime detail still come from the pod.
> 
> I also checked a few higher-level options instead of a bare pod. They don’t add real value for our test logic:
> 
> Deployment is meant to keep a workload available and always running. We create a pod, run a test, and exit, so it doesn’t help.
> 
> CronJob is meant for scheduled or recurring runs. Customers typically run latency tests once to validate hardware, not on a schedule, so it doesn’t help either.
> 
> I also checked Job which is a natural fit for one-shot work, but for a single completion, job complete is basically the same as the pod reaching succeeded. Pull, start, and runtime progress still come from the child pod, so a job only adds creating the job and then looking up that child pod to watch its status and logs, without richer status or simpler code.
> 
> I agree that a useful improvement for users is a clear progress indicator. It can sit on top of filtering specific pod events, which remains the best way to track pull, start, and completion progress, as we do now. A higher-level object is not a better alternative for that.

fair enough thank you for talking the time to look into it.

test/e2e/performanceprofile/functests/4_latency/cluster/performance_profile.go
```diff
@@ -0,0 +1,76 @@
...
+	if discovery.Enabled() && testutils.ProfileNotFound {
+		Skip("Discovery mode enabled, performance profile not found")
+	}
+}
+
+type PerformanceProfile struct {
+	Profile *performancev2.PerformanceProfile
+}
+
+func (p *PerformanceProfile) Load() {
```
Tal-or: `func (p *PerformanceProfile) Load()` change to  `func Load(p *PerformanceProfile)`

the general rule is that you want to use methods for stateful execution which most of the time here we don't care about.

see: https://www.sohamkamani.com/golang/functions-vs-methods/

test/e2e/performanceprofile/functests/4_latency/cluster/performance_profile.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func SkipIfProfileNotFoundInDiscovery() {
+	GinkgoHelper()
+	if discovery.Enabled() && testutils.ProfileNotFound {
+		Skip("Discovery mode enabled, performance profile not found")
+	}
+}
+
+type PerformanceProfile struct {
```
Tal-or: you can pass the performanceProfile directly to all the relevant functions and avoid the reciever function.

test/e2e/performanceprofile/functests/4_latency/config/consts.go
```diff
@@ -0,0 +1,34 @@
+package config
```
Tal-or: This is usually a good idea to have separate packages when you have multiple consumers.
This is not the case here since the sole consumer is the `latency.go` so everything can and should be inline.

test/e2e/performanceprofile/functests/4_latency/cluster/performance_profile.go
```diff
@@ -0,0 +1,76 @@
...
+	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/4_latency/config"
+	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/discovery"
+	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/profiles"
+
+	"k8s.io/utils/cpuset"
+)
+
+func SkipIfProfileNotFoundInDiscovery() {
```
Tal-or: Since this logic called just once. make this code inline is good enough and we don't need a function here.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -1,537 +1,117 @@
...
-
-		//At least one worker node should have cpu.Allocatable greater than the quantity requested by each test, else skip the test
-		workerRTNodesWithSufficientCpu := nodes.GetByCpuAllocatable(workerRTNodes, latencyTestCpus)
-		if len(workerRTNodesWithSufficientCpu) == 0 {
-			Skip("Insufficient cpu to run the test")
-
-		}
-		workerRTNode = &workerRTNodesWithSufficientCpu[0]
-
+		Expect(config.LoadLatencyTestEnvs()).To(Succeed())
```
Tal-or: why do we call this in `BeforeEach`? this is not going to change nor setup again during the whole test suite run.

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+func (p *TestPod) phase() corev1.PodPhase {
+	if p == nil || p.pod == nil {
+		return ""
+	}
+	return p.pod.Status.Phase
+}
+
+func (p *TestPod) checkCPUs() {
```
Tal-or: seems like overkill but anyway this is something that is suppose to be sit in unit-tests and usually we don't have unit-tests for test code.

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
```diff
@@ -0,0 +1,284 @@
...
+	podKey := fmt.Sprintf("%s/%s", p.pod.Namespace, p.pod.Name)
+	remaining := contextTimeLeft(ctx)
+
+	updatedPod, err := pods.WaitForPhase(ctx, client.ObjectKeyFromObject(p.pod), phase, remaining)
+	if updatedPod != nil {
+		p.pod = updatedPod
+	}
+	if err != nil {
+		p.logEvents()
+		Fail(podTimeoutMessage(ctx, phase, podKey, p, err))
```
Tal-or: can be replaced with `Expect(err).ToNot(HaveOccurred(), p.logEvents(), podTimeoutMessage(ctx, phase, podKey, p, err))`

Tal-or (commented):
General idea is good, but we'll have to refine the implementation a little bit.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -26,13 +26,14 @@ const (
...
-	cyclictestMaxLatency = "CYCLICTEST_MAXIMUM_LATENCY"
-	latencyTestCpus      = "LATENCY_TEST_CPUS"
+	latencyTestDelay      = "LATENCY_TEST_DELAY"
+	latencyTestRuntime    = "LATENCY_TEST_RUNTIME"
+	latencyTestSetupDelay = "LATENCY_TEST_SETUP_DELAY"
+	maximumLatency        = "MAXIMUM_LATENCY"
+	oslatMaxLatency       = "OSLAT_MAXIMUM_LATENCY"
+	hwlatdetecMaxLatency  = "HWLATDETECT_MAXIMUM_LATENCY"
+	cyclictestMaxLatency  = "CYCLICTEST_MAXIMUM_LATENCY"
+	latencyTestCpus       = "LATENCY_TEST_CPUS"
```
shajmakh: If we are doing a refactor we better also consider having the env vars public and use the same source in both 4_latency and 5_latency directories.  I understand and completely agree the suites can use some refactoring, however, we should split that refactoring from the actual solution of the bug, and that would better fit in a different enhancement PR (no functional changes) because the refactoring is not a prerequisite for the bug fix.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -230,32 +238,34 @@ func getValidValuesTests(toolToTest string) []latencyTest {
...
+	successGinkgoTimeout := "250s"
+	testSet = append(testSet, latencyTest{testDelay: "140", testRuntime: successRuntime, testSetupDelay: successSetupDelay, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
+	testSet = append(testSet, latencyTest{testDelay: "0", testRuntime: successRuntime, testSetupDelay: successSetupDelay, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 	// Covers the scenario of BZ 2055267 as well
-	testSet = append(testSet, latencyTest{testDelay: "0", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "6", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
+	testSet = append(testSet, latencyTest{testDelay: "0", testRuntime: successRuntime, testSetupDelay: successSetupDelay, testMaxLatency: untunedLatencyThreshold, testCpus: "6", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
```
shajmakh: for existing tests we should not override the default, because we must expect the tests to still work with the old scenarios (no regressions). To increase the coverage for the new functionality, we add more scenarios exercising the different values as it can fit.

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
```diff
@@ -0,0 +1,284 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/cpuset"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var latencyPodAnnotations = map[string]string{
```
shajmakh: this is used only in newTestPod(), let's please move it as a local function var instead.

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+}
+
+func (p *TestPod) Delete() {
+	GinkgoHelper()
+	if p == nil || p.pod == nil {
+		return
+	}
+	Expect(pods.DeleteAndSync(context.TODO(), testclient.DataPlaneClient, p.pod)).To(Succeed())
+}
```
shajmakh: no need for this wrapper, DeleteAndSync already handles pod is nil. simply use direct https://github.com/openshift/cluster-node-tuning-operator/blob/a9d25d502ca894272f88753f1bd7ecef82fb188e/test/e2e/performanceprofile/functests/utils/pods/pods.go#L100

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
```diff
@@ -0,0 +1,284 @@
...
+
+func (p *TestPod) logs() (string, error) {
+	return pods.GetLogs(testclient.K8sClient, p.pod)
+}
```
shajmakh: no need for this wrapper, GetLogs is anyhow a oneliner and is clear enough

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -1,537 +1,117 @@
...
+		By(fmt.Sprintf("latency test will run on worker node %q", workerRTNode.Name))
 	})
 
 	AfterEach(func() {
-		if latencyTestPod != nil {
-			Expect(pods.DeleteAndSync(context.TODO(), testclient.DataPlaneClient, latencyTestPod)).To(Succeed())
+		if pod != nil {
+			pod.DumpLogs()
+			pod.Delete()
+			pod = nil
```
shajmakh: `pod = nil` this should rather be in BeforeEach

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -1,537 +1,117 @@
...
+		performanceProfile.SkipIfOddCpuNumber(config.LatencyTestCpus)
+		workerRTNode = cluster.GetTestWorkerNode(config.LatencyTestCpus)
+		By(fmt.Sprintf("latency test will run on worker node %q", workerRTNode.Name))
 	})
 
 	AfterEach(func() {
-		if latencyTestPod != nil {
-			Expect(pods.DeleteAndSync(context.TODO(), testclient.DataPlaneClient, latencyTestPod)).To(Succeed())
+		if pod != nil {
+			pod.DumpLogs()
```
shajmakh: DumpLogs is only used here -> no need for a helper function

shajmakh (commented):
Thanks for the PR. I left few comments from initial review. enhancements to the existing code are more than welcome, however in this case unless the refactoring is highly needed for the bug fix, we better split any refactoring/enhancments to a followup PR while also think of a better design for the new version. Let's only have the parts that are crucial for the bug fix in this PR and update the commit please.

test/e2e/performanceprofile/functests/4_latency/cluster/performance_profile.go
tavital: noted.

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
tavital: Sure.

test/e2e/performanceprofile/functests/4_latency/cluster/performance_profile.go
tavital: 👍

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: Agreed,
loading the env vars in BeforeEach is unnecessary since they don’t change for the whole suite run.

That was already the pattern before this change.
I’ve moved the LATENCY_TEST_* loading to a one-time BeforeSuite in latency.go
BeforeEach now only keeps the per-spec setup (profile, node selection, skips).

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
tavital: Understandable,
I kept a local copy of the refactored version,
Ping me when you want me to resume that work @Tal-or @shajmakh.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
tavital: Done.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
tavital: I’ve updated the branch with the actual bug fix commit.
Happy to adjust further if needed.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -57,34 +58,49 @@ const (
...
+	latencyTestSetupDelay = defaultTestSetupDelay
+	maximumLatency        = defaultMaxLatency
+	latencyTestCpus       = defaultTestCpus
+	latencyTestMemory     = defaultTestMemory
 )
 
 // LATENCY_TEST_DELAY delay the run of the binary, can be useful to give time to the CPU manager reconcile loop
 // to update the default CPU pool
 // LATENCY_TEST_RUNTIME: the amount of time in seconds that the latency test should run
+// LATENCY_TEST_SETUP_DELAY is shared overhead (image pull, pod startup/CPU detect, exit). combined with RUNTIME+DELAY into one wait deadline. Default 150.
```
shajmakh: please remove `Default 150.` to avoid missing to update that fixed value and stay consistent with the rest of the env-var comments.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -57,34 +58,49 @@ const (
...
 // LATENCY_TEST_CPUS: the amount of CPUs the pod which run the latency test should request
 // LATENCY_TEST_MEMORY: the amount of memory the pod which run the latency test should request
+
+// LoadLatencyTestEnv loads LATENCY_TEST_* once for the suite.
+func LoadLatencyTestEnv() {
```
shajmakh: is called once in the BeforeSuite, let's inline it there then and remove the GinkgoHelper()

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -275,6 +291,27 @@ func getLatencyTestDelay() (int, error) {
...
 }
 
+func getLatencyTestSetupDelay() (int, error) {
+	if latencyTestSetupDelayEnv, ok := os.LookupEnv("LATENCY_TEST_SETUP_DELAY"); ok {
+		val, err := strconv.Atoi(latencyTestSetupDelayEnv)
+		if err != nil {
+			return val, fmt.Errorf("the environment variable LATENCY_TEST_SETUP_DELAY has incorrect value %q, it must be a non-negative integer with maximum value of %d: %w", latencyTestSetupDelayEnv, math.MaxInt32, err)
+		}
+		if val < 0 || val > math.MaxInt32 {
+			return val, fmt.Errorf("the environment variable LATENCY_TEST_SETUP_DELAY has an invalid number %q, it must be a non-negative integer with maximum value of %d", latencyTestSetupDelayEnv, math.MaxInt32)
```
shajmakh: or predefine that `var val int32` and only validate the err is not nil , and later that `val >0`

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -57,34 +58,49 @@ const (
...
+
+	latencyTestRuntime, err = getLatencyTestRuntime()
+	Expect(err).ToNot(HaveOccurred())
+
+	latencyTestSetupDelay, err = getLatencyTestSetupDelay()
+	Expect(err).ToNot(HaveOccurred())
+
+	perTest := PerTestTimeout()
+	suiteTotal := 3 * perTest // oslat + cyclictest + hwlatdetect
+	testlog.Infof("MAX TIME FOR SUITE (3 tests): %s", suiteTotal)
```
shajmakh: we shouldn't assume that 3 tests will run, because the user can choose to focus on part of them only:
https://docs.redhat.com/en/documentation/openshift_container_platform/4.8/html/scalability_and_performance/cnf-performing-platform-verification-latency-tests
so when this is removed, you can inline PerTestTimeout within the function createLatencyTestPod for better readability, there you can also log the timeout for debugging details

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -275,6 +291,27 @@ func getLatencyTestDelay() (int, error) {
...
+func getLatencyTestSetupDelay() (int, error) {
+	if latencyTestSetupDelayEnv, ok := os.LookupEnv("LATENCY_TEST_SETUP_DELAY"); ok {
+		val, err := strconv.Atoi(latencyTestSetupDelayEnv)
+		if err != nil {
+			return val, fmt.Errorf("the environment variable LATENCY_TEST_SETUP_DELAY has incorrect value %q, it must be a non-negative integer with maximum value of %d: %w", latencyTestSetupDelayEnv, math.MaxInt32, err)
+		}
+		if val < 0 || val > math.MaxInt32 {
+			return val, fmt.Errorf("the environment variable LATENCY_TEST_SETUP_DELAY has an invalid number %q, it must be a non-negative integer with maximum value of %d", latencyTestSetupDelayEnv, math.MaxInt32)
+		}
+		return val, nil
```
shajmakh, line 295: I think we should guard the value to at least 120, or at least warn the users that to increase the chances of the pod to complete gracefully they need to set the setup delay to at least 120s.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -460,27 +497,26 @@ func logEventsForPod(testPod *corev1.Pod) {
...
-	Expect(err).ToNot(HaveOccurred())
+func createLatencyTestPod(testPod *corev1.Pod, testName string) {
+	Expect(testclient.Client.Create(context.TODO(), testPod)).To(Succeed())
+
+	testlog.Infof("%s test will run on node %q", testName, testPod.Spec.NodeSelector["kubernetes.io/hostname"])
 
-	timeout, err := strconv.Atoi(latencyTestRuntime)
+	// One shared timeout for Running + Succeeded; remaining time shrinks as waits use it.
+	testTimeout := PerTestTimeout()
+	ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
```
shajmakh: should use the upper level contextinstead of background

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -460,27 +497,26 @@ func logEventsForPod(testPod *corev1.Pod) {
 	}
 }
 
-func createLatencyTestPod(testPod *corev1.Pod) {
-	err := testclient.Client.Create(context.TODO(), testPod)
-	Expect(err).ToNot(HaveOccurred())
+func createLatencyTestPod(testPod *corev1.Pod, testName string) {
+	Expect(testclient.Client.Create(context.TODO(), testPod)).To(Succeed())
```
shajmakh: pass the context and use it here and in other places in the function. should be done in a different commit in this PR.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -460,27 +497,26 @@ func logEventsForPod(testPod *corev1.Pod) {
...
+
+	testlog.Infof("%s test will run on node %q", testName, testPod.Spec.NodeSelector["kubernetes.io/hostname"])
```
shajmakh: what will be visible in the log for the node? is it the node name?

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -489,13 +525,40 @@ func createLatencyTestPod(testPod *corev1.Pod) {
...
+	GinkgoHelper()
+
+	updatedPod, err := pods.WaitForPhase(ctx, client.ObjectKeyFromObject(testPod), phase, contextTimeLeft(ctx))
 	if err != nil {
-		logEventsForPod(testPod)
+		logEventsForPod(updatedPod)
 	}
-	Expect(err).ToNot(HaveOccurred(), "pod %q did not reach %q phase; error: %v", podKey, corev1.PodSucceeded, err)
+	Expect(err).ToNot(HaveOccurred(), podTimeoutMessage(testName))
+}
```
shajmakh: why this helper? can't it be inlined?

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -460,27 +497,26 @@ func logEventsForPod(testPod *corev1.Pod) {
...
-	timeout, err := strconv.Atoi(latencyTestRuntime)
+	// One shared timeout for Running + Succeeded; remaining time shrinks as waits use it.
+	testTimeout := PerTestTimeout()
+	ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+	defer cancel()
+
+	runtimeSec, err := strconv.Atoi(latencyTestRuntime)
 	Expect(err).ToNot(HaveOccurred())
+	testlog.Infof("timeout for %s test is %s (RUNTIME=%d DELAY=%d SETUP_DELAY=%d)",
+		testName, testTimeout, runtimeSec, latencyTestDelay, latencyTestSetupDelay)
```
shajmakh: this log is misplaced, the calculation in PerTestTimeout() can be changed and the log would then be inconsistent (another reason to inline PerTestTimeout()).

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -489,13 +525,40 @@ func createLatencyTestPod(testPod *corev1.Pod) {
...
+	updatedPod, err := pods.WaitForPhase(ctx, client.ObjectKeyFromObject(testPod), phase, contextTimeLeft(ctx))
 	if err != nil {
-		logEventsForPod(testPod)
+		logEventsForPod(updatedPod)
 	}
-	Expect(err).ToNot(HaveOccurred(), "pod %q did not reach %q phase; error: %v", podKey, corev1.PodSucceeded, err)
+	Expect(err).ToNot(HaveOccurred(), podTimeoutMessage(testName))
+}
+
+func podTimeoutMessage(testName string) string {
```
shajmakh: can be inlined too

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -489,13 +525,40 @@ func createLatencyTestPod(testPod *corev1.Pod) {
...
+	if !ok || remaining < 0 {
+		return 0
+	}
+	return remaining
+}
+
+func waitForPhase(ctx context.Context, testPod *corev1.Pod, testName string, phase corev1.PodPhase) {
+	GinkgoHelper()
+
+	updatedPod, err := pods.WaitForPhase(ctx, client.ObjectKeyFromObject(testPod), phase, contextTimeLeft(ctx))
```
shajmakh: `contextTimeLeft(ctx)` not sure if this is good or is more restrictive than before. the code is now reducing the waiting time for pod started to run, and waits for it to finish with the remaining timeout. before this change the code was separately waiting for the pod to start( go running) which is part of the product (test in this case) we are offering, so without it there would be no running pod. 
the goal of the test is to run the pod and let it finish gracefully, if it took few extra seconds it's fine, I don't think we should be strict here.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -489,13 +525,40 @@ func createLatencyTestPod(testPod *corev1.Pod) {
...
 	if err != nil {
-		logEventsForPod(testPod)
+		logEventsForPod(updatedPod)
 	}
-	Expect(err).ToNot(HaveOccurred(), "pod %q did not reach %q phase; error: %v", podKey, corev1.PodSucceeded, err)
+	Expect(err).ToNot(HaveOccurred(), podTimeoutMessage(testName))
+}
+
+func podTimeoutMessage(testName string) string {
+	runtimeSec, _ := strconv.Atoi(latencyTestRuntime)
```
shajmakh: 3 times processing the same value is fine but in this context we can do it only once

test/e2e/performanceprofile/functests/4_latency/test_suite_latency_test.go
```diff
@@ -41,6 +42,7 @@ func init() {
 }
 
 var _ = BeforeSuite(func() {
+	latencytests.LoadLatencyTestEnv()
```
shajmakh: `Expect(testclient.ClientsEnabled).To(BeTrue())`
should be before any other check

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: The vars `latencyTestDelay`, `latencyTestCpus`, `latencyTestRuntime`, `latencyTestSetupDelay` are used only in `latency.go` (package __latency).
Inlining the load in test_suite_latency_test.go (package __latency_test) would require exporting them so the suite can set state that only __latency uses, which is redundant.
I prefer leaving them unexported in latency.go and calling LoadLatencyTestEnv() from BeforeSuite unless you see a reason to export them.

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: Fixed.

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: Predefining `math.MaxInt32` alone doesn’t let us drop the range check, we still need to validate that val is within bounds.
Could you clarify the change you had in mind?

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: Yes, it's the node name.

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: Setting the value to 120s leaves the issue unresolved,
I recommend adding a buffer and set it for at least 150s.
What do you think?

Tal-or (commented):
overall looking good. 
please address the rest of the comments from the reviewers.
also it would be nice to add tests for the new argument under https://github.com/openshift/cluster-node-tuning-operator/tree/main/test/e2e/performanceprofile/functests/5_latency_testing

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: This code doesn't exist in last commit.
Outdated.

test/e2e/performanceprofile/functests/4_latency/cluster/performance_profile.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/testpod/test_pod.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: This code doesn't exist in last commit.
Outdated

tavital:
> overall looking good. please address the rest of the comments from the reviewers. also it would be nice to add tests for the new argument under https://github.com/openshift/cluster-node-tuning-operator/tree/main/test/e2e/performanceprofile/functests/5_latency_testing

Tests added.

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: This code doesn't exist in last commit.
Outdated

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: Which approach do you think is more appropriate?
Per-test context, or one shared context for the whole Ordered suite (oslat / cyclictest / hwlatdetect)?

Tal-or (commented):
@shajmakh PTAL at:
https://github.com/openshift/cluster-node-tuning-operator/pull/1566/changes/52b21b397c097e935d2503128dde56f24c649ed6#r3672411944 
and
https://github.com/openshift/cluster-node-tuning-operator/pull/1566/changes/52b21b397c097e935d2503128dde56f24c649ed6#r3674394951

test/e2e/performanceprofile/functests/4_latency/latency.go
shajmakh: good point

test/e2e/performanceprofile/functests/4_latency/latency.go
shajmakh: right we would still need some kind of type/range check because `strconv.Atoi` returns int.

docs/performanceprofile/performance_controller.md
```diff
@@ -117,6 +117,7 @@ You can run the container with different ENV variables, but the bare minimum is
 
 - `LATENCY_TEST_DELAY` indicates an (optional) delay in seconds to be used between the container is created and the tests actually start. Default is zero (start immediately).
 - `LATENCY_TEST_RUNTIME` the amount of time in seconds that the latency test should run.
+- `LATENCY_TEST_SETUP_DELAY` overhead seconds included in the per-test timeout to Pod to reach Succeded.
```
shajmakh: can we rename it something else than delay? it is easy to confuse between the difference between the old delay and he new one, 
@Tal-or mentioned `LATENCY_TEST_TIMEOUT_BUFFER`, can we use that?

test/e2e/performanceprofile/functests/4_latency/latency.go
shajmakh, line 71: since the new env is not specific to one tool, we can add a comment for it here.

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -275,6 +279,23 @@ func getLatencyTestDelay() (int, error) {
 	return defaultTestDelay, nil
 }
 
+func getLatencyTestSetupDelay() (int, error) {
```
shajmakh: if you consider renaming the env-var please also rename the function title

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
```diff
@@ -241,6 +249,11 @@ func getValidValuesTests(toolToTest string) []latencyTest {
 	testSet = append(testSet, latencyTest{testDelay: "60", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 	testSet = append(testSet, latencyTest{testRuntime: "2", testCpus: "5", testMaxLatency: untunedLatencyThreshold, outputMsgs: []string{skip, skipOddCpuNumber}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
 
+	// LATENCY_TEST_SETUP_DELAY is added to podTimeout (runtime + delay + setupDelay) when waiting for Succeeded.
+	// Verify the default (150) and a higher value (200) are accepted and the tool still succeeds.
+	testSet = append(testSet, latencyTest{testSetupDelay: "150", testRuntime: successRuntime, testMaxLatency: untunedLatencyThreshold, testCpus: "4", outputMsgs: []string{success}, toolToTest: toolToTest, ginkgoTimeout: successGinkgoTimeout})
```
shajmakh: this is not the way we should verify that the default is 150, because this test is setting the env-var to 150 so it does not trigger the default evaluation. 
to test the default timeout for a test pod we need a test with a different design and special calulation and can also be flaky.
Or make up a scenario (not sure if that exists TBH) in which executing the latency tool actually takes more time than the default formula, and ensure that when increasing the timeout buffer , it is reflected on the waiting time for the test to finish (regardless the result of the test). 
if that can be tested then we covered this case, if the scenario does not exist or way beyond resources that we have, then let's keep the current tests.

shajmakh: To be clear, this isn't blocker for this PR, the scenario can be investigated post merging it also because QE would need to verify the fix

shajmakh:
@tavital Let's lastly improve the commit message/body. It isn't clear on the commit what why was the hardcoded value a problem and how the new fix solves it.
/jira refresh

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
tavital: A large-CPU cluster is not always available for verification. On those nodes, in-pod CPU init can consume most of the Succeeded-wait buffer even after the latency tool finishes.

To exercise that case without large-CPU hardware, temporarily replace the tool command in getLatencyTestPod (test/e2e/performanceprofile/functests/4_latency/latency.go) with a sleep that stands in for CPU-init time:

`Command: []string{
						"/bin/sh",
						"-c",
						"sleep 160",
					},`

Because 160 is greater than the default LATENCY_TEST_TIMEOUT_BUFFER (150), the Succeeded wait should time out.

Then rebuild and run from the repo root:

`make dist-latency-tests`
`./build/_output/bin/latency-e2e.test -ginkgo.v -ginkgo.focus=oslat -ginkgo.timeout=600s`

You should see a failure like:

`pod "performance-addon-operators-testing/oslat-..." did not reach "Succeeded" phase; error: context deadline exceeded. Please Increase LATENCY_TEST_TIMEOUT_BUFFER to allow the cluster to move the pod to Succeeded phase and run again`

If you instead set sleep below 160 (within the default buffer), that timeout error should not appear.

shajmakh: > Decreasing the number lower than 150 will not trigger that error.

I think it you meant the opposite? meaning having a greater buffer will not trigger the error, and by that you are verifying that the fix of adding the buffer time is working, is that true? 

Please add the command you used to test the updates version of latency.go with the pod sleep and the output.

shajmakh:
Thanks agian for the updates. Please sign the commit (`git commit -s --amend`)

test/e2e/performanceprofile/functests/4_latency/latency.go
```diff
@@ -57,15 +57,17 @@ const (
...
+	latencyTestRuntime       = defaultTestRuntime
+	latencyTestTimeoutBuffer = defaultTestTimeoutBuffer
+	maximumLatency           = defaultMaxLatency
+	latencyTestCpus          = defaultTestCpus
+	latencyTestMemory        = defaultTestMemory
 )
 
 // LATENCY_TEST_DELAY delay the run of the binary, can be useful to give time to the CPU manager reconcile loop
 // to update the default CPU pool
+// LATENCY_TEST_TIMEOUT_BUFFER: extra seconds in the wait timeout for in-pod CPU setup/init and for the latency test pod to reach Succeeded after the tool finishes. it's not for per-tool execution time.
```
shajmakh: nit: make the comment shorter or divide it into lines like done for LATENCY_TEST_DELAY.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
shajmakh: Additionally, I see there is no longer a positive test, please re-add that using a value different than the default.

test/e2e/performanceprofile/functests/4_latency/latency.go
tavital: Done.

test/e2e/performanceprofile/functests/5_latency_testing/latency_testing.go
tavital: > > Decreasing the number lower than 150 will not trigger that error.
> 
> I think it you meant the opposite? meaning having a greater buffer will not trigger the error, and by that you are verifying that the fix of adding the buffer time is working, is that true?
> 
> Please add the command you used to test the updates version of latency.go with the pod sleep and the output.

I meant `160` which is the sleep inside the pod. nice catch.

shajmakh:
/lgtm
Thanks for all the updates.

tavital:
jira/valid-reference

PR #1571, opened 2026-07-23
OCPBUGS-127113: E2E: Fix no-op Get() assertions in mixedcpus

oblau:
This PR is mixedcpus only. Split from the old 3-commit stack:
- Offlined → #1629
- PPC → #1638

`Expect(client.Get(...))` without a matcher is a Gomega no-op — API errors were swallowed (3 sites). Now `Expect(...Get...).To(Succeed())`.

Also:
- Log printed Isolated twice; now Isolated and Shared.
 - When expanding Shared from 1 CPU to a new 2-CPU set, Union the previous Shared CPU back into Isolated so it is not left unassigned.
- Keep Guaranteed pods at cpu:1. Always set full-pcpus-only=false in setup() (not only when Capacity.Cpu()<=4) so 4-CPU and larger CI can both schedule.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mixed-CPU profile setup and teardown reliability with stronger validation.
  * Corrected CPU set handling during exec-affinity tests so CPUs remain assigned appropriately.
  * Updated shared CPU logging to accurately reflect the CPUs in use.
  * Ensured profile configuration consistently applies the expected full-CPU allocation setting.

### DIFF
--- a/test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
+++ b/test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
@@ -3,16 +3,16 @@ package __performance_ppc
 import (
 	"fmt"
 	"os/exec"
-	"regexp"
+	"path/filepath"
 	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
 	performancev2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/profilecreator"
 	testutils "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils"
 	"github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/label"
-	testlog "github.com/openshift/cluster-node-tuning-operator/test/e2e/performanceprofile/functests/utils/log"
 	"k8s.io/utils/cpuset"
 	"sigs.k8s.io/yaml"
 )
@@ -89,9 +89,12 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
 			podmanArgs := append(defaultArgs, cmdArgs...)
 			session, err := ppcIntgTest.PodmanAsUserBase(podmanArgs, false, false)
 			Expect(err).ToNot(HaveOccurred(), "Podman command failed")
+
 			output := session.Wait(20).Out.Contents()
+			Expect(session).Should(gexec.Exit(0))
+
 			err = yaml.Unmarshal(output, pp)
-			Expect(err).ToNot(HaveOccurred(), "Unable to marshal the ppc output")
+			Expect(err).ToNot(HaveOccurred(), "Unable to unmarshal the ppc output")
 			reservedCpus, err := cpuset.Parse(string(*pp.Spec.CPU.Reserved))
 			Expect(err).ToNot(HaveOccurred(), "Unable to parse cpus")
 			totalReservedCpus := reservedCpus.Size()
@@ -99,8 +102,8 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
 			Expect(*pp.Spec.RealTimeKernel.Enabled).To(BeTrue())
 			Expect(*pp.Spec.WorkloadHints.RealTime).To(BeTrue())
 			Expect(*pp.Spec.NUMA.TopologyPolicy).To(Equal("restricted"))
-			Eventually(session).Should(gexec.Exit(0))
 		})
+
 		It("[test_id:41405] Verify PPC script fails when the splitting of reserved cpus and single numa-node policy is specified", func() {
 			cmdArgs := []string{
 				fmt.Sprintf("%s:%s:z", mustgatherDir, mustgatherDir),
@@ -116,14 +119,12 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
 			podmanArgs := append(defaultArgs, cmdArgs...)
 			session, err := ppcIntgTest.PodmanAsUserBase(podmanArgs, false, false)
 			Expect(err).ToNot(HaveOccurred(), "Podman command failed")
+
 			output := session.Wait(20).Err.Contents()
-			errString := "Error: failed to obtain data from flags not appropriate to split reserved CPUs in case of topology-manager-policy: single-numa-node"
-			ok, err := regexp.MatchString(errString, string(output))
-			Expect(err).ToNot(HaveOccurred())
-			if ok {
-				testlog.Info(errString)
-			}
-			Eventually(session).Should(gexec.Exit(1))
+			Expect(session).Should(gexec.Exit(1))
+
+			errString := "not appropriate to split reserved CPUs in case of topology-manager-policy: single-numa-node"
+			Expect(string(output)).To(ContainSubstring(errString), "expected error:\n%q\ngot:\n%s", errString, output)
 		})
 
 		It("[test_id:41419] Verify PPC script fails when reserved cpu count is 2 and requires to split across numa nodes", func() {
@@ -140,14 +141,12 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
 			podmanArgs := append(defaultArgs, cmdArgs...)
 			session, err := ppcIntgTest.PodmanAsUserBase(podmanArgs, false, false)
 			Expect(err).ToNot(HaveOccurred(), "Podman command failed")
+
 			output := session.Wait(20).Err.Contents()
-			errString := "Error: failed to compute the reserved and isolated CPUs: can't allocate odd number of CPUs from a NUMA Node"
-			ok, err := regexp.MatchString(errString, string(output))
-			Expect(err).ToNot(HaveOccurred(), "did not fail with Expected:%s failure", errString)
-			if ok {
-				testlog.Info(errString)
-			}
-			Eventually(session).Should(gexec.Exit(1))
+			Expect(session).Should(gexec.Exit(1))
+
+			errString := "can't allocate odd number of CPUs from a NUMA Node"
+			Expect(string(output)).To(ContainSubstring(errString), "expected error:\n%q\ngot:\n%s", errString, output)
 		})
 
 		It("[test_id:41420] Verify PPC script fails when reserved cpu count is more than available cpus", func() {
@@ -165,14 +164,13 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
 			podmanArgs := append(defaultArgs, cmdArgs...)
 			session, err := ppcIntgTest.PodmanAsUserBase(podmanArgs, false, false)
 			Expect(err).ToNot(HaveOccurred(), "Podman command failed")
+
 			output := session.Wait(20).Err.Contents()
-			errString := "Error: failed to compute the reserved and isolated CPUs: please specify the reserved CPU count in the range [1,3]"
-			ok, err := regexp.MatchString(errString, string(output))
-			Expect(err).ToNot(HaveOccurred(), "did not fail with Expected:%s failure", errString)
-			if ok {
-				testlog.Info(errString)
-			}
-			Eventually(session).Should(gexec.Exit(1))
+			Expect(session).Should(gexec.Exit(1))
+
+			errString := fmt.Sprintf("please specify the reserved CPU count in the range [1,%d]",
+				maxReservedCPUCountFromMustGather(mustgatherDir, mcpName))
+			Expect(string(output)).To(ContainSubstring(errString), "expected error:\n%q\ngot:\n%s", errString, output)
 		})
 
 		It("[test_id: 54187] PPC generates profile with PerPodPowerManagement workload hint", func() {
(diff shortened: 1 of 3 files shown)


test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -320,7 +320,8 @@ var _ = Describe("Mixedcpus", Ordered, Label(string(label.MixedCPUs)), func() {
 				By(fmt.Sprintf("Waiting when %s finishes updates", poolName))
 				profilesupdate.WaitForTuningUpdated(context.TODO(), profile)
 
-				Expect(testclient.ControlPlaneClient.Get(ctx, client.ObjectKeyFromObject(profile), profile))
+				err := testclient.ControlPlaneClient.Get(ctx, client.ObjectKeyFromObject(profile), profile)
```
Tal-or: The new pattern that's being used when there's only `error` return as value is:

`Expect(testclient.ControlPlaneClient.Get(ctx, client.ObjectKeyFromObject(profile), profile)).To(Succeed())`

test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
```diff
@@ -120,9 +119,7 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
 			errString := "Error: failed to obtain data from flags not appropriate to split reserved CPUs in case of topology-manager-policy: single-numa-node"
 			ok, err := regexp.MatchString(errString, string(output))
 			Expect(err).ToNot(HaveOccurred())
-			if ok {
```
Tal-or, line 123: That's a change in the behavior. it makes the test more restrict. is that what we want here?

oblau: This is changing behavior - but isn't this the intention here? 
If we are not asserting on `ok` then i fail to see why we need the `errString`
and `regexp.MatchString(errString, string(output))` to begin with. 
Maybe im missing something here.

Is the goal to make sure the correct error message appears, or that PPC script fails in general?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
oblau: Ok first time i run in to this, will amend my commit

test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
```diff
@@ -169,9 +164,7 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
 			errString := "Error: failed to compute the reserved and isolated CPUs: please specify the reserved CPU count in the range [1,3]"
 			ok, err := regexp.MatchString(errString, string(output))
 			Expect(err).ToNot(HaveOccurred(), "did not fail with Expected:%s failure", errString)
-			if ok {
-				testlog.Info(errString)
-			}
+			Expect(ok).To(BeTrue(), "expected error %q not found in output: %s", errString, output)
```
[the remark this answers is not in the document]
Tal-or: interesting catch @oblau worth looking into that

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -320,7 +320,8 @@ var _ = Describe("Mixedcpus", Ordered, Label(string(label.MixedCPUs)), func() {
...
+				Expect(err).ToNot(HaveOccurred())
 				testlog.Infof("new isolated CPU set=%q\nnew shared CPU set=%q", string(*profile.Spec.CPU.Isolated), string(*profile.Spec.CPU.Isolated))
```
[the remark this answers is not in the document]
Tal-or: @oblau the bot found bug. could you please fix it next time you resubmit?

test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
Tal-or: After reevaluation I think this change is ok.
the only change I would add is in the message to make it more clear:
```
Expect(ok).To(BeTrue(), "expected error %q to be found in output: %s", errString, output)

File: test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (oblau) [reply test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go]: Is the intention that the output will contain explicitly [1,3]? 
If so im not sure `regexp.MatchString` is the right tool to begin with. 
Would `Expect(string(output)).To(ContainSubstring(errString))` be better here? we already use `Expect(ppcErrorString).To(ContainSubstring("...")` on `test_id:41420`

File: test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (oblau) [reply test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go]: Done, now logging  `string(*profile.Spec.CPU.Isolated), string(*profile.Spec.CPU.Shared)`

File: test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (Tal-or) [reply test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go]: @oblau sounds good.  if you're running it locally, it's worth printing `string(output)` and see the full error message.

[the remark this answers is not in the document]
Comment (oblau) [reply test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go]: Verified locally. ContainSubstring now matches the leaf error text directly, so errString aligns with what the command actually prints (e.g. "please specify the reserved CPU count in the range [1,%d]"). No regex interpretation, no mismatch from outer error wrapping. See commit 0b00bd1b3

File: test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go [hunk 5]
```diff
@@ -577,13 +572,16 @@ var _ = Describe("Mixedcpus", Ordered, Label(string(label.MixedCPUs)), func() {
 
 					coreSiblings, err := nodes.GetCoreSiblings(ctx, workerRTNode)
 					Expect(err).ToNot(HaveOccurred())
+					// When Shared already has 1 CPU and we need 2, we replace Shared with a new pair from
```
Tal-or, line 575: But with the new logic we always have 2 shared CPUs by default, so why the comment says 
"When Shared already has 1 CPU and we need 2"?

Besides the wrong comment, we don't even need to change anything, since now 2 CPUs is the default.

oblau: Thanks for flagging this, I think there's a small mix-up, my bad if the earlier phrasing wasn't clear.

The cpu:1 -> cpu:2 change in this PR was to the pods' resource requests, for SMT alignment reasons. It's unrelated to profile.Spec.CPU.Shared, which is set separately by setup() at the top of the file, and still defaults to 1 CPU there. So updatedShared.Size() < 2 is still true in the normal case, and this code (with the oldShared/Union fix) still runs.

Worth noting too: even if we changed setup() to assign 2 shared CPUs, that only covers the case where setup() itself enables mixed CPUs. If a cluster arrives with mixed CPUs already enabled and only 1 shared CPU, setup() skips that assignment entirely, so we'd still need this fallback to grow Shared to 2.

On the comment wording: Shared could in theory be 0 as well as 1, but that's actually unreachable in practice, since IsMixedCPUsEnabled requires Shared to already be non-empty. So the comment holds as written, only the "1 CPU" case ever really happens here.
Happy to add a short note in the comment mentioning that.

Tal-or: > If a cluster arrives with mixed CPUs already enabled and only 1 shared CPU, setup() skips that assignment entirely, so we'd still need this fallback to grow Shared to 2.

This is a well defined testing environment, we're not expecting prior state and fully control it.

IOW, if we're moving from a single CPU for testing mixed-cpus feature, lets do it properly and get rid of the other ugly workarounds we had in the code, (which increasing from 1 -> 2 was one of them).

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -843,19 +841,14 @@ func setup(ctx context.Context) func(ctx2 context.Context) {
...
-			}
-			testlog.Infof("canceling SMT alignment for nodes under profile %q", profile.Name)
-			updateNeeded = true
-		}
+	// CI may have only 4 CPUs - Typical layout: reserved + shared + infra + test pod (4 cores).
+	// Pods request cpu:1; SMT alignment will not schedule that.
+	// Cancel full-pcpus-only. Apply it always so larger nodes are not a different kubelet policy.
+	profile.Annotations = map[string]string{
+		"kubeletconfig.experimental": "{\"cpuManagerPolicyOptions\": {\"full-pcpus-only\": \"false\"}}",
 	}
```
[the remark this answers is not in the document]
oblau: Leaving the replace as-is.

This is the same map assign main already used for SMT cancel; we only dropped the `Capacity.Cpu()<=4` gate so 4-CPU and larger nodes get the same kubelet policy.
Teardown reverts to initialProfile, so other annotations are not left wiped after the suite.

test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
oblau: ppr changes are not longer part of this pr- moved them to a new pr   #1638

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -50,13 +50,9 @@ const (
 	kubeletMixedCPUsConfigFile = "/etc/kubernetes/openshift-workload-mixed-cpus"
 	crioRuntimesConfigFile     = "/etc/crio/crio.conf.d/99-runtimes.conf"
 	sharedCpusResource         = "workload.openshift.io/enable-shared-cpus"
-	// the minimal number of cores for running the test is as follows:
-	// reserved = one core, shared = one core, infra workload = one core, test pod = one core - 4 in total
-	// smt alignment won't allow us to run the test pod with a single core, hence we should cancel it.
-	numberOfCoresThatRequiredCancelingSMTAlignment = 4
```
Tal-or, line 56: Maybe it would be better to set the pod request dynamically?
i.e if CPU capacity is 4 then we'll disable SMT alignment and request a single CPU for GU pod, otherwise keep SMT alignment and asks for even number of CPUs (2 * N) for the pod.

The main reason for that is to keep test closer to production environment on which the standard is to have SMT alignment enabled and request for even number of CPUs. WDYT?

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -151,7 +146,7 @@ var _ = Describe("Mixedcpus", Ordered, Label(string(label.MixedCPUs)), func() {
 		When("workloads requests access for shared cpus", func() {
 			It("verify cpu load balancing still works with mixed cpus", func() {
 				rl := &corev1.ResourceList{
-					corev1.ResourceCPU:    resource.MustParse("1"),
+					corev1.ResourceCPU:    resource.MustParse("2"),
```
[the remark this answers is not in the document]
oblau: Outdated. Pods are back to cpu:1. This extra isolated-capacity guard is not needed.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
oblau: Reverted the pod cpu:1 → cpu:2 change after finding CI is 4 CPUs, not 6.

Test here still needs 2 shared cpus so BeforeEach growing profile Shared from 1 to 2. 
On 4-CPU CI we hit the Skip here (Isolated too small after setup()).

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
oblau: Makes sense for to make the tests more production-like, however we are not running this test d\s, so we are limited to the 4 cpus on u\s. 

setup() always moves one Isolated CPU into Shared. If we start asking the pod for 2 CPUs, we also have to change Isolated, Shared, and SMT, or the pod will not schedule.
I think this is too big for this PR. I would keep it as is. We can do a dedicated change later if we still want it.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -577,13 +573,15 @@ var _ = Describe("Mixedcpus", Ordered, Label(string(label.MixedCPUs)), func() {
 
 					coreSiblings, err := nodes.GetCoreSiblings(ctx, workerRTNode)
 					Expect(err).ToNot(HaveOccurred())
+					// New Shared is taken from Isolated. Put previous Shared back into Isolated so those CPUs are not left in neither set.
+					oldShared := updatedShared
```
Tal-or: what is see here is put updatedShared into oldShared and not what the is stated in the comment.

test/e2e/performanceprofile/functests/11_mixedcpus/mixedcpus.go
```diff
@@ -838,30 +835,18 @@ func setup(ctx context.Context) func(ctx2 context.Context) {
...
-		if numOfCores <= numberOfCoresThatRequiredCancelingSMTAlignment {
-			profile.Annotations = map[string]string{
-				"kubeletconfig.experimental": "{\"cpuManagerPolicyOptions\": {\"full-pcpus-only\": \"false\"}}",
-			}
-			testlog.Infof("canceling SMT alignment for nodes under profile %q", profile.Name)
-			updateNeeded = true
-		}
+	// CI may have only 4 CPUs - Typical layout: reserved + shared + infra + test pod (4 cores).
+	// Pods request cpu:1; SMT alignment will not schedule that.
+	// Cancel full-pcpus-only. Apply it always so larger nodes are not a different kubelet policy.
```
Tal-or: it's not clear from the comment that SMT alignment and `full-pcpus-only` is the same thing.

PR #1575, opened 2026-08-04
NO-JIRA: e2e: fix: restart tuned pod after irqbalance test to prevent cpuset pollution

Tal-or:
The irqbalance test `Should not overwrite the banned CPU set on tuned restart` creates a guaranteed pod (exclusive CPUs), then restarts the tuned pod while that guaranteed pod is still running.
The new tuned process starts with a narrowed CPU affinity mask (missing the exclusive CPUs).
When the guaranteed pod is later cleaned up, the cgroup cpuset expands but the tuned process's sched_setaffinity mask stays narrow.

If the `cpu_management.go` `Ordered` block runs after `irqbalance.go`, test `test_id:87722` reads the tuned pod's stale process affinity via taskset and fails because it doesn't match the full online CPU set.

Fix: restart the tuned pod in the irqbalance test's defer cleanup after the guaranteed pod is deleted, so the tuned process picks up the current (full) default cpuset.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved performance test cleanup by restoring CPU affinity and recreating the tuned pod after checks.
  * Added readiness validation for the replacement pod before continuing, improving test reliability and reducing failures caused by incomplete pod recovery.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/irqbalance.go
@@ -223,6 +223,7 @@ var _ = Describe("[performance] IRQBalance", Ordered, func() {
 			err = testclient.DataPlaneClient.Create(context.TODO(), testpod)
 			Expect(err).ToNot(HaveOccurred())
 			defer func() {
+				GinkgoHelper()
 				if testpod != nil {
 					testlog.Infof("deleting pod %q", testpod.Name)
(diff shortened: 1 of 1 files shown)


Tal-or:
/retest  
infra issue

MarSik:
Hmm, shouldn't the real fix be in the place where tuned detects the available cpus? Because if Tuned pod starts after a guaranteed pod on a real cluster deployment and this happens then we will have a misconfigured system too, right?

PR #1582, opened 2026-08-11
OCPBUGS-105476: Update PPC help description

tavital:
# OCPBUGS-105476: Update PPC help description 
Align `ppc -h` and `ppc info -h` with the current CLI: 
describe info as cluster topology from must-gather, 
document default text output and `--json`/`--text`, 
and drop the stale `log,json` wording.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that the `info` command requires a must-gather directory path.
  * Documented that plain-text output is used by default when no format option is specified.

PR #1584, opened 2026-08-13
OCPBUGS-105458: Use Add() instead of AddRateLimited() for routine Profile enqueues

jmencak:
Change 1:

workqueue.DefaultTypedControllerRateLimiter's AddRateLimited() draws from a single 10qps/burst-100 token bucket shared by every item in the queue, regardless of kind or key.  Cascade/trigger enqueues (Pod/Node change to Profile update, per-Profile bootstrap loop, initial bootstrap event) used AddRateLimited() for ordinary enqueuing rather than error retries. During a burst of ~100 nodes joining, this exhausted the shared bucket, delaying later Profile syncs by minutes even though they were never dropped, just queued behind the backlog.

Switch these call sites to Add(), which is unthrottled and still deduped by key via the delaying queue's dirty set.  Keep AddRateLimited() only for its intended purpose: the retry-after- sync()-error path, where per-key exponential backoff is correct.

Change 2:

validateTunedCRs() called UpdateStatus() on every Tuned CR on every invocation, even when the computed status conditions were identical to what's already stored.  Compare the newly computed status against the cached copy with reflect.DeepEqual() and skip the UpdateStatus() call when nothing semantically changed.

Resolves: OCPBUGS-105458

## Summary by CodeRabbit

- **Performance**
  - Reduced unnecessary rate limiting during routine event processing, improving responsiveness for profile updates.
  - Avoided redundant status updates when no meaningful configuration change occurred.

- **Bug Fixes**
  - Preserved retry-based reprocessing after synchronization failures.
  - Improved handling of initial setup and updates triggered by workload or infrastructure changes.
  - Reduced unnecessary system activity when configuration status remains unchanged.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -125,7 +125,7 @@ func NewController() (*Controller, error) {
 	controller.mcLabelsAcrossMCP = map[string]bool{}
 
 	// Initial event to bootstrap CR if it doesn't exist.
-	controller.workqueue.AddRateLimited(wqKey{kind: wqKindTuned, name: tunedv1.TunedDefaultResourceName})
+	controller.workqueue.Add(wqKey{kind: wqKindTuned, name: tunedv1.TunedDefaultResourceName})
 
 	controller.clients.Kube, err = kubeset.NewForConfig(controller.kubeconfig)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jmencak:
/hold
re-evaluating this solution

MarSik (approved):
/lgtm

jmencak:
> /retest-required

@yanirq, I see the similar failures in  https://github.com/openshift/cluster-node-tuning-operator/pull/1582 .  3 PAO test lanes seem to be failing and it will probably require some CI investigation on Sunday/Monday.

jmencak:
Let's debug in the CI:
/test e2e-gcp-pao

PR #1585, opened 2026-08-14
CNF-25172: E2E: adjust timer_migration expectation for wrapped performance profiles

SargunNarula:
Test 28466 hardcodes `kernel.timer_migration=1` from the NTO `openshift-node-performance profile`. On clusters where ran-du-performance wraps the PP via `include=`, network-latency in the chain sets `timer_migration=0` and that value wins over NTO's override to 1.

Reuse the include-chain helpers from test 37127 to detect wrapper Tuned profiles and `expect timer_migration=0` when a wrapper is present.

## Summary by CodeRabbit

* **Tests**
  * Improved performance profile validation across directly selected and wrapped Tuned profiles.
  * Added checks for expected timer migration settings based on the active profile.
  * Tests now determine profile names dynamically, handle unavailable include-chain information, and use timeouts for more reliable execution.

### DIFF
--- a/test/e2e/performanceprofile/functests/1_performance/performance.go
+++ b/test/e2e/performanceprofile/functests/1_performance/performance.go
@@ -394,24 +394,71 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
 	})
 
 	Context("Tuned kernel parameters", Label(string(label.Tier0)), func() {
+		// When a higher-priority Tuned wraps the performance profile via include=<pp-profile>
+		// (e.g. ran-du-performance), network-latency's kernel.timer_migration=0 can win over
+		// the NTO-generated override to 1. Reused the include-chain helpers from test 37127
+		// to read each node's active profile and adjust the expectation when a wrapper is active.
 		It("[test_id:28466][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should contain configuration injected through openshift-node-performance profile", func() {
+			ctx, cancel := context.WithTimeout(context.Background(), cluster.ComputeTestTimeout(testTimeout*time.Second, RunningOnSingleNode))
+			defer cancel()
+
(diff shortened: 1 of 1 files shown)


test/e2e/performanceprofile/functests/1_performance/performance.go
```diff
@@ -394,24 +394,71 @@ var _ = Describe("[rfe_id:27368][performance]", Ordered, func() {
...
+				Eventually(func(g Gomega) {
+					tunedProfile := &tunedv1.Profile{}
+					g.Expect(testclient.DataPlaneClient.Get(ctx, profileKey, tunedProfile)).To(Succeed(),
+						"failed to get Tuned profile for node %s", node.Name)
+
+					activeProfile := tunedProfile.Status.TunedProfile
+					testlog.Infof("Node %s has active tuned profile: %q", node.Name, activeProfile)
+
+					expectedTimerMigration := "1"
+					if activeProfile != performanceProfileName {
```
MarSik, line 443: The logic of "ha, I see there is a Tuned override and for that reason timer_migration will be disabled" is not sound. The current version of RAN Tuned overrides does not do that for example (https://github.com/openshift-kni/telco-reference/blob/main/telco-ran/configuration/source-crs/node-tuning-operator/TunedPerformancePatch.yaml).

Also notice that in the previous version we actually enabled and disabled tuning migration based on workload hints even when the same set of Tuned objects were used.

If anything, you would have to parse all the included profiles and figure this out.

MarSik:
"On clusters where ran-du-performance wraps the PP via include=, network-latency in the chain" - what does this mean? ran-du-performance is a higher priority profile and includes= the performance profile. 

network-latency is always in the chain as it is included by cpu-partitioning tuned profile here https://github.com/openshift/cluster-node-tuning-operator/blob/main/assets/performanceprofile/tuned/openshift-node-performance#L18 and https://github.com/redhat-performance/tuned/blob/master/profiles/cpu-partitioning/tuned.conf#L6

Both the logic and the explanation are confusing to me.

PR #1591, opened 2026-08-18
OCPBUGS-112579: unconditionally disable fwupd-refresh.timer

bartwensley:
In RHEL 10, the fwupd-refresh.timer is enabled by default and it periodically triggers the fwupd-refresh service, which can trigger SMIs on all CPUs (including isolated CPUs) and introduce latency spikes that are unacceptable for latency sensitive workloads.

Updating the openshift-node-performance profile to stop and disable the fwupd-refresh.timer unconditionally. The fwupd-refresh service is not required and was not enabled prior to RHEL 10.

## Summary by CodeRabbit

* **Bug Fixes**
  * Performance profiles now stop and disable the firmware refresh timer to help prevent firmware activity, system interruptions, and latency spikes on isolated CPUs.
  * Existing real-time service tuning remains unchanged across supported performance profiles.

* **Tests**
  * Updated rendered profile expectations for bootstrap, ARM, CPU frequency, manual, and no-reference configurations.

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -44,11 +44,13 @@ energy_perf_bias=performance
 min_perf_pct=100
 {{end}}
 
-{{if .RealTimeHint}}
 [service]
+# Unconditionally stop and disable the fwupd firmware refresh timer. It triggers
+# the fwupd service which can cause SMIs (and latency spikes) on isolated CPUs.
+service.fwupd-refresh.timer=stop,disable
+{{if .RealTimeHint}}
 service.stalld=start,enable
 {{else}}
-[service]
 service.stalld=stop,disable
 {{end}}
 
--- a/test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-mcp/openshift-bootstrap-master_tuned.yaml
+++ b/test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-mcp/openshift-bootstrap-master_tuned.yaml
@@ -26,7 +26,9 @@ spec:
       All values are mapped with a comment where a parent profile contains them.\n#
       Different values will override the original values in parent profiles.\n\n[variables]\n#>
       isolated_cores take a list of ranges; e.g. isolated_cores=2,4-7\n\nisolated_cores=0-1\n\n\nnot_isolated_cores_expanded=${f:cpulist_invert:${isolated_cores_expanded}}\n\n\n[cpu]\n#>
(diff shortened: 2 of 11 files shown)


assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -44,11 +44,13 @@ energy_perf_bias=performance
...
 
-{{if .RealTimeHint}}
 [service]
+# Unconditionally stop and disable the fwupd firmware refresh timer. It triggers
+# the fwupd service which can cause SMIs (and latency spikes) on isolated CPUs.
+service.fwupd-refresh.timer=stop,disable
+{{if .RealTimeHint}}
 service.stalld=start,enable
 {{else}}
-[service]
```
Tal-or, line 51: I see this field was removed, is that on purpose?

bartwensley: The `[service]` plugin name was moved to above the if/else (line 47) so it only needs to appear once now.

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -44,11 +44,13 @@ energy_perf_bias=performance
 min_perf_pct=100
 {{end}}
 
-{{if .RealTimeHint}}
 [service]
+# Unconditionally stop and disable the fwupd firmware refresh timer. It triggers
+# the fwupd service which can cause SMIs (and latency spikes) on isolated CPUs.
+service.fwupd-refresh.timer=stop,disable
```
MarSik, line 50: Is the service present on RHEL 9 (even disabled)? If not then we might need to make this conditional based on the RHEL version. Otherwise this looks ok to me (and does service. work for timers?).

bartwensley: > Is the service present on RHEL 9 (even disabled)?

Yes

> does service. work for timers?

Yes

bartwensley:
The ci/prow/e2e-gcp-pao-updating-profile job is currently broken and failing for all PRs.

assets/performanceprofile/tuned/openshift-node-performance
MarSik: /lgtm then

MarSik (approved):

bartwensley:
Rebased to pick up fix for CI failure.

bartwensley:
The failed tests both passed on the original commit so must be flakey.

/retest

MarSik:
/retitle OCPBUGS-112579: unconditionally disable fwupd-refresh.timer

PR #1596, opened 2026-08-20
OCPBUGS-66217: PerformanceProfile and Tuned CRDs during bootstrap

MarSik:
Make sure both the main NTO CRDs are deployed early on bootstrap so day 0 rendering can rely on those two CRDs being available.

## Summary by CodeRabbit

* **Improvements**
  * Updated performance and node-tuning resource definitions to support required initialization during system setup.


Tal-or (approved):
/lgtm

PR #1614, opened 2026-08-25
OCPBUGS-113651: Disable timer.migration on RHCOS 10

MarSik:
The new timer management in RHEL 10 (timer wheel) makes the timer migration obsolete.

This solves two issue in one stroke - no latency for polling workloads and no issues with high resolution timers either.

## Summary by CodeRabbit

* **Bug Fixes**
  * Applied timer migration settings only on RHEL 9 systems.
  * Prevented the setting from being enabled by default on newer kernel versions, including RHEL 10.
  * Documented the setting’s scope and known timer-related effects.

* **Tests**
  * Updated performance profile validation and rendering expectations to reflect RHEL 9-specific behavior across supported profile variants.

### DIFF
--- a/assets/performanceprofile/tuned/openshift-node-performance
+++ b/assets/performanceprofile/tuned/openshift-node-performance
@@ -122,14 +122,27 @@ vm.swappiness=10
 #> rps configuration
 # net.core.rps_default_mask=${not_isolated_cpumask}
 
-# cpu-partitioning for RHEL disables it (= 0)
-# OCP is too dynamic when partitioning and needs to evacuate
+[sysctl.timer-migration]
+type=sysctl
+# Match uname with kernel versions containing
+# .el9.x86_64 or .el9_<number>, but not .el91
+uname_regex=\.el9[ _.]
+# cpu-partitioning for RHEL disables timer.migration (= 0),
+# but OCP is too dynamic when partitioning and needs to evacuate
 # scheduled timers when starting a guaranteed workload (= 1)
+# 
 # This has an unfortunate side-effect that HRTIMERs break
 # and cyclictest results will be affected, but there is no
 # way around that. Not enabling this breaks cpu pinned
 # workload isolation for example when any TCP timeout
 # and keep alive timers are present.
+#
+# This is only needed for RHEL 9 based platforms.
+# RHEL 10 kernels reworked the timer management and are
+# no longer affected by this issue.
+#
+# This whole block can be removed when the RHEL 9
+# support is dropped.
 kernel.timer_migration=1
 
 [selinux]
--- a/test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-mcp/openshift-bootstrap-master_tuned.yaml
+++ b/test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-mcp/openshift-bootstrap-master_tuned.yaml
@@ -49,16 +49,21 @@ spec:
       move them to swap cache\n#> latency-performance\nvm.swappiness=10\n\n# also
       configured via a sysctl.d file\n# placed here for documentation purposes and
       commented out due\n# to a tuned logging bug complaining about duplicate sysctl:\n#
-      \  https://issues.redhat.com/browse/RHEL-18972\n#> rps configuration\n# net.core.rps_default_mask=${not_isolated_cpumask}\n\n#
-      cpu-partitioning for RHEL disables it (= 0)\n# OCP is too dynamic when partitioning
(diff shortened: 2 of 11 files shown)


assets/performanceprofile/tuned/openshift-node-performance
bartwensley: I think this comment should be updated to indicate the difference in behaviour in RHEL 9 vs. RHEL 10 (can probably delete most of the existing comment since this is no longer going to apply). Also would be good to mention that this should be removed in OCP 5.1 (since OCP 5.1 NTO should only be running on RHEL 10).

assets/performanceprofile/tuned/openshift-node-performance
```diff
@@ -122,6 +122,9 @@ vm.swappiness=10
 #> rps configuration
 # net.core.rps_default_mask=${not_isolated_cpumask}
 
+[sysctl.timer-migration]
+type=sysctl
+uname_regex=\.el9
```
bartwensley: nit: Would this match "el91"? Would "el9_" be better?

MarSik: It would, but.. are you expecting CoreOS 91 any time soon? Btw, kernels might have just ".el9" with no suffix too and your regex would ignore those.

MarSik: @jmencak I actually wonder.. do we care about OKD? They probably have different unames too, right?

test/e2e/performanceprofile/testdata/render-expected-output/bootstrap/extra-mcp/openshift-bootstrap-master_tuned.yaml
```diff
@@ -49,13 +49,17 @@ spec:
...
-      \  https://issues.redhat.com/browse/RHEL-18972\n#> rps configuration\n# net.core.rps_default_mask=${not_isolated_cpumask}\n\n#
-      cpu-partitioning for RHEL disables it (= 0)\n# OCP is too dynamic when partitioning
-      and needs to evacuate\n# scheduled timers when starting a guaranteed workload
-      (= 1)\n# This has an unfortunate side-effect that HRTIMERs break\n# and cyclictest
-      results will be affected, but there is no\n# way around that. Not enabling this
-      breaks cpu pinned\n# workload isolation for example when any TCP timeout\n#
-      and keep alive timers are present.\nkernel.timer_migration=1\n\n[selinux]\n#>
+      \  https://issues.redhat.com/browse/RHEL-18972\n#> rps configuration\n# net.core.rps_default_mask=${not_isolated_cpumask}\n\n[sysctl.timer-migration]\ntype=sysctl\n#
+      Match uname with kernel versions containing\n# .el9 or .el9_<number>, but not
+      .el91\nuname_regex=\\.el9[ _]\n# cpu-partitioning for RHEL disables timer.migration
```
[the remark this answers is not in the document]
jmencak: @MarSik , probably missing `make render-sync`

assets/performanceprofile/tuned/openshift-node-performance
jmencak: Likely, check out https://amd64.origin.releases.ci.openshift.org/

bartwensley: @MarSik - I agree that 91 is unlikely - that's why this was marked as a nit. I don't see how ".el9" is possible as the NTO version 5.0 should only be running on OCP 4.23 or OCP 5.0 and later - right?

MarSik: Funny, OKD 4.23 is actually using CentOS 10 kernel

https://amd64.origin.releases.ci.openshift.org/releasestream/4.23.0-0.okd-scos-nightly/release/4.23.0-0.okd-scos-nightly-2026-08-22-063941

```
CentOS Stream CoreOS 10 (stream-coreos)
Package List
    kernel-6.12.0-254.el10
```

And notice there is still .el10 in the kernel name so the regex should be fine.

MarSik: @bartwensley Both 4.23 and 5.0 can be installed with the RHEL 9 kernel afaik. And .el9 is possible if you build your own kernel for example or use the kernel directly from RHEL mainline. So this is mainly to not complicate testing.

bartwensley (approved):
Thanks Martin  - looks good.

MarSik:
/verified by [test_id:28466][crit:high][vendor:cnf-qe@redhat.com][level:acceptance] Should contain configuration injected through openshift-node-performance profile

PR #1618, opened 2026-08-26
OCPBUGS-105474: Prevent offlined CPUs from being re-onlined after reboot by udev hotplug rule

tavital:
CPUs that a PerformanceProfile marks as offline are sometimes coming back online after a reboot:

The RHCOS CPU hotplug rule, shipped read-only in
`/usr/lib/udev/rules.d/40-redhat-hotplug.rules` , re-onlines CPUs on udev
`"add"` events, which races with NTO's `set-cpus-offline.service`. When the
udev rule wins, the CPUs actually offlined on the node are only a subset
of those defined in the profile's `spec.cpu.offlined`.

Fix this by shadowing the existing rule with a same-named, empty file
under `/etc/udev/rules.d`, which udev prefers over `/usr/lib`, so the rule
no longer re-onlines the offlined CPUs.

Note: We are currently in contact with the RHCOS team to remove or modify this rule upstream. 
If they agree to drop the rule, this temporary workaround can be safely reverted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented system hotplug behavior from re-enabling CPUs configured as offlined.
  * Ensured configured-offline CPUs remain offline after device events.
  * Updated generated MachineConfigs consistently across supported performance profile configurations.
  * Simplified IRQ affinity handling by removing obsolete default SMP affinity adjustments.

### DIFF
--- a/assets/performanceprofile/configs/40-redhat-hotplug.rules
+++ b/assets/performanceprofile/configs/40-redhat-hotplug.rules
@@ -1,0 +1,6 @@
+# Managed by the Node Tuning Operator. Do not edit.
+#
+# This file shadows the default read-only rule /usr/lib/udev/rules.d/40-redhat-hotplug.rules,
+# which re-onlines CPUs on udev "add" events and races NTO's set-cpus-offline
+# service, re-onlining CPUs the profile offlined. This empty file overrides that
+# rule and keeps the offlined CPUs offline. Written only when spec.cpu.offlined is set.
--- a/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
+++ b/pkg/performanceprofile/controller/performanceprofile/components/machineconfig/machineconfig.go
@@ -65,6 +65,7 @@ const (
 
 	udevRulesDir         = "/etc/udev/rules.d"
 	udevPhysicalRpsRules = "99-netdev-physical-rps.rules"
+	udevHotplugRules     = "40-redhat-hotplug.rules"
 	// scripts
 	hugepagesAllocation       = "hugepages-allocation"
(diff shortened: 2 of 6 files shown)


MarSik (commented):
I would actually make the file non-empty and put a comment explaining why the shadowing was installed into it. This will help support engineers and even us devels if we ever come across a bug or a customer case with a system where this plays a role.

Tal-or (commented):
From the PR description:
> CPUs that a PerformanceProfile marks offline could come back online after a reboot. 

Rephrase: Issue: CPUs that a PerformanceProfile marks as offline are sometimes coming back online after a reboot. 

the sentence looks like it start from the middle.


Also, can we add a comment (can be in the PR description) why didn't we remove the redhat hotplug service from RHCOS? IIRC we said that OCP doesn't support hotplug anyway

tavital:
> From the PR description:
> 
> > CPUs that a PerformanceProfile marks offline could come back online after a reboot.
> 
> Rephrase: Issue: CPUs that a PerformanceProfile marks as offline are sometimes coming back online after a reboot.
> 
> the sentence looks like it start from the middle.
> 
> Also, can we add a comment (can be in the PR description) why didn't we remove the redhat hotplug service from RHCOS? IIRC we said that OCP doesn't support hotplug anyway

A comment was added in PR description.

tavital:
> I would actually make the file non-empty and put a comment explaining why the shadowing was installed into it. This will help support engineers and even us devels if we ever come across a bug or a customer case with a system where this plays a role.

Added some comments in the file as requested, describing its role.

MarSik (approved):
/lgtm

PR #1628, opened 2026-09-02
OCPBUGS-127362: E2E: ovsDpdk remaining cases and consolidate suite

oblau:
This PR has 4 commits:
1. Merge the two Ordered ovsDpdk Describes into one parent with two
   Contexts (shared setup/teardown, fewer MCP reboots) and add test ids
2. Simplify By/Expect formatting
3. Assert OVS_DPDK_CPUS in 89987; add ovs-vswitchd restart case (89996)
4. Add remaining cases: sched domains, IRQ ban across reboot, expand
   ovsDpdk, mixedCpus coexistence, cleanup on remove; consolidate into
   shared verify helpers

## Summary by CodeRabbit

* **Tests**
  * Reworked OVS-DPDK end-to-end coverage for clearer scenarios with CPU load balancing enabled and disabled.
  * Added validation for CPU scheduling domains, IRQ isolation, service environment settings, CPU expansion, mixed CPU configurations, and node reboots.
  * Added coverage confirming cleanup when OVS-DPDK is removed and configuration persistence across service restarts.
  * Improved test cleanup and profile restoration between scenarios, with enhanced scenario identification.


test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -26,15 +28,18 @@ import (
...
+const crioRuntimesConfigFile = "/rootfs/etc/crio/crio.conf.d/99-runtimes.conf"
 
 var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDpdk), string(label.Slow), string(label.Tier2)), func() {
 	var (
-		workerRTNodes  []corev1.Node
-		profile        *performancev2.PerformanceProfile
-		initialProfile *performancev2.PerformanceProfile
+		workerRTNodes   []corev1.Node
+		baselineProfile *performancev2.PerformanceProfile
+		initialProfile  *performancev2.PerformanceProfile
```
Tal-or, line 42: Not clear what is the difference between baslineProfile and initialProfile. the name are misleading since usually both use to describe the initial state.

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -132,67 +130,67 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
 			node := &workerRTNodes[0]
 			testlog.Infof("Verifying node %s", node.Name)
 
-			By(fmt.Sprintf("Verifying kernel cmdline on node %s", node.Name))
```
Tal-or, line 135: I'm always in favor of giving extra details if possible when it cost free like in this case, i.e keep node name.

you can add a wrapper:
```
func ByF(format string, a ...any) {
  By(fmt.Sprintf(format, a...))
}
```
if you would like to save the extra wording but not critical imo.

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -132,67 +130,67 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
 
-			By(fmt.Sprintf("Verifying isolcpus includes ovsDpdk CPUs on node %s", node.Name))
+			By("Verifying isolcpus includes ovsDpdk CPUs")
 			isolcpusSet := parseCPUSetFromKernelParam(cmdlineStr, "isolcpus")
 			Expect(isolcpusSet.IsEmpty()).To(BeFalse(), "isolcpus param not found in cmdline")
 			Expect(expectedIsolatedPlusOvsDpdk.IsSubsetOf(isolcpusSet)).To(BeTrue(),
-				fmt.Sprintf("isolcpus=%s should include all isolated + ovsDpdk CPUs %s",
-					isolcpusSet.String(), expectedIsolatedPlusOvsDpdk.String()))
+				"isolcpus=%s should include all isolated + ovsDpdk CPUs %s",
+				isolcpusSet.String(), expectedIsolatedPlusOvsDpdk.String())
```
Tal-or, line 143: we can drop `.String()` calls. they'll be called implicitly.

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -132,67 +130,67 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
 
-			By(fmt.Sprintf("Verifying nohz_full includes ovsDpdk CPUs on node %s", node.Name))
+			By("Verifying nohz_full includes ovsDpdk CPUs")
 			nohzSet := parseCPUSetFromKernelParam(cmdlineStr, "nohz_full")
 			Expect(nohzSet.IsEmpty()).To(BeFalse(), "nohz_full param not found in cmdline")
 			Expect(ovsDpdkSet.IsSubsetOf(nohzSet)).To(BeTrue(),
-				fmt.Sprintf("nohz_full=%s should include ovsDpdk CPUs %s",
-					nohzSet.String(), ovsDpdkSet.String()))
+				"nohz_full=%s should include ovsDpdk CPUs %s",
+				nohzSet.String(), ovsDpdkSet.String())
```
Tal-or, line 150: ditto

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -132,67 +130,67 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
 
-			By(fmt.Sprintf("Verifying rcu_nocbs includes ovsDpdk CPUs on node %s", node.Name))
+			By("Verifying rcu_nocbs includes ovsDpdk CPUs")
 			rcuSet := parseCPUSetFromKernelParam(cmdlineStr, "rcu_nocbs")
 			Expect(rcuSet.IsEmpty()).To(BeFalse(), "rcu_nocbs param not found in cmdline")
 			Expect(ovsDpdkSet.IsSubsetOf(rcuSet)).To(BeTrue(),
-				fmt.Sprintf("rcu_nocbs=%s should include ovsDpdk CPUs %s",
-					rcuSet.String(), ovsDpdkSet.String()))
+				"rcu_nocbs=%s should include ovsDpdk CPUs %s",
+				rcuSet.String(), ovsDpdkSet.String())
```
Tal-or, line 157: ditto

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -132,67 +130,67 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
 
-			By(fmt.Sprintf("Verifying systemd.cpu_affinity excludes ovsDpdk CPUs on node %s", node.Name))
+			By("Verifying systemd.cpu_affinity excludes ovsDpdk CPUs")
 			affinitySet := parseCPUSetFromKernelParam(cmdlineStr, "systemd.cpu_affinity")
 			Expect(affinitySet.IsEmpty()).To(BeFalse(), "systemd.cpu_affinity param not found in cmdline")
 			Expect(affinitySet.Intersection(ovsDpdkSet).IsEmpty()).To(BeTrue(),
-				fmt.Sprintf("systemd.cpu_affinity=%s should not contain ovsDpdk CPUs %s",
-					affinitySet.String(), ovsDpdkSet.String()))
+				"systemd.cpu_affinity=%s should not contain ovsDpdk CPUs %s",
+				affinitySet.String(), ovsDpdkSet.String())
```
Tal-or, line 164: ditto

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -132,67 +130,67 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
+			By("Verifying kubelet reservedSystemCPUs is union of reserved + ovsDpdk")
 			kubeletConfig, err := nodes.GetKubeletConfig(ctx, node)
 			Expect(err).ToNot(HaveOccurred())
 			reservedSystemCPUs, err := cpuset.Parse(kubeletConfig.ReservedSystemCPUs)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(reservedSystemCPUs.Equals(expectedReservedSystem)).To(BeTrue(),
-				fmt.Sprintf("ReservedSystemCPUs should be %s (reserved + ovsDpdk), got %s",
-					expectedReservedSystem.String(), reservedSystemCPUs.String()))
+				"ReservedSystemCPUs should be %s (reserved + ovsDpdk), got %s",
+				expectedReservedSystem.String(), reservedSystemCPUs.String())
```
Tal-or: ditto

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -132,67 +130,67 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
 			irqConf, err := nodes.ExecCommand(ctx, node, []string{"cat", "/rootfs/etc/sysconfig/irqbalance"})
 			Expect(err).ToNot(HaveOccurred())
 			irqConfStr := testutils.ToString(irqConf)
 
 			bannedSet := getIRQBannedCPUSet(irqConfStr)
 			Expect(ovsDpdkSet.IsSubsetOf(bannedSet)).To(BeTrue(),
-				fmt.Sprintf("IRQBALANCE_BANNED_CPUS should include ovsDpdk CPUs %s, got %s",
-					ovsDpdkSet.String(), bannedSet.String()))
+				"IRQBALANCE_BANNED_CPUS should include ovsDpdk CPUs %s, got %s",
+				ovsDpdkSet.String(), bannedSet.String())
```
Tal-or, line 183: ditto

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -132,67 +130,67 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
-			By(fmt.Sprintf("Verifying default_smp_affinity excludes ovsDpdk CPUs on node %s", node.Name))
+			By("Verifying default_smp_affinity excludes ovsDpdk CPUs")
 			smpCPUSet, err := nodes.GetDefaultSmpAffinitySet(ctx, node)
 			Expect(err).ToNot(HaveOccurred())
 			testlog.Infof("default_smp_affinity on %s: %s", node.Name, smpCPUSet.String())
 			Expect(smpCPUSet.Intersection(ovsDpdkSet).IsEmpty()).To(BeTrue(),
-				fmt.Sprintf("default_smp_affinity should not have ovsDpdk CPU bits set, got CPUs %s",
-					smpCPUSet.Intersection(ovsDpdkSet).String()))
+				"default_smp_affinity should not have ovsDpdk CPU bits set, got CPUs %s",
+				smpCPUSet.Intersection(ovsDpdkSet).String())
```
Tal-or, line 191: ditto. there are more occurrences below

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -248,22 +187,181 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
 			verifyOvsDpdkIRQBanningAcrossGUPodLifecycle(context.TODO(), &workerRTNodes[0], baselineProfile, ovsDpdkSet, smtAlignmentDisabled)
 		})
+
+		It("[test_id:89992] should keep ovsDpdk CPUs outside kernel scheduling domains", func() {
+			ctx := context.TODO()
+			node := &workerRTNodes[0]
+			testlog.Infof("Verifying node %s", node.Name)
+
+			By("Verifying ovsDpdk CPUs are outside kernel scheduling domains before pod creation")
+			out, err := nodes.ExecCommand(ctx, node, []string{"/bin/bash", "-c", "cat /proc/schedstat"})
```
Tal-or, line 203: also check output is not empty/in the expected format

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -342,16 +430,130 @@ var _ = Describe("[performance] ovsDpdk CPUs", Ordered, Label(string(label.OvsDp
...
+			By("Applying profile and waiting for MCP")
+			profilesupdate.ApplyProfileAndWait(ctx, profile)
+
+			By("Verifying kubelet reservedSystemCPUs is reserved + shared + ovsDpdk")
+			expectedReservedSystem := reservedSet.Union(sharedSet).Union(ovsDpdkSet)
+			reservedSystemCPUs, err := getReservedSystemCPUs(ctx, node)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(reservedSystemCPUs.Equals(expectedReservedSystem)).To(BeTrue(),
+				"ReservedSystemCPUs should be %s (reserved + shared + ovsDpdk), got %s",
+				expectedReservedSystem.String(), reservedSystemCPUs.String())
```
Tal-or, line 532: `String()` is being implicitly called, you can drop it. remove it also from other occurrences in the code.

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
```diff
@@ -538,3 +734,114 @@ func getOvsDpdkSliceExclusiveCPUs(ctx context.Context, node *corev1.Node) (cpuse
 	}
 	return cpuset.Parse(strings.TrimSpace(testutils.ToString(cgroupCpus)))
 }
+
+func getReservedSystemCPUs(ctx context.Context, node *corev1.Node) (cpuset.CPUSet, error) {
```
Tal-or, line 738: Is there a reason why this and `getNodeIRQBannedCPUSet` returning error and the rest are containing assertions?

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
oblau: Hmm i see, Initial profile is profile before suite started. 
Baseline is a shared baseline for each context to apply (like carving the ovsdpdk cpus and other shared requirements).
Its just a baseline since context 1 adds the annotation on top of it for example.
its there so i don't repeat the the same preparation in each context's before all.

Do you have a suggestion for a better name, i can see now how it can come across as confusing ?

Tal-or: I would simply call it `profile` and then modify it inside each context.

you don't need the baslineProfile state anyway, you'll be reverting whatever you have back to `initialProfile`

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
oblau: Thanks, I left it out because some hostnames are long and the trace is harder to read when every step repeats it. Each spec already logs the node once at the start via testlog, so the name is still in the output. In most of functests, By is the step itself, and the node is printed as a log when needed. Does that work for you, or would you still rather have it on each By?

test/e2e/performanceprofile/functests/2_performance_update/ovsdpdk.go
oblau: I split them on purpose: get* only fetch (so they return error), and verify* assert the expected result. getNodeIRQBannedCPUSet is also used from Eventually after pod deletion so a failed read can retry. Happy to change it if you’d rather have the getters assert internally.

Tal-or (commented):
/ok-to-test

Tal-or (approved):
/lgtm
Thank you for working on this

@MarSik @yanirq PTAL

ffromani (commented):
/approve

trusting prior LGTM

PR #1629, opened 2026-09-02
OCPBUGS-116703: E2E: add missing matcher to offlined CPU Expect() calls

oblau:
Expect(bool) without .To() is a no-op; these tests never fail on these assertions.

## Summary by CodeRabbit

- **Tests**
  - Improved end-to-end performance profile test failures by including the expected and actual offlined CPU sets in assertion messages.
  - This provides clearer diagnostic information when the node’s CPU configuration differs from the profile configuration, making test failures easier to understand and investigate.


jmencak:
The failure `ci/prow/okd-scos-images` looks suspicious, I'll look into it.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -708,7 +708,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				Expect(err).ToNot(HaveOccurred())
 				offlinedCPUSetProfile, err := cpuset.Parse(string(offlined))
 				Expect(err).ToNot(HaveOccurred())
-				Expect(offlinedCPUSet.Equals(offlinedCPUSetProfile))
+				Expect(offlinedCPUSet.Equals(offlinedCPUSetProfile)).To(BeTrue(), "offlined CPUs mismatch: expected %q, got %q", offlinedCPUSetProfile, offlinedCPUSet)
```
ffromani, line 711: this is good. What's better would be checking the two sets for direct equality:
```
Expect(offlinedCPUSet).To(Equal(offlinedCPUSetProfile), "offlined CPUs mismatch: expected %q, got %q", offlinedCPUSetProfile, offlinedCPUSet)
```
we will probably need to either pass through `String()` or to use a custom gomega matcher.
Applies to all the following instances

Tal-or: I think converting to `%s` will call `String()` implicitly

oblau: Switched all five sites to `Expect(offlinedCPUSet).To(Equal(offlinedCPUSetProfile), ...).`

No extra `.String()` wrap and no custom matcher. Gomega `Equal` uses `reflect.DeepEqual`, same as `CPUSet.Equals()`, so comparing the two sets directly is ok as far as i understand.

also, re-verified the tests on 5.1 and they passed.

test/e2e/performanceprofile/functests/2_performance_update/updating_profile.go
```diff
@@ -708,7 +708,7 @@ var _ = Describe("[rfe_id:28761][performance] Updating parameters in performance
 				Expect(err).ToNot(HaveOccurred())
 				offlinedCPUSetProfile, err := cpuset.Parse(string(offlined))
 				Expect(err).ToNot(HaveOccurred())
-				Expect(offlinedCPUSet.Equals(offlinedCPUSetProfile))
+				Expect(offlinedCPUSet).To(Equal(offlinedCPUSetProfile), "offlined CPUs mismatch: expected %q, got %q", offlinedCPUSetProfile, offlinedCPUSet)
```
ffromani: I checked, and deeps down this does deep struct field equality check, not too differently than `reflect.DeepEqual`. Currently it's fine, but is still liable to break in the future.
If you're keen to change again:
```
  func EqualCPUSet(expected cpuset.CPUSet) types.GomegaMatcher {
  	return gcustom.MakeMatcher(func(actual cpuset.CPUSet) (bool, error) {
  		return actual.Equals(expected), nil
  	}).WithTemplate(
  		"Expected CPUs {{.Actual.String}} to equal {{.Data.String}}",
  		expected,
  	)
  }
```
Then use it like
```
Expect(offlinedCPUSet).To(EqualCPUSet(offlinedCPUSetProfile))
```

Tal-or: I'm not sure why we're not using the built in `Equals` if the operation was no-op all we gotta do is:
```
Expect(offlinedCPUSet.Equals(offlinedCPUSetProfile)).To(BeTrue(), "offlined CPUs mismatch: expected %q, got %q", offlinedCPUSetProfile, offlinedCPUSet)
```

Tal-or (approved):
/lgtm

PR #1633, opened 2026-09-08
NO-JIRA: Fix CentOS Stream image build failing in dnf history undo

jmencak:
The CentOS branch of hack/dockerfile_install_support.sh used
"dnf history undo 0" to clean up the tuned build dependencies.
That reverts every dnf transaction back to the base image state,
which requires re-downloading the old versions of the packages
that "dnf build-dep" upgraded (e.g. python3-3.9.25-2,
systemd-252-64).   As CentOS Stream 9 is a rolling release, those
old builds are no longer available in the repositories, so the
undo fails with "Cannot find rpm nevra" and the image build is
broken.

Drop the line.  The build dependencies do not need to be removed
right after the RPM build: the existing "dnf remove" of the build
tools and the final "dnf autoremove" already drop the build tools
and everything installed only as build dependencies, without
downgrading anything.

## Summary by CodeRabbit

- **Build and Packaging**
  - CentOS builds no longer automatically remove build dependencies after creating TuneD RPMs.

### DIFF
--- a/hack/dockerfile_install_support.sh
+++ b/hack/dockerfile_install_support.sh
@@ -29,7 +29,6 @@ if [[ "${ID}" == "centos" ]]; then
   dnf build-dep tuned.spec -y
   make rpm PYTHON=/usr/bin/python3
   rm -rf /root/rpmbuild/RPMS/noarch/{tuned-gtk*,tuned-utils*,tuned-profiles-compat*}
-  dnf --setopt=protected_packages= history -y undo 0  # Remove builddep
 
   cp -r /root/rpmbuild/RPMS/noarch /root/rpms
   dnf install --setopt=tsflags=nodocs -y ${INSTALL_PKGS}


Tal-or (commented):
In general what we could do is executing`RUN /bin/bash /tmp/dockerfile_install_support.sh` in the builder image or on a separate stage and then only copy the artifacts we care about. this way we should not be worrying about cleaning and also making the final image size smaller.

jmencak:
> In general what we could do is executing`RUN /bin/bash /tmp/dockerfile_install_support.sh` in the builder image or on a separate stage and then only copy the artifacts we care about. this way we should not be worrying about cleaning and also making the final image size smaller.

In theory, yes, but not really in practice, @Tal-or .  There are some restrictions on the number of stages the upstream container file can have and the number is two.  Creating a 3 stage containerfile would complicate things and I'm not sure I want to mess with that.  I don't remember the full details, but that's why things are the way they are.

Edit: [some details about 2 vs 3 stages](https://github.com/openshift/cluster-node-tuning-operator/pull/1051#issuecomment-2096741970)

Tal-or (approved):
/lgtm
TIL. Thank you @jmencak for the clarification

MarSik (approved):
/lgtm

jmencak:
This is a known issue, we're hitting 4h CI limits to pass this test.  This commit cannot cause these failures.  See: https://redhat.atlassian.net/browse/CNF-27079

PR #1635, opened 2026-09-14
Skip cyclictest latency test on non-RT kernels

tavital:
cyclictest measures real-time scheduler wake-up latency, so its pass/fail verdict is only meaningful on a PREEMPT_RT kernel, where the worst-case latency is bounded and repeatable.

On a non-RT kernel the worst-case latency is unbounded and non-deterministic, so no threshold is both correct and stable and raising it does not help. The result is a flaky, meaningless pass/fail. This is the case on aarch64 today, which has no RT kernel available.

Add IsRealTimeKernelEnabled, a profile-package helper that reads the setting directly from the PerformanceProfile spec, and use it to skip cyclictest when the profile does not enable the real-time kernel. The check runs before the pod is created. oslat and hwlatdetect are unaffected: they measure hardware and firmware jitter, which is bounded regardless of the kernel.

## Summary by CodeRabbit

* **Bug Fixes**
  * Latency testing now skips profiles that do not enable the real-time (PREEMPT_RT) kernel, avoiding inapplicable test failures.

* **Tests**
  * Added coverage for detecting whether the real-time kernel is enabled, including unset and explicitly enabled or disabled configurations.


jmencak:
okd-scos-images fails with "nevra" issues, see https://github.com/openshift/cluster-node-tuning-operator/pull/1633

tavital:
PR closed because customers in fact use non-rt with cyclictest (they should raise the max latency manually for cyclictest if reached).

tavital:
PR closed because customers in fact use non-rt with cyclictest (they should raise the max latency manually for cyclictest if reached).

PR #1638, opened 2026-09-17
OCPBUGS-126458:  E2E: PPC: replace no-op log-on-match with actual assertions (4 sites)

oblau:
Split out of #1571 (PPC slice only)

PPC failure tests never asserted the error text. After `regexp.MatchString`, `if ok { testlog.Info(...) }` only logged a hit. A miss skipped the log and the case still passed if the process exited 1.
Replace with `Wait` + `gexec.Exit`, then `ContainSubstring` on the leaf error (not outer wraps, not regex):

Also move `gexec.Exit` to right after `Wait`, and fix success-path `"marshal"` → `"unmarshal"`.

## Summary by CodeRabbit

- **Tests**
  - Improved PPC end-to-end test reliability by adding explicit session timeouts and validating command exit codes before processing output.
  - Updated error checks to use clearer substring matching and refreshed expected messages.
  - Made reserved CPU count validation adapt to values collected from the test environment.
  - Corrected misleading assertion failure text.


test/e2e/performanceprofile/functests/10_performance_ppc/ppc.go
```diff
@@ -116,14 +119,12 @@ var _ = Describe("[rfe_id: 38968] PerformanceProfile setup helper and platform a
...
-			ok, err := regexp.MatchString(errString, string(output))
-			Expect(err).ToNot(HaveOccurred())
-			if ok {
-				testlog.Info(errString)
-			}
-			Eventually(session).Should(gexec.Exit(1))
+			Expect(session).Should(gexec.Exit(1))
+
+			errString := "not appropriate to split reserved CPUs in case of topology-manager-policy: single-numa-node"
+			Expect(string(output)).To(ContainSubstring(errString), "expected error:\n%q\ngot:\n%s", errString, output)
```
Tal-or: Not sure why do we need so many new lines, but no biggy

Tal-or (approved):
/lgtm

Thank you

